PR #117, opened 2013-12-02
Sort operations in generated code

vaelen:
This update sorts the operations in each class before generating client code.

Previously the operations were not sorted, so regenerating client code could result in git changesets when there were no real changes, only a reordering of the methods within their classes.  This code sorts by path, verb (httpMethod), and finally method name (nickname).  The last of these is only a failsafe, as there should never be two methods with the same path and verb assigned to them.

An alternative method would be to sort the list according to the method name (nickname) only, which would produce output classes with properly ordered methods.  However, in my experience the method names are more likely to change than the path and verb, and so this ordering creates a more stable sort.


fehguy:
Changed in 2.1.x, this isn't relevant anymore.

PR #155, opened 2014-02-15
Package separator - PHP 5.3

yanoosh:
Php has different namespace separator - use backslash not dot. The lowest supported version of PHP has implemented namespaces. This change give opportunity to add composer.json (file used by PHP package manager).


fehguy:
This should be supported now.

PR #163, opened 2014-03-06
Upgrade Objective-C templates to AFNetworking 2.

mkantor:
I should note that I couldn't get the tests from the wordnik-api and petstore samples to run, but this passes all of the tests for the project we're integrating it into (which is unfortunately private). Coverage-wise I know we don't hit either of the `multipart/form-data` file upload code paths, so those are untested.

Any advice about how I might be able to run the tests would be appreciated. The petstore sample seems to be set up for an old version of XCode and I couldn't get code to generate off of the Wordnik API at all.


fehguy:
I've copied these into the develop 2.0 branch, which has a total reorganization of code.  Thank you for contributing this!

PR #199, opened 2014-06-05
Allow filtering endpoints for which to generate clients.

mkantor:
Adds a callback that generators can implement to filter the APIs for which to generate clients/models.

Also removes some dead code that looks like it was the beginnings of a plan to do the same thing.


fehguy:
sorry this took so long to get back to.  Please rebase if you'd like to get this supported.

PR #299, opened 2014-10-03
Preventing local variable names from colliding with query parameters in the Java template

Ccook:
Full commit message: Updated src/main/resources/Java/api.mustache to use underscores for local variables to prevent query parameter collisions (i.e the query param 'path' would collide with the local variable named path)

Came across a problem when generating Java client code for my swagger definition. One of my routes required the variable name 'path', which collided with a local variable called path. I've added underscores to each local variable to reduce the likelihood of it happening again. 

Its probably a smart idea to do it with all the templates, the only other I could do confidently is the Scala template. Let me know if I should work on that one as well.

### DIFF
--- a/src/main/resources/Java/api.mustache
+++ b/src/main/resources/Java/api.mustache
@@ -15,19 +15,19 @@ import java.util.*;
 
 {{#operations}}
 public class {{classname}} {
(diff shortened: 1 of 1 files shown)


fehguy:
will need to address this in develop_2.0

PR #338, opened 2014-12-05
APIs and Models both have classVarName as well as instanceVarName

swps:
Slight modification from https://github.com/swagger-api/swagger-codegen/pull/323 (I can't seem to re-open that pull request).

I created templates for a Java client built by Gradle, with fault tolerance from Hystrix and templated by Feign. See: https://github.com/swps/swagger-codegen-templates

I need APIs to have `classVarName` and `instanceVarName` in their mustache template context (like models do), and in addition, give models `instanceVarName` to be in alignment with APIs.

Also, allow for processing scala server pages (.ssp) files for more complicated templating that mustache can't provide.

### DIFF
--- a/bin/runscala.sh
+++ b/bin/runscala.sh
@@ -20,6 +20,8 @@ fi
 cd $APP_DIR
 SCALA_RUNNER_VERSION=$(scala ./bin/Version.scala)
 
+echo Using Scala: $SCALA_RUNNER_VERSION
+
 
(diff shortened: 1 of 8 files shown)


webron:
@swps - any chance you can do it? If so, it needs to be done against the develop_2.0 branch.

fehguy:
closing as inactive.

PR #348, opened 2014-12-14
update php swagger to use serialize to show error response 

wing328:
update php swagger to use serialize to show error response as the json response may not contain the 'message' attribute, which will result in exception


fehguy:
can you please merge this change into the develop_2.0 branch?

wing328:
sorry completely missed your reply. will merge the change into develop_2.0 by coming Fri

PR #349, opened 2014-12-14
I needed som additional features like inheritance...

bluebossa63:
- string property with format uri and byte
- object property
- inheritance (see swagger.json)
- additional arguments (model/api package)
- some NPE eliminated

I saw that you're working on implementing inheritance, too, My soulution is rather quick an dirty an I also considered to use ComposedModel. I think, the loops creating the models should be refactored completely.

Cheers

Daniele

### DIFF
--- a/src/main/resources/JavaJaxRS/ApiOriginFilter.mustache
+++ b/src/main/resources/JavaJaxRS/ApiOriginFilter.mustache
@@ -6,7 +6,7 @@ import javax.servlet.*;
 import javax.servlet.http.HttpServletResponse;
 
 public class ApiOriginFilter implements javax.servlet.Filter {
-	@Override
+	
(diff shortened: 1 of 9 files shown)


src/main/resources/JavaJaxRS/ApiOriginFilter.mustache
```diff
@@ -6,7 +6,7 @@ import javax.servlet.*;
 import javax.servlet.http.HttpServletResponse;
 
 public class ApiOriginFilter implements javax.servlet.Filter {
-	@Override
```
lanwen: Why do you remove it? It very useful annotation

fehguy:
Hi, my fault but this PR has drifted.  If you want to rebase against develop_2.0, we can review it quickly.

PR #350, opened 2014-12-17
Java, PHP, Python binding enhancements for better API request control

jebentier:
Adapting the Java, PHP, and Python bindings to account for API Keys that are of a different name than 'api_key' as well as adding cookie support in Java and Python bindings


fehguy:
can you please rebase this against develop_2.0?

PR #362, opened 2015-01-02
Introduced the isRef attribute to ModelProperty...

quenio:
in order to be able to determine whether the property was defined using $ref

That allows a custom generator (for MongoDB for example) to use this information in order to determine whether an attribute should be a reference to another document (if defined with "$ref"), or just an embedded, sub-document (if defined with "type").


fehguy:
This is doable with `instanceof` now.

PR #365, opened 2015-01-07
Parameter names with certain characters can generate invalid Java syntax

Helmsdown:
I have a header param that has a "-" character in it.
```
          name: Request-Id
          in: header
          description: Request id
          required: true
          type: string
```

The JaxRS generator was generating this

```java
 @ApiParam(value = "Request id" , required=true ) @HeaderParam("Request-Id") String Request-Id,
```

Which does not compile because "-" is not a valid character for variable names in Java.

I added a field to CodegenParameter to store a language safe version of the param name. 

I updated the JAXRS mustache templates to use the new CodegenParameter field. I did not touch any other templates because I am not currently using them. This problem will likely require updates to other templates.

I also added a set of invalid characters that each language specific codegen can fill out to specify which characters are not allowed in their variable names. I updated the JavaClientCodgen to consider "-" an invalid character for variable names.


fehguy:
sorry for the delay getting back to this.  Some thoughts.

1) `baseName` is intended to hold the original name of the param, such as `x-forwarded-for`
2) `paramName` is intended to be the "safe" name, for the language

That means, we shouldn't have to add a third name.  But we must fix the escaping logic for the names that are invalid.  I believe we need to override the `toParamName` to make it safe.

So let's look in the code to make sure that path is correct.

Helmsdown:
Ok i resynced with develop_2.0

I was unaware of baseName. I changed the JAX-RS templates to use "baseName" inside the JAX-RS header annotations. I overrode toParamName() in JavaClientCodegen to replace "-" characters with an "_" the same way that toVarName() is doing it.

fehguy:
This should be fixed now, please try develop_2.0

PR #366, opened 2015-01-07
JaxRS @ApiResponse annotations for 200 response codes are not generated 

Helmsdown:
In my swagger spec I have the following case:

```
      responses:
        200:
          description: Success - Duplicate transaction detected and previous charge returned
          schema:
            $ref: '#/definitions/Charge'
```

When I generate JaxRS resources the @ApiResponse annotation for the 200 response code is not generated. 

My solution was to add a new field, successResponse, to CodegenOperation which contains the success response object (which is normally culled during DefaultCodegen.fromOperation()).

I then use successResponse in the JaxRS template.


fehguy:
I need to look into this more before merging.

fehguy:
I apologize for the delay responding to this.  I do believe that we need to support the `succes` response type as well as all other codes.  Since this PR is old and the code has drifted (my fault), I will close it and reference a new ticket.

PR #368, opened 2015-01-09
Adding first and hasRequiredParams parameters to Codegen

amandaducrou:
I found that I needed two extra variables in the generator/mustache templates. I thought other people might find these useful as well.

"first" - the first query param, this could also be aded for the other types of params
"hasRequiredParams" - a flag to indicate the presence of required params


fehguy:
One thing I'll point out is, you can do this in the develop_2.0 branch without code changes--you can supply your own models into the model factory so special properties can be accessed.

amandaducrou:
Oh cool, I will check that out.

PR #374, opened 2015-01-13
Parameter upgrades

Helmsdown:
Make codegen resolve RefParameter objects. Make codegen aware of
parameters defined at the Path object level.


fehguy:
This logic will go here:

https://github.com/swagger-api/swagger-parser/issues/24

PR #378, opened 2015-01-20
PythonGenerator api args

lucasnad27:
This patch addresses the order of arguments with default values within a function and what `defaultValue` is set to in the generated code. If we have a `Foo` function that takes three arguments: `bar, quz, and baz` but `baz` has a defaultValue of "Something". Currently, the output looks like this:

```
def Foo(self, bar, baz=None, quz, **kwargs):
    ...
```

Any values with default values should go at the end of the arguments list and the default should be "Something" rather than null. This patch explicitly orders `requiredParams` by defaultValue and puts that value in the mustache template. Below is the resulting output

```
def Foo(self, bar, quz, baz="Something", **kwargs):
    ...
```

### DIFF
--- a/src/main/resources/python/api.mustache
+++ b/src/main/resources/python/api.mustache
@@ -31,7 +31,7 @@ class {{classname}}(object):
 
     {{newline}}
     {{#operation}}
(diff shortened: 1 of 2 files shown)


fehguy:
Hi, we have switched from scala to java in master, and I *believe* this logic should already exist.  Please try out the latest dev branch, which is develop_2.0 to see.

PR #382, opened 2015-01-20
Java Model Template Fixes

Helmsdown:
The current Java/model.mustache generates an empty enum class when your specify a numeric property because the template was using the "min" property to determine whether to generate an enum. I switched the template to use "isEnum" instead. Now only enum fields get an enum generated.

I also reformatted the template to take advantage of the the [newline trimming](https://github.com/samskivert/jmustache#newline-trimming) feature in jmustache. While this does increase the length of the template, i think it makes it more readable.

### DIFF
--- a/src/main/resources/Java/model.mustache
+++ b/src/main/resources/Java/model.mustache
@@ -6,42 +6,54 @@ package {{package}};
 import com.wordnik.swagger.annotations.*;
 {{#models}}
 
-{{#model}}{{#description}}
+    {{#model}}
+{{#description}}
 /**
- * {{description}}
- **/{{/description}}
+* {{description}}
+**/
+{{/description}}
 @ApiModel(description = "{{{description}}}")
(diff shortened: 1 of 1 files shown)


fehguy:
Thank you.  If possible, would be great to get a test put in place.  I understand it's a little tricky to do so, but take a look here:

https://github.com/swagger-api/swagger-codegen/tree/develop_2.0/src/test/scala/Java

To see what we can add so there are no regressions

Helmsdown:
So I looked at the tests written in Scala. They seem to be testing that the codegen code correctly parses a swagger model. Since I didn't change how the code parses the model does this still apply? Do you have tests that confirm that template actually created the right code?

fehguy:
this should be obsolete now.  I've added a property called `datatypeWithEnum` which will contain the model property type if it's an enum or not.  I had to do this to support the other languages which don't support enums.

PR #395, opened 2015-01-26
Fix for java client codegen for array-type responses

fleque:
Hi,
this fixes generated Java client code for array type responses. Before the fix the resulting client would have the following as the method result statement
```java 
(List<ResultClassname>) ApiInvoker.deserialize(response, "ResultClassname", ResultClassname.class);
```
Which led to an error when actually using this method (Innstance of RsultClassname was created from JSON array but could not be casted to List<ResultClassname>).
After the fix this is the result statement:
```java 
(List<ResultClassname>) ApiInvoker.deserialize(response, "List", ResultClassname.class);
```
However, I'm not sure whether this change is done at the correct location or whether it should be somewhere more basic. I've adapted the way something similar was done for `JaxRSServerCodegen`.

Best regards
Fleque

### DIFF
--- a/src/main/java/com/wordnik/swagger/codegen/languages/JavaClientCodegen.java
+++ b/src/main/java/com/wordnik/swagger/codegen/languages/JavaClientCodegen.java
@@ -1,10 +1,10 @@
 package com.wordnik.swagger.codegen.languages;
 
+import java.io.*;
+import java.util.*;
+
 import com.wordnik.swagger.codegen.*;
 import com.wordnik.swagger.models.properties.*;
-
-import java.util.*;
-import java.io.File;
 
 public class JavaClientCodegen extends DefaultCodegen implements CodegenConfig {
   protected String invokerPackage = "com.wordnik.client";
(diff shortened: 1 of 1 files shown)


fehguy:
Yes, your 2nd code snippet is the right behavior.  I'm not sure how this would happen, though, as the returnContainer should be a list.

Can you share the JSON that you're using that's creating this scenario?

fleque:
Here is an excerpt of the JSON producing the above behaviour:
```
{
  "swagger": "2.0",
  "info": {
    "version": "2015-01-20",
    "title": "test",
    "description": "Test",
    "license": {
      "name": "None"
    }
  },
  "host": "localhost",
  "basePath": "/testapi",
  "schemes": [
    "http",
    "https"
  ],
  "consumes": [
    "application/json"
  ],
  "produces": [
    "application/json"
  ],
  "paths": {
    "/authors": {
      "get": {
        "summary": "get Authors",
        "operationId": "getAuthors",
        "description": "get all authors",
        "parameters": [
          {
            "name": "filter",
            "in": "query",
            "required": false,
            "description": "some string filter",
            "type": "string"
          }
        ],
        "responses": {
          "200": {
            "description": "return list of authors",
            "schema": {
              "type": "array",
              "items": {
                "$ref": "#/definitions/Author"
              }
            }
          }
        }
      }
    }
  },
  "definitions": {
    "Author": {
      "required": [
        "authorId",
        "firstName",
        "lastName"
      ],
      "properties": {
        "authorId": {
          "type": "string"
        },
        "firstName": {
          "type": "string"
        },
        "lastName": {
          "type": "string"
        },
        "email": {
          "type": "string"
        }
      }
    }
  }
}
```

fehguy:
Hi, this has been tested in develop_2.0

PR #396, opened 2015-01-26
[WIP]  Enhancements for the Java client codegen

fleque:
Hi,
this is a pull request for work in progress enhancements I want to do to the Java client codegen. For now I have at least the need for a API-way of adding authentication for the client (basic auth and an auth-ticket as url parameter).

As a start, I've added comments to the generated operation-methods ;-)

Best regards
Fleque

### DIFF
--- a/src/main/resources/Java/api.mustache
+++ b/src/main/resources/Java/api.mustache
@@ -39,7 +39,14 @@ public class {{classname}} {
   {{#errorList}} //error info- code: {{code}} reason: "{{reason}}" model: {{#responseModel}}{{responseModel}}
   {{/responseModel}}{{^responseModel}}<none>
   {{/responseModel}}
-  {{/errorList}}  
+  {{/errorList}}
+  /**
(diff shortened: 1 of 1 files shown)


fleque:
By the way. I haven't figured out yet how to add a test for the code generation output the swagger-codegen way. Is there a framework for this in the project? (I don't know scala so I might have overlooked that in the scala tests)

fehguy:
Yes, the general plan is to test the values that are put in the model which is used by the templates.  So from here:

https://github.com/swagger-api/swagger-codegen/blob/develop_2.0/src/test/scala/Java/JavaModelTest.scala

We have tests to look at the processing of models.  This doesn't handle the operations, though, and I'm guessing that's what you're after.

Per my last comment, can you share the JSON that you're using that's causing this need?  I want to see if some existing logic isn't broken, in which case I'd rather fix it than work around it.

fehguy:
I think this PR has become obsolete.

PR #397, opened 2015-01-26
[WIP] Support Python code generation

fleque:
Hi,
I need to create a python client for my current project and therefore added this implementation to based on develop_2.0 branch. The implementation is a port of what I understood of the scala generator class (from master) and produces a working Python client (as far as my tests for my API can tell that by now).

Please check the implementation and see whether there is something missing/wrong.

I would like to add tests for the code generation. It would be great if someone could tell me how this should be done in swagger-codegen.

Best regards
Fleque


fehguy:
Thank you, merging.  Please try to add a test like such:

https://github.com/swagger-api/swagger-codegen/blob/develop_2.0/src/test/scala/Java/JavaModelTest.scala

To handle arrays, maps, etc.

Issue #406, opened 2015-01-31
Need upgrade scala version to 2.11 


netroby:
[netroby@localhost swagger-codegen]$ ./bin/validate.sh 
Please set scalaVersion := "2.11" in build.sbt and run ./sbt assembly

webron:
We're concentrating our work on the develop_2.0 branch which is just like the master, only much much better. It's also based on Java instead of scala and will help us rid of all the version compatibility support involving scala.

I'd suggest using using the develop_2.0 branch now and avoid all this hassle.

PR #407, opened 2015-01-31
Update scala to 2.11.5, See #406

netroby:
Update scala to 2.11.5, See #406


fehguy:
now that the codegen is using pure java, this isn't necessary.

PR #410, opened 2015-02-03
add .gitattributes to prevent .sh checked out as DOS line ending

bbbush:
1. requires mass line ending updates in the repo (included in this commit)
2. requires remove all working folder files and then do a reset --hard


fehguy:
Hi, what is this PR trying to do?

bbbush:
It is simply white-space changes. I tried to run it under cygwin, which requires LF line ending for shell scripts.

fehguy:
Any PRs need to be done against develop_2.0

PR #412, opened 2015-02-04
Python/PHP Multipart-Form request support

jebentier:
Adding support for submitting a multipart/form-data request.  Including full support for file uploads by passing in the filepath as the argument to the generated method.


fehguy:
Thank you.  Can you send a PR into the develop_2.0 branch for this?  That's where the majority of the codegen work is happening.

jebentier:
develop_2.0 PR: https://github.com/swagger-api/swagger-codegen/pull/416

PR #415, opened 2015-02-06
Handle reference parameters by de-referencing RefParameters

ryanw999:
I want to use the common parameters definitions in my schema, so I added some code to de-reference those parameters into regular parameters when generating code. Without this change, the code throws a ClassCastException when trying to use the common parameters.


Helmsdown:
Looks like this is a duplicate of #374.

fehguy:
Please track here: https://github.com/swagger-api/swagger-parser/issues/24

PR #417, opened 2015-02-07
Add API Version to Codegen Additional Properties

Helmsdown:
I want to be able to display the API version my static html documentation.

### DIFF
--- a/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultGenerator.java
+++ b/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultGenerator.java
@@ -54,6 +54,9 @@ public class DefaultGenerator implements Generator {
           if(license.getUrl() != null)
             config.additionalProperties().put("licenseUrl", license.getUrl());
         }
+        if(info.getVersion() != null) {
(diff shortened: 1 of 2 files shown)


fehguy:
@Helmsdown did you see the properties put into the templates automatically?  Namely:

https://github.com/swagger-api/swagger-codegen/blob/develop_2.0/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/languages/JavaClientCodegen.java#L44-L47

I'm looking for a better way to pass arbitrary options into the generator without forcing a recompilation.  One option is to pass KV map and they make it all the way into the mustache graph.  What are your thoughts, from the integration point of view?  Will you be building a special java class to configure the templates or would you be happier with a config file / arguments?

Are you using a CI server to build your clients?

Helmsdown:
@fehguy I don't think we are on the same page. I'm talking about the REST API version not the java artifact version. If you look at my pull request, you'll see I'm pulling a value out of the swagger model, and putting it into the additional properties.

kc284:
+1 for this commit, with one remark: you could rename "version" to "apiVersion" to retain the name that was originally used in Swagger 1.2.

PR #418, opened 2015-02-09
Support for HTTP Async client transport for Jersey on Scala (legacy master)

lucamilanesio:
When generating Scala client, the ApiInvoker is a class with
overridable options, avoiding the need of public vars to be
manually modified.

Additionally it is possible to plug the AsyncHttp protocol
for Jersey in order to achieve Netty support and maximum scalability.


fehguy:
this is great--we're mostly working in the develop_2.0 branch and that would be the appropriate place to merge this.  Can you send the PR there?

lucamilanesio:
@fehguy thanks for the feedback: I would keep this PR anyway for a fix on the release we are currently using and additionally I'll open a new one on develop_2.0

PR #426, opened 2015-02-10
Scala api.mustache: allow easy basePath and ApiInvoker override

lucamilanesio:
Gets rid of the smelly vars inside Scala api.mustache allowing the
override of default basePath and ApiInvoker on the constructor.

Previously the code for overriding was:

  val myApi = new MyApi()
  myApi.basePath = "http://myapi.endpoint.com"
  myApi.apiInvoker = new ApiInvoker(authScheme = "SPNEGO")

after the change everything can be simply declared in the constructor:

  val myApi = new MyApi("http://myapi.endpoint.com",
                         new ApiInvoker(authScheme = "SPNEGO"))


fehguy:
Oh scala

Issue #443, opened 2015-02-20
Required parameters not handled as required


cchafer:
Hi, 
At this time, DefaultCodegen doesn't handle rightly required parameters. All of them as marked as NOT required when trying to figure out their type.

This can lead to several problems when, like me, you want to define type type of not required items as Option[...] in Scala.

Here's a simple fix for this : 
```
diff --git a/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultCodegen.java b/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultCodegen.java
index 9d908aa..965106e 100644
--- a/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultCodegen.java
@@ -774,6 +774,8 @@ public class DefaultCodegen {
         LOGGER.warn("warning!  Property type \"" + qp.getType() + "\" not found for parameter \"" + param.getName() + "\", using String");
         property = new StringProperty().description("//TODO automatically added by swagger-codegen.  Type was " + qp.getType() + " but not supported");
       }
+
+      property.setRequired(param.getRequired());
       CodegenProperty model = fromProperty(qp.getName(), property);
       p.collectionFormat = collectionFormat;
       p.dataType = model.datatype;
```

fehguy:
Can you put in a test or two to demonstrate this in a PR to develop_2.0 branch?

webron:
@cchafer - a bit confused here. You prepared #446 but closed it, and can't find another submission. Am I missing something or did you perhaps forget? :)

cchafer:
Hi,

To be honnest, this little modification has already been merged a few weeks ago, in another PR, about CodegenParameter missing some data (can't remember exactly which one).
So this PR could be closed (I thought it already was, in fact).
Do you want me to close it ?

webron:
I'm just going over the issues and cleaning out to get a better picture of what's needed for the final release. I don't keep track over all changes in the project, but if you say this issue can be closed, that's great.

SunriseJJJ:
facing this issue right now
part of my contract (definitions section):
```
"sampleRequest": {
    "type": "object",
    "required": [
        "a",
        "b"
    ],
    "properties": {
        "a": {
            "type": "string",
            "description": "A field"
        },
        "b": {
            "type": "string",
[...]

PR #446, opened 2015-02-23
#443 - Handling required parameters during swagger type discovery

cchafer:
The problem fixed by this patch occurs when one tries to do some computation on the "requireness" of a parameter, in the DefaultCodegen.getSwaggerType.
Indeed, this method takes a Property as single parameter, and the 'required' field of this property is always set to false.
It should be true when the parameter is required (cptn obvious), and during computation of the operation.

Important note : overriding getSwaggerType (as done in the test) is definitely not the best way to handle required parameters. If one wants to add Option[...] or Optional<...> to any type, the easiest and best way is to modify the templete using {{required}}.
But this doesn't change the fact the property passed to the method is 'invalid'.


cchafer:
Sorry, did some crap. Closing this pull request, making  new one clearer.

Issue #445, opened 2015-02-23
properties with 2 chars upper case have getter and setter badly generated


herveDarritchon:
When a properties has the second character in uppercase, such as pId for example, the getter and setter are not well generated.

The getter and setter are generated like this getPId and setPId and it is not as it should be.
In fact, it should be getpId and setpId.

DefaultCodegen.java (line 441/442)
        property.getter = "get" + initialCaps(name);
        property.setter = "set" + initialCaps(name);

initialCaps is : 
public String initialCaps(String name) {
        return Character.toUpperCase(name.charAt(0)) + name.substring(1);
    }

in the case of getter and setter, we should make an evolution to the code to be compliant with the getter and setter rules.

You should use  Introspector.decapitalize from package java.beans instead of a home made function.

        property.getter = "get" +  Introspector.decapitalize(name);
        property.setter = "set" +  Introspector.decapitalize(name);

webron:
@herveDarritchon - #447, #449 and #452 resolve together resolve this issue?

herveDarritchon:
yes. it's ok. it is resolved.

webron:
Thank you!

PR #447, opened 2015-02-23
Defect (Issue #445) : Use of Introspector.decapitalize

herveDarritchon:
change the call of initialCaps for getter and setter with Introspector.decapitalize from pakage java.beans.


fehguy:
looks like this was sent in a new PR, so closing this one.

PR #449, opened 2015-02-23
Defect (Issue #445) : Call Introspector.decapitalize instead of home made function

herveDarritchon:
To correct a bug during creation of getter and setter, I change the call of initialCaps for getter and setter with Introspector.decapitalize from package java.beans.


fehguy:
Hi thanks for submitting.  Two things that would help a lot
1) Any PR should go against the development branch, not master, since master is "clean and released".  The development branch is `develop_2.0`
2) If you can generally avoid reformatting code so the changes can be more easily detected, it can be merged much more easily.

I'm closing this out so you can send a PR against the dev branch.

herveDarritchon:
sorry for the mess :)
I do my change against the develop_2.0 branch.

Is it possible to think of reformatting all the code one day, because it's easy to read ;)

thanks

PR #452, opened 2015-02-24
Defect (Issue #445) : change the call of initialCaps for getter and setter proper generation

herveDarritchon:
Getter and Setter with 2nd chars being uppercase are not generated following the Sun specification : http://download.oracle.com/otn-pub/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/beans.101.pdf?AuthParam=1424762391_ef17a82335203d9fbf095a2f819eb317
Chapter 8.8.

When you generate a getter and setter, you should get : 
fooBah --> getFooBah
mAttr --> getmAttr

That was not the cas, we got previously, which is wrong.
fooBah --> getFooBah
mAttr --> getMAttr

I have made a new function called getterAndSetterCapitalize which takes a String and return the proper capitalized name for getter and setter.


herveDarritchon:
I have a question.

I forgot to add some unit tests.
Do I have to do a new PR only with the tests or do I have do a new PR with the evolution and the tests ?

webron:
You should be able to modify this PR with the added tests.

PR #456, opened 2015-02-25
Fixed bug that prevented Objective-C code from compiling when using a Sw...

geoand:
...agger 'map' type.

In the current code, when one uses `map` as a type in the Swagger model, the Objective-C code produced uses `initWithValues` which is not corrent since `NSDictionary` does not inherit from `SWGObject`.
The correct behavior in this case seems to be to use `initWithDictionary`.

During the creation of this fix I tried to limit the impact of the code changes as much as possible.
Therefor I have not added any new values to `CodegenProperty` (instead opted for using `setter` which is unused in the case of Objective-C), nor have I altered the code flow of the other clients in any way. 
All that I did is add a new method to `DefaultCodegen` that is only called when producing Objective-C code and only when a Swagger `map` or custom user type is used.

### DIFF
--- a/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultCodegen.java
@@ -528,16 +528,20 @@ public class DefaultCodegen {
         property.isPrimitiveType = true;
     }
     else {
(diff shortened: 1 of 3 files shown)


webron:
Can you resubmit the PR against develop_2.0?

geoand:
Yes of course!
I'll do that first thing in tomorrow!

PR #460, opened 2015-02-27
Allow SPNEGO security to be enforced as system level

lucamilanesio:
By introducing the Java system property -DapiInvokerAuthScheme=SPNEGO
    it is now possible to make sure that security is enforced regardless
    of the client settings provided by the client.
    
    This is sometimes needed as the Security Officer requires policies to
    be enforced regardless of what the client code is actually going to
    provide as parameters.


fehguy:
Hi, I currently don't have intentions of another release of swagger-codegen 2.0.x.  Is this needed or can you move entirely to 2.1.x?

fehguy:
closing as we're not intending on another release of 2.0.x

PR #462, opened 2015-03-03
Feature/threepane

ponelat:
Here I have the templates from the development rig for the Threepane generator.
The only snag is that the PNG logo is being modified through supportingFiles somehow, which is probably due to my lack knowledge. The result being a corrupt PNG file after generating.. :/


fehguy:
@ponelat i fixed the logo issue, it was copying the file as a string, not a blob.

PR #466, opened 2015-03-04
Add support for maps in parameters.

cor3ntin:
Maps where not supported in operations parameters. They now are


fehguy:
Hi, the intent of `SerializableParameter` is to avoid putting unsupported types inside query, header, path, form, or cookie params.  So per the swagger spec, you cannot put maps in these fields, so I don't believe this code makes sense.  Correct me if I'm reading it wrong.

cor3ntin:
Hi. 
I guess you are right, but still
There is technically nothing preventing the serialization of a map as a body parameter.

One other issue I notices with CodeParam not being a full blown CodegenProperty is that you can't implement any kind of value checking.
For example, I looked into supporting sting-based enums in parameters but that is currently not douable. 

What is your take on that ?

fehguy:
we will get the two in sync and merged.  But I am closing this out because we should not support out-of-spec implementations.

PR #477, opened 2015-03-09
use baseName for parameter name in PHP api

wing328:
use baseName for parameters (query, header, form) in PHP api method/function (other clients like Java also use baseName instead of paramName for query, header and form parameters in api method/function )

### DIFF
--- a/modules/swagger-codegen/src/main/resources/php/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/php/api.mustache
@@ -50,20 +50,20 @@ class {{classname}} {
 
       {{#queryParams}}// query params
       if(${{paramName}} !== null) {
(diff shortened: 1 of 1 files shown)


fehguy:
this may cause things to break, since `baseName` may contain illegal characters.  `paramName` will have been _escaped_ during the generator.  Have you looked into that logic?  Run codegen with -d to see options

wing328:
I'm aware of the fact that paramName is escaped during the generator. baseName is used as a hash key e.g. `$queryParams['{{baseName}}']` (inside a single quote) and a test with the updatePetWithForm using "name+testing" as a parameter confirms the encoding is done by PHP cURL automatically:

`name%2Btesting=sampleName&status=sampleStatus`

fehguy:
got it! thanks @wing328

PR #480, opened 2015-03-09
Fix for python3 client deserialize passing wrong parameter

theFigler:
The parent objClass was being passed recursively to deserialize when it needed to be the attrType, which is the correct swagger datatype to deserialize.


fehguy:
Can you please revise to develop_2.0?

theFigler:
Created new pull request on the 2.0 branch

PR #485, opened 2015-03-11
minor bug fixes and enhancement to Swagger.php

wing328:
add logic to handle all 2xx response code
bug fix for string response
bug fix for showing error message using serialize
update php petstore sample

### DIFF
--- a/modules/swagger-codegen/src/main/resources/php/Swagger.mustache
+++ b/modules/swagger-codegen/src/main/resources/php/Swagger.mustache
@@ -129,11 +129,14 @@ class APIClient {
     if ($response_info['http_code'] == 0) {
       throw new APIClientException("TIMEOUT: api call to " . $url .
         " took more than 5s to return", 0, $response_info, $response);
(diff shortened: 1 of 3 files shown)


wing328:
Here is more information about the bug: json_decode($response)->message will result in exception since not *all* error response is in JSON format and has a property named "message" (it could be error_message, err_message, error, etc)

fehguy:
we know the response model and the error model, so we should construct the proper types, correct?

wing328:
Ideally yes. I'd a look at other API clients (e.g. Java) and it simply returns the entity (body) at the moment

```
    else {
      throw new ApiException(
                response.getClientResponseStatus().getStatusCode(),
                response.getEntity(String.class));
    }
```
One of the goals of this PR is to avoid `json_decode($response)->message` throwing exception. Later we can further enhance it with the error model defined by the user.

fehguy:
yes, that makes sense.  The PR isn't mergable, can you see about addressing that?

wing328:
pushed the change to resolve the conflicts

PR #486, opened 2015-03-11
Bug fixes for Swagger.php deserialize function

wing328:
1) syntax error `PHP Parse error:  syntax error, unexpected '[' in Swagger.php on line 223` due to the following:
`$subClass = explode(',', $inner, 2)[1];`

not sure if it's due to me not using the latest version of php, I workaround the issue by moving the [1] to the next line with a temporary variable so that php5.3 can compile it without syntax error

2) updated array comparison to case insensitive so that it works for both array of primitive type and array of model

3) updated PHP sample

### DIFF
--- a/modules/swagger-codegen/src/main/resources/php/Swagger.mustache
+++ b/modules/swagger-codegen/src/main/resources/php/Swagger.mustache
@@ -223,13 +223,14 @@ class APIClient {
       $inner = substr($class, 4, -1);
       $values = array();
       if(strrpos($inner, ",") !== false) {
-        $subClass = explode(',', $inner, 2)[1];
+        $subClass_array = explode(',', $inner, 2);
+        $subClass = $subClass_array[1];
         foreach ($data as $key => $value) {
           $values[] = array($key => self::deserialize($value, $subClass));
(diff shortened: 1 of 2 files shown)


wing328:
Confirmed the syntax error in 1 occurs with php v5.3 but not v5.5

`➜  php-client  php -l Swagger.php 
PHP Parse error:  syntax error, unexpected '[' in Swagger.php on line 223
Errors parsing Swagger.php
➜  php-client  php -v
PHP 5.3.10-1ubuntu3.14 with Suhosin-Patch (cli) (built: Sep  4 2014 07:08:49) 
Copyright (c) 1997-2012 The PHP Group
Zend Engine v2.3.0, Copyright (c) 1998-2012 Zend Technologies`

`php-client|⇒ php -l Swagger.php 
No syntax errors detected in Swagger.php
php-client|⇒ php -v
PHP 5.5.14 (cli) (built: Sep  9 2014 19:09:25) 
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2014 Zend Technologies`

Issue #495, opened 2015-03-13
"resourcePath" : "/" causes NullPointerException


burmanm:
If the generated json has:

  "resourcePath" : "/",

the code generation fails with the following error:

java.lang.StringIndexOutOfBoundsException: String index out of range: 0
    at java.lang.String.charAt(String.java:658)
    at com.wordnik.swagger.codegen.DefaultCodegen.snakeCase(DefaultCodegen.java:315)
    at com.wordnik.swagger.codegen.DefaultCodegen.toApiVarName(DefaultCodegen.java:138)
    at com.wordnik.swagger.codegen.DefaultGenerator.generate(DefaultGenerator.java:117)
    at com.wordnik.swagger.codegen.Codegen.main(Codegen.java:99)

This happens because the tag parser removes "/" and only uses "" while the snakeCase calls for charAt(0) (return Character.toLowerCase(name.charAt(0)) + name.substring(1);) which isn't there.

fehguy:
this is fixed in develop_2.0

PR #496, opened 2015-03-13
Fix NPE with resourcePath = /, issue #495

burmanm:
Assuming there are no side-effects I'm not aware off, this should fix the issue with NPE.


webron:
@burmanm - PRs should be created against the develop_2.0 branch.

burmanm:
PR 498 is against the develop_2.0 branch

PR #502, opened 2015-03-13
PHP Generated Models

jebentier:
Adding in the ability to access and generate models as/from arrays.  This makes it easier to generate response and requests models for testing binding usage as well as creating request models in subsequent code.


fehguy:
Can you please describe more as to how this change is general enough for the all models?

jebentier:
This code change better enables an end user to write tested code by making it easier to construct response and request models with full and partial mocks of API responses.

PR #505, opened 2015-03-15
update User-agent in PHP API client

wing328:
Allow developers to customize the user agent in PHP API client. If not set, default to "PHP-Swagger".


wing328:
The current behaviour for PHP API client is that no User-Agent is included in the HTTP request

PR #511, opened 2015-03-17
Java: properly add request headers

sosso:
We were running into issues with our headers not making it into our requests.  Upon some research, it appears that Jersey's header() method returns the builder object, meaning we have to do this reassignment to properly construct the request.  After making the change, our requests work as expected.

https://jersey.java.net/nonav/apidocs/1.8/jersey/com/sun/jersey/api/client/PartialRequestBuilder.html#header(java.lang.String, java.lang.Object)


fehguy:
can you please send this PR into develop_2.0?

PR #482, opened 2015-03-10
Python: fixing an issue in swagger python code generation

jebentier:
Attribute classes weren't being respected and now they are.

PR #512, opened 2015-03-17
Java: properly add request headers

sosso:
We were running into issues with our headers not making it into our requests. Upon some research, it appears that Jersey's header() method returns the builder object, meaning we have to do this reassignment to properly construct the request. After making the change, our requests work as expected.

https://jersey.java.net/nonav/apidocs/1.8/jersey/com/sun/jersey/api/client/PartialRequestBuilder.html#header(java.lang.String, java.lang.Object)

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/apiInvoker.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/apiInvoker.mustache
@@ -106,12 +106,12 @@ public class ApiInvoker {
 
     Builder builder = client.resource(host + path + querystring).accept("application/json");
     for(String key : headerParams.keySet()) {
-      builder.header(key, headerParams.get(key));
+      builder = builder.header(key, headerParams.get(key));
     }
 
     for(String key : defaultHeaderMap.keySet()) {
       if(!headerParams.containsKey(key)) {
-        builder.header(key, defaultHeaderMap.get(key));
+        builder = builder.header(key, defaultHeaderMap.get(key));
       }
     }
     ClientResponse response = null;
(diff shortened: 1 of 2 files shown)


modules/swagger-codegen/src/main/resources/Java/apiInvoker.mustache
```diff
@@ -106,12 +106,12 @@ public class ApiInvoker {
 
     Builder builder = client.resource(host + path + querystring).accept("application/json");
     for(String key : headerParams.keySet()) {
-      builder.header(key, headerParams.get(key));
+      builder = builder.header(key, headerParams.get(key));
```
xhh: I believe this is not needed, as `builder.header` sets the header value directly to the `builder` itself, right?

Please refer to https://jersey.java.net/nonav/apidocs/1.8/jersey/com/sun/jersey/api/client/PartialRequestBuilder.html#header(java.lang.String, java.lang.Object)

sosso: I do agree that it appears as though it should not be needed, but it was definitely needed in our case.  With no other changes than this reassignment, our code started working.

xhh: I see. However, I found the `header` method from the source code of jersey-client v1.7 downloaded [here](http://download.java.net/maven/2/com/sun/jersey/jersey-client/1.7/):

```java
    public T header(String name, Object value) {
        getMetadata().add(name, value);
        return (T)this;
    }
```

As you can see the method just returns itself, so theoretically speaking, assigning the result again would make no difference, right?

fehguy:
I don't see how this piece is needed--can you show it with a unit test or something?

It looks harmless but still, would like to see if it's necessary.

fehguy:
addressed in https://github.com/swagger-api/swagger-codegen/pull/482

PR #516, opened 2015-03-18
Bug fix for ruby header parameter and better code format

wing328:
- Bug fix for Ruby header parameter: the current code will only set the *last* header parameter if there're more than one header parameter

```
    {{#headerParams}}headers = {
      {{{paramName}}}: {{{paramName}}},
    }
    {{/headerParams}}
    {{^headerParams}}headers = nil
    {{/headerParams}}
```

- remove extra line break in api.mustache and model.mustache so that the code format looks better

- put single quote in parameter name to avoid special character from causing failure in compiling the auto-generated Ruby code

### DIFF
--- a/modules/swagger-codegen/src/main/resources/ruby/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/ruby/api.mustache
@@ -23,10 +23,8 @@ class {{classname}}
     
     # set default values and merge with input
     options = {
-    {{#allParams}}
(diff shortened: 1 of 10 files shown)


wing328:
To elaborate on the issue a bit, if there are 3 header parameters (var1, var2, var3), here is the current auto-generated Ruby code:

headers = {
      var1: var1,
}
headers = {
      var2: var2,
}
headers = {
      var3: var3,
}

"headers" (hash) would only contain var3 after going through all the header parameters

PR #517, opened 2015-03-18
Add parameterToString to normalize parameters for Java client

xhh:
* Convert values to string for query/header/form parameters.
* For parameter of type Date, convert to ISO 8601 format.
* Also add utility methods to parse date with ISO 8601 format.

### DIFF
--- a/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultCodegen.java
@@ -621,6 +621,8 @@ public class DefaultCodegen {
         count += 1;
         if (count < operation.getConsumes().size())
           mediaType.put("hasMore", "true");
+        else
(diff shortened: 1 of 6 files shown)


modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultCodegen.java
```diff
@@ -621,6 +621,8 @@ public CodegenOperation fromOperation(String path, String httpMethod, Operation
         count += 1;
         if (count < operation.getConsumes().size())
           mediaType.put("hasMore", "true");
+        else
+          mediaType.put("hasMore", null);
```
xhh: this change is to remove redundant trailing comma, e.g. change the following generated code

```java
    String[] contentTypes = {
      "application/json","application/xml",
    };
```

to 

```java
    String[] contentTypes = {
      "application/json","application/xml"
    };
```

PR #523, opened 2015-03-19
Add JSONModel to Obj-C client

geekerzp:
This PR will replace integrate JSONModel(https://github.com/icanzilb/JSONModel) into Obj-C client.

- JSONModel can create model from dictionary and easily export properties to dictionary.
- JSONModel supports conversion between date/datetime string and NSDate (so we no longer need SWGDate)
- JSONModel is similar to the use of @jsonproperty in Java client.


modules/swagger-codegen/src/main/resources/objc/model-body.mustache
```diff
@@ -1,83 +1,16 @@
...
-    {{/vars}}
-{{newline}}{
-    {{#vars}}_{{name}} = {{name}};
-    {{/vars}}
-
-    return self;
-}
-
--(id) initWithValues:(NSDictionary*)dict
+{{#vars}}
```
fehguy: I don't know how this can work.  Declaring keyMapper in the `{{#vars}}` loop will cause duplicate declarations, like this from the petstore example:

```objc
#import "SWGCategory.h"

@implementation SWGCategory
    
+ (JSONKeyMapper *)keyMapper
{
    return [[JSONKeyMapper alloc] initWithDictionary:@{@"id": @"_id"}];    
}
    
    // compilation error
+ (JSONKeyMapper *)keyMapper
{
    return [[JSONKeyMapper alloc] initWithDictionary:@{@"name": @"name"}];    
}
    
@end

```

modules/swagger-codegen/src/main/resources/objc/model-body.mustache
```diff
@@ -1,81 +1,12 @@
...
-            if(_{{name}} != nil) dict[@"{{baseName}}"] = {{#complexType}}[(SWGObject*){{/complexType}}_{{name}} {{#complexType}}asDictionary]{{/complexType}};
-        {{#complexType}}
-        }
-    }
-    {{/complexType}}
-    {{/vars}}
-
-    NSDictionary* output = [dict copy];
-    return output;
+    return [[JSONKeyMapper alloc] initWithDictionary:@{ {{#vars}}@"{{baseName}}": @"{{name}}"{{#hasMore}}, {{/hasMore}}{{/vars}} }];
```
geekerzp: @fehguy I fixed it, now the generated code looks like:
```objective-c
#import "SWGPet.h"

@implementation SWGPet
    
+ (JSONKeyMapper *)keyMapper
{
    return [[JSONKeyMapper alloc] initWithDictionary:@{ @"id": @"_id", @"category": @"category", @"name": @"name", @"photoUrls": @"photoUrls", @"tags": @"tags", @"status": @"status" }];    
}
    
@end
```
I tested new pet object from NSDictionary using method `initWithDictionary`:
```objective-c
    NSDictionary *petDict = @{ @"id": @123, @"name": @"test_pet",
                               @"category": @{ @"id": @123, @"name": @"test_category" },
                               @"photoUrls": @[ @"sample url" ],
                               @"tags": @[ @{ @"id": @123, @"name": @"black" }, @{ @"id": @456, @"name": @"white" } ],
                               @"status": @"pending" };
    SWGPet *pet = [[SWGPet alloc] initWithDictionary:petDict error:nil];
```
and export pet object to NSDictionary using method `toDictionary`:
```objective-c
NSDictionary *petDict =  [pet toDictionary];
```
It works fine.
I'm very sorry i didn't notice this issue.

modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/languages/ObjcClientCodegen.java
```diff
@@ -66,9 +68,8 @@ public ObjcClientCodegen() {
 
     typeMapping = new HashMap<String, String>();
     typeMapping.put("enum", "NSString");
-    typeMapping.put("Date", "SWGDate");
-    typeMapping.put("DateTime", "SWGDate");
-    // typeMapping.put("Date", "SWGDate");
+    typeMapping.put("Date", "NSDate");
+    typeMapping.put("DateTime", "NSDate");
```
geekerzp: we use `NSDate` instead of `SWGDate`, as JSONModel can covert `NSDate` to `NSString` and build
`NSDate` from `NSString` automatically.

modules/swagger-codegen/src/test/scala/Objc/ObjcModelTest.scala
```diff
@@ -173,7 +170,7 @@ class ObjcModelTest extends FlatSpec with Matchers {
     val vars = cm.vars
     vars.get(0).baseName should be ("children")
     vars.get(0).complexType should be ("SWGChildren")
-    vars.get(0).datatype should be ("NSArray*")
+    vars.get(0).datatype should be ("NSArray<SWGChildren>*")
```
geekerzp: If the item of array is a model, the declaration of model property data type should be `NSArray<SWGChildren>*`.
JSONModel use the protocal `SWGChildren` to determine the inner type of the array of the property.

geekerzp:
This pr is a little messy, i will submit a new pr looks  more tidy.

fehguy:
Even better, please pull develop_2.0.  If you cd to `samples/client/petstore/objc/` you can now run `mvn integration-test` which will run the unit tests against the generated objc client.

There is one test failing, and some errors when the stub server shuts down.  But this can be used to test the output of regenerating the objc client via `./bin/objc-petstore.sh`

geekerzp:
Thanks @fehguy , it's a new pr about JSONModel that looks more better.
https://github.com/swagger-api/swagger-codegen/pull/558

PR #527, opened 2015-03-20
Support iso8601 datetime format in Python client

geekerzp:
Use date.util lib to parse string to datetime in python.
The string can be iso8601 datetime format.
`ISO8601`(http://www.w3.org/TR/NOTE-datetime)

### DIFF
--- a/modules/swagger-codegen/src/main/resources/python/swagger.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/swagger.mustache
@@ -194,10 +194,7 @@ class ApiClient:
     if objClass in [int, long, float, dict, list, str, bool]:
       return objClass(obj)
     elif objClass == datetime:
-      # Server will always return a time stamp in UTC, but with
-      # trailing +0000 indicating no offset from UTC. So don't process
-      # last 5 characters.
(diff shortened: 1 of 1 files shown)


modules/swagger-codegen/src/main/resources/python/swagger.mustache
```diff
@@ -194,10 +194,7 @@ class ApiClient:
     if objClass in [int, long, float, dict, list, str, bool]:
       return objClass(obj)
     elif objClass == datetime:
-      # Server will always return a time stamp in UTC, but with
-      # trailing +0000 indicating no offset from UTC. So don't process
-      # last 5 characters.
-      return datetime.datetime.strptime(obj[:-5], "%Y-%m-%dT%H:%M:%S.%f")
+      return self.__parse_string_to_datetime(obj)
```
geekerzp: The datetime string that server returned not only in the format `1997-07-16T19:20:30.45`,
but also in other format e.g. `1997-07-16T19:20:30+01:00`, `1997-07-16T19:20+01:00`.
The client could be able to parse all kinds of iso8601 format string.
Additional, the server returned datetime may contain Timezone, we should convert it to UTC time.

PR #530, opened 2015-03-20
PHP Request Headers and Deserialization

jebentier:
Adding in a function for applying additional headers to an ApiClient Object.


jebentier:
Most recent commit fixed an issue with deserialization that I fixed before but didn't make it over to the latest branch.

PR #539, opened 2015-03-23
JaxRSServerCodegen Delegate Model

joegnelson:
Extended JaxRSServerCodegen to include a delegate call back model allowing the orginally generated code to go untouched


fehguy:
will consider this pattern for next release.

joegnelson:
I updated this fork to match the master. Is there a status on this request?  I am a working for a large company and we are trying to rolling out the use of swagger-codegen and swagger-ui globally but we need this change.  

Mostly we just need 'init(Runnable generate)' added to CodegenConfig so that we can write custom Codegen's with added arguments.   

see compare:
https://github.com/swagger-api/swagger-codegen/compare/master...joegnelson:master?diff=split&name=master

If it will help, i can remove the JaxRSServerCodegenDelegate.java and JaxRSServerCodegenDelegate mustache files from this fork so that the only changes are the addition of the init method.

fehguy:
@joegnelson, we have a mechanism to solve this that @olensmar has worked on.  Would love to get your input on it as I think it will solve your issue.

olensmar:
@joegnelson have a look at the https://github.com/swagger-api/swagger-codegen/tree/jaxrs-interface-generation branch for this; it generates an abstract service class for the API together with implementation and factory classes that won't be overridden when doing subsequent codegens. The actual jaxrs class uses the factory to look up the service implementation to use as a delegate...

Not sure that made any sense - let me know :-)

joegnelson:
it looks like the jaxrs-interface-generation branch is gone?  Can you take a look?

Also, im primary needs are to expose the and extend the command line arguments.  I  did this by exposing a init(Runnable generate) method on CodegenConfig.  

The delegate pattern is a second need.

olensmar:
the branch was merged into develop_2.0

in regard to command-line arguments - I extended the CLI to allow for setting system properties via -Dname=value,name2=value2,etc... which at least give some possibility to pass in custom arguments to the codegen

hope this helps!

/Ole

fehguy:
I'm going to close this out as we have a generalized support in develop_2.0

PR #540, opened 2015-03-23
rft - only codestyle, remove wildcard imports and use string-utils from apache

lanwen:
Use Oracle codestyle, brackets with every if+else.
Nothing else changed
Preparation of complexity refactoring and unit testing in future

[Review on Reviewable](https://reviewable.io/reviews/swagger-api/swagger-codegen/540)


fehguy:
Hi, the code format throughout this project is two-space indents.  Changing to four would be inconsistent for the rest of the code (and other projects).

lanwen:
This is the same for the poms? Two space indents is very hard to read on big parts of code. Why it was chosen? Is any plans to convert to indents that used by most of the other world? :)

PR #541, opened 2015-03-23
Rm duplicated information from pom, reformat indents for better readability

lanwen:
Join duplicated shade plugin declaration to two executions. 
Rm wrong declaration of incremental build in scala plugin. 
Use plugin-management configuration declaration where possible
Rm duplicated versions and groupId in module poms (as they inherited by default)

[Review on Reviewable](https://reviewable.io/reviews/swagger-api/swagger-codegen/541)


lanwen:
(rebased)

lanwen:
will be new one soon

PR #542, opened 2015-03-23
fix StringIndexOutOfBoundsException when tag is not provided

mrwanny:
when tag is not provided a StringIndexOutOfBoundsException is generated 

java.lang.StringIndexOutOfBoundsException: String index out of range: 0
	at java.lang.String.charAt(String.java:646)
	at com.wordnik.swagger.codegen.DefaultCodegen.initialCaps(DefaultCodegen.java:379)
	at com.wordnik.swagger.codegen.DefaultCodegen.toApiFilename(DefaultCodegen.java:142)
	at com.wordnik.swagger.codegen.DefaultGenerator.generate(DefaultGenerator.java:144)
	at com.wordnik.swagger.codegen.Codegen.main(Codegen.java:99)
java.lang.StringIndexOutOfBoundsException: String index out of range: 0
	at java.lang.String.charAt(String.java:646)
	at com.wordnik.swagger.codegen.DefaultGenerator.generate(DefaultGenerator.java:144)
	at com.wordnik.swagger.codegen.Codegen.main(Codegen.java:99)
java.lang.StringIndexOutOfBoundsException: String index out of range: -1
	at java.lang.String.substring(String.java:1918)
	at com.wordnik.swagger.codegen.DefaultGenerator.generate(DefaultGenerator.java:144)
	at com.wordnik.swagger.codegen.Codegen.main(Codegen.java:99)

since tags are note required, the fix only try to capitalize non emply tags

### DIFF
--- a/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultCodegen.java
@@ -376,7 +376,10 @@ public class DefaultCodegen {
   }
 
   public String initialCaps(String name) {
-    return Character.toUpperCase(name.charAt(0)) + name.substring(1);
+    if (name.length()>0)
+      return Character.toUpperCase(name.charAt(0)) + name.substring(1);
+    else
+      return name;
   }
 
   public String getTypeDeclaration(String name) {
--- a/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultCodegen.java
@@ -139,7 +139,7 @@ public class DefaultCodegen {
   }
 
   public String toApiFilename(String name) {
-    return initialCaps(name) + "Api";
+    return toApiName(name);
   }
 
   public String toApiVarName(String name) {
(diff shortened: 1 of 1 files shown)


modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultCodegen.java
```diff
@@ -376,7 +376,10 @@ public String snakeCase(String name) {
   }
 
   public String initialCaps(String name) {
-    return Character.toUpperCase(name.charAt(0)) + name.substring(1);
```
lanwen: Better use StringUtils.capitalize(name) from commons-lang3

fehguy:
There should be a check to set the tag to `default` if it does not exist.  Thus the null check isn't required?

mrwanny:
should we merge the following 2?
```public String toApiFilename(String name) {
    return initialCaps(name) + "Api";
  }```
```public String toApiName(String name) {
    if(name.length() == 0)
      return "DefaultApi";
    return initialCaps(name) + "Api";
  }```

modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultCodegen.java
mrwanny: Agree

PR #544, opened 2015-03-24
Support attribute key mapping in python client

geekerzp:


modules/swagger-codegen/src/main/resources/python/model.mustache
```diff
@@ -30,6 +35,11 @@ class {{classname}}:
             {{/vars}}{{newline}}
         }
 
+        self.attributeMap = {
```
geekerzp: The instance variable `attributeMap` is `dict`, which key is the attribute name and value is the json key in model definition.

modules/swagger-codegen/src/main/resources/python/swagger.mustache
```diff
@@ -131,10 +131,14 @@ class ApiClient(object):
       if isinstance(obj, dict):
         objDict = obj
       else:
-        objDict = obj.__dict__
+        # Convert model obj to dict except attributes `swaggerTypes`, `attributeMap`
+        # and attributes which value is not None.
+        # Convert attribute name to json key in model definition for request.
+        objDict = {obj.attributeMap[key]: val
```
geekerzp: In request, we should convert the model attribute name to the json key in model definition.
So the key in request json is same with the key in model definition.

modules/swagger-codegen/src/main/resources/python/swagger.mustache
```diff
@@ -199,8 +203,8 @@ class ApiClient(object):
     instance = objClass()
 
     for attr, attrType in instance.swaggerTypes.iteritems():
-        if obj is not None and attr in obj and type(obj) in [list, dict]:
-          value = obj[attr]
+        if obj is not None and instance.attributeMap[attr] in obj and type(obj) in [list, dict]:
+          value = obj[instance.attributeMap[attr]]
```
geekerzp: In response, we should convert the json key to model attribute name.
As the json key in response is same with the key in model definition, we should convert it to attribute name.

fehguy:
this is great, thank you

PR #547, opened 2015-03-24
add - new command line interface in distribution module

lanwen:
can replace Codegen and MetaGenerator classes. Provides a bit different cli interface
based on refactored pom from #541,
affects #349

```
usage: swagger <command> [<args>]

The most commonly used swagger commands are:
    generate   Generate code with chosen lang
    help       Display help information
    langs      Shows available langs
    meta       MetaGenerator. Generator for creating a new template set and configuration for Codegen.  The output will be based on the language you specify, and includes default templates to include.
 ```
  **generate** and **meta** commands have same as current cli options

I can remove change of default main class so that it can be merged without brake backward compatibility

[Review on Reviewable](https://reviewable.io/reviews/swagger-api/swagger-codegen/547)


fehguy:
lovely.  thank you

PR #562, opened 2015-03-27
Feature/java client

fehguy:


modules/swagger-codegen/src/main/resources/Java/apiInvoker.mustache
```diff
@@ -83,11 +84,19 @@ public class ApiInvoker {
       return "";
     } else if (param instanceof Date) {
       return formatDateTime((Date) param);
+    } else if (param instanceof Collection) {
+      StringBuilder b = new StringBuilder();
+      for(Object o : (Collection)param) {
```
lanwen: `StringUtils.join()` ?

PR #574, opened 2015-03-30
Fixed issue that query parameters double url encoded in python client

geekerzp:


modules/swagger-codegen/src/main/resources/python/swagger.mustache
```diff
@@ -103,9 +103,9 @@ class ApiClient(object):
         string -- quoted value
     """
     if type(obj) == list:
-      return urllib.quote(','.join(obj))
+      return ','.join(obj)
```
geekerzp: When calling the  method `toPathValue` on the query parameters, the query parameters will be urlencoded.
But in `swagger.mustache` 60 line,
```python
if queryParams:
      # Need to remove None values, these should not be sent
      sentQueryParams = {}
      for param, value in queryParams.items():
        if value is not None:
          sentQueryParams[param] = ApiClient.sanitizeForSerialization(value)
      url = url + '?' + urllib.urlencode(sentQueryParams)
```
will urlencode query parameters again.
The query parameter will be double urlencoded.

PR #575, opened 2015-03-30
Fixed issue about form parameters in Objc client

geekerzp:
* Fixed issue that it will throw exceptions when post form parameters without file.
  In the methods `dictionary` and `stringWithCompletionBlock`:
  ```objective-c
    NSMutableURLRequest * request = nil;
    if (body != nil && [body isKindOfClass:[NSArray class]]){
        SWGFile * file;
        NSMutableDictionary * params = [[NSMutableDictionary alloc] init];
        for(id obj in body) {
            if([obj isKindOfClass:[SWGFile class]]) {
                file = (SWGFile*) obj;
                requestContentType = @"multipart/form-data";
            }
            else if([obj isKindOfClass:[NSDictionary class]]) {
                for(NSString * key in obj) {
                    params[key] = obj[key];
                }
            }
        }
        NSString * urlString = [[NSURL URLWithString:path relativeToURL:self.baseURL] absoluteString];

        // request with multipart form
        if(file != nil) {
            request = [self.requestSerializer multipartFormRequestWithMethod: @"POST"
                                                                   URLString: urlString
                                                                  parameters: nil
                                                   constructingBodyWithBlock: ^(id<AFMultipartFormData> formData) {

                                                       for(NSString * key in params) {
                                                           NSData* data = [params[key] dataUsingEncoding:NSUTF8StringEncoding];
                                                           [formData appendPartWithFormData: data name: key];
                                                       }

                                                       [formData appendPartWithFileData: [file data]
                                                                                   name: [file paramName]
                                                                               fileName: [file name]
                                                                               mimeType: [file mimeType]];

                                                   }
                                                                       error:nil];
        }
}
  ```
If the body array only contains form parameters without file, the request object will never be initialized.
* Determine the correct request content-type and response type.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/objc/SWGApiClient.m
+++ b/modules/swagger-codegen/src/main/resources/objc/SWGApiClient.m
@@ -31,13 +31,13 @@ static bool loggingEnabled = true;
                                       diskSize: (unsigned long) diskSize {
     NSAssert(memorySize > 0, @"invalid in-memory cache size");
     NSAssert(diskSize >= 0, @"invalid disk cache size");
-    
+
     NSURLCache *cache =
     [[NSURLCache alloc]
      initWithMemoryCapacity:memorySize
      diskCapacity:diskSize
      diskPath:@"swagger_url_cache"];
-    
+
     [NSURLCache setSharedURLCache:cache];
 }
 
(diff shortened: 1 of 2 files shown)


fehguy:
I have merged this, but it still has a problem.  Please check out this test (it's disabled by naming convention):

https://github.com/swagger-api/swagger-codegen/commit/5a5bb6edbffb010c189328f10d08e18af845b95b#commitcomment-10575829

geekerzp:
Sure, let me fix it.

geekerzp:
Hi @fehguy, what's the expected behaviour when the file is not provided? Should the server return 200?
Should the client throw an exception (even though file is an optional parameter) ?

fehguy:
I think it should allow the request to be made to the server, which I would expect returns a 2xx response.  I believe right now, it'll just crash.

geekerzp:
I have tested upload without file in the objc client and the ruby client, the server returns 415 for now.
Is it the expected behaviour?

fehguy:
I believe it's returning a 415 because it's not setting `ContentType:multiplart/form-data`.  If you do it with the java client, it should be fine.

geekerzp:
Ok i see, let me fix it.

fehguy:
thank you!

PR #580, opened 2015-04-01
Fix content-type and accept in PHP API

wing328:
Update content-type for PHP API to allow multiple content-type: pick the first one or use 'application/json' as default. (implemented with reference to how Java API handles content-type)

Set "Accept" header only if "produces" is provided in the Swagger spec (json).


fehguy:
@wing328 you picked up some .cs files in this commit.  intentional?

wing328:
Sorry, I've removed those files: CSharp model files (lower case). Please kindly review.

fehguy:
@wing328 thank you again.  If you look at the ruby sample, I've added a maven profile to run unit tests :)  Would be very helpful to do the same with php if you have time.

https://github.com/swagger-api/swagger-codegen/commit/13ee40a4b8d05b27ff1d08e1c1577d3d5ea14e8e

wing328:
@fehguy Sure but I'm no expert in PHP unit testing so it might take some time before I can come up with something ready for automated unit testing.

Issue #577, opened 2015-03-31
Java CL has compiler error if model has (\n) in its definition text.


suguptaxbox:
If the description of the model has new line characters in the text, the generated client library has invalid code and hence cause compile time error.

Repro:
1. Open the Swagger Editor http://editor.swagger.io/#/
2. In the editor (assuming you have yaml editor) add the following definition to the end of the file.
```yaml
  TestingNewLineCompilerError:
    description: "First line\nSecond line."
    required:
      - message
    properties:
      message:
        type: string
```
3. Now generated the java client library.
    click on Generate Client->Java
4. Download the generated source code and open the model file for TestingNewLineCompilerError
    Open file: src/main/java/io/swagger/client/model/TestingNewLineCompilerError.java
5. Verify the code in the file and observe that it is incorrect where description of ApiModel attribute has java string on two different lines. Therefore if you compile the code it will cause compile time error because there is no '+' plus sign in between two lines.

/**
 * First line
Second line.
 **/
@ApiModel(description = "First line
Second line.")
public class TestingNewLineCompilerError  {

Actual behavior:
When compiling the generated code there is compile time error.

Expected:
When compiling the generated code there should not be any error.

fehguy:
Yes, this is a bug.  Will get it fixed.

PR #581, opened 2015-04-01
This is a fix for issue: https://github.com/swagger-api/swagger-codegen/issues/577

guptasu:
Fixed the compile time error in the generated code caused due to unescaped new line characters.

Since the description text is not being escaped, if there is a new line character, it will cause a new line to be added to the source code. This can cause compile time error if the new line character is part of a string (e.g. java string) and there is no '+' sign explicitly added to the code. Note: The code was doing escapeText for properties but not for models. This patch fixes that.


webron:
@guptasu - code contributions should be made against the develop_2.0 branch, so please resubmit against that one. Also, if you want to refer to an issue, just write something like #577 and github parses it automatically :)

PR #583, opened 2015-04-02
Determine package name from host for Java client

xhh:
Currently the generated Java client has hard-coded values for package, `groupdId`, `artifactId` and `version`:

```java
  protected String invokerPackage = "io.swagger.client";
  protected String groupId = "io.swagger";
  protected String artifactId = "swagger-client";
  protected String artifactVersion = "1.0.0";
```

This pull request determines the above fields from the host, title and version provided in swagger spec.

The integration test and petstore sample tests for Java pass without errors. If this looks fine, I'll implement it for Android generator as well.


samples/client/petstore/java/src/main/java/io/swagger/client/model/ApiResponse.java
```diff
@@ -1,64 +0,0 @@
-package io.swagger.client.model;
```
xhh: I regenerated petstore java sample by removing `samples/client/petstore/java/src/main` first, to remove the old files in different directory. Then the generation does not generate this `ApiResponse.java` file, is it still needed?

fehguy:
Hi, I don't know why we would need this, since the swagger-spec is going to have the host in it.

xhh:
@fehguy for now the generated Java SDK already has the following in `pom.xml` for any host:

```xml
  <groupId>io.swagger</groupId>
  <artifactId>swagger-client</artifactId>
  <name>swagger-client</name>
  <version>1.0.0</version>
```

and always use `package io.swagger.client.api;` for all API classes (similar for models).

This PR is to change the hardcoded values to guess from the host provided in swagger spec.

fehguy:
Got it. In that case I'm going to expose a configuration object so this whole process is completely explicit.

jfiala:
Wouldn't it be better to allow the package name to be passed in as a commandline parameter?
Especially in development you might connect to localhost instead of the "real" api server.
At least, it should be possible to override the auto-assembled "DefaultCodegen.hostToPackage"-name.

xhh:
@fehguy while making the process explicit, I think it still needs to automatically generate the package name for an implicit process, to allow the user leave the option and use the default generation process.

Also different languages would use different formats of package name. For example, Java would use "com.twitter.api" while PHP would use "twitter-php".

PR #585, opened 2015-04-02
Fixed Ruby path variable url-encoded twice, removed duplicated path variable replacement

wing328:
Currently, ruby path variable is url-encoded twice:
- by escapeString at line 33 of https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/ruby/api.mustache#L33
- by URI.encode at https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/ruby/swagger/request.mustache#L101

This PR will remove escapeString and let URI.encode to perform the encoding

Removed "Fill in the path params":  https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/ruby/swagger/request.mustache#L76

as the path params are already handled at api file: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/ruby/api.mustache#L33

As a result of this change, I've updated the test case *not* to test the variable replacement in request.rb (only test {format} replacement with xml, json)


wing328:
@fehguy I believe the above comment refers to another PR related to Objc

To illustrate the issue fixed by this PR for Ruby codegen, below is the tcpdump of getStoreById with "%" (invalid store ID)
```
GET /v2/store/order/%2525 HTTP/1.1
```
% is encoded as %25 and the above shows that % is encoded twice

After the fix, tcpdump shows the correct encoded URL below:
```
GET /v2/store/order/%25 HTTP/1.1
```

I've just updated this PR fixing the rspec error result to due the order with ID 2 no longer found. The change also puts the Store spec in a separate file (store_spec.rb)

PR #586, opened 2015-04-02
Akka scala

cchafer:
Hi,

This is a Scala client generator based on Akka and Spray.
I did start this one quite a long time ago, when the async-scala was still new, and because current drivers where insufficient for my needs.

I'm using it right now, and feel comfortable with it. Please, feel free to make remarks, critics, ..

Main features now available are : 
* Truly asynchronous : not a single blocking call, and usage of Scala Futures.
* All responses described in the swagger json can be retrieved through the API, not only the successful one.
* Authentication (ApiKey & Basic Auth, no oAuth at the moment) works out of the box
* generated API is "looking good", and documented. Because who knows if we won't need to dig into the sources ?
* enumerations work as expected
* arrays serialization works as expected (the 'collectionFormat' parameter is used, and can be csv, ssv, tsv, pipes or multi)
* responses in headers are parsed and returned
* all defined parameters are sent : in headers, path, query, body, ;..
* requests can be compressed (quite strange and not handled by all http servers, but useful in y case)
* SSL certificates verification can be disabled using the configuration file
* content-type are handled as expected by the HTTP specification. Not just using contentTypes[0], hoping it's the good one :D Detection is based on the content type (... strange sentence), and fallback on the specified one if needed.
* non required fields are of type Option[T]. Because null is devil. Warning : could look strange if you don't specify the requireness carefully in your format.
* custom headers can be specified by conf. By default, user-agent is defined.

Here are some other points of this generator :
- uses Spray http client, and Akka Futures. Spray should soon become a full part of Akka, through Akka Http, hence the name of the generator : akka-scala instead of spray-scala
- at this time, the invoker is built by the generator. The usage of an external ApiInvoker (as it's done in the JavaCodegen, for example), is definitely better since it would allow one to use one and only one invoker for several different generated APIs. This said, in this PR, models and APIs files do not depend on the ApiInvoker. It should then be quite straightforward to create a separate project only for the invoker.
- configuration is done through the akka standard configuration mechanism : https://github.com/typesafehub/config
- I can see two different ways of using Swagger. One can consider that an API can result in only 1 valid answer. This is the choice made by actual drivers. But I used to consider that several different valid answers can be returned. In this case, API should be able to return an "undetermined" type of content. This generator let's you choose (using some variable in AkkaScalaCodegen) which kind of usage fits your needs the best.
- due to Spray limitation/conception, non standard http status code need to be known and registered in the application. This is automatically done for status codes defined in the json, and you can add more custom non standard codes (like >=600) either in the configuration file or in your code (ApiInvoker.addCustomStatusCode)
- I used some Mustache Lambdas. Don't know if it's a problem, but hey ! It's quite useful and let you minimise the number of properties to add to CodegenOperations, Parameters or Properties.

And finally, a usage example:
```scala
  // for IoC : invoker.execute(r: ApiRequest) becomes r.response(invoker)
  import ApiInvoker._

  // EnumsSerializers automatically includes all seralizers for the model enums. In this case, not really useful
  // since the Pet model doesn't contain any enumeration.
  implicit val invoker = ApiInvoker(EnumsSerializers.all)

  implicit val apiKey = ApiKeyValue("some api Key")

  // get an ApiResponse, containing the expected content, the status code, and the response headers
  PetApi.getPetById(1).response.onComplete {
    case Success(ApiResponse(code, pet, headers)) =>
      log info s"Pet ${pet.name} received with status code $code !"
    case Failure(e) =>
      log error s"Some error occured : ${e.getMessage}"
  }

  // automatically map the ApiResponse to the expected type if status code or headers are not needed
  PetApi.getPetById(1).result[Pet].onComplete {
    case Success(pet) =>
      log info s"Pet ${pet.name} received !"
    case Failure(e) =>
      log error s"Some error occured : ${e.getMessage}"
  }
```

That's quite all. Thanks for your time, and sorry for my crap english !


fehguy:
Thank you for this!  It will take a bit to review the PR so apologies if it takes a bit.

webron:
@cchafer - sorry for the delay, but we're preparing for the release of the next version. Any chance you can rebase your code? There have been quite a few changes in the code base.

cchafer:
Hi, @webron,
no problem for the delay.
I'll try for the rebase, but can't guarantee any delay.

PR #611, opened 2015-04-09
Updated CSharp client to support ISO8601 format

wing328:
- Updated CSharp client to support ISO8601 format
- Reduced warnings from 22 down to only 6 for Petstore CSharp client
- Added PATCH support
- Added more comments to utility functions in ApiInvoker


webron:
It looks like you got some php changes there as well?

wing328:
Hi Ron, those PHP files (vendor, composer.lock) were removed. It was left there when I performed testing with the PHP client for another PR.

PR #618, opened 2015-04-10
Added back python 3 codegen

wing328:
The python3 template is already there. I just added the Codegen (java) based on Python (v2) codegen with the following differences:

- remove "long" from the primitive type
- mapped "long" to "int" instead of "long" (which is removed from Python3)


wing328:
This PR is based on a request from a user who still wants to use the Python3 client. If the direction is to use a single Python client for both python v2 and v3, I'm happy to close this PR.

PR #623, opened 2015-04-12
Fixed "double" response type in PHP client

wing328:
Currently, the PHP API client only looks for the following primitive types:
```
    } elseif (in_array($class, array('string', 'int', 'float', 'bool'))) {
```
This PR updates the check to include "double":
```
    } elseif (in_array($class, array('string', 'int', 'float', 'double', 'bool'))) {
```

### DIFF
--- a/modules/swagger-codegen/src/main/resources/php/APIClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/php/APIClient.mustache
@@ -282,7 +282,7 @@ class APIClient {
       $deserialized = $values;
     } elseif ($class == 'DateTime') {
       $deserialized = new \DateTime($data);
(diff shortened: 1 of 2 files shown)


jfiala:
I think this would be more convenient for future updates to make it dynamic (I can provide as PR later):

PhpClientCodeGen.java:
```java
// add double
languageSpecificPrimitives.add("double");

// provide primitives to mustache template
    String primitives = "'" + StringUtils.join(languageSpecificPrimitives, "', '") + "'";
    additionalProperties.put("primitives", primitives);
```

APIClient.mustache:
```
// before: } elseif (in_array($class, array('string', 'int', 'float', 'bool', 'double'))) {
    } elseif (in_array($class, array({{&primitives}}))) {
```

PR #629, opened 2015-04-13
Correct usage of strpos in PHP APIClient mustache template

who:
The PHP function strpos is not being used as recommended by http://php.net/strpos.

I have fixed this.


webron:
@who - you deserve a prize for choosing that username.

Issue #636, opened 2015-04-14
Objc - SWGQueryParamCollection.h missing import for Foundation.h


hyeghiazaryan:
There is no import for Foundation/Foundation.h in SWGQueryParamCollection.h which causes error when extending NSObject.

PR #637, opened 2015-04-14
Adding import for Foundation.h to SWGQueryParamCollection.h

hyeghiazaryan:
fix for https://github.com/swagger-api/swagger-codegen/issues/636

### DIFF
--- a/modules/swagger-codegen/src/main/resources/objc/SWGQueryParamCollection.h
+++ b/modules/swagger-codegen/src/main/resources/objc/SWGQueryParamCollection.h
@@ -1,3 +1,5 @@
+#import <Foundation/Foundation.h>
+
 @interface SWGQueryParamCollection : NSObject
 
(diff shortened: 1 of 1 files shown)


fehguy:
I can add this, but isn't this something typically in the *.pch files?

geekerzp:
I think we should add `#import <Foundation/Foundation.h>` in the *.h files.
When we inherit the `NSObject`, we should import the Foundation.h explicitly in the head file.
Also see https://github.com/swagger-api/swagger-codegen/blob/develop_2.0/modules/swagger-codegen/src/main/resources/objc/SWGApiClient.h#L1

hyeghiazaryan:
@fehguy

This import is usually added to .h files, you can see this import in SWGApiClient.h, SWGDate.h, SWGGile.h, SWGObject.h.
Only SWGQyeryParamCOllection.h does not have it.

fehguy:
done, thanks.

Issue #642, opened 2015-04-15
In the html client generator, newline (\n) added in the JSON under Example data 


kvelakur:
For example go to http://editor.swagger.io/ and export the default Uber API that is loaded as a static html page by selecting Generate Client -> html. 

In the generated html, all the JSONs under Example data look like this: 
`[ {\n  "image" : "aeiou",\n  "product_id" : "aeiou",\n  "description" : "aeiou",\n  "display_name" : "aeiou",\n  "capacity" : "aeiou"\n} ]`

The \n shouldn't appear here, and should probably a <\br> instead.

webron:
@kvelakur - it looks like you have a potential PR here?

kvelakur:
Yes I did:
https://github.com/swagger-api/swagger-codegen/pull/654

Removing the line fixes the Static HTML page, but I'm not sure if it affects anything else. In any case this line needs to be fixed, the second argument in the function call is not even a valid regular expression.

wing328:
@kvelakur I believe it's fixed and I could no longer repeat the issue. Please reopen the issue if it's not the case.

PR #654, opened 2015-04-16
Issue #642 - Removing the unnecesary \n added to generated example reque...

kvelakur:
...sts


fehguy:
This will break some upcoming features.  What issue is the `\n` causing for you?

kvelakur:
If you generate a static html document, the example requests in the page look like this

`[ {\n  "image" : "aeiou",\n  "product_id" : "aeiou",\n  "description" : "aeiou",\n  "display_name" : "aeiou",\n  "capacity" : "aeiou"\n} ]`

To reproduce, go to http://editor.swagger.io/ and select Generate Client -> html 

Why are the \n added explicitly as characters in a string?

PR #655, opened 2015-04-17
Set version dynamically in ruby codegen

xhh:
As discussed [here](https://github.com/swagger-api/swagger-codegen/commit/7db7b146beee6811d2effa4ddf0d3e84b1ef0e39#commitcomment-10760943)


xhh:
This is included in #662

PR #662, opened 2015-04-20
Organize the generated Ruby code into a module structure

xhh:
This PR is to make the folder/file structure of the generated Ruby code comply with the conventions of Ruby gems.

Taking Petstore as an example:

```shell
$ tree lib
lib
├── swagger_client
│   ├── api
│   │   ├── pet_api.rb
│   │   ├── store_api.rb
│   │   └── user_api.rb
│   ├── models
│   │   ├── category.rb
│   │   ├── order.rb
│   │   ├── pet.rb
│   │   ├── tag.rb
│   │   └── user.rb
│   ├── monkey.rb
│   ├── swagger
│   │   ├── configuration.rb
│   │   ├── request.rb
│   │   ├── response.rb
│   │   └── version.rb
│   └── swagger.rb
└── swagger_client.rb
```

**1. Place all files in the `lib` folder**
As suggested by [RubyGems Basics](http://guides.rubygems.org/rubygems-basics/):
> The lib directory itself normally contains only one .rb file and a directory with the same name as the gem which contains the rest of the files.

**2. Move source files in an enclosing namespace module to avoid potential conflicts**
This is a practice used by most gems (e.g. [stripe-ruby](https://github.com/stripe/stripe-ruby/blob/master/lib/stripe/account.rb)).
As in the Petstore sample, use `SwaggerClient::PetApi` and `SwaggerClient::Pet` to access API and model classes.

**3. Rename gem name of Petstore sample from `swagger-client` to `swagger_client`**
according to the naming conventions from [RubyGems guides](http://guides.rubygems.org/name-your-gem/).


fehguy:
Hi @xhh thank you for this.  I'd like to avoid the `importPath` in CodegenModel.  Can you explain that a bit more?

xhh:
@fehguy that is for the `require` statement to require/import all the models and apis, for example the [petstore sample](https://github.com/swagger-api/swagger-codegen/pull/662/files#diff-247c7c4a34e15da7187adcb8182cc832R9):

``` ruby
# Models
require 'swagger_client/models/user'
require 'swagger_client/models/category'
require 'swagger_client/models/pet'
require 'swagger_client/models/tag'
require 'swagger_client/models/order'

# APIs
require 'swagger_client/api/user_api'
require 'swagger_client/api/pet_api'
require 'swagger_client/api/store_api'
```

The [template](https://github.com/swagger-api/swagger-codegen/pull/662/files#diff-052d7ffc0291c3b80b78c02ccae058afR9) generating the above is:

``` ruby
# Models
{{#models}}
{{#model}}
require '{{importPath}}'
{{/model}}
{{/models}}

# APIs
{{#apiInfo}}
{{#apis}}
require '{{importPath}}'
{{/apis}}
{{/apiInfo}}
```

The `models` variable is used here to generate all model requires, which need to [set the actual require path](https://github.com/swagger-api/swagger-codegen/pull/662/files#diff-1e1f9c96bd64b37327fddd5ddcd93bd8R431), i.e. `importPath`. For apis, the `importPath` key is added to the [corresponding Map](https://github.com/swagger-api/swagger-codegen/pull/662/files#diff-54cc197a0b6d4a5b10bbe573abcd22c6R168).

Any other ways to generate the require/import path of all models? If there are I'm happy to make the changes.

xhh:
There is another way to require ruby files dynamically, similar to [the PHP approach](https://github.com/swagger-api/swagger-codegen/blob/e8ae9178af5a62276f09deac9805ca3ff8b2d6e8/modules/swagger-codegen/src/main/resources/php/require.mustache), which does not need to iterate the list of models/apis (so the `importPath` is not needed).

If that approach looks OK I can make the change. However, I'd like to know the concerns of avoiding `importPath` here. Is there a plan to decommission import, or because `importPath` is not used by codegen of other languages?

xhh:
@fehguy let me know what else needs to be done for this if you have time

xhh:
just rebased this

wing328:
Could build and install the Petstore Ruby gem locally without any issue/warning.
Tested with `get_pet_by_id` using the local 'swagger-client' gem without issue.

csaunders:
Here's a note about something I noticed while playing with swagger codegen:

A bit of a concern with the current Ruby API consumer is that the generated libraries aren't thread-safe. An easy way would be to move the `configuration` values over to thread locals, but a better thing would be a configuration object or client that is initialized instead.

The result is perhaps a client that requires a bit more setup but it gets rid of the class methods that rely on shared global state.

---------------------------------

I just did a cursory look through the patch and the resulting code seems more ruby-esque which is super awesome.

fehguy:
Hi @xhh I'd like to merge this but looking to see if we can avoid modifying the `CodegenModel`

xhh:
@fehguy i managed to avoid modifying `CodegenModel` by setting the key to the `models` item in this commit: https://github.com/swagger-api/swagger-codegen/commit/312f1c6f25461a8c7b01e47625cf3d9313f383f3

Issue #674, opened 2015-04-22
Spring MVC request param required field missing


marcinkiewiczblazej:
In current version even if `required` parameter is provided in query parameters it's not appearing in Spring MVC annotation (it's only used in Swagger annotation).

webron:
Fixed in develop_2.0.

PR #675, opened 2015-04-22
Fix for issue #674

marcinkiewiczblazej:
Added missing `required` parameter in Spring annotation.


webron:
Fixes #674.

fehguy:
@webron it seems like this fix only adds a newline?

webron:
@fehguy - didn't really look at the PR. Just wanted to create a link to the issue. The fix actually removes a line too, so I'm not really sure how/why it fixes anything.

marcinkiewiczblazej:
There's a problem with Github diff, if you check the file there are changes. Most important one is:
```
@RequestParam(value = "{{paramName}}"{{#required}}, required = true{{/required}}{{^required}}, required = false{{/required}}
```
So I've added required param to `@RequestParam` in both cases just to be future proof if the default settings in Spring are changed.

PR #676, opened 2015-04-22
Fixes #646: Example field handling has been fixed.

### LINKED ISSUE #646: Parsing error when a response example is of type `object`
From the swagger spec's doc: https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#example-object-example

the response example could be an `object`, e.g.
```javascript
{
  "application/json": {
    "name": "Puma",
    "type": "Dog",
    "color": "Black",
    "gender": "Female",
    "breed": "Mixed"
  }
}
```

But when I copied the above example to `modules/swagger-codegen/src/test/resources/2_0/petstore.json` and run the Java codegen, I got this error:
```shell
$ ./bin/java-petstore.sh
reading from modules/swagger-codegen/src/test/resources/2_0/petstore.json
Can not deserialize instance of java.lang.String out of START_OBJECT token
 at [Source: N/A; line: -1, column: -1] (through reference chain: com.wordnik.swagger.models.Swagger["paths"]->com.wordnik.swagger.models.Path["get"]->com.wordnik.swagger.models.Operation["responses"]->com.wordnik.swagger.models.Response["examples"])
reading from modules/swagger-codegen/src/test/resources/2_0/petstore.json
Can not deserialize instance of java.lang.String out of START_OBJECT token
 at [Source: N/A; line: -1, column: -1] (through reference chain: com.wordnik.swagger.models.Swagger["paths"]->com.wordnik.swagger.models.Path["get"]->com.wordnik.swagger.models.Operation["responses"]->com.wordnik.swagger.models.Response["examples"])
Exception in thread "main" java.lang.RuntimeException: missing swagger input or config!
        at com.wordnik.swagger.codegen.DefaultGenerator.generate(DefaultGenerator.java:54)
        at com.wordnik.swagger.codegen.cmd.Generate.run(Generate.java:77)
        at com.wordnik.swagger.codegen.SwaggerCodegen.main(SwaggerCodegen.java:33)
```

There's a similar error for example of type array as well.

tomtit:
Fixes #646: Example field handling has been fixed.
The swagger-api/swagger-core/pull/1002 has to be accepted first.

I would suggest to remove escaping of '\n' (i.e. example = example.replaceAll("\n", "\\\\n") in [ExampleGenerator](https://github.com/swagger-api/swagger-codegen/pull/676/files#diff-08b7ea6b5aaa485f71bcf31aec901b5dL40) and jsonExample = jsonExample.replaceAll("\\\\n", "\n") in [NodeJSServerCodegen](https://github.com/swagger-api/swagger-codegen/pull/676/files#diff-7a12f5f8b19af3ab6e9ee4867476611bL181)) and leave this on behalf of particular descendants of the DefaultCodegen class.
As for now only `html` and `nodejs` generators use examples, while the first outputs examples as is and has the issue reported [here](https://github.com/swagger-api/swagger-codegen/pull/654#issuecomment-93792558), and the former handles escaped '\n' properly

### DIFF
--- a/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/examples/ExampleGenerator.java
+++ b/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/examples/ExampleGenerator.java
@@ -1,19 +1,33 @@
 package com.wordnik.swagger.codegen.examples;
 
-import com.wordnik.swagger.models.*;
-import com.wordnik.swagger.models.properties.*;
-import com.wordnik.swagger.util.Json;
-
-import java.text.SimpleDateFormat;
-
-
-import javax.xml.bind.JAXBContext;
-import javax.xml.bind.JAXBException;
-import javax.xml.bind.Marshaller;
- 
-
 import java.math.BigDecimal;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import com.wordnik.swagger.models.Model;
+import com.wordnik.swagger.models.ModelImpl;
+import com.wordnik.swagger.models.properties.ArrayProperty;
+import com.wordnik.swagger.models.properties.BooleanProperty;
+import com.wordnik.swagger.models.properties.DateProperty;
+import com.wordnik.swagger.models.properties.DateTimeProperty;
+import com.wordnik.swagger.models.properties.DecimalProperty;
+import com.wordnik.swagger.models.properties.DoubleProperty;
+import com.wordnik.swagger.models.properties.FileProperty;
+import com.wordnik.swagger.models.properties.FloatProperty;
+import com.wordnik.swagger.models.properties.IntegerProperty;
+import com.wordnik.swagger.models.properties.LongProperty;
+import com.wordnik.swagger.models.properties.MapProperty;
+import com.wordnik.swagger.models.properties.ObjectProperty;
+import com.wordnik.swagger.models.properties.Property;
+import com.wordnik.swagger.models.properties.RefProperty;
+import com.wordnik.swagger.models.properties.StringProperty;
+import com.wordnik.swagger.models.properties.UUIDProperty;
+import com.wordnik.swagger.util.Json;
 
 public class ExampleGenerator {
   protected Map<String, Model> examples;
@@ -22,7 +36,7 @@ public class ExampleGenerator {
     this.examples = examples;
   }
 
-  public List<Map<String, String>> generate(Map<String, String> examples, List<String> mediaTypes, Property property) {
+  public List<Map<String, String>> generate(Map<String, Object> examples, List<String> mediaTypes, Property property) {
     List<Map<String, String>> output = new ArrayList<Map<String, String>>();
     Set<String> processedModels = new HashSet<String>();
     if(examples == null ) {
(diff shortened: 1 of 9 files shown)


pom.xml
```diff
@@ -440,7 +440,7 @@
     <swagger-parser-version>1.0.6-SNAPSHOT</swagger-parser-version>
     <scala-version>2.10.4</scala-version>
     <felix-version>2.3.4</felix-version>
-    <swagger-core-version>1.5.1-M2</swagger-core-version>
+    <swagger-core-version>1.5.2-M2-SNAPSHOT</swagger-core-version>
```
webron: @tomtit - notice that this broke the build

tomtit: That is why I've referenced the swagger-api/swagger-core/pull/1002 pull request. The  com.wordnik.swagger.models.Response from that change is required. Just don't know how to reference the develop_2.0 branch properly.

webron: Apologies, I had missed that comment. We can't depend on a SNAPSHOT here though. That means we'll have to wait until swagger-core is released and then modify this PR accordingly. Still not sure what version that's going to end up being.

modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/examples/ExampleGenerator.java
```diff
@@ -53,12 +65,10 @@ else if(property != null && mediaType.startsWith("application/xml")) {
       }
     }
     else {
```
tomtit: This piece has to be reworked as it renders examples into JSON view regardless of the `contentType` value.

modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/examples/ExampleGenerator.java
```diff
@@ -53,12 +65,10 @@ else if(property != null && mediaType.startsWith("application/xml")) {
...
-      for(String key: examples.keySet()) {
-        String value = examples.get(key);
-
-        Map<String, String> kv = new HashMap<String, String>();
-        kv.put("contentType", key);
-        kv.put("example", value);
+      for(Map.Entry<String, Object> entry: examples.entrySet()) {
+        final Map<String, String> kv = new HashMap<String, String>();
+        kv.put("contentType", entry.getKey());
+        kv.put("example", Json.pretty(entry.getValue()));
```
tomtit: This piece has to be reworked as it renders examples into JSON view regardless of the `contentType` value.

wing328: @tomtit I would recommend you to leave a comment (TODO) in the code so that someone can find out what needs to be done and potentially file a PR for that.

tomtit: @wing328, I would like to rework this piece as a part of this PR. I realized the problem with `contentType` after the PR had been created. Perhaps @fehguy or @webron could provide some additional comments here.

tomtit: Here is the note from discussion with @webron:
> If the user provided:
>"application/xml": {
>    "name": "Puma",
>    "type": "Dog",
>    "color": "Black",
>    "gender": "Female",
>    "breed": "Mixed"
>  }
>we are going to show it just like that. A JSON.
>If they want an XML there, let them put an XML inside a string.
So, the proposed changes provide exactly this functionality.

pom.xml
stevehu: @tomtit There is a Java generator that uses examples but the code cannot be merged back to the official repo as the generated code is in Java 8 but swagger-codegen has to be built in Java 7 container. 
Here is the repo
https://github.com/networknt/swagger-codegen
And here is the petstore generated
https://github.com/networknt/light-java-example/tree/master/petstore

PR #677, opened 2015-04-22
Fix PHP swagger type

wing328:
- fixed Swagger type for PHP codegen
- tested with model name `Pet.Test` and response correctly casts to `PetTest`
- commented out import test as import is not used in PHP templates

### DIFF
--- a/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/languages/PhpClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/languages/PhpClientCodegen.java
@@ -39,9 +39,6 @@ public class PhpClientCodegen extends DefaultCodegen implements CodegenConfig {
     apiTemplateFiles.put("api.mustache", ".php");
     templateDir = "php";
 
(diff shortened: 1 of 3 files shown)


maver1ck:
Hi,
Is it possible to merge this change ?
I have the same problem ? Especially when using online editor.

wing328:
@maver1ck are you referring to the PHP codegen (Generate Client => PHP) in the online editor not able to properly generate PHP client with model name containing dot ?

maver1ck:
No.
I'm referring to issue when online codegen can't generate php client when there is some boolean variables.

wing328:
@maver1ck this PR aims to fix the issue with model name containing dot (e.g. Pet.Store).

Have you tested it it confirm it will address the issue about boolean variables? Do you have the Swagger spec that you can share with us (publicly) so that I can try to repeat the issue ?

wing328:
@maver1ck looking at my PR again, I think you're referring to the typeMapping fix that adds the mapping for "boolean".

it would be nice if you can test to confirm it can address the boolean issue you've encountered. Thanks!

PR #681, opened 2015-04-24
Add form parameters support to POST operations in Java codegen

xhh:
Only PUT operations handle form parameters currently, this PR handles them for POST too.

Refactor is also done for handling query and form parameters to let jersey-client do the encoding work.


webron:
Is #712 related to this one?

xhh:
Right. This PR only adds form parameters support to POST, not to DELETE (while #712 does it for both POST and DELETE).
This one can be closed now.

PR #683, opened 2015-04-24
Java template: Make ApiInvoker more pluggable

karussell:
We needed simple key authentication and with this change you can easily extend the ApiInvoker and e.g. hook into ApiInvoker.invokeAPI to modify the queryParams. And for header based authentication you could change the headers there too.

Example usage:
```java
ApiInvoker myInvoker = new ApiInvoker() {
  public String invokeAPI(String host, String path, String method, Map<String, String> queryParams, ...) {
    queryParams.put("key", someKey);
    super.invokeAPI(host, path, method, queryParams, ...);
  }
}

SomeApi api = new SomeApi(myInvoker);
```

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/api.mustache
@@ -21,8 +21,12 @@ import java.util.HashMap;
 
 {{#operations}}
 public class {{classname}} {
-  String basePath = "{{basePath}}";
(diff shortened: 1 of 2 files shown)


webron:
@karussell - thanks for the PR, but it should be submitted against the develop_2.0 branch.

karussell:
Okay

karussell:
I've reposted it as #684 hope it is correct now :)

PR #694, opened 2015-04-28
Add missing include in SWGQueryParamCollection.h

sschiffli:


webron:
@sschiffli - thank you for the PR but those need to be made against the develop_2.0 branch. Can you resubmit?

sschiffli:
Oh my bad, looks like its already been fixed on the develop_2.0 branch.

Issue #673, opened 2015-04-22
[Suggestion] Use RestSharp to replace ApiInvoker in CSharp


wing328:
I would suggest using [RestSharp](https://github.com/restsharp/RestSharp) to replace ApiInvoker in the CSharp client.

I've tested GetPetById (GET), AddPet (POST JSON) and UpdatePetWithForm(POST FORM) using RestSharp and it works perfectly fine. To me, the code looks cleaner and easier to understand.

Here is a list of projects/SDKs using RestSharp:
https://github.com/restsharp/RestSharp/wiki/Projects-Using-RestSharp

May I know if anyone has concern using RestSharp to replace ApiInvoker in CSharp?

fehguy:
no concerns from my side.  Looks like a solid library.

wing328:
PR merged

PR #700, opened 2015-04-29
Use RestSharp for CSharp API client

wing328:
Use RestSharp to make HTTP calls. Tested with the following:
- GetPetById
- UpdatePetWithForm (confirmed by inspecting the HTTP request using tcpdump)
- AddPet

Add `DataContract` to CSharp models (similar to @jsonproperty in Java) to properly map attributes with special characters. Deserialisation of responses from server works fine (tested with Pet, a model with sub-models)


wing328:
Based on suggestion #673

PR #703, opened 2015-04-30
Added support for Basic Auth

MaorBril:


fehguy:
To be clear, this only supports basic auth for reading the specs.

modules/swagger-codegen-cli/src/main/java/com/wordnik/swagger/codegen/cmd/Generate.java
```diff
@@ -64,6 +70,11 @@ public void run() {
             input.setAuth(auth);
         }
 
+        if (isNotEmpty(basicAuth)) {
+            String authorization = new String(Base64Coder.encode(basicAuth.getBytes()));
+            input.setAuth("Authorization:" + "Basic "+ authorization);
```
lanwen: why you concatenate it twice?

webron: twice?

wing328: I think what @lanwen suggested is the following (which concatenates the strings once only):
`input.setAuth("Authorization: Basic "+ authorization);`

webron: Oh, I wish those were our issues ;)

Issue #687, opened 2015-04-24
Generated Android client has high number of referenced methods


who:
In an Android app, there is a [limit on the number of methods you can have](https://developer.android.com/tools/building/multidex.html).  This limit is ~65k.

Because of the dependencies that it imports, the android-java client generated by swagger-codegen has a significantly high number of methods referenced. We should see if this can be reduced.

Below is an example.  Note: the example relies on https://gist.github.com/who/cdb37baffa7f64d27180

```
me@box:/tmp/tmp.wMZIQvX55e$ git clone https://github.com/swagger-api/swagger-codegen.git
...
me@box:/tmp/tmp.wMZIQvX55e$ cd swagger-codegen/samples/client/petstore/android-java
me@box:/tmp/tmp.wMZIQvX55e$ mvn clean install
...
me@box:/tmp/tmp.wMZIQvX55e/swagger-codegen/samples/client/petstore/android-java$ for jar in ./target/lib/*.jar; do dexcount $jar; done | paste -s -d+ | bc
16881                         
me@box:/tmp/tmp.wMZIQvX55e/swagger-codegen/samples/client/petstore/android-java$ for jar in ./target/*.jar; do dexcount $jar; done | paste -s -d+ | bc
342
```

fehguy:
The main culprits will be JSON and HTTP support.  using gson should help considerably as we don't need all of Jackson's fancy methods for android.

who:
@fehguy Yes, swapping for gson would be good start.  The Jackson libs alone contribute 7-8k to the limit.

who:
@fehguy 

I have been looking into this. Here is what I propose: We should change the swagger-codegen templates to, wherever possible, use libraries contained within the [Android API bucket](http://developer.android.com/reference/packages.html).  If these libraries are used, they do not count against the method limit.

fehguy:
@who that's the right move.

0legg:
@fehguy @who Another reason to change android template is deprecation of built-in Apache HTTP client. I think implementing https://github.com/swagger-api/swagger-codegen/issues/465 will fix this issue as well as https://github.com/swagger-api/swagger-codegen/issues/686. I've got WIP candidate for this fix, but I'm not sure that all stuff will work as well as it supposed to be.

who:
@fehguy @0legg 

[...]

Issue #686, opened 2015-04-24
Use one client for Android and Java


karussell:
As noted in #684: please merge the client for Android and Java, e.g. via using [okhttp](https://github.com/square/okhttp/)

xhh:
:+1: okhttp seems to support PATCH, which would fix #682

xhh:
[Retrofit](https://square.github.io/retrofit/) looks nice too, it uses okhttp by default.
The ability of turning REST API into Java interfaces without the needs of implementation details would make the generated client pretty slim.

wing328:
Closing this as this is supported by the "okhttp-gson" template.

Issue #465, opened 2015-03-04
Add a Retrofit client generator


futurely:
[Retrofit](https://github.com/square/retrofit) is a very popular "type-safe REST client for Android and Java by Square, Inc".

Smartbear developed an example for GitHub API.

https://github.com/olensmar/github-retrofit-client/tree/master/src/main/java/com/olensmar/apis/github

They didn't use template. Instead, the generator is hard coded in Groovy. It should be possible to implement the same functionality with pure templates.

https://github.com/olensmar/soapui-retrofit-plugin/blob/e40722ebbb4c9c97017aa72c04d184e5cd65c4a6/src/main/groovy/com.smartbear.soapui.plugin.retrofit/RetrofitGenerator.groovy

webron:
Sounds like a great idea! Maybe @olensmar would like to contribute it? ;)

olensmar:
awesome - I'll definitely have a look!

just FYI - the plugin referred to above generates Retrofit interfaces for any REST API, i.e. the API doesn't have to be defined with a swagger spec

Panthro:
Any news on that guys?

I use swagger for all my apis, but like retrofit better for the client side.

It would be awesome to be able to generate the client automagically!

wing328:
The latest Java API client supports different HTTP client including Retrofit. To use Retrofit, please create a config.json file as follows:
```
{
  "library": "retrofit"
}
```
and pass config.json via -c in the command line

You can run `java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar config-help -l java` for more information on customizing the Java API client.

Would that meet your requirement?

fehguy:
Please reopen if you need help on this.  It's now supported as an option when generating java clients.

Panthro:
I just generated a cli based using the following command:


java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar  generate -i http://localhost:8080/v2/api-docs -l java -o myapi/client/java --library retrofit


It all went well and seemed to work, I explored the code and liked what I saw, with Oauth authentication flow and everything  in place, but when I tried a `mvn package` this happened:


[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /Users/panthro/swagger/swagger-codegen/myclient/client/java/src/main/java/io/swagger/client/api/BuildrequestresourceApi.java:[10,30] error: cannot find symbol
[...]

PR #706, opened 2015-04-30
[WIP] add template for Retrofit and Gson

0legg:
Java and Android templates holds a lot of boilerplate code and leads user to include heavyweight libraries. Moreover, Apache HttpClient is deprecated for Android and it's usage can lead user to unstable behavior.

Square Retrofit library allows to dramatically reduce source code size without growing app footprint. As a bonus, Gson is much lighter library than Jackson, while covers most of cases used in Swagger.

This PR can affect issues https://github.com/swagger-api/swagger-codegen/issues/686, https://github.com/swagger-api/swagger-codegen/issues/687, and definitely https://github.com/swagger-api/swagger-codegen/issues/465


0legg:
Current version breaks test "testLoginUser()". While being declared able to return "application/json" and "application/xml", endpoint "/user/login" returns plain string. This behavior breaks Gson parser, and I still haven't figured out elegant solution for this case. To my mind, it can be server issue (really, this is strange to return structured data all the way but plain string in rare cases).

webron:
@olensmar - feel like checking it out?

0legg:
@xirt4m Fixed in last commit, thank you

fehguy:
@olensmar  ping

olensmar:
sorry - digging in!

modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/languages/RetrofitClientCodegen.java
```diff
@@ -0,0 +1,190 @@
+package com.wordnik.swagger.codegen.languages;
+
+import com.wordnik.swagger.codegen.*;
+import com.wordnik.swagger.models.Operation;
+import com.wordnik.swagger.models.properties.*;
+
+import java.util.*;
+import java.io.File;
+
+public class RetrofitClientCodegen extends DefaultCodegen implements CodegenConfig {
```
olensmar: it seems to me you could derive from JavaClientCodegen instead and get rid of a lot of duplicate code!?

0legg: Well, I took AndroidClientCodegen as reference, and it doesn't derive from JavaClientCodegen. Default constructor adds lot of stuff to additionalProperties and supportingFiles, so any change in java codegen can break its subclasses. Yes, I can just clear these lists before setting my own values, but it can break some stuff tat will be done in future versions of DefaultCodegen

olensmar: Then perhaps we should have an AbstractJavaCodegen that all these derive from or put that shared functionality in some kind of helper (i.e. inheritance vs composition). If you don't have time for either please let me know so I can open a corresponding ticket after your PR has been merged. Thank you!

0legg: It's a good idea, I'll take a look at it in couple of minutes.

olensmar: great! let me know if I can help in any way...

0legg: I doubt how to include additional properties like invokerPackage, groupId, etc.. I think it should be specified in each codegen separately, but I want to avoid copy-pasted solution. 

Also, there are lot of methods that use class fields like sourceFolder which can't be overridden. I hope there'll not become an issue if someone wants to use structure different from "src/java/main/my/favorite/package", it's very rare case, and shouldn't be covered by codegen. Anyway, user can set this field in deriver constructor

olensmar: have a look at https://github.com/swagger-api/swagger-codegen/pull/805 - I think it provides the footwork for this?

0legg: Yes, so it'll be better to wait for changes from #805 to be applied (or not), then to make a decision about generalisation java clients

olensmar: makes sense - thank you!

olensmar:
Very cool - thank you!

cemo:
@0legg any chance for Retrofit 2?

0legg:
@cemo Will take a look at it slightly later (well, Retrofit 2 is still in beta, so I have enough time =) )

cemo:
@0legg I have actually done internal implementation based on your work. However It should be polished. :)

PR #711, opened 2015-05-01
Java: support x-www-form-urlencoded for PUT, POST, DELETE

sosso:
We use Swagger Codegen at FindTheBest for our internal API, and noticed issues with x-www-form-urlencoded requests in our Java libraries.  This patch should cover it; let me know if you see any issues.


webron:
@sosso - PRs need to be made against the develop_2.0 branch. Can you please resubmit?

PR #716, opened 2015-05-04
Fix/click on return type in dynamic html

### LINKED ISSUE #671: Return Type Href Not Actually Set
The link to the return type in the static pages is just left as `#`:
https://github.com/swagger-api/swagger-codegen/blob/618b4698c246a39824262e781ecd2b7f7a954261/modules/swagger-codegen/src/main/resources/swagger-static/operation.mustache#L16

ponelat:
Fixed certain functionality...When you click on the return type, a modal of the related model pops up.

fixes #671

### DIFF
--- a/modules/swagger-codegen/src/main/resources/swagger-static/assets/js/main.js
+++ b/modules/swagger-codegen/src/main/resources/swagger-static/assets/js/main.js
@@ -1,13 +1,8 @@
-function main() {}
-
 window.onhashchange = function() {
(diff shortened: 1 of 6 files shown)


wing328:
Tested and I'm able to view the model by clicking on the return type.
screen shot 2015-05-16 at 11 35 46 pm

One suggestion I've is to show whether the parameter is required or optional, e.g. "photoUrls : List (optional)"

PR #718, opened 2015-05-04
SpringMVC template is ignoring the required attribute in swagger definition when generating @RequestHeader annotation

swinz:
For SpringMVC, RequestHeader annotations are not setting the required attribute, even when required is false (RequestHeader uses default of true).  This change always sets the required param


webron:
@swinz - thanks, but PRs need to be made against the develop_2.0 branch. Can you please resubmit against it?

PR #720, opened 2015-05-04
Fixes a bug with Node.js example generating invalid package.json and swa...

akras14:
...gger.json

Currently string "... You can find out more about Swagger at <a href=\"http://swagger.io\">http://swagger.io</a> or ... " gets URL encoded, resuling in Node js throwing "Unable to parse" error.

Use of {{{ }}} forces mustache to skip URL encoding
http://mustache.github.io/mustache.5.html#Partials

### DIFF
--- a/modules/swagger-codegen/src/main/resources/nodejs/package.mustache
+++ b/modules/swagger-codegen/src/main/resources/nodejs/package.mustache
@@ -1,7 +1,7 @@
 {
   "name": "{{projectName}}",
   "version": "{{appVersion}}",
-  "description": "{{appDescription}}",
+  "description": "{{{appDescription}}}",
   "main": "index.js",
   "keywords": [
     "swagger"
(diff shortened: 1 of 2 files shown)


akras14:
To reproduce the bug: 

1. Clone clean repo
2. `mvn package`
3. `ava -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate   -i http://petstore.swagger.io/v2/swagger.json   -l nodejs   -o samples/server/petstore/nodejs`
4. `cd samples/server/petstore/nodejs/'
5. `npm install` will trigger the first error
6. Manually delete description from package.json
7. `npm install` will pass this time
8. `node index.js` will trigger the second error

fehguy:
To be clear, this potentially introduces some other issues, namely if there are double-quotes in the description it may cause a parse error.

PR #724, opened 2015-05-06
PHP Client Fixes

hkz-aarvesen:
Dear Swagger,

I have updated the php client generator.  I'm not personally a php guy, but have worked with our php guys to produce something that they are happy with.  I humbly submit this PR for your consideration.

Notable Functionality
================
- Modified Java code
  - The package is based on the hostname rather than being hard coded to `SwaggerPetstore`
  - Renamed `processOpts` to `preGenerate`
  - Renamed `processSwagger` to `postGenerate`
  - Passing the `swagger` object to both `pre-` and `postGenerate`
  - Added a setter method for `invokerPackage`
  - Numerous additional values depend on `invokerPackage`, so these were moved inside of the setter
- Modified mustache templates
  - Added code to avoid warnings when running php with `E_STRICT`
  - Added additional deserialization code to handle translating the JSON schema types into valid php types (e.g. "number" to "float")
  - Changes some tabs to spaces
- Modified php client layout
  - Renamed the autoloader file to just be `loader.php`.  This avoided the problem of having files with weird names like `yoyodyne\\db.php`
- Modified scala
  - Added a unit test for the hostname to package name method. This is literally the first piece of Scala I have written so YMMV.

### DIFF
--- a/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/CodegenConfig.java
+++ b/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/CodegenConfig.java
@@ -26,7 +26,7 @@ public interface CodegenConfig {
   String escapeReservedWord(String name);
   String getTypeDeclaration(Property p);
   String getTypeDeclaration(String name);
-  void processOpts();
+  void preGenerate(Swagger swagger);
   String generateExamplePath(String path, Operation operation);
 
   Set<String> reservedWords();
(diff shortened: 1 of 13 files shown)


webron:
@hkz-aarvesen - thanks for the PR, it looks like you've put a lot of work into it.

I can't comment on whether the PR will be merged (it would be reviewed by others), but we don't accept PRs against the `master` branch but rather the `develop_2.0` branch. Can you please resubmit against it?

hkz-aarvesen:
Certainly!

PR #731, opened 2015-05-07
csharp: Fixed handling of 2xx result codes != 200.

psmay:
Before this change, API calls returning successful result codes other than 200 were treated as failures. For example, one possible successful result code for a POST or PUT operation is 201 Created; this previously resulted in an exception but is now treated as a success.

A code is classified as a success in a similar way to how javax.ws.rs.core.Response (as of jersey-core 1.8.3) determines a result code's "Family".


webron:
@psmay - thanks, but PRs need to be submitted against the develop_2.0 branch. Can you please resubmit?

psmay:
Reworked into #740.

PR #732, opened 2015-05-07
Mime type fixes

Helmsdown:
Currently swagger-codegen does not honor the the global consumes and produces fields.

For example:
```yaml
consumes:
  - application/json;charset=utf-8
produces:
  - application/json;charset=utf-8
```

Results in:
```java
@Path("/somepath")
@Api(value = "/somepath", description = "some API")
public class PaymentsApi {
//operations here
}
```

With this change the following will occur:
```java
@Path("/somepath")
@Api(value = "/somepath", description = "some API")
@Consumes({ "application/json;charset=utf-8" })
@Produces({ "application/json;charset=utf-8" })
public class PaymentsApi {
//operations here
}
```


wing328:
@Helmsdown This addressed #757 as well. Thanks!

PR #747, opened 2015-05-15
A Parameter's default value is not flowing from the swagger def to the templates

### LINKED ISSUE #748: A Parameter's default value is not flowing from the swagger def to the templates
specifying this in a swagger yaml definition:

```
parameters:
  bucketSizeQueryParameter:
    name: bucketSize
    in: query
    description: Number of elements to be returned as a paged set
    default: 50
    required: false
    type: number
    format: double
```
results in the **default** value not being available to the templates. 

This change: #747 moves the value so that it is available, and uses the value to populate the RequestParam.defaultValue SpringMVC annotation

swinz:
specifying this in a swagger yaml definition:

```
parameters:
  bucketSizeQueryParameter:
    name: bucketSize
    in: query
    description: Number of elements to be returned as a paged set
    default: 50
    required: false
    type: number
    format: double
```
results in the **default** value not being available to the templates. 
This change moves the value so that it is available, and uses the value to populate the RequestParam.defaultValue SpringMVC annotation

Closes #748


modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/DefaultCodegen.java
```diff
@@ -904,6 +904,10 @@ public CodegenParameter fromParameter(Parameter param, Set<String> imports) {
       p.required = param.getRequired();
     p.jsonSchema = Json.pretty(param);
 
+    if(param instanceof QueryParameter) {
+        p.defaultValue = ((QueryParameter)param).getDefaultValue();
+    }
+
```
wing328: Is it correct to say that only the default value of query parameters (but not form parameters nor header parameters) is supported ?

fehguy: The intent is that it should be in all param types.  If not, that's an unimplemented feature.

swinz: @wing328 Yes you are correct, this PR only currently covers query params (not form or header).

wing328: would be nice if this PR can be extended to cover form and header parameters as well.

swinz: @wing328 it's now moving over default values for forms and headers, as requested.  I updated the corresponding SpringMVC templates as well, since I am using those and could test them.  (Note: Spring RequestPart, which supports Form elements, does not support default values)

wing328: Thanks @swinz! 

@fehguy do you now feel more comfortable merging this PR?

thinw002: What about in the Models? I get defaults for some types (boolean) but not for others (integer).

PR #712, opened 2015-05-03
Java: support x-www-form-urlencoded for PUT, POST, DELETE 

sosso:
We use Swagger Codegen at FindTheBest for our internal API, and noticed issues with x-www-form-urlencoded requests in our Java libraries. This patch should cover it; let me know if you see any issues.

@webron This PR isn't as clean as the one I (accidentally) made to master; do you want me to make a commit to a freshly-pulled develop_2.0?

PR #758, opened 2015-05-19
Fix test of updatePetWithForm for Android Petstore sample

xhh:


xhh:
btw, Java Petstore sample would need to be fixed as well after form parameter support is added for POST (see #712)

who:
:+1:

PR #754, opened 2015-05-18
[WIP] Support Swift code generation

tkQubo:
### TODO

- [ ] file upload
- [ ] post form data
- [ ] post array
- [ ] test code

### Dependent libraries

- [PromiseKit](https://github.com/mxcl/PromiseKit)
- [Alamofire](https://github.com/Alamofire/Alamofire)

PR #773, opened 2015-05-21
Adds isPrimitiveType flag to CodegenParameter.

chakrit:
This is required in some languages like Obj-C where there is no automatic value boxing and
handling of primitive types require extra work compared to non-primitive types.

Case in point: `BOOL` type. To assign `BOOL` into a dictionary, one needs to box it into an
`NSValue *` instance, so to build a dictionary for sending form data, for example, you
cannot do this in the template:

    dict["{{paramName}}"] = {{paramName}};

Because if the parameter ends up being of type `BOOL`, an error about boxing values will be
generated:

    BOOL boolValue = NO;
    dict["boolValue"] = boolValue;
                        ^---------------- Cannot do the assignment here.

The fix is to wrap it in `@()` like so:

    BOOL boolValue = NO;
    dict["boolValue"] = @(boolValue);

So a flag is needed in `CodegenParameter` so we can selectively emit the right boxing or
non-boxing assignment in the templates where needed.

### DIFF
--- a/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/CodegenParameter.java
+++ b/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/CodegenParameter.java
@@ -2,7 +2,8 @@ package com.wordnik.swagger.codegen;
 
 public class CodegenParameter {
   public Boolean isFormParam, isQueryParam, isPathParam, isHeaderParam,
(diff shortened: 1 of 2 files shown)


chakrit:
We have a copy of our own templates (modified from the work in #754) at the moment. I will try to add fixes to the master branch template files where applicable if I have the time.

wing328:
@chakrit please kindly submit the PR for the branch `develop_2.0` instead of `master`

Issue #730, opened 2015-05-07
CSharp code generator - http return code 201 causes an exception


boazsapir:
In the CSharp generated code, when the api call returns 201 an exception is thrown although this code does not indicate any error or problem

wing328:
Is it correct to say that the issue also occurs in the "csharp_restcsharp" PR #700 ?

boazsapir:
Yes, this is the generated code:
              if (webResponse.StatusCode != HttpStatusCode.OK)
              {
                  webResponse.Close();
                  throw new ApiException((int)webResponse.StatusCode, webResponse.StatusDescription);
              }

psmay:
See #731 for a tiny fix.

psmay:
#700 seems to remove/replace the broken code also.

wing328:
@boazsapir In the generated C# code (csharp_restsharp branch), I couldn't find any reference to `HttpStatusCode`. 

```
csharp|csharp_restcsharp ⇒ ls
bin         compile.bat src
csharp|csharp_restcsharp ⇒ grep -R -i HttpStatusCode *
csharp|csharp_restcsharp ⇒ 
```

psmay:
#731 reworked to #740. Again, this patch is not necessary if #700 / RestSharp is used instead.

boazsapir:
@wing328 - sorry for the confusion, I see now that the file ApiInvoker.cs (which had been originally created by the master code) has not been re-created when I used the csharp_restsharp code for generation, therefore the HttpStatusCode problem remained.
The reason for this is an exception that was thrown during the code generation. Looks like property types are checked against a list of reserved words,  and this is bound to fail since the list contains the words 'object', 'string', etc. (this check does not happen in the master branch since getSwaggerType() does not call toModelName(), therefore I did not open a separate issue for this). 

java.lang.RuntimeException: object (reserved word) cannot be used as a model name
	at com.wordnik.swagger.codegen.languages.CSharpClientCodegen.toModelName(CSharpClientCodegen.java:133)
	at com.wordnik.swagger.codegen.languages.CSharpClientCodegen.getSwaggerType(CSharpClientCodegen.java:174)
	at com.wordnik.swagger.codegen.DefaultCodegen.fromProperty(DefaultCodegen.java:554)
	at com.wordnik.swagger.codegen.DefaultCodegen.fromResponse(DefaultCodegen.java:866)
[...]

PR #775, opened 2015-05-22
[C#] Fix "object' type for response

wing328:
To address #730 

Tested with type set to `object` with and without $ref

### DIFF
--- a/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/languages/CSharpClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/com/wordnik/swagger/codegen/languages/CSharpClientCodegen.java
@@ -83,6 +83,7 @@ public class CSharpClientCodegen extends DefaultCodegen implements CodegenConfig
     typeMapping.put("file", "string"); // path to file
     typeMapping.put("array", "List");
     typeMapping.put("map", "Dictionary");
+    typeMapping.put("object", "Object");
 
   }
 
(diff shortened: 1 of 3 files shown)


boazsapir:
@wing328 I am trying to retest but when I use the generation code from branch develop_2.0, and  generate my CSharp code,  the following lines in do not compile:

        request.AddQueryParameter(param.Key, param.Value);
      return (Object) await restClient.ExecuteTaskAsync(request);

Looks like a mismatch with my RestSharp.dll version. I took the dll from here:
swagger-codegen/modules/swagger-codegen/src/main/resources/csharp

Which dll version should I use?

wing328:
@boazsapir do you have NuGet installed ? If yes, please get it via https://www.nuget.org/packages/RestSharp

boazsapir:
ok I will install it

wing328:
https://github.com/restsharp/RestSharp/releases/tag/105.1.0

The latest release does not contain the updated DLL. I'll follow up with the owner of RestSharp. For the time being, please use NuGet.

wing328:
@boazsapir  I've filed a PR to update the DLL to the latest version (obtained from NuGet).

Please still use NuGet to obtain the latest version of RestSharp as NuGet would get the DLL of RestSharp for different platforms/frameworks.

boazsapir:
ok, I managed to retest. The case of inline object works ok.

wing328:
@boazsapir thanks for performing the test.

Issue #779, opened 2015-05-23
Why is there no accessor to the members of the generators and the constructor define properties ?


rastaman:
Hi all,
I'm in the process of learning swagger-codegen and i'm surprised that there is no accessors on the members of the generators classes and also that some properties are defined in the constructors. 

In order to configure the generators from the swagger-codegen-maven-plugin, i had to add the accessors ( getter/setter) and move the construction of the properties (additionalProperties, supportingFiles) in a init method which is called from the construtor and the setters. 

I can submit this patch in a pull request but before i do so, i would like to know if i have missed something ? it is strange that the accessors are missing, perhaps there is a good reason ?

I also have patched the maven plugin to use reflection to configure the generators, and all the code is in the forks of the repos (swagger-codegen and swagger-codegen-maven-plugin).

Thanks in advance for any answer,
Best regards

fehguy:
@rastaman you can add accessors.  The lack of them is due to a previous mustache template engine that we were using--there's no reason not to have them.  It will have wide-spread impact across the code, though, so I suggest introducing the getters/setters and and keep the variables public for now (not changing all the places that touch them).  There are a number of PRs that should get merged first before making that sort of change, which will break all of them.

wing328:
@rastaman do you need help adding the accessors?

Please pull the latest and submit the PR against the `master`.

rastaman:
Hi wing328,
I'm ok to add the accessors again on the master but before i do so i would like to know if there is a better way than mine to refresh the value of the computed properties ? i mean if you look at this commit on my fork : https://github.com/rastaman/swagger-codegen/commit/315452628f7e238159ac4841e7ee15b302c6f4bd you can see that i had to add a "init()" method to refresh the values of some computed properties, did you think it is the right way to do it ? (look at AsyncScalaClientCodegen.java for instance)

xhh:
I think the calls to `init` could be removed by moving the `init` code to [`processOpts`](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L89) or [`preprocessSwagger `](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java#L167) (if you need to access the `swagger` object before it is processed).

rastaman:
[...]

PR #784, opened 2015-05-24
#779  Why is there no accessors...

rastaman:
Hi, here is a pull request to address the issue #779 : 
- the patch add accessors for fields of the all CodeGen classes. 
- Also it move the code wich was setting additionnalProperties and supportingFiles in the constructors to a init() method. 

The init() method, where present/needed, is called in the constructors but also in the setters of the members. So the files or properties which are computed from the value of a field can be refreshed.

The defintion and visibility of the fields of of the classes have not changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,11 +5,6 @@
 ## Overview
 This is the swagger codegen project, which allows generation of client libraries automatically from a Swagger-compliant server.  
 
-## What's Swagger?
-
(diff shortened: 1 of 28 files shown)


rastaman:
I'm wondering if i have not too much messed with the commit history... :-( let me know if it's a problem.

wing328:
I believe you need to submit the PR against the `develop_2.0` branch instead of master. Please have a look at other PRs for reference.

rastaman:
Thanks for the hint!

rastaman:
I will rework this patch for the develop_2.0 branch and submit it again, thanks for your patience!

Issue #786, opened 2015-05-25
[Suggestion] Standardise naming, exception and more


wing328:
Hi all,

I would like to propose standardising the following across all API client libraries:

- name the HTTP client as ApiClient instead of ApiInvoker, SwaggerClient, etc (For ObjC, we'll use SWGApiClient instead)
- name exception as ApiException instead of APIClientException, ErrorResponse, etc
- use a Configuration class for configuration (authentication, logging, etc)
- throw ApiException instead of returning null data for 404

This will make the API client more consistent across different languages and easier to maintain.

If you've any feedback or concern for the above, please share with us.

Thanks.

fehguy:
This is a good idea as long as the conventions are consistent with each language vernacular.

fehguy:
@wing328  can we close this?  I believe we've covered this need.

wing328:
Yes, most clients are using the standardised naming, exception, etc

PR #791, opened 2015-05-26
[Java] Make API client more pluggable

xhh:
Make ApiClient (ApiInvoker before) more pluggable for Java clients, similar to what has been done to PHP and Python clients (#778, #787), and follow suggestions from #786 on naming.

- Rename ApiInvoker to ApiClient
- Make ApiClient pluggable by allowing setting the ApiClient field of
  API classes
- Introduce a Configuration class, containing the default
  ApiClient (which is also customizable)
- Move basePath from API class to ApiClient
- Change static methods in ApiClient to instance level


who:
:+1:

olensmar:
please rebase so we can merge...

xhh:
@olensmar it is done

PR #801, opened 2015-05-27
Removing comment (// do some magic!) that is no longer necessary and …

rlo2014:
…confusing.  After olensmar refactor of jax-rs codegen, magic/implementation should be placed in the delegate/implementation as opposed to the api itself.  Any logic put in api itself will be overwritten with the next code gen, whereas the delegate implemenation will remain intact.


olensmar:
Good catch - thank you!

Issue #800, opened 2015-05-27
ObjectiveC generation


uniprof:
generation ObjectiveC code return this error

can't open objc\SWGObject.h for input
can't open objc\SWGObject.m for input
can't open objc\SWGQueryParamCollection.h for input
can't open objc\SWGQueryParamCollection.m for input
can't open objc\SWGApiClient.h for input
can't open objc\SWGApiClient.m for input
can't open objc\SWGFile.h for input
can't open objc\SWGFile.m for input
can't open objc\JSONValueTransformer+ISO8601.m for input
can't open objc\JSONValueTransformer+ISO8601.h for input

wing328:
Are you using `developer_2.0`? I did a test with `mvn clean && ./bin/objc-petstore.sh` and it works for me.

uniprof:
Have you tested it on Windows? on Mac it work fine, on Windows not.

wing328:
I've not

wing328:
I submitted a PR to fix the issue #803 

Please kindly perform a test (in Windows) to see if it fixes the issue.

fehguy:
This should be fixed now.

PR #803, opened 2015-05-28
[ObjC] Minor fix to output path

wing328:
Minor fix to output path using `File.separator` #800


uniprof:
Yes it is working now.

fehguy:
probably needs to be applied across all codegen

wing328:
@fehguy yes, will review and submit a PR later if the issue is not yet fixed in other codegen

PR #738, opened 2015-05-08
Fixes #429: Handling of composed models has been added.

tomtit:
Fixes #429: Handling of composed models has been added.

Issue #429, opened 2015-02-11
ClassCastException ComposedModel cannot be cast to ModelImpl trying to generate html from spec with $ref in definitions


DavidBiesack:
In branch develop_2.0 I was able to do a 'git pull' and got a clean 'mvn package' build.

First, some minor problem:

I followed the README.md instructions 
```
java -cp ./target/*:./target/lib/* com.wordnik.swagger.codegen.Codegen \
  -i http://petstore.swagger.wordnik.com/v2/swagger.json \
  -l java \
  -o samples/client/petstore/java
```
but had to change the classpath because Maven puts the jars in `modules/swagger-codegen/target` and `modules/swagger-codegen/target/lib` not in `target` and `target/lib` (please fix README.md)

./bin/html.sh does not work because the URL  http://petstore.swagger.wordnik.com/v2/swagger.json does not exist (that redirects to  http://petstore.swagger.io ). I think you want **http://petstore.swagger.io/v2/swagger.json** instead

I changed html.sh and now it works.

I then substituted one of our internal Swagger 2.0 documents and got an exception:
```
./bin/html.sh
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
reading from http://swagger.na.sas.com/swagger/public/api/v1/swagger.json
no format specified for integer type, falling back to int32
java.lang.ClassCastException: com.wordnik.swagger.models.ComposedModel cannot be cast to com.wordnik.swagger.models.ModelImpl
	at com.wordnik.swagger.codegen.DefaultCodegen.fromModel(DefaultCodegen.java:361)
	at com.wordnik.swagger.codegen.DefaultGenerator.processModels(DefaultGenerator.java:348)
	at com.wordnik.swagger.codegen.DefaultGenerator.generate(DefaultGenerator.java:80)
	at com.wordnik.swagger.codegen.Codegen.main(Codegen.java:91)
```

The Swagger doc uses $ref in the schema definitions as shown in the sample at https://github.com/swagger-api/swagger-ui/issues/723  Perhaps this schema $ref is what results in a ComposedModel (petstore's schema do not use allOf)

This is possibly related to https://github.com/swagger-api/swagger-js/issues/188 although that bug appears tied to JavaScript (I don't know if codegen uses swagger-js)

fehguy:
[...]

PR #804, opened 2015-05-28
Add support for handling ComposedModel 

mhardorf:
If the swagger spec file contains allOf model composition the codegenerator would fail with a class cast exception. This commit fixes the issue.


webron:
@mhardorf - can't comment on the content of the PR, but in general, we only accept PRs against the develop_2.0 branch as it is our development branch. Please resubmit it against that one.

tomtit:
Looks like the same issue as #429, that was fixed by #738 PR at the `develop_2.0` branch.

PR #809, opened 2015-05-29
[C#] Asynchronous support

Hadou1:
- Extended api.mustache for CSharp to support asynchronous calls aswell. Every generated method is now generated with a async version


wing328:
One minor feedback is to add XML documentation to interface methods

ref: https://msdn.microsoft.com/en-us/library/vstudio/z04awywx%28v=vs.100%29.aspx

fehguy:
This is great--however I'm not sure we should just replace the existing c# library with an async one.  How about a new csharp-async template set?

wing328:
@fehguy from what I understand, the PR will add an additional method (while keeping the old/existing one). For example, a method named "GetPetByIdAsync" will be added by this change while the GetPetById method will still be available.

Last time I tested with the non-async methods (GetPetById, UploadFile, UpdatePetWithStatus, etc) and they're working fine.

Hadou1:
Hi guys,

what i have done is to add a interface for the API classes and as @wing328 wrote an additional method wich is implemented Async. I would definitely integrate this in the existent templates, hence this is normal pattern for me. 

The reason of adding interfaces is to provide tha ability to work with dependency injection frameworks, eg. Autofac or Ninject

fehguy:
got it @Hadou1 @wing328.  We had some template drift, but if you can help merge the logic back I will merge the PR.

Hadou1:
OK... will do

Hadou1:
@fehguy can you please tell me how to proceed in order to complete this. I'm new to this and i'm not sure how to get the changes the best way. I forked the repo and i would now like to get those latest changes. Do i have to fork again or can i somehow pull the latest changes into my repo?

wing328:
@Hadou1 here is what I would recommend:
- git checkout develop_2.0
- git pull upstream (upstream is the official repo of swagger-codegen :https://github.com/swagger-api/swagger-codegen.git
- git checkout your_local_branch
- git rebase develop_2.0 and resolve the conflicts

Hadou1:
@wing328 Thanks for your reply. I also found that command... but when i try executing this i always get exception like this: 

$ git pull upstream
fatal: 'upstream' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

wing328:
You need to define an "upstream" (remote) first. Please refer to https://help.github.com/articles/configuring-a-remote-for-a-fork/ for more info.

Hadou1:
@wing328  Thanks.. got it

Hadou1:
@fehguy @wing328 
I now rebased the changes from develop_2.0 and fixed the conflicts. Please review/Feedback and merge...

wing328:
@Hadou1 Thanks for the quick turnaround

PR #839, opened 2015-06-07
[Perl] update deserialization to support Map and List

wing328:
- updated `deserialize` to support Map (hash) and List (array)
- added test case for StoreApi's `get_inventory`
- updated pom.xml with new tests


webron:
@wing328 - not sure when it'll be merged, but this needs to be rebased.

wing328:
@webron thanks for the reminder. i'm in the process of creating a new one instead. will close this one

PR #852, opened 2015-06-09
Fixed #837: Generation of default values for complex arrays and maps has been improved.

### LINKED ISSUE #837: Wrong generation of Map of List in java
The following definition:
```
      foo:
        type: object
        additionalProperties:
          type: array
          items:
            $ref: "#/definitions/bar"
```
will generate in java:
```java
private Map<String, List<Bar>> foo = new HashMap<String, List>() ;
```
This doesn't compile. The correct code generated should be:
```java
private Map<String, List<Bar>> foo = new HashMap<String, List<Bar>>() ;
```

tomtit:
Fixed #837: Generation of default values for complex arrays and maps has been improved.


wing328:
Tested and confirmed no longer seeing the error with the default value.

Issue #782, opened 2015-05-23
Missing RestSharp and Newton.Json files


uniprof:
the compile.bat return the following error:

src\main\csharp\IO\Swagger\Api\ApplicationsApi.cs(3,7): error CS0246: The type or namespace name 'RestSharp' could not be found (are you missing a using
        directive or an assembly reference?)

wing328:
@uniprof can you please try using NuGet to import RestSharp and Json.NET ? Let me know if you need more detail using NuGet.

uniprof:
I know how to use nuget in Visual Studio but not from the command line, I am trying to archive a server side compilation and publishing of the DLL ready for downloads. Can you help?

wing328:
You can find 2 ways to install it (command line or DLL) from the SO link below:

http://stackoverflow.com/questions/10630685/how-do-i-install-and-setup-restsharp-without-nuget

Hope it helps.

uniprof:
Thanks @wing328.
I know how to use NuGet but I don't know how to use it in the workflow of your generation flow.
The goal would be to do all at command line (batch file) without any interactions with the folders.

Using Nuget will download the folders somewhere... but then your compile.bat must find it in some particular position???

Will be good to have any kind of suggestion.

Thx

wing328:
To be clear, I was not the one creating the compile.bat but I will see what I can do to fix it or replacing it with something better.

I wonder if you can tell me more about the server side compilation. Are you using a batch file (*.bat) to compile the server side project ? Does the server side project has any other dependency (DLL) ?

Here is SO link about the location of the installed package:
http://stackoverflow.com/questions/7676640/how-do-i-specify-the-directory-where-nuget-packages-are-installed

uniprof:
@wing328 Thanks to be clear!
My server side compilation will use the Swagger codegen to query a swagger doc deployed with a .NET Web Api and to generate the client side SDK's in JAVA, ObjectiveC and C#.

My goal is to do it at runtime as the API's will change frequently in the development process and will be useful for the client's developers to have always an updated SDK for their mobile's developments.

My idea is to use the compile.bat file to make a full complete compilation process, so the following steps:

1. Get the Json from the /docs of Swagger
2. copy/paste or download the required dependency
3. compile the code in DLL, JAR or ObjectiveC library.

Hope this help you to understand better the scenario.

wing328:
I wonder if you can try the following:

[...]

PR #856, opened 2015-06-10
[C#] upgrade restsharp to 105.1.0

wing328:
The latest [RestSharp release](https://github.com/restsharp/RestSharp/releases/tag/105.1.0) does not contain the updated DLL. 

This PR uses the DLL (v105.1.0) downloaded via NuGet to replace the outdated RestSharp.dll.


who:
:+1: to merge this

By the way, @wing328, instead of including the DLLs directly in the repo, why not use NuGet?

wing328:
@who using NuGet is what I would recommend to obtain the lib for Newtonsoft and RestSharp. 

Please see https://github.com/swagger-api/swagger-codegen/issues/782 about the requirement to compile the C# SDK with the batch file (which was created before I started working on this project). I think it's still technically possible to use NuGet inside the batch file to download the libraries required for compilation but I don't have cycle to do that.

PR #859, opened 2015-06-11
Fixed #831: skipOverwrite flag has been added

### LINKED ISSUE #831: add "don't overwrite" support option
When generating files, we may not want to overwrite certain files that can be modified after generation.  Suggest adding an optional flag to "skipOverwrite" where the default is false (i.e. always overwrite).

lugaru1234:
Fixed #831: skipOverwrite flag has been added


wing328:
Tested and worked as expected.

modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
```diff
@@ -1294,6 +1295,14 @@ public String apiFilename(String templateName, String tag) {
     }
 
     public boolean shouldOverwrite(String filename) {
-        return true;
+        return !(skipOverwrite && new File(filename).exists());
```
ahgittin: this change in semantics from "should overwrite" to "should write" breaks `JaxRSServerCodegen` (and doesn't correspond with the method name); addressed in https://github.com/ahgittin/swagger-codegen/commit/8fadce27d7208564a16bad98175e210c19648a00

PR #862, opened 2015-06-12
[Java] Replace jersey-client and jackson with okhttp and gson in Java generator

xhh:
Compared with apache-httpclient (used by Android generator), jersey-client and jackson (used by Java generator), both [okhttp](https://github.com/square/okhttp/wiki/Recipes) and [gson](https://github.com/google/gson) have cleaner API and documentation, and are more friendly to Android projects (see #686).

With this change, the code generated by the Java generator is able to be used nicely in both Java (JVM) and Android environments, so it's possible to maintain only one template for several targets: Java, Android, as well as other JVM targeted languages like Scala and Groovy.
The petstore sample generated with this PR has been tested in plain JVM, Android and Scala and worked well.

Regarding Android APK files, the generated jar contains smaller method count:
with apache-httpclient and gson: [9123](https://gist.github.com/xhh/ec0396cabbd0fa2c7f65#file-apache-httpclient_gson-txt)
with okhttp and gson: [5302](https://gist.github.com/xhh/ec0396cabbd0fa2c7f65#file-okhttp_gson-txt)

**UPDATE**: added support for async HTTP call, for example:

```java
PetApi api = new PetApi();

// sync call
Pet pet = api.getPetById(new Long(1));
System.out.println(pet);

// async call
api.getPetByIdAsync(new Long(1), new ApiCallback<Pet>() {
    @Override
    public void onFailure(ApiException e) {
        System.out.println("Failure!");
        System.out.println(e.getCode());
        System.out.println(e.getMessage());
        System.out.println(e.getResponseBody());
        e.printStackTrace();
    }

    @Override
    public void onSuccess(Pet pet) {
        System.out.println("Success!");
        System.out.println(pet);
    }
});
```


webron:
That raises a bigger issue though. Do we really want to consolidate the templates? Personally, I'd never use Gson in Java over Jackson, and there are other implementations as well. People will always have a preference of one over the other. Same thing applies for any of the libraries used though (jax rs client, http components, retrofit, and so on). Not really sure what would be the right solution here. Abstraction could work, but may make it overall more complicated.

wing328:
@webron thanks for the feedback. In this case (Java/Android/Scala SDK), the use of Gson should be transparent to the developers using the SDKs since the returned JSON string is deserialized into a model. In other words, the developers do not need any knowledge with Gson, Jackson or other JSON library in order to consume the SDKs.

webron:
That's partially true, and depends on how the 'sdk' is used by the code (for example, if the models are reused). 

In any case, it forces the application to include the Gson dependency even if they prefer using Jackson in general. Granted, we do just the same with swagger-core and Jackson (which to me seems like a discrepancy between the libraries).

I don't know what's the right answer here, but this is something we need to consider. If I understand correctly, this affects only the generated Java *client* which may be fine, but I'd rather we end up with a sound reasoning here. Would hate having to revert it later because we hadn't thought of something.

PR #865, opened 2015-06-12
Improve PHP namespace support

ammmze:
This pull request updates the codegen to better support php namespacing. Previously, if I were to use a namespace such as `Swagger\Client`, the composer file became invalid, deserialization would also be incorrect, etc. This brings proper support the multiple layer namespacing.

Also namespace format was inconsistent...the vender was `StudlyCaps` and the Model space is lower case. This brings it to be more consistent...the Models are now in `Swagger\Client\Model` and the api's are in `Swagger\Client\Api`.

In addition, I've also added phpdoc block comments for the properties and methods following the [phpdoc types](http://phpdoc.org/docs/latest/guides/types.html) which allows for IDE code/type hinting to function properly.


wing328:
The integration fails:
```
SwaggerClient-php|feature/php-namespace-and-doc ⇒ ./vendor/bin/phpunit tests 

Warning: require_once(SwaggerClient.php): failed to open stream: No such file or directory in /private/tmp/swagger-codegen/samples/client/petstore/php/SwaggerClient-php/tests/PetApiTest.php on line 3

Fatal error: require_once(): Failed opening required 'SwaggerClient.php' (include_path='/private/tmp/swagger-codegen/samples/client/petstore/php/SwaggerClient-php/vendor/phpunit/php-text-template:.:') in /private/tmp/swagger-codegen/samples/client/petstore/php/SwaggerClient-php/tests/PetApiTest.php on line 3
```

Do you mind fixing it? I believe you need to update the test files with autoload.php instead of SwaggerClient.php 

To run the integration test, please run the following under `samples/client/petstore/php/SwaggerClient-php`:
```
mvn integration-test -rf :PhpPetstoreClientTests
```

ammmze:
Ah...yes...wasn't sure what all tests existed. I will fix those.

ammmze:
They pass now...however there were 2 tests that were failing that appear to be unrelated.

```
There were 2 errors:

1) PetApiTest::testApiClient
Non-static method Swagger\Client\ApiClient::addDefaultHeader() should not be called statically, assuming $this from incompatible context

swagger-codegen/samples/client/petstore/php/SwaggerClient-php/tests/PetApiTest.php:51

2) PetApiTest::testGetInventory
Undefined index: sold

swagger-codegen/samples/client/petstore/php/SwaggerClient-php/tests/PetApiTest.php:206
```

The first one comes from the methods relating to the default headers in the ApiClient are used in a static context but did not have the static keyword. Which I believe prior to php 5.4 it only printed a notice or a warning. In php 5.4 they made it more strict.

The second one, i'm not sure how it would have passed previously. The response is a map[string,int] and sold is not one of the keys in the response.

wing328:
For 1, it should be addressed by #857.

For 2, I've performed a quick fix is to add a pet with "sold" status. Please check again. 
A long term solution is to create a pet with "sold" status in the setupBeforeClass

ammmze:
Gotcha...should I revert my changes to resolve the unrelated test failures?

FYI...I fixed the static warning by marking the methods as static. I
assumed there was a reason they were originally intended to be static?

wing328:
I believe converting the method from non-static to static is the right direction for us. This would allow developers to easily extend the ApiClient class and customize (override) the methods based on their requirement.

ammmze:
Sounds good...I'll revert my changes for the unrelated issues when I get
back to the computer.

wing328:
@ammmze thanks! Please take your time and once again thanks for this PR.

ammmze:
Done :)

wing328:
@ammmze please kindly do a rebase on develop_2.0 as my PHP warning PR has been merged. Thanks!

ammmze:
@wing328 Sure thing ... all done!

PR #867, opened 2015-06-13
[Bug fix] Fix path inside JAR for non-template files

wing328:
Template files (*.mustache) uses `getCPResourcePath` to get the correct path in all platforms.

This PR also leverages `getCPResourcePath` for non-template files to address the file not found issue in Windows as Path inside JAR should use backslash.

The fix has been tested in Windows and Mac.


uniprof:
It work a charm now !!!

wing328:
@uniprof thanks for the confirmation. Sorry for taking this long as the bug is quite tricky in my opinion.

who:
:+1: to merge this

Issue #484, opened 2015-03-10
Array of string values as query parameter is not generated correctly in Java client code


kcao3:
Hello,

It seems to me the swagger-codegen version 2.1.2-M1 currently does NOT generate Java client code properly when dealing with a query parameter as an array of string values. For example, when I specified an optional query parameter as the following:
```js
{
     "name": "callerIds",
    "in": "query",
    "description": "A list contains the unique IDs the caller could pass in.",
    "required": false,
    "type": "array",
    "items": {
        "type": "string"
    },
    "collectionFormat": "multi"
}
```

I expected the tool to generate a query string in this specific this format: callerIds=abc&callerIds=xyz&callerIds=lmn to represent a list callerIds with string values("abc", "xyz", "lmn"). However, currently swagger-codegen generates the query string in this exact format:  callerIds=[abc,xyz,lmn].
I believe this is not the correct format to present an array of string value as a query parameter.

The following shows the a portion of the Java generated code from my swagger specification document: https://raw.githubusercontent.com/kcao3/test/master/MyResource.json using the swagger-codegen tool from the swagger-codegen-develop_2.0 branch.
```java
public class TopologyServiceApi {

public GetMetricsResponse getMetrics (List<String> callerIds, String sourceNodeId, String destinationNodeId) throws ApiException {
    Object postBody = null;
    ....
    
    // query params
    Map<String, String> queryParams = new HashMap<String, String>();
    
    .... 
    
    if(!"null".equals(String.valueOf(callerIds)))
      queryParams.put("callerIds", String.valueOf(callerIds));
    ...
    
    String response = apiInvoker.invokeAPI(basePath, path, "GET", queryParams, postBody, headerParams, formParams, contentType);
        ...
}
}    
```
```java
public class ApiInvoker {

    public String invokeAPI(String host, String path, String method, Map<String, String> queryParams, Object body, Map<String, String> headerParams, Map<String, String> formParams, String contentType) throws ApiException {
    ...
    
    StringBuilder b = new StringBuilder();

    for(String key : queryParams.keySet()) {
      String value = queryParams.get(key);
      if (value != null){
        if(b.toString().length() == 0)
          b.append("?");
        else
          b.append("&");
        b.append(escapeString(key)).append("=").append(escapeString(value));
      }
    }
    String querystring = b.toString();
    
    ...
    }   
}
```

sudohippie:
@webron I have fixed Java and Android templates to address this issue. Will you accept a PR?

[...]

PR #880, opened 2015-06-16
Android & Java Template Modifications to address List of Query Parameters Issue

sudohippie:
This is to solve the issue #484. Changes are for Android and Java language.

Made the following changes:

* Added a new template called QueryParam.mustache 
* Changed queryParams data structure from Map<String, String> to Set<QueryParam>
* Method signature changes to address new data structure.


sudohippie:
@who, @webron - I ran the integration tests, and noticed some issues. Modified the sample clients and their unit tests to reflect the changes to the templates.

The tests I ran were:

```
./bin/java-petstore.sh;
./bin/android-java-petstore.sh;
mvn integration-test -Pjava-client;
mvn integration-test -Pandroid-java-client;
```

xhh:
I'm negative on this PR. It adds support of `multi` `collectionFormat` but removes support of `csv` format.

The real problem here is that the current generator does not (fully) support the [`collectionFormat`](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#user-content-parameterCollectionFormat) for parameters, we should address it by supporting all of them.

sudohippie:
@xhh - What you say makes sense. I'll take a look in to codegen to make it extendable.

sudohippie:
Closing this PR. Look at #893.

Issue #890, opened 2015-06-18
[Java] Java ApiClient is using a deprecated Jersey method


who:
ApiClient is using a deprecated method, getClientResponseStatus().

See: https://jersey.java.net/apidocs/1.18/jersey/com/sun/jersey/api/client/ClientResponse.html#getClientResponseStatus()

wing328:
PR merged. Closing this issue.

PR #891, opened 2015-06-18
Replacing getClientResponseStatus() with getStatusInfo()

who:
This fixes https://github.com/swagger-api/swagger-codegen/issues/890

Integration tests have been run, and are passing.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/ApiClient.mustache
@@ -421,10 +421,10 @@ public class ApiClient {
       throw new ApiException(500, "unknown method type " + method);
     }
 
-    if(response.getClientResponseStatus() == ClientResponse.Status.NO_CONTENT) {
+    if(response.getStatusInfo() == ClientResponse.Status.NO_CONTENT) {
       return null;
     }
-    else if(response.getClientResponseStatus().getFamily() == Family.SUCCESSFUL) {
+    else if(response.getStatusInfo().getFamily() == Family.SUCCESSFUL) {
       if(response.hasEntity()) {
         return (String) response.getEntity(String.class);
       }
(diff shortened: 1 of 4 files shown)


wing328:
@who thanks! The change looks good to me.

I found that Scala is also using `getClientResponseStatus()`. Should we update that with `getStatusInfo()` as well ?

https://github.com/swagger-api/swagger-codegen/blob/eafcc432951e717a06c2bfbf7532ce979d9e611e/modules/swagger-codegen/src/main/resources/scala/apiInvoker.mustache#L144

who:
@wing328 I'm not sure - I don't have a scala environment set up to be able make changes to the scala templates with confidence.

wing328:
@who no problem. I'll take a look later on that.

modules/swagger-codegen/src/main/resources/Java/ApiClient.mustache
```diff
@@ -421,10 +421,10 @@ public class ApiClient {
       throw new ApiException(500, "unknown method type " + method);
     }
 
-    if(response.getClientResponseStatus() == ClientResponse.Status.NO_CONTENT) {
+    if(response.getStatusInfo() == ClientResponse.Status.NO_CONTENT) {
```
olivervidovic: This is incorrect. The ```response.getStatusInfo()``` returns an object implementing ```Response.StatusType``` interface, and not a ```ClientResponse.Status``` Enum, so the comparison will always fail. Because ```ClientResponse.Status``` also implements ```Response.StatusType``` interface, the correct fix is  ```response.getStatusInfo().getStatusCode() == ClientResponse.Status.NO_CONTENT.getStatusCode()```, which will compare the numeric values from the objects.

olivervidovic: https://github.com/swagger-api/swagger-codegen/pull/3963

The integration tests from Maven do not cover this case. I tried setting up rails-petstore and tried testing against it, but rails-petstore does not have a response that covers the pull request.

PR #899, opened 2015-06-22
PHP Codegen updates

nmonterroso:
Some changes for PHP generated code:

* rename APIClient.mustache to ApiClient.mustache to keep it consistent with other languages, and this case mismatch was causing an error for me locally
* change `require.php`, which includes all generated files up-front to be a [PSR-4](http://www.php-fig.org/psr/psr-4/) compliant auto-loader
* New CLI options
  * `invokerPackage` - Allow users to update the `invokerPackage` using an option
  * `rootNamespace` - define a PHP namespace root for all packages (invoker, api, and model packages)
* Fluent interface for modifying object properties
* `Model` and `Api` classes can now live in a different namespace (from the root namespace) than the `Invoker` package.
* The `Invoker` package is now api agnostic - it does not know anything about the APIs using the clients or configurations, so it's possible apis defined by multiple swagger docs share the same code in a core `Invoker` package.
* Deserialize non-2xx responses from the API as part of the `ApiException`

These changes are inspired by our needs at Wikia, where the plan is to have our central application talk to several microservices during the course of a request. We've tried to maintain backwards compatibility with previous use-cases (ex: having a static `$defaultConfiguration` object so `ApiClient`s don't need to always instantiate one), but let us know if we've missed anything.


wing328:
@nmonterroso thanks for the PR. Please kindly submit the PR for the `develop_2.0` branch

There's another PR with similar changes and it's just been merged into `develop_2.0` a few hours ago (https://github.com/swagger-api/swagger-codegen/pull/865)

Would you please pull the latest for develop_2.0 and submit a new PR ?

I'll review the change you've made and share with you my feedback later today.

nmonterroso:
Ah ok, looks like some of the work around namespacing will conflict. I'll put together a new PR on `develop_2.0` early tomorrow. Thanks!

nmonterroso:
Closing, since #902 is against the correct branch

PR #902, opened 2015-06-23
PHP Improvements

nmonterroso:
Note - Copied over from #899 to be against the `develop_2.0` branch.

Some changes for PHP generated code:

* New CLI options
  * `invokerPackage` - Allow users to update the `invokerPackage` using an option
  * `rootNamespace` - define a PHP namespace root for all packages (invoker, api, and model packages)
* Fluent interface for modifying object properties
* The `Invoker` package is now api agnostic - it does not know anything about the APIs using the clients or configurations, so it's possible apis defined by multiple swagger docs share the same code in a core `Invoker` package.
* Deserialize non-2xx responses from the API as part of the `ApiException`

These changes are inspired by our needs at Wikia, where the plan is to have our central application talk to several microservices during the course of a request. We've tried to maintain backwards compatibility with previous use-cases (ex: having a static `$defaultConfiguration` object so `ApiClient`s don't need to always instantiate one), but let us know if we've missed anything.


modules/swagger-codegen/src/main/resources/csharp/model.mustache
```diff
@@ -13,7 +13,7 @@ namespace {{package}} {
   /// {{description}}
   /// </summary>
   [DataContract]
-  public class {{classname}} {
+  public class {{classname}}{{#parent}} : {{{parent}}}{{/parent}} {
```
nmonterroso: looks like the changes here came from 5f1df9e, and are introduced as a result of me creating the original PR from `master`.

modules/swagger-codegen/src/main/resources/php/ApiClientConfiguration.mustache
```diff
@@ -0,0 +1,257 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+namespace {{invokerPackage}};
+
+class ApiClientConfiguration {
```
wing328: Would it be possible to keep the original name "Configuration"? The reason is to make it consistent with other languages (e.g. Python, C#, etc), which use the name "Configuration"

nmonterroso: Sure thing. Done in 6623e12

modules/swagger-codegen/src/main/resources/php/api.mustache
```diff
@@ -86,28 +87,29 @@ class {{classname}} {
...
       if (!is_null($_header_accept)) {
         $headerParams['Accept'] = $_header_accept;
       }
-      $headerParams['Content-Type'] = $this->apiClient->selectHeaderContentType(array({{#consumes}}'{{mediaType}}'{{#hasMore}},{{/hasMore}}{{/consumes}}));
+      $headerParams['Content-Type'] = ApiClient::selectHeaderContentType(array({{#consumes}}'{{mediaType}}'{{#hasMore}},{{/hasMore}}{{/consumes}}));
 
       {{#queryParams}}// query params
       if(${{paramName}} !== null) {
-        $queryParams['{{baseName}}'] = $this->apiClient->toQueryValue(${{paramName}});
+        $queryParams['{{baseName}}'] = ObjectSerializer::toQueryValue(${{paramName}});
```
wing328: Would the use of static methods cause any potential issue in a multi-thread program ?

nmonterroso: Hm, I don't think so, since the `ObjectSerializer` doesn't maintain any sort of state in static variables. If you prefer, though, I can modify `ObjectSerializer` to be an instanced object.

modules/swagger-codegen/src/main/resources/csharp/model.mustache
wing328: Since this PR is for PHP, I would suggest you to roll back the change for this file. 

I (or the author of https://github.com/swagger-api/swagger-codegen/commit/5f1df9e093eaf7b584c98879b4e05dcf1ac8c58f) will submit another one for develop_2.0

modules/swagger-codegen/src/main/resources/php/api.mustache
```diff
@@ -122,22 +124,38 @@ class {{classname}} {
         // for HTTP post (form)
         $httpBody = $formParams;
       }
-
-      // authentication setting, if any
-      $authSettings = array({{#authMethods}}'{{name}}'{{#hasMore}}, {{/hasMore}}{{/authMethods}});
-
+      {{#authMethods}}{{#isApiKey}}
```
wing328: if this approach and the original approach both work, I would recommend keeping the authentication logic in ApiClient to avoid repeating the code in the endpoint method and to make it consistent with other API client (e.g, Python, C#, Objc, etc)

modules/swagger-codegen/src/main/resources/csharp/model.mustache
nmonterroso: Sure, reverted in 2fced0f

modules/swagger-codegen/src/main/resources/php/api.mustache
wing328: The reason why I used the instance method instead of the static one is that I'm not 100% sure if static methods (depending on what it does) would work in all multi-thread/process environments.

In a multi-thread/process program, each thread/process should have its own ApiClient to avoid potential conflicts/issues.

modules/swagger-codegen/src/main/resources/php/api.mustache
nmonterroso: They both work, the reason I moved it into the responsibility of the *Api classes is so that the client is not coupled to any specific api. 

For us, we'll want to have our central application talking to many microservice APIs. Ideally they would all share a single `ApiClient`, so any improvements would affect all the API calls, not just the ones for the API which recently generated them. In order to do that, we had to move anything that was specific to a particular API (such as how it gets authenticated), out of the core classes.

To the point of duplicated code, I agree, though since this is generated and shouldn't be hand-modified, I think this is an acceptable amount of duplication to have spread across the APIs. I can change it if you disagree, but would really like this change to stick :)

modules/swagger-codegen/src/main/resources/php/api.mustache
nmonterroso: Ah, OK. I was basing my changes here from the `master` branch, where these functions are static ([example](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/php/APIClient.mustache#L322)). I'll change ObjectSerializer to be instanced to avoid any potential issues.

modules/swagger-codegen/src/main/resources/php/api.mustache
wing328: Thanks for sharing the use case with me (first time I heard of such use case). I think you've a valid reason to keep the authentication logic out of the ApiClient so please keep your approach as the change should be transparent to the developers using the library.

For other clients, we'll consider doing the same.

modules/swagger-codegen/src/main/resources/php/api.mustache
wing328: Thanks :smiley:

nmonterroso: `ObjectSerializer` and all its uses changed to be instanced in c91a1b9

modules/swagger-codegen/src/main/resources/php/api.mustache
nmonterroso: Thanks!

wing328:
@nmonterroso do you mind updating the php petstore sample using `mvn clean && ./bin/php-petstore.sh` and following the instruction below to run the phpunit test to see if the test case needs to be updated as well?

https://github.com/swagger-api/swagger-codegen/tree/develop_2.0/samples/client/petstore/php/SwaggerClient-php#tests

nmonterroso:
@wing328 - I've updated the sample files are updated the tests in efd5b80

wing328:
Integration test result is good
```
SwaggerClient-php|php-upgrades⚡ ⇒ ./vendor/bin/phpunit tests



PHPUnit 4.7.5 by Sebastian Bergmann and contributors.

..........

Time: 7.47 seconds, Memory: 26.00Mb

OK (10 tests, 4362 assertions)
```

nmonterroso:
Awesome! Any other feedback that needs to be addressed?

PR #914, opened 2015-06-24
Optional is used in PHP template although not available

webjunkie:
I changed it to `required` and now it works for me. Why was `optional` used and where did it come from?


wing328:
@webjunkie Good catch. I think it came from swagger 1.2, in which the `optional` tab worked before.

wing328:
Some templates are still using `optional` and I'll submit a PR to fix those.

PR #919, opened 2015-06-26
[Java, Android] Fix path variable name

wing328:
Currently, the path variable name in underscore case won't work in Java and Android client.

This PR fixed the issue by using `baseName`.

Tested with manually changing petId to pet_id and the result is good.


cadesalaberry:
@wing328 Thanks a lot for the prompt fix. Do you know when it will be part of the release ?

wing328:
@cadesalaberry there's no ETA on when this PR will be merged. 

If you urgently need the fix, I would suggest you to patch the files (2 lines of change in 2 files) locally for the time being.

cadesalaberry:
That's what I am doing :smile: 

Just to get an ETA on when I will be able to get rid of my custom code and go with the official stable release.

wing328:
:+1: 

Thanks again for reporting the issue.

PR #923, opened 2015-06-26
[Ruby] Some improvements & cleaning-up on Ruby API client

xhh:
* Add debugging switch to configuration, e.g. `config.debug = true`
* Allow customizing SSL CA certificate via configuration, e.g. `config.ssl_ca_cert = '/path/to/ca_cert'`
* Add comments for some configuration options
* Make raw HTTP response accessible via `Swagger.last_response`
* Avoid monkey-patching which might break existing code using the client
* Fix "date" type deserialization for model property
* Add config file for bin/ruby-petstore.sh to rename "swagger_client" to "petstore" for clarity

Thanks for the feedback from @godfat

The integration test result of Ruby Petstore is good:

```
...................................

Finished in 36.66 seconds (files took 0.87902 seconds to load)
35 examples, 0 failures
```


modules/swagger-codegen/src/main/resources/ruby/swagger/request.mustache
```diff
@@ -125,93 +103,60 @@ module {{moduleName}}
...
+
+        raw = Typhoeus::Request.new(self.url, request_options).run
+        @response = Response.new(raw)
+
+        if Swagger.configuration.debug
+          Swagger.logger.debug "HTTP response body ~BEGIN~\n#{@response.body}\n~END~\n"
+        end
+
+        # record as last response
+        Swagger.last_response = @response
```
godfat: This is not thread-safe, you might want to make `Swagger.last_response=` as `Thread.current[:swagger_last_response]=`?

xhh: @godfat Thanks for the suggestion, updated in this commit: https://github.com/swagger-api/swagger-codegen/commit/6f1dd5ce3523069ecad37548953be8e5dd828a7c

modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java
```diff
@@ -219,7 +219,11 @@ public Reader getTemplate(String name) {
             bundle.put("models", allModels);
             bundle.put("apiFolder", config.apiPackage().replace('.', File.separatorChar));
             bundle.put("modelPackage", config.modelPackage());
-            bundle.put("authMethods", config.fromSecurity(swagger.getSecurityDefinitions()));
+            List<CodegenSecurity> authMethods = config.fromSecurity(swagger.getSecurityDefinitions());
+            if (!authMethods.isEmpty()) {
```
lakeba: AuthMethods can be null, please fix it. 

if (authMethods != null && !authMethods.isEmpty()) {
                bundle.put("authMethods", authMethods);
                bundle.put("hasAuthMethods", true);
            }

xhh: Thanks for pointing it out, will fix it soon.

PR #924, opened 2015-06-26
datetype catched to avoid AttributeError

bletail:


wing328:
@bletail thanks for the PR. Please submit it against the "develop_2.0" branch instead.

PR #931, opened 2015-06-29
generate oauth2 fields in the data for the templates

### LINKED ISSUE #347: [develop_2.0] Does swagger codegen 2.1 generates OAuth2 dedicated methods ?
I see in swagger 2.0 spec that security schemes such as OAuth2 are available.
In the petstore.json file from the develop_2.0 branche,  oauth2 is defined in the "securityDefinitions" part but I don't see any dedicated method in android-java client code generated. I thought I would have found a method calling the authorizationUrl defined below in one of the generated classes :    
 "petstore_auth": {
      "type": "oauth2",
      "authorizationUrl": "http://petstore.swagger.wordnik.com/api/oauth/dialog",
      "flow": "implicit"
    }

cbornet:
Fixes #347

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenSecurity.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenSecurity.java
@@ -1,5 +1,7 @@
 package io.swagger.codegen;
 
+import java.util.Set;
+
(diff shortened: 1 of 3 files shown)


wing328:
Tested and code generation works fine as expected.

PR #933, opened 2015-06-29
datetype catched to avoid AttributeError

bletail:


wing328:
@bletail thanks for the PR. Do you have a sample Swagger spec that we can use to repeat the issue (so as to have a better understanding of what this PR fixes as other languages may run into similar issue) ?

bletail:
@wing328 In fact this problem was only due to the python function __parse_string_to_datetime.
This function can't take a None parameter that sometime I had. The AttributeError has to be catched with this part of code

geekerzp:
Fixed in this commit https://github.com/geekerzp/swagger-codegen/commit/bb30e2c98970c4514efea5717359357b730e1d7e.

Does it look good? @bletail

bletail:
@geekerzp Yes it does thank you

PR #940, opened 2015-07-02
Sanitize API class name in Jax RS server codegen

### LINKED ISSUE #939: JAX-RS generator produces invalid Java code when there are dashes in the resource URL
If a "path" contains a dash, e.g.

```
"paths": {
      "/online-community": {
          "get": {
          ...
   }
}
```
and __jaxrs__ is used as the "language", the generated Java classes still contain the dashes:

```
public class Online-communityApiServiceImpl extends Online-communityApiService {
}
```
and the generation is unusable. Is there a way to remove dashes or even better force the replacement of "*snake-case*" style in URLs by "*CamelCase*" style in Java/JAX-RS?

xhh:
This PR is to fix #939

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JaxRSServerCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JaxRSServerCodegen.java
@@ -167,6 +167,15 @@ public class JaxRSServerCodegen extends JavaClientCodegen implements CodegenConf
     }
 
     @Override
+    public String toApiName(String name) {
+        if (name.length() == 0) {
+            return "DefaultApi";
+        }
+        name = name.replaceAll("[^a-zA-Z ]+", "_");
+        return camelize(name) + "Api";
+    }
+
+    @Override
     public String apiFilename(String templateName, String tag) {
 
         String result = super.apiFilename(templateName, tag);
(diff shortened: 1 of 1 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JaxRSServerCodegen.java
```diff
@@ -167,6 +167,15 @@ public void addOperationToGroup(String tag, String resourcePath, Operation opera
     }
 
     @Override
+    public String toApiName(String name) {
+        if (name.length() == 0) {
+            return "DefaultApi";
+        }
+        name = name.replaceAll("[^a-zA-Z ]+", "_");
```
olensmar: what about numbers? it's probably unusual but that doesn't mean we should remove them!?

xhh: i copied that line from [here](https://github.com/xhh/swagger-codegen/blob/jax-rs-classname/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java#L389) which is used to generate the `classname` in other languages
but i agree with you, let me update it to allow numbers

olensmar:
thank you!

PR #943, opened 2015-07-02
[C#] Use RestSharp's mechanism for path parameters in URL templates

psmay:
Previously, a simple find-and-replace had been used to substitute path parameters into the path. Among other omissions, this resulted in special characters such as `/` being left unescaped.

The RestSharp request object, as it turns out, expects a path template in the same format as provided by Swagger (with param names in braces), to be filled in using `AddParameter()`. In this edit, the code now uses this mechanism. The form parameter values are now passed to `CallApi*()` as a `Dictionary<String, String>`, where they are added to the request before submission. Since this was already how query and form parameters (etc.) were implemented, the resulting code is more consistent with itself than before.


wing328:
@psmay thanks for the PR. Do you mind rebasing the PR on the latest develop_2.0 branch?

psmay:
@wing328 Updated, but not tested (I'm not at my work machine, which is where the C# happens). It is logically sound, I think, but the result really should be run through a compiler/IDE to make sure there aren't any typos.

wing328:
@psmay no problem. I'll run some tests to make sure. Thanks again for the PR and have a nice weekend.

wing328:
Unit test results are good

PR #946, opened 2015-07-03
Copy properties from parent models defined with allOf

### LINKED ISSUE #936: ClassCastException: io.swagger.models.RefModel cannot be cast to io.swagger.models.ModelImpl
Hi,

I am trying to generate static documentation for my API definition

`java -XX:MaxPermSize=256M -Xmx1024M -jar modules\swagger-codegen-cli\target\swagger-codegen-cli.jar generate -i swagger.json -l html -o target`

using both `master` & `develop_2.0` branch I have the following exception
```
java.lang.ClassCastException: io.swagger.models.RefModel cannot be cast to io.swagger.models.ModelImpl
        at io.swagger.codegen.DefaultCodegen.fromModel(DefaultCodegen.java:532)
        at io.swagger.codegen.DefaultGenerator.processModels(DefaultGenerator.java:452)
        at io.swagger.codegen.DefaultGenerator.generate(DefaultGenerator.java:125)
        at io.swagger.codegen.cmd.Generate.run(Generate.java:124)
        at io.swagger.codegen.SwaggerCodegen.main(SwaggerCodegen.java:35)
```

I have isolated my issue. It comes from the definition of a model using `allOf` and 2 references to other models.

Here is a simplified json swagger api definition to reproduce:

```json
{
    "swagger": "2.0",
    "info": {
        "title": "My demo API",
        "description": "My demo API documentation",
        "version": "1.0.0"
    },
    "host": "myserver",
    "schemes": [
        "http"
    ],
    "basePath": "/resources/v1",
    "paths": {
        "/clients/{clientId}": {
            "get": {
                "summary": "get info on a specific client",
                "tags": [
                    "Clients"
                ],
                "parameters": [
                    {
                        "in": "path",
                        "name": "clientId",
                        "type": "integer",
                        "required": true,
                        "format": "int64"
                    }
                ],
                "produces": [
                    "application/json"
                ],
                "responses": {
                    "200": {
                        "description": "retrieve data of the corresponding client",
                        "schema": {
                            "$ref": "#/definitions/Client"
                        }
                    }
                }
            }
        }
    },
    "definitions": {
        "Client": {
            "allOf": [
                {
                    "$ref": "#/definitions/Person"
                },
                {
                    "$ref": "#/definitions/Address"
                }
            ]
        },
        "Person": {
            "required": [
                "name"
            ],
            "properties": {
                "name": {
                    "type": "string"
                }
            }
        },
        "Address": {
            "required": [
                "postalCode",
                "city"
            ],
            "properties": {
                "street": {
                    "type": "string"
                },
                "postalCode": {
                    "type": "integer"
                },
                "city": {
                    "type": "string"
                }
            }
        }
    }
}
```

xhh:
This PR copies model properties of the parent models that are defined with the `allOf` key (with value of either `$ref` or `properties`), i.e. for composed models.

It also fixes #936.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConfig.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConfig.java
@@ -65,6 +65,8 @@ public interface CodegenConfig {
 
     CodegenModel fromModel(String name, Model model);
 
+    CodegenModel fromModel(String name, Model model, Map<String, Model> allDefinitions);
+
     CodegenOperation fromOperation(String resourcePath, String httpMethod, Operation operation, Map<String, Model> definitions);
 
     List<CodegenSecurity> fromSecurity(Map<String, SecuritySchemeDefinition> schemes);
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -506,6 +506,10 @@ public class DefaultCodegen {
     }
 
     public CodegenModel fromModel(String name, Model model) {
+        return fromModel(name, model, null);
+    }
+
+    public CodegenModel fromModel(String name, Model model, Map<String, Model> allDefinitions) {
         CodegenModel m = CodegenModelFactory.newInstance(CodegenModelType.MODEL);
         if (reservedWords.contains(name)) {
             m.name = escapeReservedWord(name);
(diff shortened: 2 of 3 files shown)


webron:
Does this mean the end result would be flattened classes rather than one exhibiting inheritance?
Also, I see that this doesn't handle the `discriminator` yet (which is fine). Is the intend here to provide better support for composition rather than inheritance?

xhh:
@webron Yes, the end result would be flattened classes as all properties of parents would be copied and made as if they are the current model's own properties.
This is to provide better support for composition (`allOf`) as some clients do not handle inheritance (yet). 

For other clients that do handle it, e.g. Java client uses something like `class Dog extends Animal` when the Dog model's `allOf` refers to Animal, I think we need to further improve the support, either remove the `extends` or do not copy Animal's properties to Dog. We might have to also consider `discriminator` here.
However, it does work/compile to overwrite attributes of parent class, though not elegant.

webron:
Thanks for the clarification. This is a tricky part of the spec, and may have a different effect on clients VS server. 

If we want to keep a 1:1 representation of spec:code, this may not fully work for servers, however, for it to work we need to have better support for it in the server libraries (I'm thinking of swagger-core but it applies to others as well). For clients, it may be less of an issue.

Composition, at least for now, can (and probably should) be translated to a flattened structure in *most* languages.

Inheritance (which is based on the discriminator field) needs to be represented eventually as actual extended classes, although the tricky part here is with multiple inheritance which JSON Schema allows but some languages don't.

Obviously there are limitations now, but it would be useful to keep the metadata of the origin of the property in the property itself to allow more flexibility in the future. I'm okay with it as-is right now. :)

Issue #949, opened 2015-07-05
Java client uses dated jersey client libraries


fehguy:
In integrating a jersey 2 server with the swagger java client, I ran into issues with the client library being incompatible with the server and causing a host of issues.  There should be a jersey 2.x-compatible client for java

webron:
This may conflict with #862 if it ends up being merged.

fehguy:
I think the right solution is to keep everything consistent except the ApiClient class, and let there be an option for each invoker

phauer:
I run into the same problems. The generated API library uses jersey-client 1.1.8 and dropwizard-core (0.8.0) uses jersey-client 2.16. This causes a lot of NoSuchMethodError when I use my dropwizard application (for startup in a test) in conjunction with the generated API library in the same project. Due to the different groupId (com.sun.jersey and org.glassfish.jersey.core), I can't try to force a certain version (which probably won't work anyway).

Unfortunately, this issue really prevents the usage of the generated API library. At least I can't think of any workaround except starting the dropwizard application via the command line in my test. But I really like the idea behind swagger-codegen.

wing328:
@phauer the latest java client allows you to specify the http library (see the `library` option):
```
swagger-codegen|master ⇒ java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar config-help -l java 

CONFIG OPTIONS
	modelPackage
	    package for generated models

	apiPackage
	    package for generated api classes

	invokerPackage
	    root package for generated code

	groupId
	    groupId in generated pom.xml

	artifactId
	    artifactId in generated pom.xml

	artifactVersion
	    artifact version in generated pom.xml

	sourceFolder
	    source folder for generated code

	localVariablePrefix
	    prefix for generated code members and local variables

	serializableModel
	    boolean - toggle "implements Serializable" for generated models

	library
	    library template (sub-template) to use:
	    <default> - HTTP client: Jersey client 1.18. JSON processing: Jackson 2.4.2
	    jersey2 - HTTP client: Jersey client 2.6
```
would that resolve the issue you reported above?

phauer:
@wing328 That looks great. Thanks!
Unfortunately, there are still version conflicts between dropwizard 0.8.0 and the client generated by swagger-codegen 2.1.3:
Jersey 2.16 (dropwizard) vs Jersey 2.8 (swagger-codegen)
Jackson 2.5.1 vs Jackson 2.4.2

[...]

PR #950, opened 2015-07-05
updated client to use jersey2, rebuilt samples

fehguy:
This PR add support for #949.  Before merging, there are some decisions to make.

1) The jersey client is quite heavy for this use case.  Sure it's fluent, etc., but we can't really take advantage of that for this codegen (as-is)
2) There will be users who still need jersey 1.x client support because they're using a combined client/server with jersey 1.x, which is very popular
3) There will be more client libraries needed (plain apache-http 3.x, 4.x, etc).

How should we support these options?  We could have an uber-java client target that takes the supporting library as an argument, where you could specify jersey1, jersey2, commons-http-3.x, etc.  There are other options such as synchronous, asynchronous (supported by Jersey2 via `Future`).  Or we can have many client options when generating.

Before merging, we should come up with a plan for the above, it'll be the same for each client language, to some degree.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/ApiClient.mustache
@@ -1,49 +1,42 @@
 package {{invokerPackage}};
 
-import com.fasterxml.jackson.core.JsonGenerator.Feature;
 import com.fasterxml.jackson.databind.*;
-import com.fasterxml.jackson.annotation.*;
-import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 
-import com.sun.jersey.api.client.Client;
-import com.sun.jersey.api.client.ClientResponse;
-import com.sun.jersey.api.client.config.ClientConfig;
-import com.sun.jersey.api.client.config.DefaultClientConfig;
-import com.sun.jersey.api.client.filter.LoggingFilter;
-import com.sun.jersey.api.client.WebResource.Builder;
-import com.sun.jersey.multipart.FormDataMultiPart;
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.client.Invocation;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.Response;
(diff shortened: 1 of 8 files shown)


nmonterroso:
Is there anything I can do to help this PR along? I was recently thinking of making this change, since one of our projects uses Jersey 2.x but we want to use code generated from Swagger docs (and so we ran into a ton of issues). Super happy to see this was already on your radar, so let me know if there's anything I can do to help :)

fehguy:
@nmonterroso it looks like we're going to make a slight change, which is to have the underlying transport independent from the client templates.  So we'll get this merged after a bit more work.

modules/swagger-codegen/src/main/resources/Java/ApiClient.mustache
```diff
@@ -61,10 +55,9 @@ public class ApiClient {
     setUserAgent("Java-Swagger");
 
     // Setup authentications (key: authentication name, value: authentication).
-    authentications = new HashMap<String, Authentication>();{{#authMethods}}{{#isBasic}}
-    authentications.put("{{name}}", new HttpBasicAuth());{{/isBasic}}{{#isApiKey}}
-    authentications.put("{{name}}", new ApiKeyAuth({{#isKeyInHeader}}"header"{{/isKeyInHeader}}{{^isKeyInHeader}}"query"{{/isKeyInHeader}}, "{{keyParamName}}"));{{/isApiKey}}{{#isOAuth}}
-    authentications.put("{{name}}", new OAuth());{{/isOAuth}}{{/authMethods}}
+    authentications = new HashMap<String, Authentication>();
```
nmonterroso: We were experimenting with using this branch to get up-and-running with jersey2, and thought this hard-coded change might be here because of development in trying to get the authentication pieces working with jersey2. This is a friendly reminder that this specific part of the changeset should probably be reverted, as well as the change to [`basePath`](https://github.com/swagger-api/swagger-codegen/pull/950/files#diff-3714cb3e3c7bbb1ce24de92a3fb9f01bR40)

fehguy: this should be addressed in #1046

fehguy:
closed in favor of #1046

PR #953, opened 2015-07-07
Python/Python3 datetime and date type fixed

bletail:


wing328:
Tested and confirm `date` works in both python2 and python3

Issue #964, opened 2015-07-08
Python: 'module' object has no attribute url


EntilZha:
I am writing some code which uses the generated code libraries and am running into a stack trace that looks like this (on develop_2.0, not using 2.1.3 since it doesn't include the patch for incorrect directory):

```python
Traceback (most recent call last):
  File "/Users/prodriguez/Code/datascience-api/docs-tests/tests/integration/test_discovery_api.py", line 26, in test_offset
    response_0 = self.discovery_api.discovery_v1_photoboards_get()
  File "/Users/prodriguez/Code/datascience-api/docs-tests/tests/integration/datascience_api/apis/discovery_api.py", line 182, in discovery_v1_photoboards_get
    response='DiscoveryResponse', auth_settings=auth_settings)
  File "/Users/prodriguez/Code/datascience-api/docs-tests/tests/integration/datascience_api/api_client.py", line 101, in call_api
    post_params=post_params, body=body)
  File "/Users/prodriguez/Code/datascience-api/docs-tests/tests/integration/datascience_api/api_client.py", line 236, in request
    return RESTClient.GET(url, query_params=query_params, headers=headers)
  File "/Users/prodriguez/Code/datascience-api/docs-tests/tests/integration/datascience_api/rest.py", line 218, in GET
    return cls.IMPL.GET(*n, **kw)
  File "/Users/prodriguez/Code/datascience-api/docs-tests/tests/integration/datascience_api/rest.py", line 148, in GET
    return self.request("GET", url, headers=headers, query_params=query_params)
  File "/Users/prodriguez/Code/datascience-api/docs-tests/tests/integration/datascience_api/rest.py", line 123, in request
    r = self.agent(url).request(method, url,
  File "/Users/prodriguez/Code/datascience-api/docs-tests/tests/integration/datascience_api/rest.py", line 70, in agent
    url = urllib3.util.url.parse_url(url)
AttributeError: 'module' object has no attribute 'url'
```

Basically the problem is that the code has this call:

`urllib3.util.url.parse_url`

It should be:

`urllib3.util.parse_url`

On a different note, this and a couple other bugs I have found (like the incorrect directory name) point to that Python code generation seems second class and not well maintained. How trustworthy can I be of the code to be correct? I am using it in a production system so would like to get a feeling for that, and if it is not super well maintained, are there some low hanging fruit to help out with?

geekerzp:
Hi @EntilZha 
[...]

PR #965, opened 2015-07-08
Fix for issue #964, correctly call urllib3.util.parse_url

EntilZha:
Fixed the issue that I started at #964. I don't know what process there is for testing, but it fixed the crashing code I am using internally at my company. It would probably be helpful to have upstream so making a PR


wing328:
@EntilZha is this PR still needed given that @geekerzp has made the change in https://github.com/geekerzp/swagger-codegen/commit/eed58bc313897c4db5d0c956ef7d4a71b8dc8f9c ?

EntilZha:
Don't think so, you can go ahead and close since they are the same thing.

PR #968, opened 2015-07-09
Cleaned up TypeScript code generator

mhardorf:
* Changed modelPackage to 'model'
* Added mapping of Map and DateTime
* Switched to use contextPath
* Changed non-required parameters to be non-required in TypeScript


wing328:
Tested and able to generate code without issues.

PR #971, opened 2015-07-09
[ObjC] Response Headers are now passed back to the caller & unused variable warnings have been fixed.

mad102190:
- [ObjC] Response headers are now passed down from the network call to the caller in the same completionBlock as the data/error objects
- [ObjC] Removed an unused "value" variable in SWGApiClient.
- [ObjC] Added check for formParams and set it to nil if its object count is 0. This also removes unused variable warnings.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/objc/SWGApiClient-body.mustache
+++ b/modules/swagger-codegen/src/main/resources/objc/SWGApiClient-body.mustache
@@ -268,7 +268,7 @@ static bool loggingEnabled = true;
         for(NSString * key in [queryParams keyEnumerator]){
             if(counter == 0) separator = @"?";
             else separator = @"&";
-            NSString * value;
+            
             id queryParam = [queryParams valueForKey:key];
             if([queryParam isKindOfClass:[NSString class]]){
                 [requestUrl appendString:[NSString stringWithFormat:@"%@%@=%@", separator,
--- a/modules/swagger-codegen/src/main/resources/objc/SWGApiClient-body.mustache
+++ b/modules/swagger-codegen/src/main/resources/objc/SWGApiClient-body.mustache
@@ -483,7 +483,7 @@ static bool loggingEnabled = true;
                             authSettings: (NSArray *) authSettings
                       requestContentType: (NSString*) requestContentType
                      responseContentType: (NSString*) responseContentType
-                         completionBlock: (void (^)(id, NSError *))completionBlock {
+                         completionBlock: (void (^)(NSDictionary *, id, NSError *))completionBlock {
     // setting request serializer
     if ([requestContentType isEqualToString:@"application/json"]) {
         self.requestSerializer = [AFJSONRequestSerializer serializer];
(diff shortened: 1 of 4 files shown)


wing328:
@mad102190 thanks for the PR to remove variable warnings and add support to obtain the headers.

If I understand correctly, the change would require developers to update their source code as the API method now returns the HTTP header as well. To make the change backward compatible, may I suggest storing the HTTP header in a new property (e.g. httpHeader) in ApiClient instead? (clients in some other languages have used similar approach so this would make the developer experience more consistent)

mad102190:
I can write it that way, sure. But would that have any concurrency issues? For example, if I store the responseHeaders for request_A and then in the completionBlock of request_A I try to access the property `[[api apiClient] httpResponseHeaders]`, wouldn't it be possible that request_B has already changed the value of the responseHeaders?

mad102190:
I made the requested changes in 1c822aa. It's fully backwards compatible now.

wing328:
@mad102190 thanks! To answer your question, the suggestion would not work if a single ApiClient is shared with many API class. To workaround the limitation, the developer should declare an ApiClient per thread to avoid concurrency issue.

In your case, do you mind sharing with me what information you want to obtain from the HTTP header ?

I look at other iOS SDKs, e.g. Twitter, but couldn't find any way to allow the developers to access the headers that we can reference to.

wing328:
@mad102190 thanks again for the PR. I ran the integration test and got the following error on my machine:
```
2015-07-14 23:53:00.291 SwaggerClient[1594:12617] got id 8
2015-07-14 23:53:00.291 SwaggerClient[1594:12617] added 8 to request queue
2015-07-14 23:53:02.586 SwaggerClient[1594:12617] removing request id 8
2015-07-14 23:53:02.798 SwaggerClient[1594:12617] [JSONModel.m:253] Incoming data was invalid [SWGPet initWithDictionary:]. Keys missing: {(
    name
)}
2015-07-14 23:53:02.805 SwaggerClient[1594:12617] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSArrayM insertObject:atIndex:]: object cannot be nil'
*** First throw call stack:
(
	0   CoreFoundation                      0x000000010b1d7c65 __exceptionPreprocess + 165
```

I wonder if you can run a test to see whether you get the same error.

To run the integration test, please run the following command under `samples/client/petstore/objc`
```
```

samuelg:
Am I correct in assuming that for this to work, each thread would have to initialized its own client and use:
```
- (id) initWithApiClient:(SWGApiClient *)apiClient
```

to ensure that  a shared client is not fetched from the pool? Documenting this would be a good idea.

geekerzp:
@samuelg  Yes, you should initialize an api client for each thread. 
A shared client from the pool may occur some issues in multiple threads environment.

wing328:
@samuelg thanks for the feedback. Agreed that the documentation can be improved. 

I've submitted https://github.com/swagger-api/swagger-codegen/pull/1070 with an updated README. 

If you encountered any multi-thread issue with the ObjC SDK, please open a new issue and we'll look into it.

wing328:
@mad102190 please kindly rebase on the latest develop_2.0 branch

webron:
@mad102190 - as @wing328 mentioned, this needs a rebase, but also a check into the integration tests which are failing. I'm going to close it now due to the upcoming move to development on the master branch (watch for the announcement). Feel free to resubmit when you can spend the time to tackle the issues raised here, we'd love to be able to merge it in.

mad102190:
I've re-implemented this from the current master branch as per @wing328 and @webron's requests. Please see #1189

PR #986, opened 2015-07-17
Don't change param names when generating retrofit interfaces

hoeggi:
At the moment param names are changed from snake_case to camelCase when generating Retrofit interfaces.

So when generating client code for this spec:
```js
parameters: [
    {
        name: "grant_type",
        in: "formData",
        description: "client_credentials",
        required: true,
        type: "string",
        default: "client_credentials"
    }
]
```
your body ends up like this:
```js
{
    grandType: "XXXX"
}
```
when it actually should look like this:
```js
{
    grand_type: "XXXX"
}
```

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RetrofitClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RetrofitClientCodegen.java
@@ -116,8 +116,9 @@ public class RetrofitClientCodegen extends DefaultCodegen implements CodegenConf
 
     @Override
     public String toParamName(String name) {
-        // should be the same as variable name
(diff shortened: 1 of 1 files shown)


xhh:
@hoeggi as in the [headerParams template](https://github.com/swagger-api/swagger-codegen/blob/develop_2.0/modules/swagger-codegen/src/main/resources/retrofit/headerParams.mustache), the unchanged field name is available via `baseName`. I've submitted #987 to fix it by updating the template

wing328:
@hoeggi I also prefer using `baseName` to fix the issue as other templates (Java, Android, etc) also use `baseName` for the field name (original name)

wing328:
#987 has been merged into `develop_2.0`. Closing the issue as we've not heard from you.

PR #987, opened 2015-07-20
Retrofit: fix form field naming by using baseName

xhh:
This PR fixes the issue described in #986, by changing `paramName` to `baseName` in the template of formParams, like other templates e.g. headerParams.


wing328:
Ran `'mvn clean package` on the sample for Retrofit without issue

PR #991, opened 2015-07-20
rename objc resource files 

qswinson:
The objc generation templates contain several files that do not end in .mustache so that they are not included as resources in the generated jar. This fix renames the files so that the maven pom will package them as resources into the jar.


wing328:
@qswinson thanks for the PR. I believe the issue should have been addressed in `develop_2.0` branch. May I know if you've performed a test with the `develop_2.0` branch ?

wing328:
Closing this as we've not heard from you. Please open a new issue if you still encounter the same problem in `develop_2.0` branch.

PR #995, opened 2015-07-21
[C#] RFC: Refactoring/DRY changes

psmay:
This change is to the C# `api` and `ApiClient` templates.

The setup of the sync and async versions of the same operation is factored out to a new common setup function. This function creates, sets up, and returns a new builder object that has been preset with the given parameters. With the parameters handled in the setup method and most of the lifting performed by the builder, repetition of code is reduced considerably. Much of the actual work is moved out of the specific `api` template and into the more general `ApiClient` template, where it may be easier to maintain or customize.

The builder handles/replaces all of the dictionaries that had been previously created within the method for collecting parameters. The parameters passed to the setup method are added to the builder using an appropriate method (such as `builder.AddQueryParam("tags", tags)` or `builder.AddPathParam("petId", petId)`). The builder performs the null check and conversions (i.e. `ParameterToString()`), so those parts no longer appear in the individual APIs.

After all desired parameters are collected, the builder object presents four methods for making an API call:
-   `void Call()`, void sync
-   `Task CallAsync()`, void async
-   `TResult Call<TResult>()`, non-void sync
-   `Task<TResult> CallAsync<TResult>()`, non-void async

Each only differs slightly from the others. All four start by creating a new `RestRequest` and loading it with the parameters collected by the builder. The call itself is made (as sync or async) and the resulting `IRestResponse` recovered. If the status code is 0 or >=400, an exception is thrown. For the non-void methods, the response payload is deserialized to the type given in the generic type parameter, then returned; otherwise, the method simply completes.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
@@ -18,7 +18,7 @@ namespace {{packageName}}.Client
     /// </summary>
     public class ApiClient
     {
-        private readonly Dictionary<String, String> _defaultHeaderMap = new Dictionary<String, String>();
+        private readonly Dictionary<String, String> defaultHeaderMap = new Dictionary<String, String>();
   
         /// <summary>
         /// Initializes a new instance of the <see cref="ApiClient" /> class.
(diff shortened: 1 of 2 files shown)


modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -47,93 +47,180 @@ namespace {{packageName}}.Client
...
+            public void AddAuthSettings(params String[] authMethod)
+            {
+                authSettings.AddRange(authMethod);
+            }
+
+            private RestRequest CreateRequest()
+            {
+                var request = new RestRequest(path, method);
+
+                var queryParamsCopy = new Dictionary<String, String>(queryParams);
```
psmay: These copies of `queryParams` and `headerParams` are made before `UpdateParamsForAuth()` edits them in place. This is solely for keeping the builder object valid for repeat requests or requests after more options are added. Naturally, for this application such defensiveness is probably overkill. If we ensure that each builder is treated as valid for a maximum of one call, the copies can be removed and `*ParamsCopy` replaced with `*Params`.

fehguy:
I actually think this is a good direction, but it will likely break all client signatures so we have to hold for now.  Keeping open.

wing328:
The C# API client in 2.3.0 has been refactored. Please review and make enhancement on top of it. Thanks.

PR #1001, opened 2015-07-22
C# .Net 2.0 generator and template

who:
This PR adds a C# .Net 2.0 generator. The current C# generator is generating .Net 4.5 output, which will not work in places where .Net compatibility has a ceiling of 2.0. The main use case that I'm aiming for in this PR is the ability to generate DLLs that can be used as managed libraries in Unity3D.

Below is a demo of using a generated .Net 2.0 Petstore DLL in Unity 3d. I'm using the Swagger petstore's `/user` API to log a simple game score of the user.

[IMAGE ALT TEXT HERE](http://www.youtube.com/watch?v=QF77jJYuIvY)


who:
@wing328 Do you have any questions on this PR that I can address?

wing328:
@who thanks again for the new template.

Tested and generated code without issue.

PR #1005, opened 2015-07-23
Sort model objects by inheritance

mhardorf:
Required for typescript generation, but seems like a general thing.


wing328:
Tested and compiled without issue.

PR #1006, opened 2015-07-23
[Swift] Add podspec template

ngs:


Edubits:
Thanks for the PR @ngs, I was planning to add this myself as well. A few things I'm wondering about in the template:
- to be able to add the Pod to Specs Repo a few properties are obligatory, like s.license (this can be tested by running `pod spec lint`)
- s.source: shouldn't it be configurable? Or do we want them to be adjusted in a private fork of the repo? In that case what about adding empty values for these.

ngs:
Hi,

To pass the podspec lint, we should set correct source information and some meta information like `s.license`.

But we can use the podspec that was generated by this template from our `Podfile` of Xcode projects.

Should we pass these information with `cliOptions` or something?

wing328:
@ngs thanks for the PR. I would prefer passing additional information required by Swift packaging (cocoapods) via `cliOptions`.

For license, I would recommend defaulting to Apache license v2.

ngs:
@wing328 Thanks, I'm adding `cliOptions` to support each [Podspec] properties.

Why do you recommend `Apache License 2` as default?

[Podspec]: https://guides.cocoapods.org/syntax/podspec.html

wing328:
@ngs my rationale is that this project (swagger-codegen) is licensed under Apache License v2. I'm no expert in software licensing but my guess is that the output of this project should also follow (or inherit) the same licensing agreement.

ngs:
@wing328 got it. added `cliOptions` for podspec 80b536320251f1f07debcc1419b596ac4aad6d25

wing328:
Tested and found no issue with the podspec
```
swagger-codegen|podspec⚡ ⇒ ruby -c /private/var/tmp/ngs/swagger-codegen/samples/client/petstore/swift/PetstoreClient.podspec
Syntax OK
```

PR #1009, opened 2015-07-24
[Python] Allow to set cookie in python(3) API constructor.

onovy:


wing328:
@onovy thanks for submitting the PR. Please close this one and submit a new one against `develop_2.0` branch instead.

Do you mind sharing a bit more about your use case of cookie? Is that for authentication/authorization?

FYI. The Python (not Python3) SDK has been upgraded to support both python v2 and v3. (Python3 template was created before the upgrade)

onovy:
done.

use-case is for cookie based auth.

PR #1010, opened 2015-07-24
[Python] Allow to set cookie in python(3) API constructor.

onovy:


wing328:
Tested and compiled without issue.

PR #1015, opened 2015-07-25
Conditional copy properties from parent

tandrup:
Pull request #946 broke the TypeScript generation with inheritance where a superclass defined an enum. This pull request makes it up to each language to decide the behavior appropriate for the language.


wing328:
@tandrup thanks for the PR.

Do you mind giving an example on how `allOf` (model composition, not inheritance) broke the Typescript code generation?

FYI. https://github.com/swagger-api/swagger-codegen/pull/1005 is also related to inheritance and its author happens to be the one who generously contributed the typescript template.

tandrup:
@wing328 Sure, assume you have this model:
```
{
  "definitions": {
    "Pet": {
      "discriminator": "petType",
      "properties": {
        "name": { "type": "string" },
        "petType": { "type": "string" },
        "huntingSkill": {
          "type": "string",
          "enum": ["clueless", "lazy", "adventurous"]
        }
      },
      "required": ["name", "petType"]
    }
  },
  "Cat": {
    "allOf": [
      {
        "$ref": "#/definitions/Pet"
      },
      {
        "properties": {
          "miaw": { "type": "integer" }
        }
      }
    ]
  },
  "Dog": {
    "allOf": [
      {
        "$ref": "#/definitions/Pet"
      },
      {
        "properties": {
          "packSize": { "type": "integer" }
        }
      }
    ]
  }
}
```

In the code by @mhardorf the enum values are represented as an enum for each field. Since there's no notion in the swagger spec of shared enum's.

If the generated code does not use inheritance we will end up with a `Cat.huntingSkillEnum` and a `Dog. huntingSkillEnum`. But it's not valid typescript to assign a enum of one type to another. You can't do:
```typescript
var x = Cat.huntingSkillEnum.lazy
x = Dog.huntingSkillEnum.clueless
```

With inheritance we would only end up with one enum.

tandrup:
BTW. I'm working together with @mhardorf on the TypeScript generation, so I know about the sorting PR  #1005. That pull request targets another problem. TypeScript does not handle it well if the child class comes before the parent class in the code.

mhardorf:
@tandrup looks good

Issue #905, opened 2015-06-23
spring-mvc config.json server should have an option for configPackage


cosbor11:
see title

wing328:
Are you referring to the following options ?

```
swagger-codegen|master ⇒ java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar config-help -l spring-mvc

CONFIG OPTIONS
	modelPackage
	    package for generated models

	apiPackage
	    package for generated api classes

	invokerPackage
	    root package for generated code

	groupId
	    groupId in generated pom.xml

	artifactId
	    artifactId in generated pom.xml

	artifactVersion
	    artifact version in generated pom.xml

	sourceFolder
	    source folder for generated code
```

cosbor11:
Yes

wing328:
would you please elaborate a bit on the configPackage ? What does it do/change exactly ?

I checked in [JavaClientCodegen.java] (https://github.com/swagger-api/swagger-codegen/blob/develop_2.0/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java) but couldn't find any configPackage

cosbor11:
Sorry, i wasnt very clear...It's the output folder name.
"io.swagger.configuration" would like to be able to rename to
com.mycompany.api.config. similarly you can do this to the other folders.

herveDarritchon:
In fact, it seems that the Swagger CodeGen is not able to override the configPackage variable set in the code even if you put a configPackage in the config json file.

Look at my example :

If I use the standard example : PetStore

If I run the script `./bin/spring-mvc-petstore-server.sh`with a modification in the run command line such as `ags="$@ generate -t modules/swagger-codegen/src/main/resources/JavaSpringMVC -i modules/swagger-codegen/src/test/resources/2_0/petstore.json -c /Users/herveDarritchonworkspace/ms-admin-data-api/src/main/swagger/config.json -l spring-mvc -o samples/server/petstore/spring-mvc"`

Please find the config.json file content :
```
{
  "groupId":"com.orange.dsifapi.microservices",
  "artifactId":"admindata",
  "artifactVersion":"0.0.1",
  "modelPackage":"com.orange.dsifapi.microservice.admindata.domain",
  "apiPackage":"com.orange.dsifapi.microservice.admindata.api",
  "invokerPackage":"com.orange.dsifapi.microservice.admindata.invoker",
[...]

PR #1017, opened 2015-07-26
Defect (Issue #905) : Add code to be able to override the configPackage …

herveDarritchon:
Based on Issue #905 I made a fix bug to allow by configuration to override the configPackage variable from config json file.

It was not possible because this variable is new and only used by Spring MVC so it needs code to override the default configuration.

I choose to put the new code in the spring mvc generation class (SpringMVCServerCodegen.java) because it is not used by the parent class.


wing328:
Tested and generated the sample code for spring mvc server without issue.

PR #1022, opened 2015-07-27
fixed a error rule for parsing `)`

SparrowJang:
test：
```python
import re
re.match('dict\(([^,]*), (.*)\)','dict(str, dict(str, dict(str,dict(str))))').group(2)
```
result:
```python
dict(str, dict(str,dict(str)))
```


wing328:
@SparrowJang thanks for the PR. Do you have the swagger spec that you can share with me so that I can repeat the issue and confirm the fix ?

SparrowJang:
`swagger.yaml` response:
```yaml
responses:                                                                      
  200:
    description: messages...
    schema:
      type: object
      additionalProperties:
        type: object
        additionalProperties:
          type: string
```
The response parse error, when regexp rule hasn't fixed.
```python
re.match('dict\((.*), (.*)\)','dict(str, dict(str, dict(str,dict(str))))').group(2)
```
error result:
```
str)
```

PR #1029, opened 2015-07-28
fixed generated code to go to target/generated-sources, added maven-c…

olensmar:
…odegen-plugin to generated jaxrs project together with initial swagger definition, fixed overwrite flag to kick in only if file exists, fixed initial readme


fehguy:
looks good to me!

olensmar:
we need to remember to bump the version of the embedded maven-plugin in the templates when we release or change version

fehguy:
OK this broke a lot of things.  You can't write to `target/generated-sources` unless it's part of the build process.  I'm reverting this and we can reconsider in the future.

PR #1031, opened 2015-07-29
Fix issues in POM configuration

Toilal:
Thanks to Intellij IDEA that has spotted thos issues in maven configuration.


Toilal:
But it cause the build to fail ... i'm investigating.

Toilal:
Updating to latest scala maven plugin has fixed the build issue.

PR #1036, opened 2015-07-30
[C#] add back RestSharp and Newtonsoft.Json DLL

wing328:
Added back the DLL to address the following error:
```
can't open csharp/Newtonsoft.Json.dll for input
can't open csharp/RestSharp.dll for input
```
Integration test looks good:
screen shot 2015-07-30 at 10 25 12 am


wing328:
@who I added back the DLL to csharp template to avoid the error and to make compile.bat works again. 

README updated with instruction to download the latest version of the DLL via NuGET.

Please let me know if you've any concern.

PR #1041, opened 2015-08-03
Support to array of enums

Rogercf:
Added support to array of enums in the DefaultCodegen (it used to generate an array of strings instead) and updated the java template to use it.
Tested only in Java and it is working fine. I don't think there will be issues with the other languages (they won't use my changes and will just ignore it).


wing328:
@Rogercf please rebase the PR based on the latest develop_2.0

Rogercf:
@wing328 sorry for the late response
@fehguy thanks for merging

PR #1045, opened 2015-08-05
[PHP] APIClient connectors

harnash:
This change introduces `APIConnectors` which are used by `APIClients`
to send requests to API servers. 
`APICurlClient` facilitates current cURL
functionality and separates connection specific logic into separate
class to allow different *backends* such as Guzzle or adding some
platform specific logic for all or some API's/resources. `APIConnector` responsibility is to make requests to an given server, process response and decode it if needed.

Samples regenerated with `mvn clean && ./bin/php-petstore.sh`

Tests:
```
./vendor/bin/phpunit tests




PHPUnit 4.7.7 by Sebastian Bergmann and contributors.

............

Time: 6.37 seconds, Memory: 5.75Mb

OK (12 tests, 277 assertions)
```


wing328:
@harnash thanks for the PR. If the goal is to generate PHP SDKs with customised HTTP library (e.g. cURL,  Guzzle), shall we create a separate ApiClient.mustache for each HTTP library and use the CLI option `-c` to determine which one to use?

harnash:
That would be nice! I see at least two places where one could define `ApiConnector` used by `ApiClient` one is as you mentioned in `ApiClient::__construct` and perhaps another one is in `Configuration` which I think is a better place. What do you think?

wing328:
Please take a look at #1046 for Java. The PR let the user pick the HTTP library (Jersey v1 or v2) and the files need to be customised for different HTTP library are `ApiClient.mustache` and `pom.xml`. PHP client can follow similar approach.

harnash:
Thanks. I will take a look at it. For the time being please do not merge this PR.

webron:
@harnash - any update?

webron:
Closing this due to lack of activity. If you still want us to merge it, please follow @wing328's advice and resubmit against the `master` branch.

harnash:
Right. Sorry got my hands full. I will come back with updated code.

Issue #993, opened 2015-07-21
How to catch errors when using swagger-codegen as a library


hgwood:
I've managed to generate code using `DefaultGenerator`. But `DefaultGenerator#generate` catches all exceptions and prints the stacktrace without rethrowing. This makes it impractical for the caller code to know if the generation was successful or not. `DefaultGenerator` was clearly not intended to be used outside of a CLI case.

How am I supposed to use swagger-codegen as Java library if not through `DefaultGenerator`? Or is it that error management was not tackled yet?

Is that latter case, I'm willing to make a patch. The patch would have to change the `Generator` interface to include a `throws` clause, except if you'd rather have (documented) unchecked exceptions.

xhh:
Agreed that the behavior of hiding exceptions would be confusing.
As in [`DefaultGenerator#generate`](https://github.com/swagger-api/swagger-codegen/blob/develop_2.0/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java#L50):

```java
public List<File> generate() {
    if (swagger == null || config == null) {
        throw new RuntimeException("missing swagger input or config!");
    }
    /* ... */
    List<File> files = new ArrayList<File>();
    try {
        /* generating code ... */
    } catch (Exception e) {
        e.printStackTrace();
    }
    return files;
}
```

In cases of null swagger/config, RuntimeException would be thrown, while exceptions happening in the process of code generation are hidden (print stack trace and then get ignored).
I think it would be better to throw the exception again after `e.printStackTrace()`, as long as there's no particular reasons not to do it.

fehguy:
This should be handled now, please reopen if not.

PR #1052, opened 2015-08-05
Default Generator Status

tiffanyperumpail:
Change to DefaultGenerator:
When the generate() method throw an exception, the exception is caught and a stack trace is printed. Due to the way this exception is caught, it is difficult to detect if code generation has failed when we aren't generating code through the command line. 

For this reason, I added enum CodeGenStatus, so a status can be set for each instance of the DefaultGenerator.

This change is inspired by our needs at Wikia, where we are using Swagger to generate code progammatically and would like to easily detect when generation fails.

Please let us know if we've missed anything or didn't take any factors into account with this change.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -456,7 +456,7 @@
         </repository>
     </repositories>
     <properties>
-        <swagger-parser-version>1.0.9-SNAPSHOT</swagger-parser-version>
+        <swagger-parser-version>1.0.9</swagger-parser-version>
         <scala-version>2.11.1</scala-version>
         <felix-version>2.3.4</felix-version>
         <swagger-core-version>1.5.0</swagger-core-version>
--- a/pom.xml
+++ b/pom.xml
@@ -456,7 +456,7 @@
         </repository>
     </repositories>
     <properties>
-        <swagger-parser-version>1.0.9</swagger-parser-version>
+        <swagger-parser-version>1.0.9-SNAPSHOT</swagger-parser-version>
         <scala-version>2.11.1</scala-version>
         <felix-version>2.3.4</felix-version>
         <swagger-core-version>1.5.0</swagger-core-version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -456,7 +456,7 @@
         </repository>
     </repositories>
     <properties>
-        <swagger-parser-version>1.0.9-SNAPSHOT</swagger-parser-version>
+        <swagger-parser-version>1.0.9</swagger-parser-version>
```
wing328: Thanks for the PR. May I know why you need to use swagger parser 1.0.9 instead of 1.0.9-SNAPSHOT ?

tiffanyperumpail: My changes don't depend on using swagger parser 1.0.9 instead of 1.0.9-SNAPSHOT; I made that change because the Travis CI failed to build until I bumped the version of swagger-parser to 1.0.9. I could change it back if necessary.

screen shot 2015-08-06 at 10 48 44 am

who: @wing328 There appears to be a problem with one of the dependencies, even with a fresh, unmodified clone. I filed a bug here: https://github.com/swagger-api/swagger-codegen/issues/1055

fehguy: I have no idea why that snapshot was deleted--I've redeployed it so 1.0.9 should now be available.

wing328:
@hgwood would this PR meet your requirement to better catch errors for DefaultGenerator? (ref: #993)

hgwood:
Yes it would. FYI, [here's](https://github.com/restlet/swagger-codegen/commit/253e95c911e8032cbfe4a896572e6bbaf84b0f46) what we currently have at Restlet.

However, there are [many other places](https://github.com/swagger-api/swagger-codegen/search?utf8=%E2%9C%93&q=printStackTrace) in swagger-codegen where exceptions are swallowed. So I don't trust it to indicate failure even after this PR is merged.

To really fix #993 without too much work, *all* `e.printStackTrace` could be replaced by `throw RuntimeException(e)`, and try-catch blocks could be added in main classes (`io.swagger.codegen.SwaggerCodegen` and `io.swagger.codegen.Codegen`), where `e.printStackTrace` is expected.

Alternatively, another system of error handling could be put in place, maybe with checked exceptions, enum return values (like in this PR) or monads.

In conclusion, although this PR is interesting and probably useful in the short term, it does not fix the whole problem. I ultimately think the dev team needs to decide how to handle errors project-wide. Once that's done, I'd be willing to help with the implementation, and I'm sure @tiffanyperumpail would be too.

Issue #1061, opened 2015-08-07
Java 'path' query string parameter causes compile error


RMarinaro:
Using `path` as a query string parameter causes the java client to generate something like this
```
public void somemethodGet (String path) throws ApiException {
    Object postBody = null;
    
    // verify the required parameter 'path' is set
    if (path == null) {
       throw new ApiException(400, "Missing the required parameter 'path' when calling fileopsDeletePost");
    }
    
    // create path and map variables
    String path = "/fileops/delete".replaceAll("\\{format\\}","json");
```

```
Variable 'path' is already defined in the scope
```

Unfortunately, I am unable to change the query string parameter of the API I'm using (dropbox). A possible fix would be adding local parameters to the [reserved list](https://github.com/swagger-api/swagger-codegen/blob/develop_2.0/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java#L34) but that would make _path an invalid query string parameter.  Another solution would be to qualify local parameters in the mustache file with some 'codegen' prefix.  If there are any other workarounds I'd be happy to go that route instead. 

Thanks!

wing328:
@RMarinaro thanks for reporting the issue. I think other clients (e.g. PHP, Python) may encounter similar issue. My suggested solution is to name the local variable differently (e.g. with _Swagger prefix). Naming collision may still occur but less likely.

RMarinaro:
Thanks for the quick reply @wing328, is there any way to add this to the language configuration? e.g. "variablePrefix": "_swagger"? Otherwise the best practice would be to fork, and tweak the mustache file?

wing328:
I agree that ideally the local variable prefix should be configurable. If it's a bit too much work, I'll implement that as a day 2 requirement. Would that be ok ?

RMarinaro:
I'm happy to take a look at it (if you guys are accepting pulls) for at least the java stuff, but I'm not sure where these config files are parsed. Can you point me in the right direction and let me know the general strategy for adding a new json configuration parameter?

webron:
We thrive on PRs.

RMarinaro:
Is there any way to run the integration tests locally, or do contributors just open up PRs? I apologize if this is answered somewhere else. Thanks!

wing328:
@RMarinaro sorry for missing your question. To run the test locally, please run `mvn verify -Psamples`

PR #1064, opened 2015-08-08
fix for #1061 java

RMarinaro:
The only problematic spot was the api.mustache file, although if more parameterized templates were to evolve in the other mustache files we could definitely add {{localVariablePrefix}} to those too.  For now I didn't want to add any extra noise to the template files which didn't need the prefixes.


wing328:
@RMarinaro thanks for the PR.

Integration test looks good:
```
[INFO] BUILD SUCCESS
[INFO] Total time: 33.003 s
[INFO] Finished at: 2015-08-10T18:46:50+08:00
[INFO] Final Memory: 20M/249M
```

I'll merge this PR tomorrow (Tue)

RMarinaro:
Thanks! Someone should probably think about adding an integration test for java with a 'path' query string parameter.  I'll look into adding this feature for the other languages in a week or so.

Issue #996, opened 2015-07-21
JaxRSServerCodegen overrides JavaClientCodegen's getTypeDeclaration() method with an incorrect implementation


Helmsdown:
So JavaClientCodegen has a good implementation of getTypeDeclaration() which correctly handles MapProperty objects.

```java
    public String getTypeDeclaration(Property p) {
        Property inner;
        if(p instanceof ArrayProperty) {
            ArrayProperty mp1 = (ArrayProperty)p;
            inner = mp1.getItems();
            return this.getSwaggerType(p) + "<" + this.getTypeDeclaration(inner) + ">";
        } else if(p instanceof MapProperty) {
            MapProperty mp = (MapProperty)p;
            inner = mp.getAdditionalProperties();
            return this.getSwaggerType(p) + "<String, " + this.getTypeDeclaration(inner) + ">";
        } else {
            return super.getTypeDeclaration(p);
        }
    }
```

JaxRSServerCodegen overrides this method with and drops some important code that handles MapProperty objects.

```java
    public String getTypeDeclaration(Property p) {
        Property inner;
        if(p instanceof ArrayProperty) {
            ArrayProperty mp1 = (ArrayProperty)p;
            inner = mp1.getItems();
            return this.getSwaggerType(p) + "<" + this.getTypeDeclaration(inner) + ">";
        } else if(p instanceof MapProperty) {
            MapProperty mp = (MapProperty)p;
            inner = mp.getAdditionalProperties();
            return this.getTypeDeclaration(inner);
        } else {
            return super.getTypeDeclaration(p);
        }
    }
```

The JavaClientCodegen implementation seems like it would work when generating the JaxRS template. 

Can we just get rid of the overriding method in JaxRSServerCodegen? WDYT?

olensmar:
sounds reasonable to me - @fehguy - any objections?

wing328:
@Helmsdown thanks for the suggestion. There's a similar PR submitted but for `SpringMVCServerCodegen`

wing328:
PR for SpringMVCServerCodegen: https://github.com/swagger-api/swagger-codegen/pull/961

Helmsdown:
Included a fix for this in #1067

Helmsdown:
Fixed via #1067. Closing.

PR #1067, opened 2015-08-10
Codegen should honor path level parameters

Helmsdown:
Currently the codegen ignores Path level params. For example, the following Path and Operation would get converted into a Codegen operation with no parameters, even though there is a path level header parameter.

```yaml
swager: 2.0
paths:
  /people :
    parameters:
      name: "Authorization"
      in: "header"
      description: "authorization ticket"
      required: true
      type: "string"
    get:
       description: Get all the people entities
       response:
         200:
           schema:
             $ref: #/definitions/People
```

Additionally, included a fix for #996 while I was at it.

As always feedback on my code is appreciated.


wing328:
Tested with parameters added to the path level in the spec and got the expected result.

PR #1068, opened 2015-08-11
Add @Generated annotation to all generated Java templates

Helmsdown:
The Java API has an [annotation](http://docs.oracle.com/javase/7/docs/api/javax/annotation/Generated.html) for marking classes that are generated. 

It is often useful to know which class generated the code and and the date-time it was generated at. 

I added a new sub-template in each Java templates directory (i.e. Java, JavaJaxRS, JavaInflector and JavaSpringMVC) and updated each Java class template in the aforementioned directories to use it. 

As always feedback is appreciated.


wing328:
Testing with Java and SpringMVC looks good

PR #1076, opened 2015-08-13
improve static html; add template guards for collections. 

DavidBiesack:
Update static html template to improve navigation, with Up links and jumps between
methods and modules. Added guards around different parameters, to eliminate
empty sections.

See https://gist.github.com/DavidBiesack/a9048b151cd7935c8d76 for sample output from the Uber example shipped with swagger-editor

I also updated the swagger-core dependency to build with most recent as of 2015-08-13

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
@@ -30,6 +30,23 @@ public class CodegenOperation {
     public List<Map<String, String>> examples;
     public ExternalDocs externalDocs;
 
+    private boolean nonempty(List<CodegenParameter> params)
+    {
+        return params != null && params.size() > 0;
+    }
+    public boolean getHasBodyParam() {
+        return nonempty(bodyParams);
+    }
+    public boolean getHasQueryParams() {
+        return nonempty(bodyParams);
+    }
+    public boolean getHasHeaderParams() {
+        return nonempty(bodyParams);
+    }
+    public boolean getHasPathParams() {
+        return nonempty(bodyParams);
+    }
+
     // legacy support
     public String nickname;
 }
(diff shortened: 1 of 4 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
```diff
@@ -30,6 +30,23 @@
...
+    public boolean getHasQueryParams() {
+        return nonempty(bodyParams);
+    }
+    public boolean getHasHeaderParams() {
+        return nonempty(bodyParams);
+    }
+    public boolean getHasPathParams() {
+        return nonempty(bodyParams);
+    }
+
```
wing328: @DavidBiesack thanks for the PR to improve the HTML documentation. Instead of creating the function above to see if there's any body/query/header/path parameters, have you considered using `.length` provided by mustache template? 

Here is an example:
```
    {{#queryParams.length}}
    List<Pair> {{localVariablePrefix}}queryParams = new ArrayList<Pair>();
    {{/queryParams.length}}
```
The queryParameter variable would be declared once only if there's at least one query parameter.

DavidBiesack: Thanks, I was not aware one could use .length in this way (I don't see any uses of that in the existing .mustache files). 

I had seen several `has*` boolean fields (`hasConsumes, hasProduces, hasMore`...) in CodegenOperation.java and wanted to continue that since `{{#hasConsumes}}` seems like a useful pattern. But those fields require initializing in distant classes, which felt wrong. The methods felt like a better implementation.

The property methods I provided are safer in that they work even if the collection is null (most of the other fields are public but not marked final, so there are no guarantees that they are always non-null -- hopefully not, but the method is more robust).

wing328: @DavidBiesack ok. your approach is definitely fine with me.

FYI. I found the .length approach in https://www.vodori.com/blog/helpful-mustache-template-tips/. Hope you also find it useful.

Issue #1050, opened 2015-08-05
[Java] Generated code does not compile when using enums in inherited models


who:
Given this spec snippet

```
{
    "SuperObject": {
        "type": "object",
        "properties": {
            "thingType": {
                "type": "string",
                "enum": [
                    "FOO",
                    "BAR"
                ]
            }
        }
    },
    "SubObject": {
        "allOf": [
            {
                "$ref": "#/definitions/SuperObject"
            },
            {
                "type": "object",
                "properties": {
                    "thingType": {
                        "type": "string",
                        "enum": [
                            "FOO",
                            "BAR"
                        ]
                    },
                    "anotherProperty": {
                        "type": "string"
                    }
                }
            }
        ]
    }
}
```

The swagger-codegen output for java will not compile. It generates an enum for both the parent and the child class, and generates a getter&setter for each. However, this will not compile in java, because the return type of the enum getter in the subtype is not equal to the enum type in the parent class, and therefore the override is not proper.

who:
Here's a repro:

```
me@box:/tmp/tmp.SUB54PNhX$ cd `mktemp -d /tmp/tmp.XXXXXXXXX`
me@box:/tmp/tmp.SUB54PNhX$ git clone https://github.com/swagger-api/swagger-codegen.git -q .
me@box:/tmp/tmp.SUB54PNhX$ git checkout develop_2.0
me@box:/tmp/tmp.SUB54PNhX$ mvn clean install 
me@box:/tmp/tmp.SUB54PNhX$ java -jar ./modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -i https://gist.githubusercontent.com/who/2ad48a523ce91b646ba7/raw/e9cda73c4bd523fa04bfb59484d6fbfebf75baa2/repro_1051.json -l java -o output
me@box:/tmp/tmp.SUB54PNhX$ cd output; mvn compile;
```

Repro output:
```
[INFO] Scanning for projects...
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for io.swagger:swagger-java-client:jar:1.0.0
[WARNING] 'build.plugins.plugin.version' for org.codehaus.mojo:build-helper-maven-plugin is missing. @ line 68, column 15
[WARNING] 
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING] 
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING] 
[...]

PR #1094, opened 2015-08-20
Java support inheritance

ahgittin:
as per #1015 (and #946) this is now optional and disabled by default. this results in extra unnecessary code, masking fields in some cases, and breaks code generated for jax-rs in certain cases (enums with inheritance).

as noted by @xhh at #1050 this is not a perfect fix, but it improves the situation significantly (restoring `master` branch behaviour, as the current situation is a regression; my spec builds fine with master but not with `develop_2.0` ?).


wing328:
@ahgittin do you mind sharing the spec with us assuming it can be shared publicly?

ahgittin:
@wing328 here's a cut-down test case whose `-l jaxrs` makes for invalid java without this, valid java with this:

```
swagger: '2.0'
info:
  title: buggy inheritance for java
  version: "0.0.1"
definitions:
  Super:
    properties:
      my_enum:
        type: string
        enum: [ A, B ]

  Sub:
    allOf:
    - $ref: '#/definitions/Blueprint'
    - type: object
    properties:
      another:
        type: string

paths:
  /home:
    get:
      summary: with develop_2.0 prior to #1094 the java generated fails
```

ahgittin:
also noticed this had some strange commits; i've rebased so it's just the one-line change

webron:
FWIW, that's not inheritance, that's composition. In this case, `Sub` and `Super` have no parent-child relationship. The end result in this case would be that `Sub` would include the same properties as `Super` but will not inherit from it.

ahgittin:
sorry @webron i mis-typed the example, it's now correct, and it definitely is implemented as inheritance (maybe it shouldn't be):

```
public class Sub extends Super {
  
  public enum MyEnumEnum {
     A,  B, 
  }
`

redeclaring the enum breaks compilation as `Super.getMyEnum()` returns the parent's `MyEnumEnum` so `Sub.getMyEnum()` has an incompatible return type

HTH

webron:
Not sure what's changed, but as long as there's no `discriminator`, there's no inheritance (from the spec's point of view).

wing328:
If we look at the java client in the "master" branch, we'll find that it implements `allOf` using `extends` in Java, which is not the correct implementation as Ron has pointed out that `allOf` is model composition (*not* inheritance)

@xhh previously attempt to fix `allOf` with #946 (copy parent properties) but it does not implement `discriminator` (should be implemented as `extends` in Java) due to time constraint. 

@tandrup introduced `supportsInheritance` to let people choose the behaviour with #1015.

Before someone from the community implements `discriminator`, we can merge this PR for the time being to keep the old (master) behaviour if no one has any concern.

webron:
I'd suggest putting a hold on the `discriminator` implementation as there's a technical detail with the spec we need to resolve first. It'll be done with in the upcoming weeks, and will come with an implementation (or suggestion for implementation). Sorry for the trouble on this one.

ahgittin:
closing and will resubmit against master as per change to swagger PR process

ahgittin:
closed in favour of #1120

steverice:
> there's a technical detail with the spec we need to resolve first

@webron  are you referring to https://github.com/swagger-api/swagger-spec/issues/403?

webron:
That is correct. It should be resolved soon.

PR #1105, opened 2015-08-22
[Java] Adding an option so java models can implement Serializable

who:
I found today that models are not cacheable because they do not implement `java.io.Serializable`.  I have added this as an option.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
@@ -35,6 +35,8 @@ public class JavaClientCodegen extends DefaultCodegen implements CodegenConfig {
     protected String artifactVersion = "1.0.0";
     protected String sourceFolder = "src/main/java";
     protected String localVariablePrefix = "";
+    protected Boolean serializableModel = false;
+    
     public JavaClientCodegen() {
         super();
         outputFolder = "generated-code/java";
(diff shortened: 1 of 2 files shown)


wing328:
Test result looks good for both serializableModel set to true/false
```
Tests run: 33, Failures: 0, Errors: 0, Skipped: 0

[INFO] BUILD SUCCESS
[INFO] Total time: 24.948 s
[INFO] Finished at: 2015-08-23T17:08:34+08:00
[INFO] Final Memory: 13M/245M
```

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
```diff
@@ -129,6 +132,12 @@ public void processOpts() {
         if (additionalProperties.containsKey("localVariablePrefix")) {
             this.setLocalVariablePrefix((String) additionalProperties.get("localVariablePrefix"));
         }
+        
+        if (additionalProperties.containsKey("serializableModel")) {
+            this.setSerializableModel(Boolean.valueOf((String)additionalProperties.get("serializableModel")));
```
fehguy: This is causing a ClassCastException because the `serializableModel` value is a Boolean

wing328: @fehguy I tested again with the `serializable_java_models` branch but couldn't repeat the `ClassCastException`. (test cases: serializableModel no set, serializableModel set to true, serializableModel set to false)

I discovered another issue that `Serializable` is used even though `serializableModel` is set to false (previously I tested with true and no value set). I'll submit a PR to address it.

fehguy: I think the `JavaClientCodegen.java` needs this at line 142:

```
this.setSerializableModel(Boolean.valueOf(additionalProperties.get("serializableModel").toString()));
```

wing328: OK. will incorporate that in the fix and perform testing.

wing328: Fixed and tested in #1114

PR #1109, opened 2015-08-23
update files as rebase from #838

fehguy:
This is a rebase of #838 which was done against outdated code.


fehguy:
@wing328 can you see that this is OK?

webron:
LGTM

fehguy:
thanks

PR #1114, opened 2015-08-24
[Java] fix serializableModel (boolean)

wing328:
Fix serializableModel in config.json for Java

Tested with no serializableModel in config, serializableModel set to false and serializableModel set to true. Only serializableModel set to true outputs "implements Serializable" (and the corresponding import)


fehguy:
Merging this, and setting swagger-core to 1.5.3 instead of 1.5.4-SNAPSHOT

PR #1098, opened 2015-08-21
Fix possible NullPointerException

Edubits:
getWrapped() returns a Boolean which can be null

PR #1121, opened 2015-08-24
Fix possible NullPointerException

Edubits:
As requested in #1098 re-submitting this PR with the master branch.

Issue which is solved: getWrapped() returns a Boolean which can be null

With https://zeef.io/api/swagger.json the problem is there. Since the merge from develop_2.0 to master it does not display the NPE anymore while generating the (in my case Swift) library, but it generates an incomplete one.


wing328:
Was able to repeat the issue with the master and confirmed the PR fixed it.

Edubits:
Thanks @wing328!

wing328:
@Edubits btw, I noticed that the spec you provided has reference to `#/definitions/APIProfile` but I don't see APIProfile defined. Same for `APIUser` and `APISubject`.

Edubits:
@wing328 Thanks for the note! We're aware of this: e.g. APIUser should be User. The bug causing it is fixed in a later version of Swagger (JaxRS), only we've got a few other things to fix before we can use it.

PR #1065, opened 2015-08-08
Develop 2.0 - support binary input and output

boazsapir:
This is my implementation of Java code generation, including the generic (not language specific) support needed, for binary data in the request body and in the response.
This implementation is based on the agreed spec, as discussed in https://github.com/swagger-api/swagger-spec/issues/50  and depends on my pull request to swagger-core which was merged on Aug-6 (https://github.com/swagger-api/swagger-core/pull/1335)

PR #1124, opened 2015-08-24
support binary input and output

boazsapir:
This is my implementation of Java code generation, including the generic (not language specific) support needed, for binary data in the request body and in the response.
This implementation is based on the agreed spec, as discussed in swagger-api/swagger-spec#50 and depends on my pull request to swagger-core which was merged on Aug-6 (swagger-api/swagger-core#1335)
This PR replaces #1065

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/api.mustache
@@ -46,16 +46,16 @@ public class {{classname}} {
 {{/allParams}}   * @return {{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}void{{/returnType}}
    */
   public {{#returnType}}{{{returnType}}} {{/returnType}}{{^returnType}}void {{/returnType}}{{nickname}} ({{#allParams}}{{{dataType}}} {{paramName}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) throws ApiException {
-    Object {{localVariablePrefix}}postBody = {{#bodyParam}}{{paramName}}{{/bodyParam}}{{^bodyParam}}null{{/bodyParam}};
+    Object {{localVariablePrefix}}postBody = {{#bodyParam}}{{^isBinary}}{{paramName}}{{/isBinary}}{{#isBinary}}null{{/isBinary}}{{/bodyParam}}{{^bodyParam}}null{{/bodyParam}};
+    byte[] {{localVariablePrefix}}postBinaryBody = {{#bodyParam}}{{#isBinary}}{{paramName}}{{/isBinary}}{{^isBinary}}null{{/isBinary}}{{/bodyParam}}{{^bodyParam}}null{{/bodyParam}};
     {{#allParams}}{{#required}}
-    // verify the required parameter '{{paramName}}' is set
-    if ({{paramName}} == null) {
-       throw new ApiException(400, "Missing the required parameter '{{paramName}}' when calling {{nickname}}");
-    }
-    {{/required}}{{/allParams}}
-
+     // verify the required parameter '{{paramName}}' is set
+     if ({{paramName}} == null) {
+        throw new ApiException(400, "Missing the required parameter '{{paramName}}' when calling {{nickname}}");
+     }
+     {{/required}}{{/allParams}}
     // create path and map variables
-    String {{localVariablePrefix}}path = "{{path}}".replaceAll("\\{format\\}","json"){{#pathParams}}
+    String {{localVariablePrefix}}path = "{{{path}}}".replaceAll("\\{format\\}","json"){{#pathParams}}
       .replaceAll("\\{" + "{{baseName}}" + "\\}", {{localVariablePrefix}}apiClient.escapeString({{{paramName}}}.toString())){{/pathParams}};
 
     // query params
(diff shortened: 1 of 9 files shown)


modules/swagger-codegen/src/main/resources/Java/api.mustache
```diff
@@ -86,12 +86,26 @@ public class {{classname}} {
...
     {{/returnType}}{{^returnType}}
-    {{localVariablePrefix}}apiClient.invokeAPI({{localVariablePrefix}}path, "{{httpMethod}}", {{localVariablePrefix}}queryParams, {{localVariablePrefix}}postBody, {{localVariablePrefix}}headerParams, {{localVariablePrefix}}formParams, {{localVariablePrefix}}accept, {{localVariablePrefix}}contentType, {{localVariablePrefix}}authNames, null);
+    {{localVariablePrefix}}apiClient.invokeAPI({{localVariablePrefix}}path, "{{httpMethod}}", {{localVariablePrefix}}queryParams, {{localVariablePrefix}}postBody, {{localVariablePrefix}}postBinaryBody, {{localVariablePrefix}}headerParams, {{localVariablePrefix}}formParams, {{localVariablePrefix}}accept, {{localVariablePrefix}}contentType, {{localVariablePrefix}}authNames, null);
     {{/returnType}}
+    {{/isBinary}}
+
+    {{/isDefault}}
+    {{/responses}}
```
xhh: @boazsapir the above changes seem to make integration tests fail for the Petstore sample:

```shell
mvn clean package
./bin/java-petstore.sh
cd samples/client/petstore/java/default
mvn clean test
```

```
Results :

Failed tests:   testFindPetsByStatus(io.swagger.petstore.test.PetApiTest)
  testFindPetsByTags(io.swagger.petstore.test.PetApiTest)

Tests in error:
  testUpdatePetWithForm(io.swagger.petstore.test.PetApiTest): {"code":1,"type":"error","message":"Pet not found"}
  testDeletePet(io.swagger.petstore.test.PetApiTest): {"code":1,"type":"error","message":"Pet not found"}
  testCreateAndGetPet(io.swagger.petstore.test.PetApiTest): {"code":1,"type":"error","message":"Pet not found"}
  testUpdatePet(io.swagger.petstore.test.PetApiTest): {"code":1,"type":"error","message":"Pet not found"}

Tests run: 33, Failures: 2, Errors: 4, Skipped: 0
```

In the `addPet` method of `PetApi`, the `apiClient.invokeAPI` statement is missing, so no HTTP request is sent at all.

boazsapir:
@xhh the failure is because addPet (also updatePet, updatePetWithForm, deletePet ) do not have a default response in their definition, only error responses. See this code in DefaultCodegen.java:
    private Response findMethodResponse(Map<String, Response> responses) {

        String code = null;
        for (String responseCode : responses.keySet()) {
            if (responseCode.startsWith("2") || responseCode.equals("default")) {
                if (code == null || code.compareTo(responseCode) > 0) {
                    code = responseCode;
                }
            }
        }
        if (code == null) {
            return null;
        }
        return responses.get(code);
    }

If I add 
          "200": {
            "description": "OK"
          },
to the responses section of addPet, updatePet, updatePetWithForm, deletePet - then the test succeeds.
Is this a valid situation to have an operation without a default response?

xhh:
I think you're right that at least a successful or default response should be defined. According to [swagger spec](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#responses-object):
> it is expected from the documentation to cover a successful operation response and any known errors

The thing is, swagger-codegen did work before even with this "invalid" spec of petstore, so either we should keep tolerant to this case in swagger-codegen, or fix the petstore spec to make it work.

webron:
"SHOULD" - the definition for the pet store is valid. Could be improved, but valid.

boazsapir:
ok, I will fix the template to support also the case of no default response

modules/swagger-codegen/src/main/resources/Java/api.mustache
boazsapir: @xhh while trying to test my fix I encountered a new problem in JavaClientCodegen.PostProcessModels, which blocks me from testing.

row 341: 
               List<String> values = (List<String>) allowableValues.get("values");

'values' is set to null and is later referenced (row 344).

This is the definition that causes the crash:

       "DerivationInput": {
            "description": "Key derivation input",
            "required": [
                "length",
                "data"
            ],
            "properties": {
                "length": {
                    "description": "Derived key size",
                    "type": "integer",
                    "maximum": 8192,
                    "minimum": 1
                },
                "data": {
                    "description": "Data for key derivation",
                    "type": "string",
                    "maxLength": 256
                }
            },
            "type": "object"
        },

xhh: sorry for that, my bad. I believe it is fixed via #1144, could you add a null check when testing?

boazsapir: I merged with the upstream branch to get fix #1144 and this solved the problem.
Now I encounter a new problem:
I have this enumeration defined:

                "cloud": {
                    "description": "cloud provider",
                    "type": "string",
                    "enum": [
                        "ec2",
                        "vmware",
                        "openstack"
                    ]
                },

The generated code has a syntax error: missing return type (should be void):

public enum CloudEnum {
  eC2("ec2"), VMWARE("vmware"), OPENSTACK("openstack");

  private String value;

  StatusEnum(String value) {
    this.value = value;
  }

  @Override
  public String toString() {
    return value;
  }
}

Do you know anything about it?

xhh: oops, `StatusEnum` should be `CloudEnum` there, I'll submit a PR to fix it

xhh: submitted #1168 to fix it

boazsapir: submitted PR #1169 to fix the original problem

PR #1126, opened 2015-08-24
[Java] Support special characters in enum values

### LINKED ISSUE #1102: Need to avoid language keywords in enums
I've been trying out some Java API client code gen on a few APIs and encountered some issues with the enum generation due to the way the values are placed directly inside a Java enum. I hit both keyword issues and enum value starting with a numeric, both of which are not allowed in Java which means the code generated fails to compile.

One example I was trying to generate a client API for is the Runscope API:
- "interval" on https://www.runscope.com/docs/api/schedules starts with a numeric
- "emails.notify_on" on https://www.runscope.com/docs/api/environments uses "switch" which is a Java keyword

I could not find anything in the Swagger Spec to define this, but perhaps I missed something. Otherwise perhaps this should be dealt with at the code generation level somehow. For example, something like this while just returning the JSON property as a String (using getValue below):

```
public enum IntervalEnum {
  onem("1m"),  fivem("5m"),  fifteenm("15m");

  private String value;

  public IntervalEnum(String value) {
    this.value = value;
  }

  public String getValue() {
    return value;
  }
};
```

I'm pretty new to this project so wanted some feedback before working on a pull request.

xhh:
Fix #1102: "Need to avoid language keywords in enums"

* Use upper case for enum names to comply with [convention](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html)
* Prepend enum name with underscore if it starts with number

Example code:

```java
public enum StatusEnum {
  AVAILABLE("available"), PENDING("pending"), SOLD("sold"), SWITCH("switch"), _1M("1m");

  private String value;

  StatusEnum(String value) {
    this.value = value;
  }

  @Override
  public String toString() {
    return value;
  }
}
```

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
@@ -318,6 +318,31 @@ public class JavaClientCodegen extends DefaultCodegen implements CodegenConfig {
         return codegenModel;
     }
 
+    @Override
+    public Map<String, Object> postProcessModels(Map<String, Object> objs) {
+        List<Object> models = (List<Object>) objs.get("models");
+        for (Object _mo : models) {
+            Map<String, Object> mo = (Map<String, Object>) _mo;
+            CodegenModel cm = (CodegenModel) mo.get("model");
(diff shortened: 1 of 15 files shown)


c-knowles:
@xhh Looks great to me and will solve the issues with Java keywords in the enums, thanks not only the quick response but the pull too!

Issue #990, opened 2015-07-20
responseHeader annotations handling


jeusdi:
I believe, actually the responseHeader annotations can not be translated on code.

wing328:
@jeusdi Would you please elaborate on the issue ? is it related to spring-mvc ?

jeusdi:
Hi @wing328.

Take a look at [here](https://gist.github.com/56bba54643c509e81e31.git).

As you can see, I've implemented two update operation, one on `POST`, and the other one on `PUT`.
The difference is I'm using `POST` verb to create a new resource on my system, and `PUT` to update it. 

So, when it's created, I return the object id using the `Location` header, or I'd like to use another customized one, for example `InstanceId`. As you can see I'm using `ResponseHeader` annotation to describe this.

I'm generating a C# client, however, I don't know:

* Is this information added on mustache objects?  
* How can issue this information on templates...

wing328:
@jeusdi  Currently, the response header is not used by the auto-generated API client. There's a way to get the HTTP response header from the last HTTP call via `RestClient`, would that meet your requirement?

ivanmartinvalle:
@wing328 I am having the same issue, specifically with the generated Java client. I need to access the value of the Location header on POST. What would be the best way to move forward? Modify the generator so that all response objects implement an interface that has the headers and status code of the response?

wing328:
@ivanmartinvalle For Java client, if we add a member variable (e.g. httpHeader) in `ApiClient` class to store the HTTP header of the last HTTP response, would that meet your requirement ?

ivanmartinvalle:
@wing328 That would cause issues in a multi-threaded environment, but is probably fine for now, especially since it's probably best practice to spin up a new instance of a client per HTTP request anyway.

Adding a member variable of all HTTP response headers AND the HTTP status code (both are really important) would indeed meet my requirements.

wing328:
@ivanmartinvalle yes, for API client in other languages (e.g. PHP), we also recommend an instance of ApiClient per thread to avoid any potential multi-thread issue.

For Java, we're working on updating the template to support different version of Jersey library (and support other HTTP libraries in the future). We'll work on the support of HTTP response headers and status code after that.

evgenynacu:
Hi.
I have a proposal. Take a look at https://gist.github.com/evgenynacu/72c0d3300a4cadc3dca414910064a524

[...]

PR #1127, opened 2015-08-24
Add status code and response headers of the last request to java and c#

ivanmartinvalle:
Per #990 

I would like to add the ability to return the status code and response headers from the previous API call from java and csharp. The python client already provides this ability via the last_response variable. This is necessary for our API, as we need the value of the Location header after a POST request.

Approach: Before returning the response body, set the non-body parts of the response on ApiClient.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/ApiClient.mustache
@@ -49,6 +49,9 @@ public class ApiClient {
 
   private Map<String, Authentication> authentications;
 
+  private int statusCode;
+  private Map<String, List<String>> responseHeaders;
+
   private DateFormat dateFormat;
 
   public ApiClient() {
@@ -78,6 +81,20 @@ public class ApiClient {
   public ApiClient setBasePath(String basePath) {
     this.basePath = basePath;
     return this;
+  }
+
+  /**
+   * Gets the status code of the previous request
+   */
+  public int getStatusCode() {
+    return statusCode;
+  }
+
+  /**
+   * Gets the response headers of the previous request
+   */
+  public Map<String, List<String>> getResponseHeaders() {
+    return responseHeaders;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


ivanmartinvalle:
@wing328 I would like to get your suggestion merged, but it appears that there is a failure on one of the scala tests. It happens with and without my changes locally (and they definitely don't look related). Any insight as to what I can do to move forward?

fehguy:
@ivanmartinvalle I'm seeing the tests pass via travis-ci.  Can you run off a clean master branch?  What is the error you're getting?

ivanmartinvalle:
@fehguy the last commit before mine is "Merge pull request #1123 from wing328/python_typo".

The error is in swagger-scala-client
Tests in error: 
  UserApi should fetch a user(UserApiTest): 12 was not equal to 1

wing328:
@ivanmartinvalle my commit (#1123) does not change the Scala client. Let me look into that as I suspect it's due to data change in the server side (petstore server)

wing328:
@ivanmartinvalle the user record looks good:
```
{
  "id": 1,
  "username": "user1",
  "firstName": "first name 1",
  "lastName": "last name 1",
  "email": "email1@test.com",
  "password": "XXXXXXXXXXX",
  "phone": "123-456-7890",
  "userStatus": 1
}
```

let me reopen this PR to trigger the test again and see if the error persists.

wing328:
@ivanmartinvalle now the tests passed.

wing328:
@ivanmartinvalle thanks for the PR to add status and response header to API client of Java and C#.

Testing looks good.

wing328:
UPDATE: we've replaced this approach (varaibles to store the last status code and response headers) with an additional method WithHttpInfo, e.g. https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/java/okhttp-gson/src/main/java/io/swagger/client/api/PetApi.java#L123-L137

PR #1132, opened 2015-08-25
[Python] Add option `verify_ssl` to switch SSL/TLS verification

geekerzp:
If set `swagger_client.configuration.verify_ssl = True`, it will verify the SSL/TLS request.
If set `swagger_client.configuration.verify_ssl = False`, it will not verify the SSL/TLS request.

If you want to customize the certificate file to verify the peer:
set `swagger_client.configuration.ssl_ca_cert = "path/to/certfile"`

Tested without issue for both python2 and python3:
```
________________________________________ summary ________________________________________
  py27: commands succeeded
  py34: commands succeeded
  congratulations :)
```


wing328:
Tested with both http and https and both test cases passed.

PR #1138, opened 2015-08-27
Expose more configuration items in the CLI

Helmsdown:
Per @fehguy  expose the following configuration in the CLI:

1. apiPackage
2. modelPackage
3. instantiationTypes
4. groupId
5. invokerId
6. artifactId
7. artifactVersion
8. additionalProperties
9. typeMapping
10. languageSpecificPrimitives
11. importMapping

I also did the following:

1. Added unit test coverage here and there
2. Refactored some duplicate code into common logic (and added unit tests for it).
3. Moved a business logic test (for skipOverwrite feature) to the `swagger-codegen` module because tests in the CLI should be testing that the CLI correctly invokes the `swagger-codegen` module and not what the swagger-codegen module does with those options.
4. Started an effort to reduce the number of duplicated "magic strings" throughout the project (see CodegenConstants.java). I figured it would be better to have string constants used in many files centrally located in one file. My changes were not exhaustive because of the scope of the "problem". I tactically targeted the strings that were not "language specific" (e.g. "apiPackage", "invokerPackage", "artifactVersion"). 

As always feedback on my code is greatly appreciated.


wing328:
@Helmsdown do mind doing a rebase? I'll take a look this weekend and let you if I've any feedback

Helmsdown:
No problem. Will do it when I get into the office today.

Helmsdown:
@wing328 Rebase done. Have a good weekend.

wing328:
@Helmsdown thanks for writing test cases to cover the change. All my tests look good.

devorbitus:
@Helmsdown do you know how to actually use some of these options like typeMapping and languageSpecificPrimitives? I cant seem to find an example and I am trying to use Short and java.sql.Timestamp

Helmsdown:
@devorbitus So each of the options has it usefulness. The README has not yet been updated with the documentation for all of the new CLI options so you'd have to browse the code to see how each of the many options is used. 

Can you please elaborate on your usecase of Short and Timestamp. Can you give an illustrative snippet of your swagger spec, and a corresponding expectation of the generated code?

devorbitus:
We are in the process of creating data microservices for some of our DB2 tables. I have modified the swagger code generator to have some additional input properties as well as generate the database calling layer as well. The problem comes when I try to parse a swagger 2.0 specification that has model definitions that have properties that match the actual DB2 table like below...

This is a definition that is able to work properly however the data types of the properties don't match the table and have to be converted later.
```json
},
"definitions":{
    "SbaActionCode": {
            "required": [
                "sbaActionCode",
				"lastUpdateUserId",
				"lastUpdateTimestamp"
            ],
            "x-tableName": "SBA2_ACTN_CD",
            "properties": {
                "lastUpdateUserId": {
                    "type": "string",
                    "x-columnName": "LAST_UPD_SYSUSR_ID",
                    "x-creatable": true,
                    "x-updateable": true,
                    "x-autoId": true,
                    "example": "sample lastUpdateUserId",
                    "description": "THE SYSTEM USER IDENTIFIER OF THE LAST USER TO UPDATE A COLLECTION OF DATA."
                },
                "sbaActionCode": {
                    "type": "integer",
                    "format": "int32",
                    "x-columnName": "SBA_ACTN_CD",
                    "x-creatable": true,
                    "x-updateable": true,
                    "x-primaryKey": true,
                    "example": "1",
                    "description": "REPRESENTS TYPES OF SUPPLIER BUYING AGREEMENT ACTIONS. EXAMPLES ARE NEW AND CHANGE."
                },
                "lastUpdateTimestamp": {
                    "type": "string",
                    "format": "date",
                    "x-columnName": "LAST_UPD_TS",
                    "x-autoTs": true,
                    "x-creatable": true,
                    "x-updateable": true,
                    "description": "THE DATE AND TIME WHEN A COLLECTION OF DATA WAS LAST UPDATED."
                }
            },
            "description": "REPRESENTS TYPES OF SUPPLIER BUYING AGREEMENT ACTIONS. EXAMPLES ARE NEW AND CHANGE."
        }
}
````
Which creates this kind of object...

```java
...
public class SbaActionCode {
  private String lastUpdateuserId = null;
  private Integer sbaActionCode = null;
  private Date lastUpdateTimestamp = null;
  ...
}
```

Ideally, we would rather have something like this...
```json
},
"definitions":{
    "SbaActionCode": {
            "required": [
                "sbaActionCode",
				"lastUpdateUserId",
				"lastUpdateTimestamp"
            ],
            "x-tableName": "SBA2_ACTN_CD",
            "properties": {
                "lastUpdateUserId": {
                    "type": "string",
                    "x-columnName": "LAST_UPD_SYSUSR_ID",
                    "x-creatable": true,
                    "x-updateable": true,
                    "x-autoId": true,
                    "example": "sample lastUpdateUserId",
                    "description": "THE SYSTEM USER IDENTIFIER OF THE LAST USER TO UPDATE A COLLECTION OF DATA."
                },
                "sbaActionCode": {
                    "type": "short",
                    "x-columnName": "SBA_ACTN_CD",
                    "x-creatable": true,
                    "x-updateable": true,
                    "x-primaryKey": true,
                    "example": "1",
                    "description": "REPRESENTS TYPES OF SUPPLIER BUYING AGREEMENT ACTIONS. EXAMPLES ARE NEW AND CHANGE."
                },
                "lastUpdateTimestamp": {
                    "type": "sqlTimestamp",
                    "x-columnName": "LAST_UPD_TS",
                    "x-autoTs": true,
                    "x-creatable": true,
                    "x-updateable": true,
                    "description": "THE DATE AND TIME WHEN A COLLECTION OF DATA WAS LAST UPDATED."
                }
            },
            "description": "REPRESENTS TYPES OF SUPPLIER BUYING AGREEMENT ACTIONS. EXAMPLES ARE NEW AND CHANGE."
        }
}
````

So that we can have our models generated like this...
```java
import java.sql.Timestamp;
...
public class SbaActionCode {
  private String lastUpdateuserId = null;
  private Short sbaActionCode = null;
  private Timestamp lastUpdateTimestamp = null;
  ...
}
```

fehguy:
This is fairly easy to configure.  Have you looked at the `typeMapping` options in the CLI?

```
java -jar swagger-codegen-cli.jar help generate

<snip>
        --type-mappings <type mappings>
            sets mappings between swagger spec types and generated code types in
            the format of swaggerType=generatedType,swaggerType=generatedType.
            For example: array=List,map=Map,string=String
```

You should be able to map `Integer` to `Short` and `Date` to `Timestamp`.

devorbitus:
But I don't want to map all integers to short just some of them. Is it still the same?

devorbitus:
Since type mappings were moved to the configuration file I attempted to send the following configuration file...
```json
{
    "verbose":true,
    "modelPackage": "com.homedepot.mm.rd.sbadata.model",
    "apiPackage": "com.homedepot.mm.rd.sbadata",
    "invokerPackage": "com.homedepot.mm.rd.sbadata.invoke",
    "systemProperties" : {
        "swagger.codegen.jaxrs.sourcefolder" : "src",
        "swagger.codegen.jaxrs.exceptionpackage" : "com.homedepot.mm.rd.sbadata.exception",
        "swagger.codegen.jaxrs.dbpackage" : "com.homedepot.mm.rd.sbadata.db",
        "swagger.codegen.jaxrs.resources" : "com.homedepot.mm.rd.sbadata.resources",
        "swagger.codegen.jaxrs.impl.source" : "com.homedepot.mm.rd.sbadata.impl",
        "swagger.codegen.jaxrs.factory.source" : "com.homedepot.mm.rd.sbadata.factory",
        "swagger.codegen.jaxrs.webxmlpath" : "WebContent/WEB-INF",
        "swagger.codegen.jaxrs.jndi.datasource" : "SupplierSupplierMerch.1",
        "domain": "homedepot.com",
        "debugParser": true
    },
    "typeMappings": {
    	"timestamp":"Timestamp"
    },
    "importMappings" : {
        "ApiResponseMessage" : "com.homedepot.mm.rd.sbadata.ApiResponseMessage",
        "Timestamp": "java.sql.Timestamp"
    }
}
```
However, it seemed to fail every time due to the property type of 'timestamp' not being preconfigured into the parser.

devorbitus:
@fehguy Ok, I think I understand my problem now. I think I was just thinking about things wrong. Previously, our web services were strongly typed in java but when serialized we would have to have converters for properties like BigDecimal and others converting to JSON. What was never considered was that the payload itself was actually JSON and JSON has limitations on what can be displayed. There is no difference between the number at the heart of BigDecimal and a double once it becomes JSON. I kept envisioning strongly typed APIs that match the database source instead of the actual JSON payload that gets transmitted over the wire. I still think I would like to be able to generate models that are strongly typed to match the database but the swagger definition would match the value that get's output from the serialization.

I don't think the existing API will allow me to have certain properties type mapped while leaving others alone.

fehguy:
@devorbitus there are still two options here.  First, `BigDecimal` and `double` can be differentiated when generating a client.  It's also advantageous to do so in some cases, since as you said, JSON has numeric limitations.

The solution I see is as follows:

1)  We update swagger-core to parse `BigDecimal` as the following:

```
{
  "type": "number",
  "format": "string"
}
```
A string will preserve all decimal precision, and can be mapped back to `BigDecimal`

2)  You use a vendor extension in the properties which you need to convert to `BigDecimal` as a hint to the code generator.

We're actually implementing (1) for an integration now, so you should see that supported by default shortly.

webron:
@fehguy, isn't it the other way around?

```json
{
  "type": "string",
  "format": "number"
}
```

fehguy:
Yes, I typed it backwards.  Updating my comment.

Issue #1143, opened 2015-08-27
JavaClientCodegen: NullPointerException when using minimum/maximum


ePaul:
**Summary**
When using the `min` or `max` constraint for a property, `JavaClientCodegen.postProcessModelMaps` throws an exception.

This happens with current master (also using current master of swagger-core).

**Reproducible by**

bug-trigger.yaml:
```
swagger: '2.0'
info:
  title: Swagger Codegen bug trigger
  description: |
     This demonstrates a bug in swagger-codegen.
  version: 0.0.1
basePath: /api
definitions:
  manual_discount:
    type: object
    properties:
      percent_value:
        type: number
        minimum: 0.0
        maximum: 100.0
```
command line (with debugger):

    java -XX:-OmitStackTraceInFastThrow \
    -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=8000 \
    -jar /home/paulo/.m2/repository/io/swagger/swagger-codegen-cli/2.1.4-SNAPSHOT/swagger-codegen-cli-2.1.4-SNAPSHOT.jar \
    generate -i src/main/resources/api/bug-trigger.yaml -l java \
    -o target/generated-sources/swagger-codegen

This finishes without ouputting anything or writing any file.

**Analysis**
A breakpoint in DefaultGenerator.generate, line 322 shows a NullPointerException.

This exception is thrown in JavaClientCodegen.postProcessModels(), line 335: Here we try to iterate over a `null` list. That null list comes from `allowableValues.get("values")`, which for a string property could contain the values from an `enum` constraint. In this case, `allowableValues` is just a map with the `min` and `max` constraints.

**Excepted behavior**
This should generate a file for class `ManualDiscount` with a `percentValue` property.

**Workaround**
Don't use `min` or `max`.

ePaul:
The code throwing the exception was added in #1126.
Adding another null-check here should be enough, I'll try to create a pull request for that.

wing328:
PR merged

PR #1144, opened 2015-08-27
Add null check to fix issue #1143

ePaul:
* fix issue #1143: add another null check.

  The problem was trying to iterate over a null list.

----

Here is an example yaml file which breaks before the fix, and runs after the fix (though I did not check if the generated classes are fully sensible).
```
swagger: '2.0'
info:
  description: |
  version: 0.0.1
basePath: /api
definitions:
  manual_discount:
    type: object
    properties:
      percent_value:
        type: number
        minimum: 0.0
        maximum: 100.0
paths: {}
```


xhh:
:+1:

Issue #1146, opened 2015-08-28
jaxrs/spring-mvc: API methods have annotations with line breaks in their strings


ePaul:
**Summary**
When a `description` of a response has a line break in the swagger yaml (or json) and we generate for `jaxrs`, an annotation with a broken string literal is generated, which is not compilable in Java.

**Reproducible by**
bug-trigger.yaml:
```
swagger: '2.0'
info:
  title: Example API
  description: bla bla
basePath: /api
definitions:
  example:
    type: object
    properties: {}
paths:
  /example-path:
    put:
      summary: create or update the example.
      responses:
        201:
          description: |
            Example was created.

            No content.
```
Generation using the maven plugin:

			<plugin>
				<groupId>io.swagger</groupId>
				<artifactId>swagger-codegen-maven-plugin</artifactId>
				<version>2.1.4-SNAPSHOT</version>
				<executions>
					<execution>
						<goals>
							<goal>generate</goal>
						</goals>
					</execution>
				</executions>
				<configuration>
					<inputSpec>src/main/resources/api/bug-trigger.yaml</inputSpec>
					<language>jaxrs</language>
					<output>target/generated-sources/swagger-codegen</output>
					<addCompileSourceRoot>false</addCompileSourceRoot>
					<apiPackage>de.zalando.einkaufshelden.example.api</apiPackage>
					<modelPackage>de.zalando.einkaufshelden.example.model</modelPackage>
				</configuration>
			</plugin>

Generated file ExamplePathApi.java (there are other files generated, but this one shows the problem):
```
package de.zalando.einkaufshelden.example.api;

import de.zalando.einkaufshelden.example.model.*;
import de.zalando.einkaufshelden.example.api.ExamplePathApiService;
import de.zalando.einkaufshelden.example.api.factories.ExamplePathApiServiceFactory;

import io.swagger.annotations.ApiParam;

import com.sun.jersey.multipart.FormDataParam;


import java.util.List;
import de.zalando.einkaufshelden.example.api.NotFoundException;

import java.io.InputStream;

import com.sun.jersey.core.header.FormDataContentDisposition;
import com.sun.jersey.multipart.FormDataParam;

import javax.ws.rs.core.Response;
import javax.ws.rs.*;

@Path("/example-path")


@io.swagger.annotations.Api(value = "/example-path", description = "the example-path API")
[...]

PR #1148, opened 2015-08-28
Fix for Issue #1146: escape operation summary and response message

ePaul:
This fixes issue #1146, getting rid of the line breaks in the strings, which broke the Java compilability.
We trim the string before escaping, because trailing `\n` are not useful for anything.

Also, fix a minor indentation problem in the generated file.

-----------

Here is an example yaml file which generates a non-compilable class before the fix and compilable classes after the fix. (The relevant part is only the path, not the definitions.)

```
swagger: '2.0'
info:
  title: Example API
  description:
      bla bla
      
      and one more line.
basePath: /api
definitions:
  example:
    description:
      This is an object with a ...
      
      ... multiline description.
    type: object
    properties:
      foo:
        description: |
          even foo has a
          multi-line description.
        type: string
paths:
  /example-path:
    put:
      summary:
         create or update the example.
         
         Another line.
      responses:
        201:
          description: |
            Example was created.

            No content.
          schema:
            $ref: "#/definitions/example"
```

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JaxRSServerCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JaxRSServerCodegen.java
@@ -118,17 +118,22 @@ public class JaxRSServerCodegen extends JavaClientCodegen implements CodegenConf
         co.baseName = basePath;
     }
 
+    @Override
     public Map<String, Object> postProcessOperations(Map<String, Object> objs) {
         Map<String, Object> operations = (Map<String, Object>) objs.get("operations");
         if (operations != null) {
             List<CodegenOperation> ops = (List<CodegenOperation>) operations.get("operation");
             for (CodegenOperation operation : ops) {
                 List<CodegenResponse> responses = operation.responses;
+                operation.summary = escapeText(operation.summary.trim());
+
                 if (responses != null) {
                     for (CodegenResponse resp : responses) {
                         if ("0".equals(resp.code)) {
                             resp.code = "200";
                         }
+
+                        resp.message = escapeText(resp.message.trim());
                     }
                 }
                 if (operation.returnType == null) {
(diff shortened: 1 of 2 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JaxRSServerCodegen.java
```diff
@@ -118,17 +118,22 @@ public void addOperationToGroup(String tag, String resourcePath, Operation opera
...
     }
 
+    @Override
     public Map<String, Object> postProcessOperations(Map<String, Object> objs) {
         Map<String, Object> operations = (Map<String, Object>) objs.get("operations");
         if (operations != null) {
             List<CodegenOperation> ops = (List<CodegenOperation>) operations.get("operation");
             for (CodegenOperation operation : ops) {
                 List<CodegenResponse> responses = operation.responses;
+                operation.summary = escapeText(operation.summary.trim());
```
ePaul: This `.trim()` breaks when an operation has no summary, like in this example:

```
swagger: '2.0'
info:
  title: Example API
  description:
      bla bla
      
      and one more line.
basePath: /api
paths:
  /example-path:
    put:
      responses:
        201:
          description: |
            Example was created.

            No content.
```

(The same problem happens some lines down when one response has no message.)

Can we put the `.trim()` inside the `escapeText()` method?

`escapeText()` is called from those places (all in DefaultCodegen), in addition to these here:

* fromOperation for the summary and notes (Oops, this means it is actually not needed in this line ... that's what I get from trying to solve multiple problems in one step, whithout verifiying that they actually are problems)
* fromModel (for the description)
* fromParameter (for the description)
* fromProperty (for the description)

I guess I could also add it into DefaultCodegen.fromResponse (for the description, again).
And from my judgement, in all of these places leading or trailing whitespace are not significant, and can be removed before escaping.

ePaul:
I'll do a new pull-request which fixes this correctly.

PR #1150, opened 2015-08-28
jaxrs: escape line breaks in response description

ePaul:
This is the fix for #1146: line breaks in response descriptions are now escaped, so they don't create invalid Java string literals anymore.

In addition, fix the indentation in the generated API operation method.

**I did only test this with with the jaxrs generator (and did run the tests which are included in the build). Please check whether this breaks anything for other languages.**

-----------

I guess there should be some test for this ... it looks like currently all the existing tests are written in Scala. As I don't know Scala at all, I won't try to spend time trying to add a Scala test for this. Should I add a Java test instead, or can we find some other contributor for writing the test?

Here is an example yaml file which generates a non-compilable class before the fix and compilable classes after the fix. 

```
swagger: '2.0'
info:
  title: Example API
  description:
      bla bla
      
      and one more line.
basePath: /api
paths:
  /example-path:
    put:
      summary:
         create or update the example.
         
         Another line.
      responses:
        201:
          description: |
            Example was created.

            No content.
```

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -963,7 +963,7 @@ public class DefaultCodegen {
         } else {
             r.code = responseCode;
         }
-        r.message = response.getDescription();
+        r.message = escapeText(response.getDescription());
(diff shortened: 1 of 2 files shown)


wing328:
Integration test with PHP and Ruby work fine.

Issue #1142, opened 2015-08-27
Java generator generates model classes with dashes in the name


ePaul:
**Summary**
When a `definitions` name contains a dash (`-`), the model class generated from this is named with a dash, too. This is not valid Java code.

This happens with current master of swagger-codegen, and I think it happened also with version 2.1.3.

**Reproducible by**

bug-trigger.yaml:
```
swagger: '2.0'
info:
  title: Swagger Codegen bug trigger
  description: |
     This demonstrates a bug in swagger-codegen.
  version: 0.0.1
basePath: /api
definitions:
  example-parent:
    type: object
    properties:
      foo:
        type: string
paths: {}
```
command line:

    java -jar /home/paulo/.m2/repository/io/swagger/swagger-codegen-cli/2.1.4-SNAPSHOT/swagger-codegen-cli-2.1.4-SNAPSHOT.jar generate -i src/main/resources/api/bug-trigger.yaml -l java -o target/generated-sources/swagger-codegen/

generated Example-parent.java (just the start):

```
package io.swagger.client.model;




import io.swagger.annotations.*;
import com.fasterxml.jackson.annotation.JsonProperty;


@ApiModel(description = "")
@javax.annotation.Generated(value = "class io.swagger.codegen.languages.JavaClientCodegen", date = "2015-08-27T17:21:13.173+02:00")
public class Example-parent   {
```

Due to the dash in file name and class name, this can't be compiled.

**Expected behavior**
The names of classes should be sanitized so the generated identifiers are actually valid Java identifiers.

**Workaround**
Use only names which are already valid identifiers in your target language, e.g. without any dashes when using Java.

ePaul:
#939 (Fixed with #940) had a similar issue, but for dashes in paths, not dashes in model definitions.

I guess similar problems could occur for predefined responses or parameters. Those should all be treated the same.

ePaul:
By merging #1164 + #1163 this is now solved for models and spring-mvc API entry points.
I'm not sure whether this occurs somewhere else, then a similar approach can be used.

I'm therefore closing this issue.

PR #1152, opened 2015-08-28
Issue #1142: replace dashes with underscores before camelizing

ePaul:
Replace dashes with underscores before camelizing.
This gets rid of dashes in model class names (at least for Java).

This is still only a partial fix for #1142: dashes in path names (for spring-mvc) still result in dashes in the Api class names.

**I did test this only with spring-mvc and jaxrs (and did run the tests included in the build, they all pass). I do not know how this effects other languages, specifically those which actually allow dashes in class names. Please verify that this doesn't break anything.**

Obviously this breaks APIs which use names only differing by `-` vs. `_`, but similar cases (`_` versus camelCase) already exist.

Maybe doing this in a Java-specific class (JavaClientCodegen) might be a better idea?

--------

Here is a file which produces a broken model class (Example-parent) before the change and a "good" one (ExampleParent) after the change:
```
swagger: '2.0'
info:
  description: |
  version: 0.0.1
basePath: /api
definitions:
  example-parent:
    type: object
    properties:
      foo:
        type: string
paths:
  example-path:
    get:
      responses:
        200:
          description: "blub"
```

(The path still causes a broken file with spring-mvc, but is okay for jaxrs.)

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -1357,6 +1357,9 @@ public class DefaultCodegen {
     }
 
     public static String camelize(String word, boolean lowercaseFirstLetter) {
+        // Replace all dashes with underscores (which will be handled later).
+        word = word.replace('-', '_');
+
         // Replace all slashes with dots (package separator)
         Pattern p = Pattern.compile("\\/(.?)");
         Matcher m = p.matcher(word);


wing328:
@ePaul thanks again for the PR. I agree that fixing it in the Java-specific class is better.

I had a PR (#1139) to fix variable name using `sanitizeName`. 

Shall we reuse `sanitizeName` to fix the mentioned issue in `toModelName` and `toModelFilename`?

ePaul:
Yes, this sounds good. I'll try to create an alternate pull request using this.

ePaul:
Alternate solution as proposed by @wing328 is in Pull request #1163.

PR #1169, opened 2015-09-01
Handle correctly the case of no default response (see comment in #1124)

boazsapir:


wing328:
Tested with Java, Android, Scala, Ruby, PHP, ObjC and results look good.

PR #1182, opened 2015-09-02
Refactor codegen to support an "uber" meta model for user defined values

Helmsdown:
Per @fehguy "Create some “meta model” in the codegen to expose these options in some JSON format.  Then we can use the model to show options in the online codegen".

Here's what I did:

1. Create a class called CodegenConfigurator. This contains all codegen configuration properties a user would want to manipulate. An instance could be created by deserializing a JSON file or by creating an instance and  populating from CLI or Maven plugin parameters. It also has a convenience method for creating a ClientOptInput class which is THE object DefaultGenerator.java needs to generate code.
2. Refactor Generate.java. If the -c/--config param is specified, deserialize the file specified by that param into a CodegenConfigurator instance, and then populate with other CLI params (if any). If the -c/--config param is not specified construct a new CodegenConfigurator and populate with CLI params. This means it is now possible to put all configuration parameters into a json file (previously you had to specify lang, outputDir, inputSpec, etc always via a CLI param). 
3. Added a `preprocessSwagger(Swagger swagger)` method to CodegenConfig interface, and updated DefaultGenerator to invoke it before doing any of the codegen work.
4. Added logic to DefaultGenerator to skip generation of models that have a import mapping defined. @fehguy had meant this feature to already be there. 

Logical next steps (not done in the PR)
1. Refactor Maven plugin to delegate to CodegenConfigurator
2. Expose more options in the Maven Plugin
3. README updates


wing328:
CI failed due to ruby rspec test failed. Re-run the test and it's fine now. Closing and re-opening this PR to trigger the CI tests again to see if the error goes away.

wing328:
@Helmsdown thanks for the PR. I will work on "3. README updates" tomorrow (Tue) if no one volunteers to do that.

PR #1185, opened 2015-09-03
[Obj-C] Added support for enums in Objective-C

mad102190:
Enums are now generated in the model-header for the class that they're part of.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CSharpClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CSharpClientCodegen.java
@@ -137,11 +137,12 @@ public class CSharpClientCodegen extends DefaultCodegen implements CodegenConfig
 
     @Override
     public String apiFileFolder() {
-        return (outputFolder + File.separator + sourceFolder + File.separator + apiPackage()).replace('.', File.separatorChar);
+
+        return outputFolder + File.separator + (sourceFolder + File.separator + apiPackage()).replace('.', File.separatorChar);
     }
 
     public String modelFileFolder() {
-        return (outputFolder + File.separator + sourceFolder + File.separator + modelPackage()).replace('.', File.separatorChar);
+        return outputFolder + File.separator + (sourceFolder + File.separator + modelPackage()).replace('.', File.separatorChar);
     }
 
     @Override
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ObjcClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ObjcClientCodegen.java
@@ -28,6 +28,10 @@ public class ObjcClientCodegen extends DefaultCodegen implements CodegenConfig {
     protected String podName = "SwaggerClient";
     protected String podVersion = "1.0.0";
     protected String classPrefix = "SWG";
+    protected String authorName = "Swagger";
+    protected String authorEmail = "apiteam@swagger.io";
+    protected String license = "MIT";
+    protected String gitRepoURL = "https://github.com/swagger-api/swagger-codegen";
 
     public ObjcClientCodegen() {
         super();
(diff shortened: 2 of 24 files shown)


wing328:
I ran the integration test for objc petstore sample and got an error as follows:
```
Testing failed:
	Property has a previous declaration
	Typedef redefinition with different types ('enum StatusEnum' vs 'enum StatusEnum')
	Property has a previous declaration
** TEST FAILED **


The following build commands failed:
	CompileC /Users/williamcheng/Library/Developer/Xcode/DerivedData/SwaggerClient-apzcfldaexzmkaemrbxncayxmgan/Build/Intermediates/Pods.build/Debug-iphonesimulator/Pods-SwaggerClient_Example-SwaggerClient.build/Objects-normal/x86_64/SWGConfiguration.o /var/tmp/mad102190/swagger-codegen/samples/client/petstore/objc/SwaggerClient/SWGConfiguration.m normal x86_64 objective-c com.apple.compilers.llvm.clang.1_0.compiler
(1 failure)
[INFO] BUILD FAILURE
[INFO] Total time: 01:11 min
[INFO] Finished at: 2015-09-04T23:03:38+08:00
[INFO] Final Memory: 11M/245M
[ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:1.2.1:exec (xcodebuild-test) on project ObjcPetstoreClientTests: Command execution failed. Process exited with an error: 65 (Exit value: 65) -> [Help 1]
[ERROR] 
[ERROR] 
```

I wonder if you can take a look.  To run the integration test, please run the following under `samples/client/petstore/objc/SwaggerClientTests`:
```
```

mad102190:
Weird. I'll take a look.

mad102190:
So from what I can tell, I think this is actually due to the fact that the swagger.json defines the enum multiple times for different classes. This causes the enum to be generated with the same name (StatusEnum) in multiple different classes (in this case, SWGPet and SWGOrder). 

Possible solutions/workarounds:
- Name the enums differently in the json (i.e. PetStatusEnum, OrderStatusEnum). This seems like it makes the most sense to me.
- In the codegen, prepend the name of the class to the enum name (i.e. SWGPetStatusEnum, SWGOrderStatusEnum, etc).
- Define all the enums in one SWGEnums.h header file, but this would add more work as it would need to check if an enum already exists and would fail if an enum is defined multiple times in the json with different possible values.

That's all I can think of right now. Do you have any other ideas on how to get around this issue @wing328 ?

wing328:
@mad102190 thanks for the detailed investigation.

I think your 2nd approach is similar to the existing Java approach. Here is the code for Pet.java:
```
public class Pet   {

  private Long id = null;
  private Category category = null;
  private String name = null;
  private List<String> photoUrls = new ArrayList<String>();
  private List<Tag> tags = new ArrayList<Tag>();

public enum StatusEnum {
  AVAILABLE("available"), PENDING("pending"), SOLD("sold");
```
StatusEnum is put inside a class so it won't conflict with other enum definition.

Ideally all enum should be put into a single file SWGEnums.h as you've pointed out in the 3rd approach but it would take too much effort.

I would go with your 2nd approach for the time being before anyone has cycle to implement the 3rd approach.

mad102190:
The thing is, these enums need to be defined in the header files so that they can be accessed externally, and in obj-c, this will cause name conflicts if they're named the same. So I agree that the 2nd option works for now, it just causes the enums' names to be fairly convoluted (`SWGPetStatusEnum` or `SWGOrderStatusEnum` instead of something like `PetStatus` or `OrderStatus` which would need to be dictated by the swagger.json). 

This approach allows it to work (albeit messily) without having to make any modifications to the swagger.json naming conventions already put in place, but takes away the ability to explicitly name the enums to a specific name (for example, if I specifically wanted an enum called `PetStatus`, it would be code-gen'd into `SWGPetPetStatus` in this approach).

mad102190:
Furthermore, enums in objective-c can't be represented as NSString objects (or have string values at all), only NSIntegers. So the problem is now that the swagger.json defines them as strings and the API endpoints expect a string parameter to compare against. So if an api is called with a pet whose status property is simply `available` (which maps to 0) instead of @"available", there will be a type issue. Maybe I should add methods (like the keyMapper) that map the enum integer values to strings before being sent, but this seems like it might be a lot more involved than I think it is.

wing328:
Given the limitation of enum in ObjC. I would suggest we implement the enum support using the current Python approach, which adds logic in the property setter to validate the value against the enumerated values. 

Here is an example: [pet.py#L194](https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/python/swagger_client/models/pet.py#L194) and the [corresponding template](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/python/model.mustache#L76)

It's not a perfect solution but it serves the purpose of validation against the input value.

Not sure how easy/difficult to implement something similar in ObjC but I think this approach is worth considering.

wing328:
@mad102190 thanks again for the PR on this. We'll work on providing better support of enum for ObjC client in v2.2.0.

judos:
Hi there,
So if I see it correctly there is no support for Enums yet for objc right?

PR #1187, opened 2015-09-03
[Swift] Update generated code to Swift 2.0

Edubits:
What does this PR change:
- Updated templates to Swift 2.0 and Xcode 7
 - First parameter of a method has an explicit name, as the # syntax is not supported anymore
 - Some global functions are now methods
 - Alamofire library updated to their Swift 2 version => result return type, use ErrorType instead of NSError
- Add some missing checks for usePromiseKit
- Add a missing date format
- Fix the generated comments for non-PromiseKit version

What still has to be done:
- PromiseKit should be updated to a version that supports Swift 2 (not released through CocoaPods yet)
- Test with the GMs of Swift 2 and Xcode 7 and update code if needed (as soon as they are available)


Edubits:
The GM is released by Apple, so no more changes to Swift 2 are to be expected.

> Test with the GMs of Swift 2 and Xcode 7 and update code if needed (as soon as they are available)

Seems to work fine without additional changes.

wing328:
@Edubits thanks for the PR to bring Swift to 2.0.

I'm no expert in Swift. If a developer using Swift 1.x, do they need to upgrade their Swift code to 2.0 in order to use the SDK or the SDK (for Swift 2.0) is backward compatible for Swift 1.x?

Edubits:
> I'm no expert in Swift. If a developer using Swift 1.x, do they need to upgrade their Swift code to 2.0 in order to use the SDK or the SDK (for Swift 2.0) is backward compatible for Swift 1.x?

Unfortunately Swift 2 is neither backward nor forward compatible. Swift 1.x can be used up to Xcode 6.4 (and will not be supported on OS X El Capitan), from Xcode 7 on only Swift 2 can be used. See also https://developer.apple.com/swift/blog/?id=32.

When looking at libraries in the wild most libraries moved in their latest version to Swift 2. Some still support Swift 1.2, but only in an older branch. If you want to use a standalone in Swift 2 all libraries should be using Swift (or Objective-C) and vica versa. In my opinion it's therefore best if a Swift 2 version is available as soon as possible.

wing328:
Thanks for the detailed explanation. If no one has further comment or question, I'll merge this tomorrow.

If later anyone wants the Swift 1.x support, we'll suggest using the [v2.1.3](https://github.com/swagger-api/swagger-codegen/releases/tag/v2.1.3) released a month ago.

Edubits:
Thanks for the merge, @wing328!

PR #1189, opened 2015-09-03
[Obj-C] Added ability to access HTTP Response headers 

mad102190:
- Added a property on the ApiClient class that is set before any operation's completionBlock. This allows access to the HTTP response headers by calling `.apiClient.HTTPResponseHeaders` in any of the API-bodies' completion blocks. 

  - Note: It is recommended to only have one ApiClient per thread to ensure that there is no concurrency issues with the HTTPResponseHeaders property across multiple threads.

- Also added a `+sharedAPI` class method on the API-Body classes that allows quick initialization of a singleton API instance. I added this because calling `+apiWithHeader:key:` would not actually add the header and key values unless the singleton was previously nil, and it was the only way to retrieve an existing singleton instance (i.e. if I wanted to access the singleton multiple times, I would have to call `+apiWithHeader:key:` each time, but the values I passed in for the header and key would only be set the first time, and ignored on each subsequent call). This makes it easier to access the singleton, and also makes it more clear that `+apiWithHeader:key:` is designed to initialize a new API instance, and not retrieve an existing one. In order to add a header to an existing singleton instance, call `[[SWGExampleApi sharedAPI] addHeader:@"exampleHeader" forKey:@"exampleKey"];`

### DIFF
--- a/modules/swagger-codegen/src/main/resources/objc/ApiClient-body.mustache
+++ b/modules/swagger-codegen/src/main/resources/objc/ApiClient-body.mustache
@@ -9,6 +9,13 @@ static bool cacheEnabled = false;
 static AFNetworkReachabilityStatus reachabilityStatus = AFNetworkReachabilityStatusNotReachable;
 static void (^reachabilityChangeBlock)(int);
 
+
+@interface {{classPrefix}}ApiClient ()
+
+@property (readwrite, nonatomic) NSDictionary *HTTPResponseHeaders;
+
+@end
+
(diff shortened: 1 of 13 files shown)


wing328:
@mad102190 thanks for the PR to provide status code and HTTP header via ApiClient (similar to C# and Java). Integration test results look good:
```
[INFO] BUILD SUCCESS
[INFO] Total time: 28.742 s
[INFO] Finished at: 2015-09-04T16:38:54+08:00
[INFO] Final Memory: 11M/245M
```

PR #1198, opened 2015-09-06
Add Dart support

yissachar:
Adds support for Dart Swagger codegen. The code is mostly based on the Java codegen. Since Dart runs in both the browser and on the Dart VM, there is a `browserClient` option that specifies what type of code to generate. Additionally, there are `pubName`, `pubVersion`, and `pubDescription` arguments which all specify how the generated Dart library will be identified as a dependency when used in other Dart libraries.


wing328:
@yissachar thanks for the PR. I'm sure Dart developers would appreciate your effort to add the Dart API client for Swagger-Codegen.

When you've time, I wonder if you can make a minor change to `bin/dart-petstore.sh`. Currently the permission is 
```
-rw-r--r--   1 williamcheng  wheel   794 Sep  7 23:56 dart-petstore.sh
```
would be nice if you can do a "chmod 755 bin/dart-petstore.sh" to make it executable (so that its permission is consistent with other sh files)

yissachar:
Hmm, for me the permission shows up as `-rwxr-xr-x`, which is the same as all the other sh files.

wing328:
@yissachar thanks for the prompt reply. I'll merge it for the time being and submit a PR to update it, if needed.

wing328:
@yissachar thanks again for the PR. My friend and I both found that the dart-petstore.sh is not executable (in Mac) so I submitted #1200 to update the permission

yissachar:
Odd, still don't understand how that happened, but thanks for fixing it!

PR #1201, opened 2015-09-08
Fix HEAD path item is missing.

### LINKED ISSUE #1156: HEAD path item is missing
If I put HEAD methods in my API definition they don't appear in generated HTML.

An example. If I use the minimal example and just change the method from GET to HEAD:

```yaml
---
swagger: '2.0'
info:
  version: 0.0.0
  title: Simple API
paths:
  /:
    head:
      responses:
        200:
          description: OK
```

Then I generate static HTML:

```sh
java -jar ./swagger-codegen/modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -i minimal.yaml -o ~/docs
```

The method is not part of the generated documentation.

According to the specification HEAD should be a valid path item: http://swagger.io/specification/#pathItemHead

Bug? Or am I doing something wrong?

arnested:
Fixes #1156.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java
@@ -413,6 +413,7 @@ public class DefaultGenerator extends AbstractGenerator implements Generator {
         for (String resourcePath : paths.keySet()) {
             Path path = paths.get(resourcePath);
             processOperation(resourcePath, "get", path.getGet(), ops, path);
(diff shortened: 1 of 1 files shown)


wing328:
@arnested thanks for the PR. Tested with PHP client and result looks good
```
    public function addPet($body=null)
    {


        // parse inputs
        $resourcePath = "/pet";
        $resourcePath = str_replace("{format}", "json", $resourcePath);
        $method = "HEAD";
        $httpBody = '';
        $queryParams = array();
        $headerParams = array();
        $formParams = array();
```

PR #1213, opened 2015-09-10
[C#] Api.mustache - Remove the parameter in the BasePath getter

LeoYReyes:
The parameter in the BasePath getter is not needed.


wing328:
@LeoYReyes thanks for the PR. The fix looks good to me.

PR #1238, opened 2015-09-16
unicode type forgotten in api_client.mustache for python

DariusMR:


wing328:
@DariusMR thanks for the PR. There're merge conflicts. Please kindly rebase on the latest master.

DariusMR:
hi
I've rebased on the latest master and all checks have passed

wing328:
@DariusMR thanks. Is it correct to say that the change (unicode) works for both Python 2.x and Python 3.x?

DariusMR:
I don't think so because literal strings are unicode by default in Python3.

wing328:
@DariusMR is there a way to make the change compatible with both Python 2.x and Python 3.x? Ideally we want to maintain one single Python client for Python 2.x and Python 3.x

DariusMR:
I think it's to hard to do that :
*A non-exhaustive list of features which are only available in 3.x releases and won't be backported to the 2.x series:*
    strings are Unicode by default
    clean Unicode/bytes separation
    exception chaining
    function annotations
    syntax for keyword-only arguments
    extended tuple unpacking
    non-local variable declarations
source : https://wiki.python.org/moin/Python2orPython3

wing328:
@DariusMR can you share the swagger spec for repeating the issue fixed by this PR?

DariusMR:
yes of course : 
http://beta.app.mapotempo.com/api/swagger_doc
https://app.mapotempo.com/api/swagger_doc

wing328:
@DariusMR thanks. We'll look into that.

Our goal is to generate a Python API client that works with Python 2.x and Python 3.x, and it should be doable, e.g. [Dropbox Python SDK](https://github.com/dropbox/dropbox-sdk-python)

DariusMR:
ok ok
and do you want to do with this commit ? Before the future Python API client ? Will you take this commit ?

wing328:
If this commit would break Python 3.x, then we should look for better solution that works with both Python 2.x and Python 3.x

geekerzp:
@DariusMR I think you can do this for compatibility:
```python
import sys

if sys.version_info >= (3,0):
    unicode = str
```

So the code:
```
isinstance(obj, (str, int, float, bool, tuple, unicode))
```
will work in both python2 and python3.

frodrigo:
`unicode` is a python 2 specific type, in python 2 there is also `str`, but in python 3 only `str`.

DariusMR:
Here is a version which is both python2 and python3 compatible

wing328:
@DariusMR thanks. We'll review and merge accordingly.

PR #1246, opened 2015-09-17
Use the baseName when referring to path and query parameters.

laurentvdl:
Changes the generated API classes to use the baseName instead of the parameter name: for a url specified as /resource/{resource_id}, the following incorrect typescript was generated:

```typescript
var path = this.basePath + '/resource/{resource_id}';

path = path.replace('{' + 'resourceId' + '}', String(resourceId));
```
which obviously fails to replace the parameter.

Query parameters have a similar problem.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/TypeScript-Angular/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/TypeScript-Angular/api.mustache
@@ -49,7 +49,7 @@ module {{package}} {
 
 {{/queryParams}}
 {{#headerParams}}
-            headerParams['{{paramName}}'] = {{paramName}};
+            headerParams['{{baseName}}'] = {{paramName}};
 
 {{/headerParams}}
             var httpRequestParams: any = {
--- a/modules/swagger-codegen/src/main/resources/TypeScript-Angular/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/TypeScript-Angular/api.mustache
@@ -27,7 +27,7 @@ module {{package}} {
             var path = this.basePath + '{{path}}';
 
 {{#pathParams}}
-            path = path.replace('{' + '{{paramName}}' + '}', String({{paramName}}));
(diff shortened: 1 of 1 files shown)


wing328:
@laurentvdl thanks for the PR. The change looks good to me. I'll merge it tomorrow (Friday) if no one has further question or feedback on the change.

wing328:
@laurentvdl one feedback from me. Shall we update the header parameter to use {{baseName}} instead in this PR?
Change
```
             headerParams['{{paramName}}'] = {{paramName}};
```
to
```
             headerParams['{{baseName}}'] = {{paramName}};
```

And shall it support form parameters ( {{formParams}} ) as well?

laurentvdl:
Indeed, updated the PR for the headerParams.
As formParams seem to be unsupported in the angular-ts generator, I'll send a separate PR with an implementation.

wing328:
@laurentvdl thanks. We'll merge this PR for the time being while you're working on a PR to support formParams.

wing328:
@laurentvdl To implement form parameter support in angular-ts generator, I would suggest you to take a look at how node-ts generator implements it: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/TypeScript-node/api.mustache#L191

FYI. I filed #1261 to fix parameter name in node-ts generator based on your fix for angular-ts generator

laurentvdl:
What would be the correct way to check if formParams (any array/list) is empty ?
(I would like to check at generate time, not runtime)

I tried to no avail: 

```typescript
{{#formParams.length}}
            var formsParams: any = {};
{{/formParams.length}}
```

wing328:
`formParams.length` should work as mentioned in my comment on another PR before: https://github.com/swagger-api/swagger-codegen/pull/1076#r37143346

One approach/workaround is to declare the Form parameter hash anyway:https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/ruby/api.mustache#L56

PR #1250, opened 2015-09-17
Add support for enum list to retrofit template

### LINKED ISSUE #1248: [Retrofit] Support list of enum
The type in the attribute is ```List<String>```, it should be ```List<FooEnum>```
The enum class is not generated for list of enums
```java
    @ApiModelProperty(value = "")  
    @SerializedName("foo")
    private List<String> foo= null;
  
    public List<FooEnum> getFoo() {
        return foo;
    }
    public void setFoos(List<FooEnum> foo) {
        this.foo= foo;
    }
```

cbornet:
Fix #1248


cbornet:
Not needed anymore with #1256

PR #1252, opened 2015-09-17
Fix NPE in JavaClientCodegen when model has no properties of its own

### LINKED ISSUE #1141: JavaClientCodegen chokes when a sub class has no own properties.
**Summary**
Current master chokes when trying to generate code where allOf is used and the "child" type has no properties of its own.

**Reproducible by**

bug-trigger.yaml:

```
swagger: '2.0'
info:
  description: |
  version: 0.0.1
basePath: /api
definitions:
  example_parent:
    type: object
    properties:
      foo:
        type: string
  example_child:
    allOf:
      - $ref: "#/definitions/example_parent"
      - type: object
        properties: {}
```

Command line:

    java -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=8000 -jar \
    /home/paulo/.m2/repository/io/swagger/swagger-codegen-cli/2.1.4-SNAPSHOT/swagger-codegen-cli-2.1.4-SNAPSHOT.jar \
    generate -i src/main/resources/api/bug-trigger.yaml -l java \
    -o target/generated-sources/swagger-codegen

This will only generate ExampleParent.java, not ExampleChild. There is no error message (I guess caused by a similar issue as #1140, but for the -cli). Putting a breakpoint at DefaultGenerator line 322 (the catch block in generate()) shows that there is a NullPointerException.

**Analysis**
The exception is thrown in JavaClientCodegen.fromModel, where `codegenModel.hasEnums` is accessed as boolean, but is `null` at this point (for `example_child`).
`DefaultCodegen.addVars` omits setting them for this model object, because there are no properties.

I don't quite understand why those fields in CodegenModel are `Boolean` (i.e. nullable) and not just `boolean`.

**Expected behavior**
There should be Java files generated for both model definitions, the second one having no properties of its own.

**Workaround**
Have some property in the "sub type", as well:

```
swagger: '2.0'
info:
  description: |
  version: 0.0.1
basePath: /api
definitions:
  example_parent:
    type: object
    properties:
      foo:
        type: string
  example_child:
    allOf:
      - $ref: "#/definitions/example_parent"
      - type: object
        properties:
          foo:
            type: string
```

seanbright:
A model that inherits from another model and does not add its own properties
causes JavaClientCodegen to crash silently with an NPE.

Fixes #1141

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -1439,6 +1439,8 @@ public class DefaultCodegen {
             }
         } else {
             m.emptyVars = true;
+            m.hasVars = false;
+            m.hasEnums = false;
         }
     }
 


wing328:
@seanbright do you mind sharing the swagger spec for repeating the issue? thanks.

seanbright:
@wing328 the referenced bug (#1141) has a spec. Is that good enough or do you want me to include it here?

wing328:
@seanbright yup. I didn't realise it's the same issue. Will use that one to repeat and confirm the fix.

Once again thanks for submitting a PR to address the issue.

wing328:
Confirmed the fix
```
swagger-codegen|model-with-no-props ⇒ java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate \
  -i /var/tmp/npe.yml \
  -l java \
  -o /var/tmp/java/test
reading from /var/tmp/npe.yml
writing file /var/tmp/java/test/src/main/java/io/swagger/client/model/ExampleParent.java
writing file /var/tmp/java/test/src/main/java/io/swagger/client/model/ExampleChild.java
```

PR #1254, opened 2015-09-18
[Obj-C] removed extraneous logs unless logging is specifically enabled in the [SWGConfiguration sharedConfig] instance

mad102190:


modules/swagger-codegen/src/main/resources/objc/ApiClient-body.mustache
```diff
@@ -75,7 +75,9 @@ static void (^reachabilityChangeBlock)(int);
         [config.loggingFileHanlder writeData:[message dataUsingEncoding:NSUTF8StringEncoding]];
     }
 
-    NSLog(@"%@", message);
+    if ([[{{classPrefix}}Configuration sharedConfig] debug]){
```
wing328: @mad102190 would it be better to check the debug flag at the beginning of the method (logResponse) instead and exit the method if debug is false?

mad102190: I assumed it was designed so that you could write logs to a file and not to the console output. I thought the purpose of the debug flag was just for console output, but you could set that off and still set a loggingFile to have the output be written there. Is this not correct?

wing328: In other clients, the default output is the console. One can change it to a file instead but the debug flag must be switched on in order for the log to be written to the debug file.

wing328: We'll revise the logic for file logging later.

PR #1256, opened 2015-09-18
Set retrofit as a java library

### LINKED ISSUE #1247: [Retrofit] Support special characters in enum values
Need to port https://github.com/swagger-api/swagger-codegen/pull/1126 to retrofit client generator

### LINKED ISSUE #1248: [Retrofit] Support list of enum

cbornet:
This way retrofit will benefit of the improvement of the java gen and will not be left behind.

Fix #1247
Fix #1248


wing328:
Tested with Java-okhttp, Java-retrofit and PHP and results are good.

cristiangarciascmspain:
Why?
I tried this new generated code..Is not the right code to be used on android..
I'll use the v2.1.3 with my custom mustache to make the retrofit interface to use Observable<> ( RxJava ), I think is a bad option for me to use an old version, but until I find out how to create an appropriate generator I have no option.

wing328:
@cristiangarciascmspain what issue did you encounter when using it?

Java client supports another library `okhttp`, which should work well in Android. Have you tried that?

PR #1260, opened 2015-09-20
[Python] remove python3 generator

wing328:
Python3 generator is no longer needed as it's replaced by Python generator, which can generate Python API client compatible with both python 2.x and 3.x.


cbornet:
python3 is still present in bin/all.sh

wing328:
@cbornet my bad. I'll file a PR later to remove it.

Issue #1229, opened 2015-09-14
Java: enum element names need to be sanitized


ePaul:
**Summary**
When generating Java classes from swagger, for String properties marked with `enum` a (local) enum class will be generated, with the enum constant names taken from the swagger enum values.
The result does not compile when using names which are not valid Java identifiers.

**Reproducible**

```yaml
swagger: '2.0'
info:
  title: Swagger Codegen bug trigger
  description: |
     This demonstrates a bug in swagger-codegen.
  version: 0.0.1
basePath: /api
definitions:
  example:
    type: object
    properties:
      foo:
        type: string
        format: uri
        enum:
          - https://purchase-order.einkaufshelden.zalan.do/purchase-order-status/INITIAL
          - https://purchase-order.einkaufshelden.zalan.do/purchase-order-status/ORDERED
          - https://purchase-order.einkaufshelden.zalan.do/purchase-order-status/CHANGE
          - https://purchase-order.einkaufshelden.zalan.do/purchase-order-status/CLOSED
paths: {}
```

This generates this Example.java (beside others):
```java
package de.zalando.einkaufshelden.example.model;


import io.swagger.annotations.*;
import com.fasterxml.jackson.annotation.JsonProperty;


@ApiModel(description = "")
@javax.annotation.Generated(value = "class io.swagger.codegen.languages.JaxRSServerCodegen", date = "2015-09-14T14:33:19.379+02:00")
public class Example  {
  
  public enum FooEnum {
     https://purchase-order.einkaufshelden.zalan.do/purchase-order-status/INITIAL,  https://purchase-order.einkaufshelden.zalan.do/purchase-order-status/ORDERED,  https://purchase-order.einkaufshelden.zalan.do/purchase-order-status/CHANGE,  https://purchase-order.einkaufshelden.zalan.do/purchase-order-status/CLOSED, 
  };
  private FooEnum foo = null;

  
  /**
   **/
  @ApiModelProperty(value = "")
  @JsonProperty("foo")
  public FooEnum getFoo() {
    return foo;
  }
  public void setFoo(FooEnum foo) {
    this.foo = foo;
  }

  

  @Override
  public String toString()  {
    StringBuilder sb = new StringBuilder();
    sb.append("class Example {\n");
    
    sb.append("  foo: ").append(foo).append("\n");
    sb.append("}\n");
    return sb.toString();
  }
}
```
The enum constants here have not the valid names:
```java
  public enum FooEnum {
[...]

PR #1264, opened 2015-09-21
[Java] Improvements on enum var name in Java client

xhh:
* [Nicer normalization on enum var name](https://github.com/swagger-api/swagger-codegen/issues/1229#issuecomment-141129940). For example:  
  `http://example.com/status/OK` => `HTTP_EXAMPLE_COM_STATUS_OK`  
  (was using `HTTPEXAMPLE_COMSTATUSOK` before)
* [Place each enum var in a separate line](https://github.com/swagger-api/swagger-codegen/issues/1229#issuecomment-141134764)
* [Truncate enum name by removing common prefix when present](https://github.com/swagger-api/swagger-codegen/issues/1229#issuecomment-141925993)

The integration test of the Petstore sample is fine:

```
Results :

Tests run: 33, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/enumClass.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/enumClass.mustache
@@ -1,5 +1,6 @@
 public enum {{datatypeWithEnum}} {
-  {{#allowableValues}}{{#enumVars}}{{name}}("{{value}}"){{^-last}}, {{/-last}}{{#-last}};{{/-last}}{{/enumVars}}{{/allowableValues}}
+  {{#allowableValues}}{{#enumVars}}{{name}}("{{value}}"){{^-last}},
+  {{/-last}}{{#-last}};{{/-last}}{{/enumVars}}{{/allowableValues}}
 
   private String value;
 
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
@@ -351,20 +351,34 @@ public class JavaClientCodegen extends DefaultCodegen implements CodegenConfig {
                 Map<String, Object> allowableValues = var.allowableValues;
 
                 // handle ArrayProperty
-                if(var.items != null) {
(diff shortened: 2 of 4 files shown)


modules/swagger-codegen/src/main/resources/Java/enumClass.mustache
```diff
@@ -1,5 +1,6 @@
 public enum {{datatypeWithEnum}} {
-  {{#allowableValues}}{{#enumVars}}{{name}}("{{value}}"){{^-last}}, {{/-last}}{{#-last}};{{/-last}}{{/enumVars}}{{/allowableValues}}
+  {{#allowableValues}}{{#enumVars}}{{name}}("{{value}}"){{^-last}},
+  {{/-last}}{{#-last}};{{/-last}}{{/enumVars}}{{/allowableValues}}
```
ePaul: You should do the same change in the JavaInflector, JavaJaxRS and JavaSpringMVC templates (there the code is in model.mustache).

xhh: Yes, I'll try to submit a PR later for that.

cbornet:
What is the result of the truncation if the enums are "bar" and "baz" ?

xhh:
@cbornet that will generate something like this (tried and confirmed it):

```java
public enum StatusEnum {
  BAR("bar"),
  BAZ("baz");

  private String value;

  StatusEnum(String value) {
    this.value = value;
  }

  @Override
  public String toString() {
    return value;
  }
}
```

The implementation is [here](https://github.com/swagger-api/swagger-codegen/pull/1264/files#diff-79d6f9708f6ea5ba956fcbc53919c452R391):

```java
private String findCommonPrefixOfVars(List<String> vars) {
    String prefix = StringUtils.getCommonPrefix(vars.toArray(new String[vars.size()]));
    // exclude trailing characters that should be part of a valid variable
    // e.g. ["status-on", "status-off"] => "status-" (not "status-o")
    return prefix.replaceAll("[a-zA-Z0-9]+\\z", "");
}
```

cbornet:
OK. Great !

PR #1270, opened 2015-09-22
Fix #1145: sanitize model name in PHP client

### LINKED ISSUE #1145: Generated PHP classes sometimes contain invalid chars
If I reference the following in my Swagger schema:

```json
"parameters": [
   {
      "name": "cat",
      "in": "body",
      "description": "The JSON payload for the new cat",
      "required": true,
      "schema": {
         "$ref": "cat-new.json"
      }
   }
]
```

Then it will generate a model file called `Cat-new.php`, with a `Cat-new` class inside, which is [prohibited](http://php.net/manual/en/language.oop5.basic.php). Ideally we should be generating `CatNew` rather than `Cat_New`.

xhh:


wing328:
Tested and result looks good:
```
PHPUnit 4.8.9 by Sebastian Bergmann and contributors.

.............

Time: 6.33 seconds, Memory: 14.75Mb

OK (13 tests, 2164 assertions)
```

PR #1271, opened 2015-09-22
implement form parameter support using Angular $httpParamSerializer

laurentvdl:
Follow up on #1246.

I ended up extending the CodegenOperation as formParams.length does not work as expected.

Note that this changes the constructor signature of API classes.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
@@ -77,4 +77,13 @@ public class CodegenOperation {
         return nonempty(pathParams);
     }
 
+    /**
+     * Check if there's at least one form parameter
+     *
+     * @return true if any form parameter exists, false otherwise
+     */
+    public boolean getHasFormParams() {
+        return nonempty(formParams);
+    }
+
 }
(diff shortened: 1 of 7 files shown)


wing328:
@laurentvdl thanks for the PR. Do you mind running `./bin/typescript-node-petstore.sh` to update the Petstore sample to make the change easier to review?

For constructor signature, I noticed you introduced a required parameter `httpParamSerializer`. Would it be possible to make the change backward compatible, e.g. creating another constructor instead of updating the existing one?

laurentvdl:
Making $httpParamSerializer an optional last constructor param makes it backwards compatible at the risk of getting undefined errors at runtime. (because the param is now optional)

wing328:
If $httpParamSerializer is not provided, can we assign a "default" serializer for it ?

From what I understand, the $httpParamSerializer is only used for form parameters but not every REST API has form parameters.

laurentvdl:
$httpParamSerializer is only used when formParams are present, so that is covered.
There is only a minor issue if you use form parameters but forget to pass in all constructor arguments.

I considered chaning the constructor signature based on hasFormParams, but that feels awkward.

wing328:
@laurentvdl thanks for the explanation. I'll let the community to review as well and merge this on coming Friday if there's no further feedback or question.

Issue #1266, opened 2015-09-21
Improve exception handling, particularly in the maven codegen plugin


bedag-moo:
Currently, `DefaultCodegen.generate()` catches all exceptions, and communicates success or failure to its caller through the public field `DefaultCodegen.status`. This field is ignored by the maven codegen plugin, i.e. a build continues even if code generation failed.

Ideally, `DefaultCodegen.generate()` would let exceptions propagate to its caller. The CodeGenMojo could then wrap the exception into a `MojoExecutionExecutionException` to tell maven about the failed build.

bedag-moo:
... and whatever you do, please do not discard exceptions without logging them (as the catch block in `DefaultCodegen.generate()` currently does). It sucks to have to debug the maven codegen plugin to find out where the swagger documentation was incomplete, as in #1267.

wing328:
@bedag-moo I agree with you that the exception handling can be further improved. I'm working on the improvement and hopefully will have a PR by the end of this week.

bedag-moo:
I have also started work on this, and have fixed the exception handling in the areas that concern me. Let me test it some more, and I'll send a pull request.

wing328:
Ok. Thanks @bedag-moo

ches:
Seems like this can be closed with #1276 merged.

ches:
Does it also satisfactorily fix #1267?

bedag-moo:
I did not systematically check the code for other instances where exceptions are discarded, but the ones I stumbled across have been fixed, so closing it ok with me.

PR #1276, opened 2015-09-22
Better diagnostics when something goes wrong

bedag-moo:


bedag-moo:
This fixes #1266

wing328:
@bedag-moo thanks for the PR. I read through the change and it looks good to me.

I'll perform some tests tomorrow using the swagger spec provided by other users who also filed issues with exception handling.

PR #1331, opened 2015-10-02
Csharp comments

qswinson:
this fixes the build warnings for xmldoc comments for c#. Generic return types are removed from the documentation so that the docs can at least be generated.


wing328:
@qswinson thanks for fixing the issues with the comments.

PR #1340, opened 2015-10-06
[objc] Add casts that avoid method resolution errors for `count`

ches:
Because the `data` that we're deserializing is of type `id` (essentially untyped), it's possible to have method resolution clashes without explicitly casting here once we've sufficiently parsed a type. I had this issue with a pagination container model, for instance, which has a field named `count` that conflicts with the property of the same name on `NSArray` or `NSDictionary`, preventing the code from compiling until I made this change.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/objc/ApiClient-body.mustache
+++ b/modules/swagger-codegen/src/main/resources/objc/ApiClient-body.mustache
@@ -311,9 +311,10 @@ static void (^reachabilityChangeBlock)(int);
                                  range:NSMakeRange(0, [class length])];
 
     if (match) {
+        NSArray *dataArray = data;
         innerType = [class substringWithRange:[match rangeAtIndex:1]];
 
-        resultArray = [NSMutableArray arrayWithCapacity:[data count]];
+        resultArray = [NSMutableArray arrayWithCapacity:[dataArray count]];
         [data enumerateObjectsUsingBlock:^(id obj, NSUInteger idx, BOOL *stop) {
                 [resultArray addObject:[self deserialize:obj class:innerType]];
             }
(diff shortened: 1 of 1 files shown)


wing328:
@ches thanks for the PR. Do you mind sharing the spec for reproducing the issue?

ches:
The actual spec is private and is rather complicated to reduce to a minimal example (the API is [JSON API]-compliant and the Swagger spec models basically the entire JSON Schema of the spec's container structures, etc.), but I think this part is enough to create a reproduction case:

```yaml
definitions:
  PaginatedResourcesResponse:
    description: A successful collection response with pagination metadata.
    allOf:
      - $ref: '#/definitions/SuccessResponseCollection'
      - properties:
          meta:
            type: object
            properties:
              pagination:
                $ref: '#/definitions/PaginationMetadata'

  PaginationMetadata:
    properties:
      count:
        description: Number of results in current page.
        type: integer
      total:
        description: Total number of results available.
        type: integer
      total_pages:
        type: integer
      per_page:
        type: integer
      current_page:
        type: integer
      links:
        type: object
        properties:
          next:
            title: URL for the next page of results.
            type: string
            format: url
          previous:
            title: URL for the previous page of results.
            type: string
            format: url
```

Nevermind the elided `SuccessResponseCollection`, substitute anything for that. The salient part is that `PaginationMetadata` has a property called `count`. Without this patch, `data` is an ObjC `id`, and the compiler can't figure out which `count` method should be dispatched, it clashes with the `count` properties on `NSArray`, `NSDictionary`, etc.

[JSON API]: http://jsonapi.org/format/

wing328:
@ches thanks for the details in reproducing the issue. I'll use it to confirm and let you know if I've further questions on the PR.

ches:
cc @geekerzp perhaps who has done much of the ObjC client work it seems.

wing328:
Was able to repeat the issue and integration test in your branch looks good:
```
[INFO] BUILD SUCCESS
[INFO] Total time: 27.464 s
[INFO] Finished at: 2015-10-09T16:40:35+08:00
[INFO] Final Memory: 11M/245M
[INFO] BUILD SUCCESS
[INFO] Total time: 27.464 s
[INFO] Finished at: 2015-10-09T16:40:35+08:00
[INFO] Final Memory: 11M/245M
```

ches:
Thanks! I think I will have some more fixes for ObjC on the way soon :smiley:

wing328:
Thanks @ches 

To run the ObjC integration test locally, please execute the following commands in the project root directory:
```
./bin/objc-petstore.sh
cd samples/client/petstore/objc/SwaggerClientTests 
```

PR #1358, opened 2015-10-08
The replacement of '{}' charachters in operationId, nickname

gaalandr:
The replacement of '{}' charachters was introduced, to get rid of '{}' characters in java names. This code is misplaced here, 'camelize' and 'sanitizeName' methods of 'DefaultCodegen' will take care of this.


fehguy:
If this is indeed covered elsewhere, can you please make sure there's a test to confirm it?

Issue #1349, opened 2015-10-07
Error in node.js index.mustache template


oYo-fr:
In nodejs file index.js when generating a node.js server, the variable serverPort is never used.
In file /modules/swagger-codegen/src/main/resources/nodejs/index.mustache, which seems to be the template for this file, I think we should have this (lines 34 to 37):

  http.createServer(app).listen(serverPort, function () {
    console.log('Your server is listening on port %d (http://localhost:%d)', serverPort, serverPort);
    console.log('Swagger-ui is available on http://localhost:%d/docs', serverPort);
  });

Instead of this:

  http.createServer(app).listen({{serverPort}}, function () {
    console.log('Your server is listening on port %d (http://localhost:%d)', {{serverPort}}, {{serverPort}});
    console.log('Swagger-ui is available on http://localhost:%d/docs', {{serverPort}});
  });

... notice the removal of {{ }} surrounding the "serverPort" variable.

wing328:
@oYo-fr  I agree that the variable `serverPort` should be used instead of `{{serverPort}}`. Do you have cycle to submit a PR for the change?

oYo-fr:
Hi wing328,

I would love to make a pull request but I simply don't know how to do it :(

I suck at git :-/

PR #1360, opened 2015-10-08
[NodeJS] update index.js to use the var serverPort

wing328:
Fix https://github.com/swagger-api/swagger-codegen/issues/1349


oYo-fr:
\o/ Yeay !

PR #1372, opened 2015-10-10
Add colllectionFormat support to retrofit

### LINKED ISSUE #1364: [Java-retrofit] Support collectionFormat in retrofit
```json
                   {
                        "name": "foo",
                        "in": "query",
                        "type": "array",
                        "items": {
                            "type": "string",
                        },
                        "collectionFormat": "csv"
                    }
```
Generates 
```java
@Query("foo") List<String> foo
```
Which in retrofit has the behavior of ```multi```.

cbornet:
Fix #1364

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/retrofit/CollectionFormats.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/retrofit/CollectionFormats.mustache
@@ -30,7 +30,7 @@ public class CollectionFormats {
         
         @Override
         public String toString() {
-            return String.join(",", params);
+            return StringUtil.join(params.toArray(new String[0]), ",");
         }
         
     }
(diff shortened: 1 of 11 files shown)


modules/swagger-codegen/src/main/resources/Java/libraries/retrofit/CollectionFormats.mustache
```diff
@@ -0,0 +1,95 @@
...
+            return params;
+        }
+
+        public void setParams(List<String> params) {
+            this.params = params;
+        }
+        
+        @Override
+        public String toString() {
+            return String.join(",", params);
```
wing328: @cbornet  Shall we use the join method in StringUtil.java instead as the static method `join` in the String class is only available in Java8 but not Java7 (which is the minimum Java requirement for Retrofit)?

cbornet:
@wing328 you're right. I will update.

wing328:
@cbornet when you've time, I wonder if you can do a rebase on the latest master to resolve the merge conflicts.

PR #1373, opened 2015-10-12
Fix bug in handling 204 "No Content" responses from the server in Android & Java generated clients

estiller:
This commit fixes a bug in handling 204 "No-Content" responses from a server in Android & Java generated clients. The bug caused 204 responses in the client to throw an ApiException, while 204 responses are completely valid.

Basically, the old code was structured like this:
```java
if (code == 204)
    responseString = "";
else if(code >= 200 && code < 300) {
    // Other situations handled here
    return responseString;
}
throw new ApiException(code, responseString);
```

While my fix does the following:
```java
if (code == 204) {
    responseString = "";
    return responseString;
}
else if(code >= 200 && code < 300) {
    // Other situations handled here
    return responseString;
}
throw new ApiException(code, responseString);
```

The rest of the files in this pull request are a result of rerunning the sample code generation.


estiller:
For some reason, regenerating the entire Android-Java sample client causes the CI to fail, so I undid it and regenerated only the sample ApiInvoker.java file which I have changed.
This has nothing to do with the fix I made, and is a separate issue.

Thanks!

wing328:
@estiller thanks for the PR. The fix looks good to me. I'll merge it tomorrow (Tue) if no one has further question or comment.

Issue #1353, opened 2015-10-08
[Java][Android] Make android a sub template of the Java generator


cbornet:
This way android would benefit from the enhancements of the Java generator.
Also I wonder if the android specific files could not be added to all the Java templates so that you can choose which libraries you want to use for your android app ?

wing328:
@cbornet currently the `okhttp-gson` is compatible with Java, Android and Scala. What  other HTTP libraries you would like to add for Android client in the Java template? 

Personally I welcome the support of more libraries in the Java template.

cbornet:
@wing328 yes, my point is that all java libraries could be made compatible with android and that they could have a android-build.gradle to generate an android apk.
And the android generator could be made as a java sub-template renamed "apache-httpclient-gson" for instance (or even completely removed because who still uses apache-httpclient ? ;-))

cbornet:
@wing328 Now that all java libs support android, what should be the next step : remove android codegen and document that java codegen supports android or move it to java apache-httpclient-gson ?

wing328:
@cbornet thanks for the effort on that. Much appreciated. 

Last time when I sunset the Python3 generator, I updated the README to inform the users that this generator won't be updated/patched any more and moving forward they should use the Python generator, which works with both Python 2.x and 3.x. Not sure how many users actually read the updated README but I suggest we also do the same this time.

Even better, we can post in the Google Group to inform the community about sunsetting Android generator in the future and they should use the Java generator as a replacement instead. (To me, one month of notification is enough)

Also if we remove the Android generator, users will no longer see "Android" under the "Generate Client" dropdown menu. Maybe we should rename the Java generator to something else (e.g. Java & Android).

cbornet:
OK. I just discovered that there are some things missing in the java libs to be fully compatible with android (like a Manifest file). I will fix this ASAP

cbornet:
There is also an issue with javax annotations Generated which is not part of Android by default

cbornet:
@wing328 There is no README file in android templates and I don't see the point of adding one now just to say that it is deprecated. I have submitted a PR to deprecate it in the editor (although it is rejected by CI and I don't really know why...). 
So where else should we notice the user that android-java is deprecated ? I looked at getHelp() method of Codegen but it seems it is not used anymore...

karussell:
@cbornet Is android-java deprecated? If so, I would point this out while calling the command with this option

[...]

PR #1377, opened 2015-10-12
Add gradle files with android support to java templates

cbornet:
See #1353


wing328:
Test results (default, jersey2, okhttp-gson, retrofit) are good.

PR #1379, opened 2015-10-13
REST client code generation bug fix proposal

gwen15:
Hello,

here is a two bug fix proposal in the client code generation:
- compilation problem with the Long instance initialization in Java code generator (before: "Long id =1;" now: "Long id=1l;") 
- generation problem with the creation of the paths in both C# implementation (the creation of the package path has a side-effect on the full path, even the existing local path containing a dot character) 

Hope this help

Gwen


wing328:
@gwen15 thanks for the PR to improve the Java and C# code generators. Do you mind rebasing on the latest master to resolve the merge conflicts?

gwen15:
Here is the rebased fix

PR #1380, opened 2015-10-13
Apply global security to operations when necessary

xhh:
See #1365


xhh:
To make this PR work, it is needed to get swagger-api/swagger-parser#116 fixed.

PR #1383, opened 2015-10-13
Updates to TypeScript Templates

### LINKED ISSUE #1334: Improve TypeScript Template
The templates in https://github.com/swagger-api/swagger-codegen/tree/master/modules/swagger-codegen/src/main/resources/TypeScript-Angular should be updated to use TS 1.6 best practices:
- No more use var keyword. Use let and cost instead (TsLint Config Value: no-var-keyword)
- Use namespace keyword instead of module keyword (TsLint Config Value:  no-internal-module)
- Why are classes used for models? I think interfaces would be better

I'd make a pull request if desired.

aersamkull:
Resolves #1334 . If it needs some improvements, please give me feedback

### DIFF
--- a/modules/swagger-codegen/src/main/resources/TypeScript-Angular/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/TypeScript-Angular/api.mustache
@@ -3,7 +3,7 @@
 /* tslint:disable:no-unused-variable member-ordering */
 
 {{#operations}}
-module {{package}} {
+namespace {{package}} {
     'use strict';
 
 {{#description}}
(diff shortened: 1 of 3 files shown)


wing328:
@aersamkull thanks for the PR. The CI build failure is likely caused by other swagger components (e.g. swagger-parser) that we're looking into at the moment.

I wonder if you can also run `./bin/typescript-angular-petstore.sh` and `./bin/typescript-node-petstore.sh` to update the Petstore code sample. Then please follow the readme in sample directory to run the unit testing for TypeScript (AngularJS, NodeJS) Petstore.

aersamkull:
Now I did, however I am getting noImplicitAny errors. I don't think I did change something there. Has this happened before?

wing328:
Last time I ran it, the test passed but it was a while ago. 

To confirm, can you please pull the latest master and run the test under the sample folder for typescript to see if you get the same error?

aersamkull:
Yes, the error already existed. But it is trivial to fix. Just have to change line 18 of api.mustache in TypeScript-Angular:
```TypeScript
constructor(private $http: ng.IHttpService, basePath?: string, private $httpParamSerializer?: (d: any) => any) {
```
The important thing is to change `$httpParamSerializer?: (any) => any` to `$httpParamSerializer?: (d: any) => any`

wing328:
@aersamkull thanks for identifying the issue and including the fix in your PR.

Issue #1336, opened 2015-10-05
C#: Invalid code with multiline XML comments


artem-dudarev:
Multiline XML comments lead to invalid C# code which does not compile.
Swagger spec:
```json
{
    "swagger": "2.0",
    "paths": {
        "/api/test": {
            "get": {
                "description": "Line1\r\nLine2",
                "operationId": "Test"
            }
        }
    }
}
```

Generated code:
```c#
namespace IO.Swagger.Api
{
    public interface IDefaultApi
    {
        /// <summary>
        /// Line1
\nLine2
        /// </summary>
        /// <returns></returns>
        void Test ();
    }
}
```

Expected:
```c#
namespace IO.Swagger.Api
{
    public interface IDefaultApi
    {
        /// <summary>
        /// Line1
        /// Line2
        /// </summary>
        /// <returns></returns>
        void Test ();
    }
}
```

artem-dudarev:
The quick workaround is to replace \r with \\\r by adding the following line to the escapeText method in DefaultCodegen.java:
```java
output = output.replace("\r", "\\r");
```
It will not produce the correct XML comment, but at least the code will compile.

wing328:
@artem-dudarev To me, your solution is good enough as the code will at least compile. Do you mind submitting a PR for the fix?

PR #1389, opened 2015-10-14
Escape \r in XML comments.

artem-dudarev:
Workaround for #1336.


modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
```diff
@@ -137,6 +137,7 @@ public String escapeText(String input) {
         if (input != null) {
             input = input.trim();
             String output = input.replaceAll("\n", "\\\\n");
+            output = output.replace("\r", "\\r");
```
wing328: @artem-dudarev thanks for the PR. Should it be replaced with "\\\\r" similar to the replaceAll for \n and \" (with \\\n and \\\" respectively)?

PR #1392, opened 2015-10-15
Add authorization scope data to CodegenSecurity

nbruno:
Currently CodegenSecurity represents authorization scopes as a collection of strings, which makes code generation difficult because it is lacking the typical "hasMore" attribute. In addition, it also omits the description field of the scope.

This pull request adds the missing information to CodegenSecurity. To demonstrate its use, I've also modified the Java JAXRS templates to include Authorization annotations when the necessary data is present. 

For testing, I used [this spec](https://gist.github.com/nbruno/3b7b6608de6df456a5c1) which is based on the [security spec](https://github.com/swagger-api/swagger-editor/blob/master/app/spec-files/security.yaml) from the swagger editor team. The spec has:

1. An operation with no security
2. An operation with multiple security definitions, one of which is OAuth with multiple scopes
3. An operation with one OAuth security definition, with one scope. 

From what I could tell, only the retrofit library is currently using scope information. I updated it in a separate commit, again using the more familiar "hasMore" syntax.


wing328:
@nbruno thanks a lot for this PR. We'll apply similar changes to other templates.

cbornet:
I don't see the interest of hasMore compared to https://github.com/samskivert/jmustache#-first-and--last

nbruno:
@cbornet Functionally I'm sure they're equivalent, however the rest of the codebase relies upon hasMore, so I opted for maintaining the current convention.

fehguy:
@nbruno this syntax is only for jmustache so we've opted for the `hasMore` functionality.

cbornet:
@fehguy But swagger-codegen uses jmustache, isn't it ? You mean that you want to be able to change to another mustache engine ?

fehguy:
More concerned with using "standard" mustache constructs so users can more easily understand and customize templates

cbornet:
OK, understandable.
@nbruno you should also update the jax-rs petstore template 
```./bin/jaxrs-petstore-server.sh```

nbruno:
Since I have some familiarity with Spring, I updated its template to include similar information. I also regenerated the examples for JAXRS and SpringMVC as suggested by @cbornet

nbruno:
CI build failed because of changes to [swagger.io petstore user1](http://petstore.swagger.io/v2/user/user1). I reverted the values for user1 back to those that the tests expect.

fehguy:
unfortunately relying on the items in the petstore api is not very good because people delete/mutate them :)  you might want to create some objects before running your integration tests.

nbruno:
If they were my ITs I would :) They're from the scala client sample. I just wanted to clarify that the previous CI build failure wasn't a result of the changes in this PR.

fehguy:
sure :)

Issue #1212, opened 2015-09-10
HTML Output does not include Form Data


rince1013:
Using 2.1.3
Running swagger-codegen to generate api documentation, the static html output does not include form data parameters - like files

Steps to reproduce:
Run swagger codegen to generate a static html document
```
java -jar modules\swagger-codegen-cli\target\swagger-codegen-cli.jar generate -i http://petstore.swagger.io/v2/swagger.json -l html -o c:\petstore
```
The html document will not show the form data paramaters
![image](https://cloud.githubusercontent.com/assets/3579529/9794246/99e68be2-57af-11e5-8d56-1a41c85f5200.png)

The relevant section from the swagger.json file
```json
"/pet/{petId}/uploadImage":{  
         "post":{  
            "tags":[  
               "pet"
            ],
            "summary":"uploads an image",
            "description":"",
            "operationId":"uploadFile",
            "consumes":[  
               "multipart/form-data"
            ],
            "produces":[  
               "application/json"
            ],
            "parameters":[  
               {  
                  "name":"petId",
                  "in":"path",
                  "description":"ID of pet to update",
                  "required":true,
                  "type":"integer",
                  "format":"int64"
               },
               {  
                  "name":"additionalMetadata",
                  "in":"formData",
                  "description":"Additional data to pass to server",
                  "required":false,
                  "type":"string"
               },
               {  
                  "name":"file",
                  "in":"formData",
                  "description":"file to upload",
                  "required":false,
                  "type":"file"
               }
            ],
            "responses":{  
               "200":{  
                  "description":"successful operation",
                  "schema":{  
                     "$ref":"#/definitions/ApiResponse"
                  }
               }
            },
            "security":[  
               {  
                  "petstore_auth":[  
                     "write:pets",
                     "read:pets"
                  ]
               }
            ]
         }
      },
```

wing328:
@rince1013 I'm able to repeat the issue. Do you mind submitting a PR to fix it?

wing328:
PR merged

PR #1397, opened 2015-10-15
Fixes #1251, #1212: Form parameters has been added to the html template

### LINKED ISSUE #1251: HTML output inconsistent with Swagger Editor preview
As per title. For my [swagger.yml](https://gist.github.com/alexhawdon/5acb0ea555bb608aa303) definition here is the Swagger UI preview for a given method:
upload_17_9_2015_at_16_54_35
and here is the same method in the HTML output:
upload_17_9_2015_at_16_56_59

Note the missing parameters.

Thanks for any assistance.

lugaru1234:
Fixes #1251, #1212 : Form parameters has been added to the html template

### DIFF
--- a/modules/swagger-codegen/src/main/resources/htmlDocs/index.mustache
+++ b/modules/swagger-codegen/src/main/resources/htmlDocs/index.mustache
@@ -87,6 +87,13 @@
     </div>  <!-- field-items -->
     {{/hasQueryParams}}
 
+    {{#hasFormParams}}
+    <h3 class="field-label">Form parameters</h3>
+    <div class="field-items">
+      {{#formParams}}{{>formParam}}{{/formParams}}
+    </div>  <!-- field-items -->
+    {{/hasFormParams}}
+
     <!-- Remove Return type... unclear where this comes from;
          for our swagger.json files, it is always empty and there is no boolean guard or hasReturnType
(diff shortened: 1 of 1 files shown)


modules/swagger-codegen/src/main/resources/htmlDocs/index.mustache
```diff
@@ -87,6 +87,13 @@
     </div>  <!-- field-items -->
     {{/hasQueryParams}}
 
+    {{#hasFormParams}}
+        <h3 class="field-label">Form parameters</h3>
+        <div class="field-items">
+          {{#formParams}}{{>formParam}}{{/formParams}}
+        </div>  <!-- field-items -->
```
webron: For the sake of consistency, this should be indented as the previous cases.

wing328:
Output looks good
```
    <h3 class="field-label">Form parameters</h3>
    <div class="field-items">
      <div class="param">additionalMetadata (optional)</div>

      <div class="param-desc"><span class="param-type">Form Parameter</span> &mdash; Additional data to pass to server </div><div class="param">file (optional)</div>

      <div class="param-desc"><span class="param-type">Form Parameter</span> &mdash; file to upload </div>
    </div>  <!-- field-items -->
```

PR #1400, opened 2015-10-16
add library in the additional option

testn:
To make it configurable via maven task.


wing328:
@testn thanks for the PR. The change looks good. Will merge it after it passes the CI

PR #1404, opened 2015-10-16
Add default initializer to Swift model classes.

tomekc:
Current Swift model classes cannot be simply instantiated from user's code. I added default parameterless initializer to create objects to use in POST requests, for example.


wing328:
@tomekc thanks for the PR. I found similar use of `public init() {}` in other Swift SDKs.

If no one has any question, we'll merge this on coming Monday (10/19).

PR #1408, opened 2015-10-16
Fix syntax issue when no auth methods defined

kkrauth:
When no authentication methods are defined, the generated code does not compile due to where the authMethods closing tag is located. Moved the closing tag further down and opening tag further up so that the generated code is correct both when authentication are and aren't present.


wing328:
@kkrauth thanks for the PR to address the issue. I was able to repeat the issue and used your PR to fix it.

We'll also check other templates to see if there're similar issues and file a PR to fix if found any.

PR #1410, opened 2015-10-17
Feature/inline models

fehguy:
This feature adds support for generating clients with inline models.  It will "flatten" the schema so that nested `ObjectProperty` and `Model` objects will be generated as separate classes.

There are optimizations for languages which support anonymous inner models--this is not present here.  In addition, there will be support for "reducing" the inline model generation when there is repetition (i.e. exact same response signature).  Again, currently not implemented.

@wing328 if you have time to review, I'd greatly appreciate it!


wing328:
@fehguy sure, I'll run some tests this weekend.

Have a nice weekend.

wing328:
@fehguy I've tested with https://gist.github.com/wing328/359166515dae7f489ce5 and here is the result:

- tested the inline model with name already in use - passed
- tested inline models with parameters (body) - passed
- tested inline models with responses - passed
- tested inline models with nested inline model - passed
- tested inline models with nested model ($ref) - passed
- tested the inline model with incorrect type/format - passed (same behaviour as a proper defined model)
- tested the inline model without type:object - passed
- tested the array of inline model - FAILED
- tested the map of inline model - FAILED

I couldn't get the array of inline model and map of inline model to work.

fehguy:
@wing328 thanks!  I'll look at those two types now...

fehguy:
OK I think they're all supported now.

wing328:
Thanks @fehguy and I'll run some tests later.

wing328:
In `master`, I got the following for map of `Pet` (using $ref):
```
  /**
   * Returns pet inventories by status
   * @return Map<String, Pet>
   */
  public Map<String, Pet> getInventory () throws ApiException {
```
But in this PR, I got the following instead for map of `Pet` (inline model):
```
  /**
   * Returns pet inventories by status
   * @return InlineResponse2002
   */
  public InlineResponse2002 getInventory () throws ApiException {
```
For array of `Pet` (inline model) I got something similar:
```
  /**
   * Finds Pets by tags
   * Muliple tags can be provided with comma seperated strings. Use tag1, tag2, tag3 for testing.
   * @param tags Tags to filter by
   * @return InlineResponse200
   */
  public InlineResponse200 findPetsByTags (List<String> tags) throws ApiException {
```
and `InlineResponse200` is defined as follows:
```
public class InlineResponse200 extends ArrayList<InlineResponse2001>  {
```

fehguy:
thanks @wing328 let me look into that more.  I think the logic is flattening array models when they don't need to be.

fehguy:
@wing328 updated, tests added.  Thanks for the input on this

jhitchcock:
Hey All, 

Not sure if this is an issue or just change of behavior. But pre this change I had my schema structured like the below, basically a wrapper, with an array of customers, re-using the customer model using a ref. Since this change I'm getting it flattened to inline_response_xxxx

Whilst this will probably be fine for generating libraries, I lose the ability to figure out in moustache which model will eventullay be returned (So for example, generating a style of documentation and using jsonSchema property (and resolving its refs to show an output schema)) 

```javascript
"schema":{
                     "type":"object",
                     "properties":{
                        "customers":{
                           "type":"array",
                           "items":{
                              "$ref":"#/definitions/customer"
                           }
                        }
                     }
                  },
```

Is this intentional, in this case? I have got round it by commenting out the flattener for now

PR #1414, opened 2015-10-20
Raw Binary for File Output

dgreenbean:
Changing Stream deserialization to use raw bytes from response, rather than UTF-8 decoded text.

Specifically, when an endpoint returns a PDF as a byte stream in the response, decoding as UTF-8 (as in RestResponse.Contents) will cause the results to be mangled.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
@@ -214,12 +214,14 @@ namespace {{packageName}}.Client
         /// <summary>
         /// Deserialize the JSON string into a proper object.
         /// </summary>
-        /// <param name="content">HTTP body (e.g. string, JSON).</param>
+        /// <param name="response">The HTTP response.</param>
         /// <param name="type">Object type.</param>
-        /// <param name="headers"></param>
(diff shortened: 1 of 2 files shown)


wing328:
@dgreenbean thanks for the PR. Do you mind sharing the Swagger spec to repeat the issue?

dgreenbean:
@wing328 This is a simple sample Swagger spec to reproduce the issue:
{"swagger":"2.0","info":{"version":"v1","title":"My Application V1"},"host":"localhost:52776","schemes":["http"],
"paths":{"/v1/sample":{"get":{
    "tags":["Sample"],
    "operationId":"Sample_Get",
    "consumes":[],
    "produces":["application/pdf"],
    "responses":{"200":{"schema":{"format":"binary","type":"file"}}},
    "deprecated":false
}}},"definitions":{"Object":{"type":"object","properties":{}}},"securityDefinitions":{"basic":{"type":"basic","description":"Basic HTTP Authentication"}}}

When the response is a PDF file, the current master branch will return a UTF-8 decoded string of the PDF contents.  This mangles the PDF.  When processed with this change, the original PDF bytes are returned in the stream.

This change would only be breaking if someone was already relying on the UTF-8 decoding of streams for file responses.

wing328:
All tests in C# petstore passed.

PR #1394, opened 2015-10-15
Deserialize the response body in case of an ApiException

akkie:
Currently it's not possible to use custom models in the ApiException because the body gets not deserialized to Json and therefore the object deserializer cannot deserialize it to a custom model. This pull request uses the same mechanism as used to deserialize the response body in case of a 2xx status code.

It removes also a lot of unnecessary whitespace at the end of the lines.

PR #1417, opened 2015-10-21
ApiException doesn't use deserialized data

akkie:
This is a fix for #1394. I had not merged the data correctly after the rebase


wing328:
@akkie thanks for the quick fix.

Issue #1391, opened 2015-10-14
API Factory and Impl classes are generated in incorrect folders on Windows


rbhunt:
When generating a JAX-RS service on Windows the factory and impl classes are written to \factories\io\swagger\api and \impl\io\swagger\api instead of \io\swagger\api\factories and io\swagger\api\impl respectively.

Here's the output from running codegen.  Looks like the use of \ as a separator in the package names could be confusing the code inserting factories and impl into the paths?

*writing file generated-code/jaxrs/src/gen/java/impl/io\swagger\api\SecurityApiServiceImpl.java
writing file generated-code/jaxrs/src/gen/java/factories/io\swagger\api\SecurityApiServiceFactory.java
writing file generated-code/jaxrs/src/gen/java/impl/io\swagger\api\RegisterApiServiceImpl.java
writing file generated-code/jaxrs/src/gen/java/factories/io\swagger\api\RegisterApiServiceFactory.java*

Edit: This was observed using 2.1.3

wing328:
@rbhunt can you please pull the latest version and see if you still encounter the same issue?

rbhunt:
Yes, just cloned and built the project and encountered the same issue.

wing328:
@rbhunt Can you share the spec for reproducing the issue?

rbhunt:
Sure thing, have provided a cut down sample spec that reproduces the issue here:

https://drive.google.com/file/d/0B5B0A4cdndFMNjdQbTJleGZ5Rkk/view?usp=sharing

Would have just attached it here but don't seem to have permission - I'm new to GitHub so sorry if I've missed something obvious.  Here's the command I'm using to generate:

java -jar C:/swagger-codegen/modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -l jaxrs -i sample.yaml

wing328:
@rbhunt I've tried to fix the issue and would need your help to test the fix (under Windows environment) as follows:
```
git clone https://github.com/wing328/swagger-codegen.git
cd swagger-codegen
git checkout jaxrs_fix_path_separator
mvn clean package
```
then run the command you provided to generate the code for JaxRS to see if the same issue persists. 

If the fix works, I'l submit a PR.

rbhunt:
Now seeing an exception:

[...]

PR #1423, opened 2015-10-22
Fix for #1391

rbhunt:
Replace File.separator with '/' to fix Windows path generation and avoid regular expression issues - see #1391 for details of the issue.

Root cause was inconsistent use of File.separator vs hard-coded '/' characters in the code, compounded by File.separator not playing nice in regular expressions on Windows (where it becomes an un-escaped backslash).

All tests are still passing but someone with much better knowledge of the project than me will need to review and assess any wider impacts of the change.


wing328:
@rbhunt thanks for the PR to fix the issue in the Windows environment. We'll review and let you know if we've any questions. 

Thanks again for the contribution.

wing328:
Tested Jax-RS in both windows and mac.

Tested with C# as well.

Files are generated in the correct folder with the fix.

Issue #1420, opened 2015-10-21
Improve naming of generated enums in Swift


tomekc:
I think it would be great, if enum values in generated Swift models would follow convention, i.e.

CapitalizedWordsInSentence

instead of whatever strings are defined in the API.
I would happily try to implement this, if only anyone point me to relevant class :)

xhh:
Hi @tomekc, here in [model.mustache](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/swift/model.mustache#L16) you can see that `{{enum}}` is used to generate enum variable names:

```
public enum {{datatypeWithEnum}}: String { {{#allowableValues}}{{#values}}
    case {{enum}} = "{{raw}}"{{/values}}{{/allowableValues}}
}
```

And in [SwiftCodegen.java](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java#L239) here is how `enum` is set:

```java
for (String value : values) {
  Map<String, String> map = new HashMap<String, String>();
  map.put("enum", StringUtils.capitalize(value));
  map.put("raw", value);
  swiftEnums.add(map);
}
```

That is, enum names are capitalized currently. You could submit a PR to change it there. As a reference, you could take a look at PR [1264](https://github.com/swagger-api/swagger-codegen/pull/1264/files) on how enum name is normalized in Java client codegen (search for `toEnumVarName`).

tomekc:
I just submitted PR #1436.

By the way, SwiftCodegen has two-space indents unlike everything else. Maybe reformat it some time?

wing328:
@tomekc thanks for the PR and I agree with you that we should reformat it later.

wing328:
PR merged

PR #1436, opened 2015-10-24
Swift enum names capitalization follows Swift convention.

tomekc:
Here is a solution for #1420 that I have reported. I introduced unit test for SwiftCodegen which documents the behavior by the way.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
@@ -15,6 +15,7 @@ import io.swagger.models.properties.MapProperty;
 import io.swagger.models.properties.Property;
 import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang.WordUtils;
 
 import javax.annotation.Nullable;
 import java.util.*;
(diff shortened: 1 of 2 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
```diff
@@ -250,6 +251,16 @@ public CodegenProperty fromProperty(String name, Property p) {
...
   }
 
+    public String toSwiftyEnumName(String value) {
+        // Prevent from breaking properly cased identifier
+        if (value.matches("[A-Z][a-z0-9]+[a-zA-Z0-9]*")) {
+            return value;
+        }
+        char[] separators = {'-', '_', ' '};
+        return WordUtils.capitalizeFully(StringUtils.lowerCase(value), separators).replaceAll("[-_ ]", "");
+    }
```
wing328: There are 2 methods that can help you sanitize and camelize the Swift Enum name:

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L1662

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L1549

Here is how CSharp uses these 2 methods for variable naming.

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CSharpClientCodegen.java#L148

tomekc:
I put camelize() in place of my implementation and it failed miserably. For example it doesn't convert all-caps names like ENTRY_NAME or ENTRY.

By the way, CSharp generator ignores Enums and generates plain property of given type.

So I would really like to keep my implementation.

wing328:
@tomekc ok. Just wanted to let you know there are utility functions that may help in your case.

wing328:
@tomekc and thanks for writing test cases for the change. Later we might move the test file to the "swift" folder instead.

PR #1441, opened 2015-10-26
Fix some bugs

mach-kernel:
Hey all,

I have been out of sync with the main development branch, but I want to submit some contributions, namely:

- Properly working OAuth 2 for Python, Ruby, PHP
- Nested container serialization support for Python, Ruby, Java (e.g a list of maps or similar)
- Upon HTTP 201 responses, Python, Ruby will return the Location header.
- Java's `Unit$` model has `getLocationHeader()` to assist with header operations
- For resources in the `path`, if a URL is provided, it will be used to resolve the location of that resource (useful for HAL style APIs, since the schema limits you to only one path parameter iirc)

I hope that the project can make use out of some of the things that we have had to fix for our SDKs here at Dwolla.

Thank you for your time.


fehguy:
thanks!  There are a bunch of merge errors but maybe @wing328 has a way to compare quickly?

wing328:
Yes, I'll review those changes and create PR manually instead.

PR #1446, opened 2015-10-27
[PHP] add OAuth support 

wing328:
- added OAuth support
- better handle of $ in model name

ref: https://github.com/swagger-api/swagger-codegen/pull/1441
credit: @mach-kernel

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PhpClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PhpClientCodegen.java
@@ -333,6 +333,9 @@ public class PhpClientCodegen extends DefaultCodegen implements CodegenConfig {
         // Note: backslash ("\\") is allowed for e.g. "\\DateTime"
         name = name.replaceAll("[^\\w\\\\]+", "_");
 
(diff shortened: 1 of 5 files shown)


wing328:
After reading the RFC for OAuth 2.0 (http://tools.ietf.org/html/rfc6749), I will rename `auth_token` to `access_token` instead so please put this PR on hold for the time being.

mach-kernel:
@wing328 I noticed that some of the libraries use `auth_token` instead of `access_token` so I wanted to keep this consistent. I do agree, however, that keeping to the RFC is what we should do. I'm going to update this in my branch and test it - you should see the commits appear in the initial PR for the languages I initially mentioned.

Is this OK? :+1:

wing328:
Yes, replacing auth_token with access_token sounds good to me

FYI. Here are some OAuth2 libraries I'd a look at:

https://github.com/intridea/oauth2
https://github.com/adoy/PHP-OAuth2
https://github.com/joestump/python-oauth2

so I think access_token is the preferred name to use.

mach-kernel:
@wing328 I went ahead and did this for ruby, php, java, and python, check the original PR for the addl. commits, thanks!

PR #1455, opened 2015-10-28
fixes #1454 by mapping format to number

### LINKED ISSUE #1454: map string-formatted numbers to `number`
For client generation, it's often preferable to use strings to avoid precision loss with numbers across programming languages.  To support this, the following string format is introduced:

```json
{
  "type": "string",
  "format": "number"
}
```

The different languages can choose to treat this appropriately, but for Java, we should expect this to be represented as `BigDecimal`.

fehguy:


fehguy:
closing in favor of new PR with better support.

fehguy:
See https://github.com/swagger-api/swagger-codegen/pull/1787

Issue #1448, opened 2015-10-27
JAX-RS default definition value 


SalehAly:
I am trying  to set the default value for  the definitions properties, but its not working when generating JAX-RS, works for integer when i specify the format but tried it for strings and boolean it does not work.. any ideas?
Example JSON:
`

  "definitions": {

     "Permission": {
         "properties": {
             "read": {
                 "type": "boolean",
                 "default": false
           }
        }
     }
  }`
`

Example Generated Models

 ```private Boolean read = null; ```

wing328:
@SalehAly thanks for reporting the issue. Can you help confirm you're using the latest version (2.1.4), which was released 2 days ago?

SalehAly:
@wing328 yes first tried with 2.1.3 didnt work then it i tried 2.1.4 didnt work neither.

wing328:
@SalehAly Ok. I'll take a look.

wing328:
PR (submitted by @xhh) merged, please try again.

SalehAly:
works now :+1:

PR #1457, opened 2015-10-28
[Java] Assign default value for primitive properties

xhh:
Assign default value (if present)  for primitive properties in Java-related client and server codegen. See  #1448
For example:

```java
Integer pageSize = 100;
Long max = 10000l;
Float myFloat = 1.2f;
Double myDouble = 3.4d;
String message = "Hello";
Boolean published = false;
StatusEnum status = StatunEnum.AVAILABLE;
```


cbornet:
I think this PR generates a regression fot enum lists. I get the error:
```
default value of property "expressions" is not in allowed values: new ArrayList<ExpressionsEnum>()
```

cbornet:
Confirmed: if I comment [this block](https://github.com/xhh/swagger-codegen/commit/2595992df7112c13d5c2e98899d230fa546e197b#diff-79d6f9708f6ea5ba956fcbc53919c452R467), it works again

xhh:
@cbornet thanks for the report! I've made a fix for it in PR #1546

cbornet:
@xhh You're welcome. Thanks for your quick fix !

Issue #1435, opened 2015-10-24
Generated Java client API not thread safe


fabiankessler:
The codegen generates a DefaultApi class and an ApiClient class.
Both classes have zero Javadoc.

Intuitively, I would assume that an API is a thread safe service API. That's not the case here.
DefaultApi looks fine, but it uses the ApiClient which a) is mutable and b) stores state in class fields, for example the responseHeaders.

Either document how it's supposed to be used, and that it's not thread safe, or refactor.

wing328:
@fabiankessler thanks for the feedback.

To ensure thread-safety, please create an instance of ApiClient per thread and let us know if you still experience any thread-safety issue.

Agreed with you that the documentation can be improved and we'll work on that.

fabiankessler:
The generated ApiClient uses a com.sun.jersey.api.client.Client and that says in the Javadoc:

> Methods to create instances of {@link WebResource} are thread-safe. Methods
> that modify configuration and or filters are not guaranteed to be
> thread-safe.
> The creation of a <code>Client</code> instance is an expensive operation and
> the instance may make use of and retain many resources. It is therefore
> recommended that a <code>Client</code> instance is reused for the creation of
> {@link WebResource} instances that require the same configuration settings.

You're right, with the current code of the generated ApiClient it's best to create (and keep) one per Thread. One for all is not safe. One per use is terribly slow (because the Client underneath gets thrown away).

For my use case, where I stick the ApiClient into a singleton service, I have refactored the code to be thread safe. I don't want to use ThreadLocal, and prefer having just one Client to share.

My changes:
* made ApiClient immutable: created an ApiClient.Builder for the configuration options with defaultHeaderMap, debugging, basePath, json and so on
* removed the authentications stuff (don't use it)
* removed responseHeaders and statusCode. instead the invokeAPI() method returns a Response object containing the ClientResponse, plus the result (generic type T). 
* hostMap is static and a ConcurrentHashMap

Here it is: https://gist.github.com/fabiankessler/a6cb51e247a8c450d4aa

But I think for the general use case, with authentication capabilities, the one-per-thread is better. Still, you could use the Builder, it looks more cleaned up than having setters all over the place.

wing328:
[...]

PR #1463, opened 2015-10-29
Java jersey2: reuse a shared Client instance

xhh:
According to [the feedback](https://github.com/swagger-api/swagger-codegen/issues/1435#issuecomment-151606194) from @fabiankessler

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
@@ -45,6 +45,7 @@ import {{invokerPackage}}.auth.OAuth;
 
 {{>generatedAnnotation}}
 public class ApiClient {
+  private Client client;
   private Map<String, Client> hostMap = new HashMap<String, Client>();
   private Map<String, String> defaultHeaderMap = new HashMap<String, String>();
   private boolean debugging = false;
@@ -68,6 +69,8 @@ public class ApiClient {
 
     // Set default User-Agent.
     setUserAgent("Java-Swagger");
+
+    buildClient();
 
     // Setup authentications (key: authentication name, value: authentication).
     authentications = new HashMap<String, Authentication>();{{#authMethods}}{{#isBasic}}
@@ -203,6 +206,7 @@ public class ApiClient {
    */
   public ApiClient setDebugging(boolean debugging) {
     this.debugging = debugging;
+    buildClient();
     return this;
   }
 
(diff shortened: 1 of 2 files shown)


chenjianjx:
Better to make the ApiClient stateless, otherwise it is not thread-safe to use it.  Currently the request and the response body are not member variables of ApiClient, but "responseHeaders" is.    Need to make the headers a per-request thing.

wing328:
@chenjianjx thanks for the feedback. The recommendation is to use an instance of ApiClient per thread to avoid any potential multi-threading issue.

chenjianjx:
@wing328 then there would be performance issues, wouldn't there?  User sends a request => creates a new ApiClient => Creates a new socket ?   Opening and closing sockets are expensive unless there is a connection pool like those for JDBC.

wing328:
Very good point about performance. 

Your concern should be addressed by the connection pool, which should be something supported by the HTTP library (e.g. Retrofit, Jersey, etc). 

Is it correct to say that you already ran into performance issue using an instance of ApiClient per thread?

fabiankessler:
The one per thread works to some extent in theory. But still

* it's not practical, developers won't do it
* still a loss depending on how things are used. Multiple requests that should be on the same Thread could be on different ones.

All you need to do is make the ApiClient immutable. I gave an example on this thread https://github.com/swagger-api/swagger-codegen/issues/1435 linking to https://gist.github.com/fabiankessler/a6cb51e247a8c450d4aa I had moved all mutability out of it.

That version of the ApiClient returns the result AND the response headers at once, using a simple wrapper object. Your generated PHP client does exactly the same, the fetcher method returns the content and the headers in an array with 2 items.

wing328:
@fabiankessler I agree with you that the solution I recommended before is not perfect and thanks for taking time to refactor ApiClient.

When you've time, do you mind submitting a PR with the change you proposed?

Ref: https://github.com/swagger-api/swagger-codegen/wiki/Guidelines-for-Contribution

wing328:
@fabiankessler @chenjianjx based on the feedback, I've refactored the C# (not Java) API client so that the ApiClient is reusable across multiple API class (e.g. PetApi). PR: https://github.com/swagger-api/swagger-codegen/pull/1643

Later we'll do the same for Java API client.

wing328:
@fabiankessler @chenjianjx a few enhancements have been made to the PR and now both ApiClient and Api classes should be thread safe.

ApiResponse class is added to store the status code and header.

Again I would appreciate your feedback on the proposed change.

chenjianjx:
has a similar PR for java been integrated? I actually need it soon.

PR #1479, opened 2015-10-31
Added support for sortParamsByRequiredFlag config option to C# and Python

sfurholm:


wing328:
@sfurholm thanks for the PR to add thje sortParamsByRequiredFlag option to CLI for CSharp and Python.

I noticed that you also use `sortParamsByRequiredFlag` in the Python mustache template. Can you please elaborate on that? Can you share a swagger spec to reproduce the issue so that we can compare the output before and after adding the `sortParamsByRequiredFlag` tag?

sfurholm:
Yes, the reason I use sortParamsByRequiredFlag in the Python mustache template is for backwards compatibility, so if sortParamsByRequiredFlag is true (the new default), the required params still appear in the api method signatures along with `**kwargs` for optional params, but if sortParamsByRequiredFlag is false, only `**kwargs` will appear in the signature. Without checking for the flag in the template the signature would stay the same (with required params first) because the params are still flagged as required.

Also the flag is used in the template to change the documentation which is output (so the example in the code follows the method signature format).

I can try to make up a swagger spec to reproduce the issue for comparison if the above explanation isn't clear...

wing328:
Thanks for the explanation.

PR #1481, opened 2015-10-31
Update BaseObject.mustache

dvz5:
Fix deserialization.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/perl/BaseObject.mustache
+++ b/modules/swagger-codegen/src/main/resources/perl/BaseObject.mustache
@@ -36,28 +36,30 @@ sub TO_JSON {
     return $_data;
 }
 
-# from json string
+# from Perl hashref
 sub from_hash {
     my ($self, $hash) = @_;
+
     # loop through attributes and use swagger_types to deserialize the data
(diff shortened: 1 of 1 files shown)


wing328:
@dvz5 thanks for the PR. Can you tell us a bit more on the issue by providing an example on what's broken?

If you've cycle, it would great if you can add test case under `samples/client/petstore/perl/t`
(To run the test case, please run `mvn integration-test -rf :PerlPetstoreClientTests` under `samples/client/petstore/perl`)

dvz5:
OK, thanks, I've been away from coding for some years and am completely new
to git and maven, that should get me started with the tests!

I'm running the code against a Swagger spec from Bookeo (
http://www.bookeo.com/apiref/swagger.json). The generated code correctly
retrieves resources (I can dump out the json which is populated with data)
but after deserialization, most fields in the resulting object are not
populated. But a few are. The code in from_hash is looking up fields in the
Perl hash using the wrong keys. The hash keys are the json attribute names,
not the more perlish field names specified as values in the attribute maps
in each object class. In many cases in the petshop the two are the same
which probably explains why the existing tests don't pick this up.

wing328:
Thanks for the explanation. Unit testing is ok.

I'll create a task to add more test cases for deserialization (community contribution)

PR #1484, opened 2015-11-01
Tests for deserialization

dvz5:
Checking correct mapping of JSON field names to object field names.


wing328:
@dvz5 thank you for adding more test cases! :beer:

PR #1485, opened 2015-11-02
Add oauth support for Ruby client

xhh:
ref: #1441
credit: @mach-kernel

see also #1446


xhh:
integration test is good:

```
.........................

Finished in 57.57 seconds (files took 1.57 seconds to load)
25 examples, 0 failures
```

PR #1486, opened 2015-11-02
Addes default headers for TypeScript Angular

### LINKED ISSUE #1460: TypeScript and Authentication
Hi
We use the TypeScript Angular Code Gen and oAuth2 for Authentication. We should therefore be able to set the Authorization header globally. I think this is not currently possible and should be very straightforward to implement.

Any thoughts?

aersamkull:
Fixes #1460. I have tested it with ./bin/typescript-angular-petstore.sh and npm test.
The old pull request had an error which caused it to modify defaultHeaders on each request, which is fixed now.


modules/swagger-codegen/src/main/resources/TypeScript-Angular/api.mustache
```diff
@@ -21,17 +22,24 @@ namespace {{package}} {
...
+        }
+
 {{#operation}}
 
         public {{nickname}} ({{#allParams}}{{paramName}}{{^required}}?{{/required}}: {{{dataType}}}, {{/allParams}}extraHttpRequestParams?: any ) : ng.IHttpPromise<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}{}{{/returnType}}> {
-            let path = this.basePath + '{{path}}';
-
-{{#pathParams}}
-            path = path.replace('{' + '{{baseName}}' + '}', String({{paramName}}));
+            const path = this.basePath + '{{path}}'{{#pathParams}}
```
aersamkull: I have done this because its more functional style not to modify the path string. Forgive me this little refactoring :)

wing328:
The CI error is due to petstore.swagger.io:
```
503 Service Unavailable

No server is available to handle this request.
```
will review after the issue with petstore.swagger.io is gone.

aersamkull:
Thank you! I think petstore.swagger.io is online again...

wing328:
@aersamkull when you've time, I wonder if you can apply the same fix to `typescript-node` as well.

PR #1494, opened 2015-11-02
Add android manifest to java codegen

cbornet:
See #1353

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
@@ -42,7 +42,8 @@ public class JavaClientCodegen extends DefaultCodegen implements CodegenConfig {
     protected String groupId = "io.swagger";
     protected String artifactId = "swagger-java-client";
     protected String artifactVersion = "1.0.0";
-    protected String sourceFolder = "src/main/java";
+    protected String projectFolder = "src/main";
+    protected String sourceFolder = projectFolder + "/java";
     protected String localVariablePrefix = "";
     protected boolean fullJavaUtil = false;
     protected String javaUtilPrefix = "";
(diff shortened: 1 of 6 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
```diff
@@ -42,7 +42,8 @@
     protected String groupId = "io.swagger";
     protected String artifactId = "swagger-java-client";
     protected String artifactVersion = "1.0.0";
-    protected String sourceFolder = "src/main/java";
+    protected String projectFolder = "src/main";
+    protected String sourceFolder = projectFolder + "/java";
```
wing328: shall we use File.separator instead of /  in projecctFolder and sourceFolder?

cbornet: You're right. This is a copy/pasting of what is done in android-java generator. I will update the PR

cbornet: But note that  sourceFolder is already "src/main/java"(without File.separator) in current java codegen

wing328: Yes, that needs to be fixed as well. We'll file a PR.

cbornet: OK. I'll do it in this PR.
Shouldn't java 7  [Path](http://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html) be used instead of having  File.separator everywhere ?

cbornet:
@wing328 I did the change. Note that #1517 will probably conflict.

wing328:
@cbornet thanks I'll run some tests in Windows platform to make sure (previously we'd issue with file separator in Windows platform and I assume you're not using Windows as your development environment)

cbornet:
Obviously, I'm not developping on Windows ;-).
BTW, looking at #1517, it seems there are other path fixes that need to be done. So maybe #1517 should be merged first. It will not be hard to rebase this one after.
Also, I'm not sure things such as
```java
new SupportingFile("auth/ApiKeyAuth.mustache", authFolder, "ApiKeyAuth.java"))
```
are OK on Windows so maybe they should also be fixed (as part of #1517 ?)

wing328:
Tested in mac, windows. back tested jaxrs as well. No issue found.

PR #1499, opened 2015-11-03
Accept arrays as arguments to collection parameters in PHP client.

arnested:
This patch makes it possible to use arrays as arguments for parameters defined as arrays of strings -- and respects the `collectionFormat`.

Until now you had to implode the string to an array yourself before calling the API.

I've been thinking about also adding type hints to array for those parameters in method argument list. But since that would break backwards compatibility it should probably be hidden behind a command line option or something.

NB: This needs swagger-parser 1.0.12.

### DIFF
--- a/samples/client/petstore/php/SwaggerClient-php/lib/ObjectSerializer.php
+++ b/samples/client/petstore/php/SwaggerClient-php/lib/ObjectSerializer.php
@@ -162,6 +162,39 @@ class ObjectSerializer
     }
 
     /**
+     * Serialize an array to a string.
+     *
+     * @param array  $collection       collection to serialize to a string
+     * @param string $collectionFormat the format use for serialization (csv,
+     * ssv, tsv, pipes, multi)
+     *
+     * @return string
+     */
+    public function serializeCollection(array $collection, $collectionFormat, $allowCollectionFormatMulti=false)
+    {
+        if ($allowCollectionFormatMulti && ('multi' === $collectionFormat)) {
+            // http_build_query() almost does the job for us. We just
+            // need to fix the result of multidimensional arrays.
+            return preg_replace('/%5B[0-9]+%5D=/', '=', http_build_query($collection, '', '&'));
+        }
+        switch ($collectionFormat) {
+            case 'pipes':
+                return implode('|', $collection);
+
+            case 'tsv':
+                return implode("\t", $collection);
+
+            case 'ssv':
+                return implode(' ', $collection);
+
+            case 'csv':
+                // Deliberate fall through. CSV is default format.
+            default:
+                return implode(',', $collection);
+        }
+    }
+
+    /**
      * Deserialize a JSON string into an object
      *
      * @param mixed  $data       object or primitive to be deserialized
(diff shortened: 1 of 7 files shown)


arnested:
Tests running fine:
```
PHPUnit 5.0.0 by Sebastian Bergmann and contributors.

.............                                                     13 / 13 (100%)

Time: 6.49 seconds, Memory: 16.00Mb

OK (13 tests, 1102 assertions)
```

samples/client/petstore/php/SwaggerClient-php/lib/ObjectSerializer.php
```diff
@@ -162,6 +162,37 @@ public function toString($value)
...
+     * @param array  $collection       collection to serialize to a string
+     * @param string $collectionFormat the format use for serialization (csv,
+     * ssv, tsv, pipes, multi)
+     *
+     * @return string
+     */
+    public function serializeCollection(array $collection, $collectionFormat, $allowCollectionFormatMulti=false)
+    {
+        if ($allowCollectionFormatMulti && ('multi' === $collectionFormat)) {
+            return http_build_query($collection, null, '&');
```
xhh: I think this does not handle the "multi" collectionFormat correctly. Assuming a parameter "foo" has an array value of 2 elements, "bar" and "baz", in the `$collection` here, then this is expected to be included in the URL:
`foo=bar&foo=baz`
However, the implementation of `serializeCollection` seems to return string `0=bar&1=baz`, which will be included into the URL via [ApiClient.php](https://github.com/arnested/swagger-codegen/blob/collectionFormat/samples/client/petstore/php/SwaggerClient-php/lib/ApiClient.php#L175) and the URL would contains:
`foo=0=bar&amp;1=baz`

arnested: Yes, you're right, @xhh, I haven't account for multidimensional arrays. That's a bug.

I'll look into -- either I have to do query string building myself or I'll postprocess the result of `http_build_query()` and weed out the encoded `[]`.

xhh: FYI I just submitted a PR #1509 to support collectionFormat in Ruby client, which is using [Typhoeus](https://github.com/typhoeus/typhoeus) to make HTTP requests and it handles array parameters in the same way with the `multi` collectionFormat. So the [`build_collection_param`] method returns the (array) parameter directly.
I've no idea how PHP handles it, but hope this helps.

arnested:
I changed the pull request to handle @xhh comments about multidimensionale arrays.

The pull request has been rebased on HEAD of master and the petstore sample has been regenerated.

fehguy:
@wing328 please merge if this looks good to you...

wing328:
The change definitely looks good.

CI test result:
```
PHPUnit 4.8.21 by Sebastian Bergmann and contributors.

...............

Time: 7.33 seconds, Memory: 6.00Mb

OK (15 tests, 288 assertions)
[INFO] BUILD SUCCESS
[INFO] Total time: 01:18 min
[INFO] Finished at: 2016-01-04T15:38:35+08:00
[INFO] Final Memory: 11M/245M
```

PR #1502, opened 2015-11-03
Rebuild perl petstore client after previous updates

dvz5:
At least 2 previous updates were committed without rebuilding the perl
petstore client. One was my fault (BaseObject.pm changes), the other
changes come from an upstream update or updates  I can't identify.


wing328:
@dvz5 thanks for the PR.

One test case in t/01_pet_api.t  failed. Partial result below:
```
ok 30 - get the proper photoUrl from get_pet_by_id
ok 31 - get the null response from update_pet_wth_form
ok 32 - get the updated status after update_pet_with_form
not ok 33 - The thing isa WWW::SwaggerClient::Object::ApiResponse
#   Failed test 'The thing isa WWW::SwaggerClient::Object::ApiResponse'
#   at t/01_pet_api.t line 73.
#     The thing isn't defined
ok 34 - get the null response from delete_pet
ok 35 - throw 404 error about pet not found after delete
ok 36 - array query param processed correctly
ok 37 - The object isa WWW::SwaggerClient::Object::Pet
# Looks like you failed 1 test of 37.
```

I don't see similar issue in master.

If you run the integration test `mvn integration-test -rf :PerlPetstoreClientTests` locally in your machine, do you encounter the same issue?

dvz5:
They all pass for me. Do you have a file samples/client/petstore/perl/lib/WWW/SwaggerClient/Object/ApiResponse.pm ? That file appeared for me when I regenerated the perl petstore, along with a bunch of minor changes in other files. My guess was that these changes came from upstream changes to the swagger definition, and that the perl client hadn't been rebuilt since. 

When that file appeared, that test case started to fail for me. It originally tested for undef, but with the new ApiResponse.pm file in place, the return value changed to an ApiResponse object. That seemed reasonable, so I modified the test. 

Is there a manifest file somewhere that lists the files that should be in the perl petstore? If there is, I haven't updated it.

dvz5:
OK, my mistake. I've been using http://petstore.swagger.io/v2/swagger.json instead of modules/swagger-codegen/src/test/resources/2_0/petstore.json as the input spec for building the petstore. I'm now using perl-petstore.sh and the ApiResponse.pm file has disappeared. I'll see if I can grok git enough to fix this PR.

wing328:
Thanks for fixing it. Now all 92 test cases (woo!!) passed for me.

PR #1519, opened 2015-11-06
remove duplicate tags, replace with extensions in templates

### LINKED ISSUE #1518: jaxrs multiple tags causes startup failure
Because tags are used to group operations and paths, when there are multiple tags it will cause errors starting the service.

fehguy:
fixes #1518

This moves the tags and puts them in extensions.  Only the first tag is used for naming the file:

```json
{
  "x-tags" : [ {
    "hasMore" : "true",
    "tag" : "fun"
  }, {
    "tag" : "stuff"
  } ]
}
```

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JaxRSServerCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JaxRSServerCodegen.java
@@ -1,23 +1,12 @@
 package io.swagger.codegen.languages;
 
-import io.swagger.codegen.CodegenConfig;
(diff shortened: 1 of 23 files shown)


fehguy:
@wing328 do you mind verifying this fix?

wing328:
@fehguy ok. I'll run some tests later today.

wing328:
@fehguy without the fix, jaxrs petstore can't start due to multiple tags

used debugOperations to confirm the `tags` are moved to `x-tags` and jaxrs petstore can start without issue.

fehguy:
Thanks @wing328

Issue #1527, opened 2015-11-06
Java Client Does Not Format date-time As Per Swagger Spec


alexbeardsley:
**This bug completely breaks any service which strictly enforces [RFC3339 formatting](http://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14)** (please reference the [SimpleDateFormat JavaDoc](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html)).

In ApiClient.mustache the constructor is the following:

```java
// Use ISO 8601 format for date and datetime.
// See https://en.wikipedia.org/wiki/ISO_8601
DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");

// Use UTC as the default time zone.
dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
```

date-time is supposed to follow  RFC3339, however this outputs 2015-11-06T06:49:57.483+0000 which is *NOT*  RFC3339 compliant. It does not output what was likely the desired behavior: 2015-11-06T06:49:57.483Z.

The problem is that `Z` is a reserved constant to indicate the RFC 822 time zone. It is getting interpreted, so `dateFormat.format(new Date())` always has +0000 at the end, not the literal string "Z". This could be solved by replacing `Z` with `XXX` (ISO-8601 time zone formatting as of Java 7) or by putting the `Z` in single quotes like the `T` is to indicate that the literal string "Z" should always be appended and not an interpreted time zone.

However this is made more complicated because ApiClient delegates date formatting to Jackson in the JSON class when serializing and deserializing data for POST requests. Jackson uses the same date pattern by default when WRITE_DATES_AS_TIMESTAMPS is disabled on the ObjectMapper, meaning it has the same problem where a POST request containing a date-time will always have +0000 instead of "Z". By default Jackson uses `StdDateFormat` which does fuzzy parsing of dates so the difference in time zone formatting does not break when deserializing.

You can manually set a DateFormat on the ObjectMapper, however that inherently disables the fuzzy parsing. Simply putting `yyyy-MM-dd'T'HH:mm:ss.SSSXXX` as the format will not work because in  RFC3339 the fractional time (SSS) is optional, so if the response does not contain it then the formatter will throw an exception.

[...]

PR #1528, opened 2015-11-06
Modified dateformat to parse UTC ISO 8601 dates correctly.

### LINKED ISSUE #1134: Android client doesn't use ApiInvoker.parseDateTime()
Symptom: I generated an Android client using a Swagger spec that uses 'date-time' on a model:

```
    "definitions": {
        "User": {
            "properties": {
                       <... >
                        "createdAt": {
                            "type": "string",
                            "format": "date-time"
                        },
            <...>
```
The server returns a string for "createdAt" field that should be accepted based on ApiInvoker's parseDateTime().
   (ex: "2015-08-21T20:36:44.000Z" ) 

The ApiInvoker.parseDateTime() is never called. Instead, gson's DefaultDateTypeAdapter's deserialize method is called, and an exception is thrown. (It doesn't expect milliseconds).

samuelg:
This enables the android client to parse dates in ISO 8601 format correctly. This fixes #1134

### DIFF
--- a/modules/swagger-codegen/src/main/resources/android-java/apiInvoker.mustache
+++ b/modules/swagger-codegen/src/main/resources/android-java/apiInvoker.mustache
@@ -69,7 +69,7 @@ public class ApiInvoker {
    * ISO 8601 date time format.
    * @see https://en.wikipedia.org/wiki/ISO_8601
    */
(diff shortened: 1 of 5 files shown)


boxed:
This doesn't look correct to me. The old code seems to support passing some ISO8601 formats like those ending with "+2" AND "Z". The new code hardcodes that the string must end with Z.

xhh:
Here's related discussion: #1527, #1534
FYI. the `okhttp-gson` library version of Java client does support Android as well and it takes care of various formats of ISO 8601.

PR #1537, opened 2015-11-09
Code refactoring: ensureUniqueParams default value has been moved from description to the defaultValue field

lugaru1234:
Code refactoring: ensureUniqueParams default value has been mooved from description to the defaaultValue field

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConstants.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConstants.java
@@ -41,7 +41,7 @@ public class CodegenConstants {
     public static final String SORT_PARAMS_BY_REQUIRED_FLAG_DESC = "Sort method arguments to place required parameters before optional parameters.";
 
     public static final String ENSURE_UNIQUE_PARAMS = "ensureUniqueParams";
-    public static final String ENSURE_UNIQUE_PARAMS_DESC = "Whether to ensure parameter names are unique in an operation (rename parameters that are not). Default: true";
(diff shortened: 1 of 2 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConstants.java
```diff
@@ -41,7 +41,7 @@
     public static final String SORT_PARAMS_BY_REQUIRED_FLAG_DESC = "Sort method arguments to place required parameters before optional parameters.";
 
     public static final String ENSURE_UNIQUE_PARAMS = "ensureUniqueParams";
-    public static final String ENSURE_UNIQUE_PARAMS_DESC = "Whether to ensure parameter names are unique in an operation (rename parameters that are not). Default: true";
+    public static final String ENSURE_UNIQUE_PARAMS_DESC = "Whether to ensure parameter names are unique in an operation (rename parameters that are not).";
```
wing328: shall we keep "Default: true"" so that users know the default value (true) for `ensureUniqueParams` ?

lugaru1234:
@wing328 Sorry for delay. I just moved it to separate field here https://github.com/lugaru1234/swagger-codegen/commit/87786f01452b66bb66ced662200fef6b7143f830#diff-0c1b6d04d7c2c41f87f7c710a5610d88R448 so users will see default value in the help (It is read here https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/CliOption.java#L85) The same was done for other options before https://github.com/swagger-api/swagger-codegen/commit/dcd8302f4216596d7af64d534f70b2c7db85fbae

PR #1541, opened 2015-11-09
[Retrofit] Don't retry if the access token doesn't change

cbornet:
This is to avoid an infinite loop if the server always sends an invalid token.


wing328:
Test cases passed. (I'll file a PR to update the Petstore sample later)

If you've time, I wonder if you can help apply similar fix to the Retrofit2 template.

cbornet:
Oh yes, sorry for the petstore sample miss.
I will also do it for retrofit2. Also when retrofit 2 goes to non-beta version, I guess we can remove the retofit 1 lib.

PR #1547, opened 2015-11-10
Generated python code massively cleaned up

boxed:
Highlights:
- Generated code is MUCH MUCH more pythonic 
- OAuth2 works (via rauth)
- uses requests instead of urllib3, which removes a LOT of code

There are probably still some rough spots, but the old code was quite terribly broken so I think this is a massive improvement :P


wing328:
The Python Petstore test failed. Please refer to samples/client/petstore/python/README.md for more information about the unit tests.

You will need http://docs.python-guide.org/en/latest/dev/virtualenvs/ to run tests for Python2 and Python3.

boxed:
Huh... there are tests for the petstore but they aren't run by travis? That's a bit unfortunate.

Anyway, I'll look into it!

boxed:
There we go. Now it works. Note though that the oauth auth is commented out in configuration.py because the oauth thing on the petstore example doesn't work with rauth. I didn't manage to figure this out, but oauth works fine for the usage I have so it's still a lot better than before.

modules/swagger-codegen/src/main/resources/python/api.mustache
```diff
@@ -1,31 +1,8 @@
 # coding: utf-8
 
-"""
-{{classname}}.py
-Copyright 2015 SmartBear Software
```
wing328: @webron @fehguy any concern from Smartbear perspective to remove the software license from the auto-generated Python code?

modules/swagger-codegen/src/main/resources/python/api.mustache
```diff
@@ -47,7 +24,7 @@ class {{classname}}(object):
             self.api_client = config.api_client
 {{#operation}}
 
-    def {{nickname}}(self, {{#sortParamsByRequiredFlag}}{{#allParams}}{{#required}}{{paramName}}, {{/required}}{{/allParams}}{{/sortParamsByRequiredFlag}}**kwargs):
+    def {{nickname}}(self{{#sortParamsByRequiredFlag}}{{#allParams}}, {{paramName}}{{#required}}{{/required}}{{^required}}=None{{/required}}{{/allParams}}{{/sortParamsByRequiredFlag}}, callback=None):
```
wing328: Using your approach will make the method signature really long if there're many optional method arguments. Shall we keep the current approach using **kargs??

modules/swagger-codegen/src/main/resources/python/api.mustache
```diff
@@ -75,60 +52,48 @@ class {{classname}}(object):
...
-        del params['kwargs']
-
-{{#allParams}}
-{{#required}}
-        # verify the required parameter '{{paramName}}' is set
-        if ('{{paramName}}' not in params) or (params['{{paramName}}'] is None):
-            raise ValueError("Missing the required parameter `{{paramName}}` when calling `{{nickname}}`")
-{{/required}}
-{{/allParams}}
-
```
wing328: shall we keep the check to ensure required parameters is not `none`?

modules/swagger-codegen/src/main/resources/python/model.mustache
```diff
@@ -1,117 +1,63 @@
...
+        self.swagger_type_by_variable_name = {
+            {{#vars}}'{{name}}': '{{{datatype}}}',
+            {{/vars}}
         }
-
-        self.attribute_map = {
-            {{#vars}}'{{name}}': '{{baseName}}'{{#hasMore}},
-            {{/hasMore}}{{/vars}}
+        
+        self.external_name_by_attribute_name = {
```
wing328: The variable name `swagger_types` and `attribute_map` are also used in other programming languages (PHP, Ruby, etc). Not sure why it's necessary to rename those.

modules/swagger-codegen/src/main/resources/python/model.mustache
```diff
@@ -1,117 +1,63 @@
...
-{{#description}}        {{{description}}}{{/description}}
-
-        :return: The {{name}} of this {{classname}}.
-        :rtype: {{datatype}}
-        """
         return self._{{name}}
 
     @{{name}}.setter
     def {{name}}(self, {{name}}):
-        """
```
wing328: What's the benefit of removing the docstring, which explains the property?

modules/swagger-codegen/src/main/resources/python/rest.mustache
```diff
@@ -48,66 +20,39 @@ except ImportError:
...
-
-    def __init__(self, pools_size=4):
-        # urllib3.PoolManager will pass all kw parameters to connectionpool
-        # https://github.com/shazow/urllib3/blob/f9409436f83aeb79fbaf090181cd81b784f1b8ce/urllib3/poolmanager.py#L75
-        # https://github.com/shazow/urllib3/blob/f9409436f83aeb79fbaf090181cd81b784f1b8ce/urllib3/connectionpool.py#L680
-        # ca_certs vs cert_file vs key_file
-        # http://stackoverflow.com/a/23957365/2985775
-
-        # cert_reqs
-        if Configuration().verify_ssl:
```
wing328: There's a requirement from developers to ignore SSL/Cert error when using the Python API client in their development/test environment.

wing328:

modules/swagger-codegen/src/main/resources/python/api.mustache
frol: `None` is a value, though an empty value, so passing such a value via GET/POST parameters would look like passing an empty string on a server side. Thus, if there is a check for None, then there should be also a check for an empty string.

modules/swagger-codegen/src/main/resources/python/rest.mustache
```diff
@@ -48,66 +20,39 @@ except ImportError:
...
+    def __init__(self):
+        self.sessions = {
+            None: requests.Session(),
+        }
+        
+    def session(self, auth_settings=None):
+        if auth_settings is None:
+            auth_settings = [None]
+        assert len(auth_settings) == 1
+        auth_type = configuration.auth_settings().get(auth_settings[0], configuration.auth_settings().values()[0])['type']
```
frol: XXX

frol:
Overall, this PR looks hacky to me due to the code style and lack of docstrings. While the author claimed the PR introduce "Generated code is MUCH MUCH more pythonic", I don't find it such. There were made big changes (some are questionable) in the generated code, but some features (mostly related to SSL) were lost, which is bad. Unlimately, it would be awesome if the generated code would not produce any warnings from `pylint`.

wing328:
@frol 

Thanks for the feedback. I remember we did use pep8 and pylint to reivew the python code and made some changes based on the suggestions. However, we didn't have resource to work on every items in the report.

Is that something you have cycle to contribute to help bring the Python API client to the next level?

modules/swagger-codegen/src/main/resources/python/api.mustache
wing328: @frol thanks for the feedback. I think empty string should be a valid value, which means the user/developer can intentionally set a parameter to empty string, e.g.
`status=&name=abc` (status is a required form parameter)
vs
`name=abc` (status is an optional form parameter)

boxed:
Maybe I should explain why it's more pythonic: functions that take 3 arguments actually take three arguments and not **kwargs. There's less unneeded juggling of variables. It uses requests which, at least for my use case, made SSL work (the old code seemed not to work at all for that). It removes docstring that are devoid of any useful information. 

That's what I remember from the top of my head. This was a while ago and I've since moved on so I won't be spending more time on this. 

I agree 100% that this PR does not represent solid production ready code, but the old code was worse :P

frol:
@wing328 I will try to find some time for it. I still think that this PR can be polished, so I will start from here.

@boxed Thank you for your work!

PR #1548, opened 2015-11-10
The swift code generator can't handle endpoints that return an array of objects

conet:
I'm not really familiar with swift, but was using swagger to specify the API I'm working on and found it useful to generate client code based on the API specification. It seems to me that the current state of the generated swift code can't handle an array of objects at the root of the response. This patch fixes the issue, feel free to improve it.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/swift/Models.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift/Models.mustache
@@ -118,6 +118,10 @@ class Decoders {
                 fatalError("formatter failed to parse \(sourceString)")
             } {{#models}}{{#model}}
 
+			// Decoder for [{{{classname}}}]
+            Decoders.addDecoder(clazz: [{{{classname}}}].self) { (source: AnyObject) -> [{{{classname}}}] in
+                return Decoders.decode(clazz: [{{{classname}}}].self, source: source)
+            }
 			// Decoder for {{{classname}}}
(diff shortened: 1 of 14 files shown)


conet:
Are there some problems with the integration tests? I doubt that this change could trigger a failure in the scala client.

wing328:
@conet thanks for the PR. Let me look into the error in the integration tests.

wing328:
@conet I've fixed the test and all checks now pass. I also filed a PR #1550 to improve the Scala tests.

Thanks for the PR to fix deserialization for array of objects.

Do you mind performing the followings?
1) run `./bin/swift-petstore.sh` to update test cases
2) add a test case for http://petstore.swagger.io/#!/pet/findPetsByStatus, which returns an array of Pet(Object), under `samples/client/petstore/swift` (currently, there's no unit test for Swift Petstore client)

conet:
OK, I'll update the PR with the changes you suggested.

conet:
@wing328 I updated the client but I'm not able to create a test case without a test project, unfortunately I lack the necessary time and knowledge about swift to setup a test project for the Petstore client.

wing328:
@conet that's ok. We would appreciate if you can add a test project to test Swift Petstore client later when you've time

conet:
@wing328 I would be glad to add a testcase for this once a test project is created. I'm not sure how to create a test project myself, but I could try this later.

wing328:
@conet thanks. 

I'll merge this PR by tomorrow (Thur) to see if anyone in the community has any feedback on this PR.

PR #1560, opened 2015-11-12
Adds null type

kwizzn:
PhpClientCodegen does not include "null" as a valid type. As a result, the resulting ObjectSerializer.php fails on null values in API responses. Would you accept this PR?

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PhpClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PhpClientCodegen.java
@@ -61,6 +61,7 @@ public class PhpClientCodegen extends DefaultCodegen implements CodegenConfig {
                         "DateTime",
                         "mixed",
                         "number",
+                        "null",
                         "void",
                         "byte")
         );
@@ -89,6 +90,7 @@ public class PhpClientCodegen extends DefaultCodegen implements CodegenConfig {
         typeMapping.put("array", "array");
         typeMapping.put("list", "array");
         typeMapping.put("object", "object");
+        typeMapping.put("null", "null");
         typeMapping.put("DateTime", "\\DateTime");
 
         cliOptions.add(new CliOption(CodegenConstants.MODEL_PACKAGE, CodegenConstants.MODEL_PACKAGE_DESC));


wing328:
@kwizzn thanks for the PR. Can you please share the swagger spec and code sample to repeat the issue?

kwizzn:
Yeah sure: this is how we run codegen:

```
./run-in-docker.sh generate \
  -i https://console.xtractor.io/scheme/api/1.4.0.json \
  -l php \
  -c config.json
```

Using this config: https://github.com/Organiceme/xtractor.io-sdk-php/blob/master/build/config.json

kwizzn:
Just to be accurate: The codegen process does not fail. The error occurs in the API request:

```
/Users/chris/workspacePHP Fatal error:  Class 'null' not found in /path/lib/ObjectSerializer.php on line 211
PHP Stack trace:
PHP   1. {main}() /path/examples/test.php:0
PHP   2. Organizeme\Xtractor\SemanticsApi->getDocumentSemantics() /path/examples/test.php:14
PHP   3. Organizeme\Xtractor\ObjectSerializer->deserialize() /path/lib/SemanticsApi.php:202
PHP   4. Organizeme\Xtractor\ObjectSerializer->deserialize() /path/lib/ObjectSerializer.php:221
PHP   5. Organizeme\Xtractor\ObjectSerializer->deserialize() /path/lib/ObjectSerializer.php:221
PHP   6. Organizeme\Xtractor\ObjectSerializer->deserialize() /path/lib/ObjectSerializer.php:191
```

wing328:
@kwizzn thanks for the additional information about the issue. I found that the model `SemanticsResultMap` has the following attribute defined:
```
dates: {
  type: "array",
  description: "Extracted dates. Objects contained in this array have one of the following types: "date", "date_interval".",
  items: [
    {
      $ref: "#/definitions/date"
    },
    {
      $ref: "#/definitions/date_interval"
    }
  ]
},
```
Based on my understanding, `items` definition (array of $ref) is not valid, which result in the following swagger types in the PHP model file:
```
        'dates' => 'null[]',
```
The `payment` attribute in that model is also invalid

wing328:
@kwizzn in other words, there should not be any null type in the swagger type mapping if the swagger spec is valid

kwizzn:
Wow, thanks @wing328 for taking such a deep look into our code. I understand your point and in fact, it's not only the date but also the payment attribute. I'm confident that we'll resolve this next week by changing our spec. I'll close this issue then. Thanks again for your time!

PR #1563, opened 2015-11-12
Fix typo in swift error variable name.

conet:
I'm not sure if this was intentional but error seems better than erorr.


wing328:
Thanks. Coincidentally I submitted a PR to address that 3 hours earlier. 

Will merge your PR instead.

Issue #1570, opened 2015-11-14
[TypeScript][Node] url and basePath in the constructor seem duplicated


wing328:
Using the Petstore example, basePath is set to `http://petstore.swagger.io/v2`
(ref: [api.ts#L121](https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/typescript-node/api.ts#L121))

But `url` is also used as indicated below:
```
        const path = this.url + this.basePath + '/user';
```
(ref: [api.ts#L162](https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/typescript-node/api.ts#L162))

so it seems to me basePath should be just `v2` while `url` should be `http://petstore.swagger.io`

or shall we simply just use basePath (set to `http://petstore.swagger.io/v2`) and remove url ? (other API clients, e.g. Java, PHP, just have `basePath`)

currently `npm test` is broken due to this

wing328:
PR merged into master.

PR #1571, opened 2015-11-15
[TypeScript][Node] remove url API, fix test case

wing328:
- removed `url` from API and only use `basePath` moving forward (consistent with other languages)
- fixed test case
- added OAuth support


wing328:
For #1570

PR #1584, opened 2015-11-18
added connexion python server generator

fehguy:


modules/swagger-codegen/src/main/resources/flaskConnexion/README.mustache
```diff
@@ -0,0 +1,27 @@
+# Swagger generated server
+
+## Overview
+This server was generated by the [swagger-codegen](https://github.com/swagger-api/swagger-codegen) project. By using the
+[swagger-spec](https://github.com/swagger-api/swagger-core/wiki) from a remote server, you can easily generate a server stub.  This
+is an example of building a swagger-enabled JAX-RS server.
```
wing328: I think "JAX-RS" should be replaced with "Python Flask (connexion)" instead.

fehguy: thanks, busted for copy/paste!

PR #1587, opened 2015-11-18
Inject JAX-RS SecurityContext information into JAX-RS server implementation

nickcmaynard:
To access SecurityContext information from generated `*.api.impl` classes, it must have access to a SecurityContext object.

The factory/service/impl pattern in use in the JAX-RS server means the standard injection patterns won't work.

This PR injects the standard JAX-RS SecurityContext object into the signature of every operation, passing this information down into the implementation class.


nickcmaynard:
Closed - bad commit sigs.

Issue #258, opened 2014-09-09
Generated classes should override equals and hashCode


mauhiz:
... especially if they are to be used in a Set (uniqueItems-type container)

dantran:
Any plan for this feature? it is marked fo 2.1.0-M2 but we are now at 2.1.4.

wing328:
@dantran I think I can come up with something for C# this weekend.

wing328:
PR for C# merged into master. Please report if you encounter any issue using the new methods

wing328:
https://github.com/swagger-api/swagger-codegen/pull/1605 submitted by @xhh for Java API client.

wing328:
PR merged

PR #1601, opened 2015-11-21
[C#] Add Equal and GetHashCode to models

wing328:
For #258 

Test cases (including new one) passed


wing328:
The change uses `SequenceEqual` from `System.Linq` to compare list of objects so I cannot apply the same change to C# .net 2.0 generator.

cc @who

PR #1617, opened 2015-11-26
[Java] fix issue when model name is "File"

### LINKED ISSUE #1561: Cannot generate a definition with name  "File"
When I try to generate the model of a custom made File definition, the file model does not get generated. I tried with Java and angular typescript client, it does not work also, not sure if this is intended ?

```
 "File": {
      "properties": {
        "id": {
          "type": "string"
        },
        "created": {
          "type": "string",
          "format": "date-time"
        }
      }
    }
```

wing328:
To fix #1561.

Tested with manually adding a model with name "File" to PetStore.


wing328:
Closed this for the time being as there's a better approach.

PR #1618, opened 2015-11-26
Do not resolve non-model body parameter as inline model

xhh:
This PR fixes the issue of treating non-model body parameter as inline model, for example:

```json
 {
    "name": "InputBinaryData",
    "in": "body",
    "required": true,
    "schema": {
        "type": "string",
        "format": "binary"
    }
}
```


wing328:
Thanks @xhh for the fix and updated test cases to cover the issue moving forward.

@webron that's the fix to the issue with "binary" body parameter.

cbornet:
:+1: I also have this issue with simple "string" body
```json
 {
    "name": "InputString",
    "in": "body",
    "required": true,
    "schema": {
        "type": "string"
    }
}
```

wing328:
@cbornet thanks for reviewing the PR. Merged.

Issue #1075, opened 2015-08-13
Code generation for Slim PHP framework


Zyles:
Please consider adding code generation for server in Slim PHP framework.

wing328:
@Zyles currently, we have code generator for Silex in `develop_2.0` branch. Here is the code samples for Petstore: https://github.com/swagger-api/swagger-codegen/tree/develop_2.0/samples/server/petstore/silex/SwaggerServer 

We do not have anything for Slim PHP framework at the moment. Is that something you would like to contribute? We (community) can help you on that and the code generator for Slim PHP framework would be a good starting point.

wing328:
@Zyles if you can provide the implementation of Slim PHP framework for the following endpoints in Petstore:
[addPet](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.json#L29)
[getPetById](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.json#L216)
[updatePetWithForm](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.json#L263)
[deletePet](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.json#L314)

We should be able to reverse-engineer the code back into the mustache template. Thanks in advance.

Zyles:
Not sure how to do this. Also it might be best to wait for Slim 3 to be released as 2 will soon be deprecated anyway.

jfastnacht:
[Slim Framework 3.0.0 RC2](https://github.com/slimphp/Slim/tree/3.0.0-RC2) is released. Is someone already working on this issue or do you wait for a full release?

wing328:
@jfastnacht personally I don't think we need a full release and I'm not aware of anyone working on it at the moment. Do you have cycle to help implement it? If yes, I can work with you on it.

jfastnacht:
@wing328 I'd like to give it a try if no one else wants to rush it. It might take some time.

jfastnacht:
I've made a fork and work on the Slim support in a seperate branch at https://github.com/EduGraph/swagger-codegen/tree/issue-1075 . I've already added basic Slim support based on silex-PHP. It's still WIP! Just wanted to mention that I'm on it.

wing328:
[...]

PR #1619, opened 2015-11-26
[PHP] Slim Framework v3 - Server Generator #1075

jfastnacht:
Pull Request for Issue #1075 . [Slim Framework v3](https://github.com/slimphp/Slim/tree/3.x) added as a server generator.


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SlimFrameworkServerCodegen.java
```diff
@@ -0,0 +1,195 @@
...
+import io.swagger.models.properties.ArrayProperty;
+import io.swagger.models.properties.MapProperty;
+import io.swagger.models.properties.Property;
+
+import java.io.File;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+
+public class SlimFrameworkServerCodegen extends DefaultCodegen implements CodegenConfig {
```
wing328: Instead of extending `DefaultCodegen`, another approach is to extend PhpClientCodegen so as to leverage the typeMapping, languageSpecificPrimitives, etc that are already defined in PhpClientCodegen.

This is not to say you need to update the code to extend PhpClientCodegen but something worth sharing with you.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SlimFrameworkServerCodegen.java
```diff
@@ -0,0 +1,195 @@
...
+                        "DateTime",
+                        "mixed",
+                        "number")
+        );
+
+        instantiationTypes.put("array", "array");
+        instantiationTypes.put("map", "map");
+
+        // ref: https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#data-types
+        typeMapping = new HashMap<String, String>();
```
wing328: You can consider reusing the type mapping in https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PhpClientCodegen.java#L77

For example, for file, we use `\\SplFileObject`

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SlimFrameworkServerCodegen.java
```diff
@@ -0,0 +1,195 @@
...
+    public String apiFileFolder() {
+        return (outputFolder + "/" + apiPackage()).replace('/', File.separatorChar);
+    }
+
+    public String modelFileFolder() {
+        return (outputFolder + "/" + modelPackage()).replace('/', File.separatorChar);
+    }
+
+    @Override
+    public String getTypeDeclaration(Property p) {
```
wing328: I would recommend reusing the one in https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PhpClientCodegen.java#L228

wing328: You may want to override ` public String getTypeDeclaration(String name)` as well.

Ref: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PhpClientCodegen.java#L246

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SlimFrameworkServerCodegen.java
```diff
@@ -0,0 +1,195 @@
...
+        return toModelName(type);
+    }
+
+    public String toDefaultValue(Property p) {
+        return "null";
+    }
+
+
+    @Override
+    public String toVarName(String name) {
```
wing328: For variable and method parameter name, what about using camelCase instead? Ref: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#6-closures uses camelCase for variable naming.

jfastnacht: I've taken this out of PhpClientCodegen too, but the field there is initiated with "snake_case":
https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PhpClientCodegen.java#L33

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SlimFrameworkServerCodegen.java
jfastnacht: In https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PhpClientCodegen.java#L92 is a second DateTime with capital D and T, but none of them fit the case sensitive dateTime specified in the swagger-spec. I'm not quite sure whether this is case sensitive, but I guess not. I've taken out the second DateTime.

wing328:
@jfastnacht for model.mustache, what about reusing https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/php/model.mustache, which has better support for enum, serialization, etc ?

PR #1627, opened 2015-11-27
[Obj-C] Fixed invalid Objective-C code generation.

farkasseb:
Hi!

I was creating the following swagger definition.
```yaml
  ChangePasswordRequest:
    type: object
    required:
      - currentPassword
      - newPassword
    properties:
      currentPassword:
        type: string
        format: password
        description: The current password in plain text.
      newPassword:
        type: string
        format: password
        description: The new password in plain text.
```

Then I tried to generate the Objective-C client from it. Unfortunately it gave me a non-compiling source code for this object.
```objc
@property(nonatomic) NSString* newPassword;
```
`SwaggerClient/SWGChangePasswordRequest.h:22:32: Property follows Cocoa naming convention for returning 'owned' objects`

[Turns out](http://stackoverflow.com/a/6327547/3503324) Objective-C doesn't like properties starting with 'new'.

One way to fix it (what this PR contains) is specify the getter's name.
```objc
@property(nonatomic , getter=getNewPassword) NSString* newPassword;
```

So I tried to extend the existing .mustache file, but I couldn't find any extensible interface for CodegenProperty, so I had to add a new field. In my opinion there should be a dynamic field/method that allows us to extend this property with any fields. What do you think?

This pull request is just an illustration, I am not proud of the coding style, just wanted to show the problem with a quick solution.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenProperty.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenProperty.java
@@ -5,7 +5,7 @@ import java.util.Map;
 
 public class CodegenProperty {
     public String baseName, complexType, getter, setter, description, datatype, datatypeWithEnum,
-            name, min, max, defaultValue, baseType, containerType;
+            name, min, max, defaultValue, baseType, containerType, otherSettings;
 
     public String unescapedDescription;
 
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ObjcClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ObjcClientCodegen.java
@@ -27,6 +27,7 @@ public class ObjcClientCodegen extends DefaultCodegen implements CodegenConfig {
     public static final String GIT_REPO_URL = "gitRepoURL";
     public static final String LICENSE = "license";
     protected Set<String> foundationClasses = new HashSet<String>();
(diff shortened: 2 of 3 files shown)


fehguy:
You should be able to use the `toVarName` method in ObjcClientCodegen to do this.  You can detect invalid names and escape them accordingly.

wing328:
@farkasseb Would prepending _ to the variable name (i.e. _newPassword) resolve the issue?

farkasseb:
@wing328 Unfortunately NO. This was the first thing I tried. Xcode still saw it as invalid.
@fehguy You are right. I tried it, but then I couldn't decide how to rename the variable (speaking in general terms). Overriding the getter seemed a more logical solution to me.

wing328:
@farkasseb what about prepending "var" (i.e. varNewPassword)?

There's already a mapping between variable name and its actual name used in the spec (using toVarName as suggested by Tony). The mapping aims to handle reserved words and we can leverage the same solution for variable starting with "new".

farkasseb:
@wing328 ok. Sounds good to me. I will do it soon.

wing328:
@farkasseb sorry I need to address similar and have filed #1644 to fix the issue.

I wonder if you can help perform a test to confirm the fix works for you as well. Thanks.

wing328:
#1644 merged. Closing this one.

@farkasseb thanks again for reporting the issue and contributing a fix.

farkasseb:
I am really glad that it is fixed. Thanks for the fast reactions.

PR #1647, opened 2015-12-01
Fixed xml validation in web.xml

evigeant:
A simple fix to fix XML schema validation on the web.xml file


wing328:
@evigeant I saw the change to move the load-on-startup xml tag from line 42 to 52. What issue does it try to solve? Is there any error message you can provide?

Ref: https://github.com/swagger-api/swagger-codegen/wiki/Guidelines-for-Contribution

evigeant:
According to the schema(http://java.sun.com/xml/ns/javaee/web-app_3_0.xsd), load-on-startup needs to be after init-param. My editor reports errors for failed schema validation so that's how I found it.

So the only goal here is to make this a valid web.xml file.

wing328:
@thanks for the explanation and contribution. PR merged

PR #1669, opened 2015-12-05
Authorization values for accessing swagger URL were added into input options

tomtit:


tomtit:
Closing this PR as it's possible to use the `spec` node instead of the `swaggerUrl`.

Issue #1294, opened 2015-09-24
Document client SDK


jpangburn:
I'm trying to generate a Javascript SDK for my API and find that there's no good way to generate documentation for how a user would use that SDK to connect to my API.  The swagger doc shows HTTP methods and URLs but those don't matter to the client SDK user- what matters to them is what methods they call in the SDK, and what parameters those methods take.
It seems like swagger-codegen has all the data available needed to generate such a static HTML documentation for a generated SDK.  This should also generate doc for a Javascript SDK too please.

wing328:
@jpangburn for JS SDK, do you actually mean http://github.com/swagger-api/swagger-js?

Currently, swagger-codegen does not offer a JS (client-side) code generator.

jpangburn:
Hi,
I think swagger-js can’t work because they generate a dynamic client.  Swagger-codegen has the right architecture to be generating a client SDK in different languages, and from the same data structure that you generate that code you can also generate the documentation for it.  At least that’s how it appears to me.  If you disagree, of course reject this.

thanks,
Jesse

wing328:
We'll mark this as a feature request (there's a similar request before) and we're looking for contribution from the community to implement the JS SDK in swagger codegen.

jfiala:
Added PR #1671, here are the details of the currently added changes:

Add basic Javascript SDK support (based on the Java templates):
* Allow to generate model classes
    * including comments
    * including property-level comments + type hints
    * including getters & setters + comment type hints

* Allow to generate API classes
    * including comments 
    * including JSDoc for params & return 

Sidenote: the Java templates have not yet been completely cleaned up, this is the first attempt to get Javascript support for Swagger-Codegen.

Sample model output: 
```javascript
/**
 * User object
 **/
function User() { 
  var self = this;
  
  /**
   * First name of the user
   * datatype: String
   **/
   self.firstname = null;

  
  /**
   * get First name of the user
   * @return {String}
   **/
  self.getFirstname = function() {
    return self.firstname;
  }
  
  /**
   * set First name of the user
   * @param {String} firstname
   **/
  self.setFirstname = function (firstname) {
    self.firstname = firstname;
  }
  

}
```

Sample API class output
```javascript
/*
 * @javax.annotation.Generated(value = "class io.swagger.codegen.languages.JavascriptClientCodegen", date = "2015-12-05T17:11:50.105+01:00")
 */
function UsercrudApi() {
	var self = this;
  
  /**
   * read User by ID
[...]

PR #1671, opened 2015-12-05
Add basic Javascript SDK support

jfiala:
see also issue #1294 "Document client SDK"

## Add basic Javascript SDK support
 (based on the Java templates):
* Allow to generate model classes
    * including comments
    * including property-level comments + type hints
    * including getters & setters + comment type hints
    * generate inline Enumeration classes to allow Enumeration usage
    * add required parameters to default constructor
    * add .toJson() to print complete model including nested models

* Allow to generate API classes
    * including comments 
    * including JSDoc for params & return  

* Allow for both AMD + plain JS usage


jfiala:
Only added new class "JavascriptClientCodegen", but it seems all classes implementing CodegenConfig are tested using their name.

However, I ran it using the full class invocation:
-l io.swagger.codegen.languages.JavascriptClientCodegen

The shorthand "javascript" didn't work:
 -l javascript

Please advise how to add a new language only by implementing interface CodegenConfig without breaking the tests.

wing328:
@jfiala thanks for the PR to add Javascript SDK :)

To add a new language, please add the following line to `modules/swagger-codegen/src/main/resources/META-INF/services/io.swagger.codegen.CodegenConfig`:

```
io.swagger.codegen.languages.JavascriptClientCodegen
```

I did a test and no longer see the same error after adding the line above.

jfiala:
added configuration entry in io.swagger.codegen.CodegenConfig, however it could be picked up automatically as it implements the interface CodegenConfig

wing328:
Agree with you that ideally it should be picked up automatically.

wing328:
@jfiala thanks again for your contribution of the JS SDK. I'm sure the community would appreciate your effort :beers:

PR #1673, opened 2015-12-07
[Javascript] add bin and petstore samples

wing328:
- added bin/javascript-petstore.sh
- added bin/windows/javascript-petstore.bat
- added petstore samples for JS


wing328:
cc @jfiala

PR #1690, opened 2015-12-09
[Javascript] Fix syntax issue when generating Javascript client operation without parameters

thibaultclem:
When generating Client JS for an operation without parameters, an unnecessary comma is added before the callback parameter:
```
self.logoutUser = function(, callback) {
```
You can see the error directly on the generating UserApi.js:
https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/javascript/src/scripts/rest/api/UserApi.js#L233

My PR add a hasParams check before adding the comma and the PetStore example generating with the fix.


wing328:
@thibaultclem thanks for the PR. The fix looks good to me.

PR #1696, opened 2015-12-10
Add library option in maven-plugin

Mumutu:
Add library option to make following configuration available
```
           <configuration>
                <inputSpec>swagger.yaml</inputSpec>
                <language>java</language>
                <library>retrofit</library>
                <configOptions>
                   <sourceFolder>src/main/java</sourceFolder>
                </configOptions>
            </configuration>
```

### DIFF
--- a/modules/swagger-codegen-maven-plugin/src/main/java/io/swagger/codegen/plugin/CodeGenMojo.java
+++ b/modules/swagger-codegen-maven-plugin/src/main/java/io/swagger/codegen/plugin/CodeGenMojo.java
@@ -95,6 +95,12 @@ public class CodeGenMojo extends AbstractMojo {
     private String language;
 
     /**
+     * Client language may has a library
+     */
+    @Parameter(name = "library", required = false)
+    private String library;
+
+    /**
      * Path to separate json configuration file.
      */
     @Parameter(name = "configurationFile", required = false)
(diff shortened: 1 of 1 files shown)


fehguy:
Because the `library` option applies only to the java target, I don't think it's appropriate like this.

Mumutu:
..But the config has a library getter/setter in `io.swagger.codegen.CodegenConfig`
```
    void setLibrary(String library);
    /**
     * Library template (sub-template).
     */
    String getLibrary();
```
So other language won't have a sub-template in the future? if then in my opinion it maybe weird the library getter/setter in the interface `io.swagger.codegen.CodegenConfig` ..

fehguy:
Yes, I agree that probably shouldn't be there, and it should only exist in the projects that need that specific support.

rdelaplante:
The swagger-codegen-maven-plugin wiki page says the configOptions parameter is "a map of language-specific parameters", and lists Java specific parameters in the next section:

https://github.com/swagger-api/swagger-codegen/tree/master/modules/swagger-codegen-maven-plugin

So, the library parameter could be supported like this:

```xml
<configuration>
    <inputSpec>swagger.yaml</inputSpec>
    <language>java</language>
    <configOptions>
        <sourceFolder>src/main/java</sourceFolder>
        <library>retrofit</library>
    </configOptions>
</configuration>
```

This is important for people who don't want to use the default library.  I want to use jersey2 and it seems like this is not yet possible with swagger-codegen-maven-plugin.

PR #1699, opened 2015-12-10
Add support for vendor extensions to CodegenModel/Property

cromwellian:
Fix issue with modelFileTemplates supporting the notion of
multiple templates per model, but not providing a way to
actually output multiple files per model. Added a
toModelFilename() call which takes a template filename
argument to allow subclasses to special case the output
filename.

This is patch to support a model generator extension that
generates @AutoValue models as well as persistence entities
as separate files.

[Review on Reviewable](https://reviewable.io/reviews/swagger-api/swagger-codegen/1699)

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConfig.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConfig.java
@@ -97,6 +97,8 @@ public interface CodegenConfig {
 
     String toModelFilename(String name);
 
+    String toModelFilename(String templateFilename, String name);
+
     String toModelImport(String name);
 
     String toApiImport(String name);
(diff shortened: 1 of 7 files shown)


wing328:
@cromwellian thanks for the PR. Can you share a Swagger spec that we can use to test the change?

cromwellian:
I'm about to go to bed, but I'll send one tomorrow morning. Essentially, one of the use cases, is I have an extension of the Java generators that generates JPA persistence entities separate from the models, so I not only need to generate a Customer.java, but a CustomerEntity.java file, so I have a model.mustache and a jpaentity.mustache both run during model generation.

it has methods like CustomerEntity.fromModel(customer) and CustomerEntity.toModel() to convert to and from for easy persistence.

In order to allow customization, I needed to allow you to configure/override things like ID columns (@Id), @Table name, and relations like @OneToMany, so I added extensions like

x-jpa-id: auto (auto generated), adds @Id/@GeneratorType in mustache file

and

x-jpa-relation: one-to-many

I guess if these are properly abstracted into the future than can be reused by other persistence frameworks (e.g. Ruby)

I'll send an updated patch with an example file tomorrow. I expect eventually I'll send a patch to swagger-codegen with my Java generator changes.

wing328:
@cromwellian please rebase to resolve the merge conflicts

You may want to review https://github.com/swagger-api/swagger-codegen/pull/1768 (merged into master), which adds support of vendor extension to model and property

wing328:

PR #1703, opened 2015-12-10
Simplify Jersey and Jersey2 client serialization

evigeant:
This PR changes the java Jersey and Jersey2 clients in order to let the framework (Jersey) do the serialization and deserialization of objects including JSON data.

This results in much less code, and also allows easier customization of data types in the generated clients. Specifically, if one wants to use an `java.io.InputStream` instead of a `byte[]`, then the parameter type or the return type can be changed without touching the serialization code (since Jersey knows how to serialize most basic types).


wing328:
@evigeant thanks for the PR. We'll review and let you know if we've any questions.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
```diff
@@ -244,6 +243,7 @@ public void processOpts() {
             // the "okhttp-gson" library template requires "ApiCallback.mustache" for async call
             supportingFiles.add(new SupportingFile("ApiCallback.mustache", invokerFolder, "ApiCallback.java"));
             supportingFiles.add(new SupportingFile("ApiResponse.mustache", invokerFolder, "ApiResponse.java"));
+            supportingFiles.add(new SupportingFile("JSON.mustache", invokerFolder, "JSON.java"));
```
xhh: The `JSON.mustache` template is required by the `jersey2` library as well. Currently there would be a compilation error of "can not find symbol (JSON)" if cleaning and regenerating the jersey2 petstore sample:

```
mvn clean package

rm -rf samples/client/petstore/java/jersey2
git checkout samples/client/petstore/java/jersey2/src/test

cd samples/client/petstore/java/jersey2
mvn test
```

evigeant: The PR modifes the Jersey2 client in a similar way, so the JSON is not needed in that case either. It is now only necessary for okhttp-gson, hence I moved it there in JavaCodegen.

I have regenerated both clients (jersey and jersey2) in the PR and all tests pass with the new clients. I had to remove the unit tests on JSON since that class no longer exists in jersey and jersey2 case.

xhh: Looking at [here in ApiClient](https://github.com/evigeant/swagger-codegen/blob/feature/simple-jaxrs-serialization/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache#L551) and [JSON](https://github.com/evigeant/swagger-codegen/blob/feature/simple-jaxrs-serialization/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/JSON.mustache), obviously the `JSON` class still exists in the jersey2 template, right?

evigeant: You are right, I removed entirely from the jersey1 code, but not from the jersey2 code, I will update the PR soon.

Issue #1705, opened 2015-12-10
CSharp Clients Do Not Format query parameter date-time As Per Swagger Spec


mirthy:
The CSharp Client serializes the date-time parameters incorrectly.  The format chosen is not ISO 8601 / RFC3339 compatible.

The Universal Sortable ("u") Format Specifier was used instead of the Round-trip ("O", "o") Format Specifier for converting parameters. [Microsoft Documentation on the Round-trip format specifier ("o"). ](https://msdn.microsoft.com/en-us/library/az4se3k1(v=vs.110).aspx#Anchor_8)

Currently: 2008-04-10 13:30:00Z  (note the missing T separator)
Should be: 2008-04-10T13:30:00.0000000

See:
https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache#L159

This bug completely breaks any service which strictly enforces RFC3339 formatting. 

There is a bit of deserializing code that's parsing it correctly using that same Round Trip format but it's unrelated here: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache#L212

wing328:
PR merged into master

PR #1706, opened 2015-12-11
[C#] Fix date-time serialization to be ISO 8601

mirthy:
Fix the date-time serialization so that it's actually ISO 8601 compliant. Currently, it generates a date-time string that's close but will fail on stricter services.
Added unit tests to verify the change on the client side
Issue #1705


wing328:
@mirthy thanks for the PR with test cases to cover the change.

May I suggest adding an option in the configuration object (e.DateTimeFormat) so that developers can customize the date time format if the default one ("o") does not meet their requirement?

mirthy:
Ok, sure, I'll take a look at adding that now.

wing328:
Thank you @mirthy

PR #1710, opened 2015-12-11
spring-mvc-j8: Uses Java 8 interface default and provide async callback

icha024:
This template is mainly for Maven code-gen plugin use-case. When Swagger
spec on **existing** project changed there is no need to manually copy/paste
the new functions from the generated client. This will provide a default
(empty) implementation to existing impl and user just need to override
it.

Because it is an interface, an actual implementation will be needed to
actuate a service end-point (don't forget the @Controller tag on the impl class!).


wing328:
@icha024 thanks for the PR. Instead of creating another template for SpringMVC Java 8, what about adding an option to the existing springmvc template to support the Java 8 feature in this PR?

In the Java API client, we allow users to choose an HTTP library (Jersey 1.x, Jersey, 2.x, okhttp, retrofit, retrofit2) and I think we can use similar approach here.

icha024:
Great idea! I'll look into doing it that way instead

PR #1718, opened 2015-12-15
[C#] Fixed ArgumentOutOfRangeException. Optimized string concatenation.

artem-dudarev:
If the list is empty you get the ```ArgumentOutOfRangeException``` when calling ```flattenString.Remove(flattenString.Length - 1)``` because it tries to remove a separator from the empty string.
Also using a ```StringBuilder``` is the prefered way to concatenate strings in a loop.


wing328:
@artem-dudarev thanks for the PR. The change looks good to me.

Will merge it shortly.

wing328:
Next time please update the Petstore sample as mentioned in https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

PR #1725, opened 2015-12-15
[C#] fix filename and Content-Disposition parsing on FileStream

hacki11:
Having a service which returns a FileStream:
@ApiOperation(value = "file", response = java.io.File.class)

the generated ApiClient.cs includes handling FileStreams:
var regex = new Regex(@"Content-Disposition:.*filename=['""]?([^'""\s]+)['""]?$");

but should be Content-Diosposition=...

Additionally the regex is applied onto: headers.ToString()
which has the value: "System.Collections.Generic.List`1[RestSharp.Parameter]"

Therefor, a filename will never be correctly parsed.


wing328:
@hacki11 thanks for the PR. Would you please update the Petstore sample (C#) as mentioned in https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md ?

PR #1726, opened 2015-12-16
Bumped Alamofire library (dependency for Swift client). Alamofire 3.1.x is compatible with previous version and resolves a few annoying bugs, like proper handling of empty HTTP responses.

tomekc:


wing328:
@tomekc thanks for the PR and would be nice if you can update the Swift sample as mentioned in https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

tomekc:
Sure, thanks for pointing it out. I also updated Cartfile mustache I forgot about.

tomekc:
And excuse me that lengthy title. I've used command line 'hub' tool for first time :)

PR #1731, opened 2015-12-17
[jaxrs] Updated generator to use additional properties instead of system properties

evigeant:
In the java jaxrs server generation code, Java system properties were used to configure the output folder, the code packages and the implementation output folder.

There are already facilities to configure the output folder and the packages, so these were used instead of the system properties. For the implementation package, the additional properties were used instead of system properties as this seems to be the standard mechanism to pass additional configuration to the generator.

Java system properties are very hard to set from a maven build and this prevented me from confguring the implementation folder correctly from my maven build.


wing328:
@evigeant thanks for the PR. The change looks good to me as it replaces system property with additional configuration to make the user experience more consistent.

I'll merge the PR tomorrow.

wing328:
@fehguy do you want to make similar change to the Java inflector codegen? Ref: https://github.com/swagger-api/swagger-codegen/blob/55fece195d35103f3bd9396f7780fd3c386bcf2e/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaInflectorServerCodegen.java#L27

fehguy:
@wing328 let me get back on this--I need to make sure the calling syntax is either back compatible or very well documented for each client

fehguy:
OK we should merge this.  @evigeant could you please rebase?

evigeant:
I will rebase this soon, I'm not home currently

evigeant:
@wing328 I merged master so that this branch could be merged automatically, however, the Travis CI build failed with an error in the Petstore Clojure client unrelated to what I was fixing.

What should I do now? Should I attempt to fix the clojure error? Or is it  random timeout error and I should push a fake commit so that Travis CI builds again?

wing328:
@evigeant I saw that you've pushed another commit and the CI passed. We'll review and approve this PR shortly. Thanks for the contribution.

wing328:
Tests look good.

Issue #1540, opened 2015-11-09
'additionalProperties' breaks iOS codegen


samskiter:
additionalProperties breaks objective-c code generation as it causes generated classes to inherit from NSMutableDictionary, rather than from JSONModel. This would be fine but the generated class doesn't implement `initWithDictionary` either so the check inside SWGApiClient to allow it to be deserialised won't work:

```
   // model
    Class ModelClass = NSClassFromString(class);
    if ([ModelClass instancesRespondToSelector:@selector(initWithDictionary:error:)]) {
        return [[ModelClass alloc] initWithDictionary:data error:nil];
    }
```

wing328:
@samskiter can you share the swagger spec us to repeat the issue ?

samskiter:
hey, here's a schema I used:
```
{
    "id": "sitesGETresponse.schema.json#",
    "$schema": "http://json-schema.org/draft-04/schema#",
    "title": "Site list",
    "description": "Response to site list request",
    "type": "object",
    "properties": {
        "title": {
            "type": "string"
        }
    }, 
    "additionalProperties": {
        "type": "object",
        "properties": {
            "title": {
                "type": "string"
            },
            "count": {
                "type": "integer"
            },
            "uuid": {
                "type": "string"
            }
        },
        "required": ["uuid"]
    }
}
```

samskiter:
(see also https://github.com/swagger-api/swagger-codegen/issues/453#issuecomment-154570630)

wing328:
OK. It's caused by #1470 (map of inline model in this case)

samskiter:
Ok, I'll take your word for it - on the face of it it looks like a different bug. So the generated code should implement initWithDictionary? And inherit from NSMutableDictionary as it currently does?

It might also be worth looking at supporting the recently introduced templates to objective-c. I think it's only available for arrays and dictionaries, but better than nothing ! :)

wing328:
Btw, to workaround the issue for the time being, you can define the model properly in the `definitions` section of your spec.

samskiter:
you mean like this?

```
{
    "id": "sitesGETresponse.schema.json#",
    "$schema": "http://json-schema.org/draft-04/schema#",
    "title": "Site list",
    "description": "Response to site list request",
    "type": "object",
    "properties": {
        "title": {
            "type": "string"
        }
    }, 
    "additionalProperties": {
        "$ref" : "otherfile.json"
}
```

Otherfile.json:

```
{
[...]

PR #1734, opened 2015-12-17
Fix array/map inline response/model

### LINKED ISSUE #1470: Inline models with array/map
Inline models with array/map are not generated correctly. Here is the spec to repeat the issue:

https://gist.githubusercontent.com/wing328/359166515dae7f489ce5/raw/4c9f39efe09c0cf7220864e6476d9253bc2e6e80/petstore_inline_model.json.

For array of inline models:
```
  /**
   * Finds Pets by tags
   * Muliple tags can be provided with comma seperated strings. Use tag1, tag2, tag3 for testing.
   * @param tags Tags to filter by
   * @return List<Object>
   */
  public List<Object> findPetsByTags (List<String> tags) throws ApiException {
    Object postBody = null;
    byte[] postBinaryBody = null;
```
The correct return type is `List<InlineResponse200>` where InlineResponse200 is the inline model created for the response.

For map of inline models:
```
  /**
   * Returns pet inventories by status
   * @return InlineResponse2001
   */
  public InlineResponse2001 getInventory () throws ApiException {
    Object postBody = null;
    byte[] postBinaryBody = null;
```

The correct return type is `Map<String, InlineResponse200>` where InlineResponse200 is the inline model created for the response)

wing328:
To fix #1470, #1540

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/InlineModelResolver.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/InlineModelResolver.java
@@ -13,9 +13,13 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public class InlineModelResolver {
     private Swagger swagger = null;
     private boolean skipMatches = false;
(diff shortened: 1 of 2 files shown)


wing328:
Fixing the CI tests

wing328:
cc @fehguy

fehguy:
Thanks, I'll get through this shortly

PR #1742, opened 2015-12-19
Spring-MVC config "j8-async": Uses async servlet & Java 8 interface

icha024:
default

This template is mainly for Maven code-gen plugin use-case - when
Swagger spec on existing project changes there is no need to manually
copy/paste the new functions from the generated client. This will
provide a default (empty) implementation to existing impl and user just
need to override the stub implementation.

Because it generates an interface instead of a concrete stub, an
implementation will be needed to actuate a service end-point. And don't
forget to put @Controller on the implementation!


wing328:
@icha024 thanks for the PR but the CI tests failed. Here is the log: https://api.travis-ci.org/jobs/97884873/log.txt?deansi=true

Please review and let us know if you need help addressing the issue reported by CI.

I would suggest you to add `./bin/spring-mvc-j8-async-petstore-server.sh` to generate the output to `samples/server/petstore/spring-mvc-j8-async` using the new option so that we can use CI to test the output (as specified in .travis.yml)

icha024:
@wing328 All done

wing328:
@icha024 thanks but I don't see `./bin/spring-mvc-j8-async-petstore-server.sh` and the sample. Did you do `git add` to those new files?

icha024:
@wing328 bin scripts are now added

wing328:
@icha024 when running 'mvn test` in the sample folder, I got the errors with the sample. Here is part of the error message:
```
[INFO] Compiling 16 source files to /private/var/tmp/pr/icha024/swagger-codegen/samples/server/petstore/spring-mvc/target/classes
[ERROR] COMPILATION ERROR : 
[ERROR] /private/var/tmp/pr/icha024/swagger-codegen/samples/server/petstore/spring-mvc/src/main/java/io/swagger/api/UserApi.java:[48,3] illegal start of type
[ERROR] /private/var/tmp/pr/icha024/swagger-codegen/samples/server/petstore/spring-mvc/src/main/java/io/swagger/api/UserApi.java:[48,11] = expected
[ERROR] /private/var/tmp/pr/icha024/swagger-codegen/samples/server/petstore/spring-mvc/src/main/java/io/swagger/api/UserApi.java:[48,19] ';' expected
[ERROR] /private/var/tmp/pr/icha024/swagger-codegen/samples/server/petstore/spring-mvc/src/main/java/io/swagger/api/UserApi.java:[48,40] <identifier> expected
[ERROR] /private/var/tmp/pr/icha024/swagger-codegen/samples/server/petstore/spring-mvc/src/main/java/io/swagger/api/UserApi.java:[48,41] ';' expected
[ERROR] /private/var/tmp/pr/icha024/swagger-codegen/samples/server/petstore/spring-mvc/src/main/java/io/swagger/api/UserApi.java:[48,52] illegal start of type
[ERROR] /private/var/tmp/pr/icha024/swagger-codegen/samples/server/petstore/spring-mvc/src/main/java/io/swagger/api/UserApi.java:[48,53] <identifier> expected
```
Did you get similar error when running `mvn test`?

I would also suggest updating `/bin/spring-mvc-petstore-j8-async-server.sh` to output the generated code into `samples/server/petstore/spring-mvc-j8-async` instead

icha024:
@wing328 The spring-mvc j8-async code is now located in it's own sample directory with Java 8 enforced via Maven plug-in.

cbornet:
@icha024 If I understand correctly, a concrete class annotated by @Controller needs to be added manually or the controller will not be created/scanned. Is that correct ?
Then I think the @Controller on the interface is useless.

icha024:
@cbornet Yes that is correct, it's not needed. In hindsight it should've been removed and replaced with a comment instead.

BTW, if you are interested in using this with Maven, here's the plugin snippet that might be useful: 
http://www.clianz.com/2016/05/29/java-mvc-swagger-gen/

cbornet:
In jaxrs generators the implementation class is generated if it doesn't exists. Would that work with your plugin ? It would be useful for the first generation.

icha024:
For the first generation that's ok, but the challenge we have is keeping the spec and code synced when spec changes (often pushed by a different team) - so an interface would work better for that. The 'default' implementation part is a nice to have.

cbornet:
I mean that in jaxrs, there is an interface and an implementation. When you regenerate, the implementation is not overridden.

icha024:
Cool, didn't know that. That should work

PR #1770, opened 2015-12-27
add basic GUI app for swagger-codegen

championswimmer:
This is nowhere ready to be merged right now. Only putting this PR up to see if there is a need for this or will this be accepted. If yes, I'll work further on this. 

Right now only allows selecting a input file, and output directory, and language. 

Signed-off-by: Arnav Gupta <championswimmer@gmail.com>

### DIFF
--- a/modules/swagger-codegen-gui/pom.xml
+++ b/modules/swagger-codegen-gui/pom.xml
@@ -1,0 +1,103 @@
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <parent>
+        <groupId>io.swagger</groupId>
+        <artifactId>swagger-codegen-project</artifactId>
+        <version>2.1.5-SNAPSHOT</version>
+        <relativePath>../..</relativePath>
+    </parent>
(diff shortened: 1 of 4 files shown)


wing328:
Thanks @championswimmer. Personally I don't see any drawback adding a GUI for swagger codegen :)

webron:
@championswimmer - just saw your message on IRC. Not sure if you're aware of it, but both the editor and the code generation can be ran in a closed environment that doesn't have access to the internet. The codegen has the 'generator' module which is what's deployed at http://generator.swagger.io/ and the editor, which can be installed locally, can be configured to run against the local deployment.

My main concern with adding another GUI module is that it would be difficult to maintain (for us) and would end up rotting.

championswimmer:
@webron I understand that, and hence I wanted to make sure if this is needed.

I'll present my personal case - I know that they can be ran in a closed environment, but you do need to setup node and download quite a few npm dependencies, then you also have to setup and run a Java webserver for the generator. 

Instead, I have found using the codegen-cli much more simple. And hence was trying to achieve GUI functionality from that angle  - a portable, cross-platform dependency-free executable jar - that I can easily drop my swagger.yaml into, and it spews out the code.

webron:
@championswimmer - don't get me wrong, I'm not saying the idea isn't good. I just don't know if it would do it good to be part of this project rather than as a standalone project that you (or someone else) maintains.

fehguy:
Thanks--I'm going to close this out in favor of the swagger-generator or swagger-codegen-cli.  Thank you!

PR #1775, opened 2015-12-28
Android volley client generator

Shyri:
Hi!

I noticed there was no generator for Android using Volley library.

(See http://developer.android.com/training/volley/index.html)

I made it with the same functionality that has Android and Java generator at least. Oauth2 support is also to be done. hope it to be useful.

Thanks


Shyri:
I also made this project in order to test the generation of samples available on the online editor

https://github.com/Shyri/swagger-android-volley-test

fehguy:
Thank you for making this.  @wing328 can we use the `-library` config with Android to pick this up?

wing328:
@fehguy yes, technically it should be possible.

@Shyri thanks for the contribution. The PR contains merge conflicts. Please rebase on the latest master when you've time.

Shyri:
@wing328 Merge conflict resolved, also travis passed. Thank you both

cbornet:
The current android lib uses the outdated Apache client lib so IMO the generator using Volley could replace the old one or at least be the default. Also, the java libs can be used for Android (especially the ones based on okhttp and retrofit) so there should be a way to mention that somewhere.

wing328:
+1 for @cbornet suggestion

If a developer needs the outdated Apache client lib for whatever reason, they can use stable release 2.1.4.

In other words, v2.1.5 will have [Android Volley](https://github.com/mcxiaoke/android-volley) (which is pretty popular, active according to the Github page) as the default HTTP lib in the Android client.

wing328:
As discussed with @fehguy we'll eventually merge this into android template using the `library` option.

Merging this for the time being and we'll have another PR to merge that later.

modules/swagger-codegen/src/main/resources/android-volley/build.mustache
```diff
@@ -0,0 +1,91 @@
...
+    httpclient_version = "4.3.3"
+    junit_version = "4.8.1"
+}
+
+dependencies {
+    compile "io.swagger:swagger-annotations:$swagger_annotations_version"
+    compile "com.google.code.gson:gson:$gson_version"
+    compile "org.apache.httpcomponents:httpcore:$httpclient_version"
+    compile "org.apache.httpcomponents:httpmime:$httpclient_version"
+    testCompile "junit:junit:$junit_version"
```
xhh: in this template there's no volley dependency, but in the build.gradle file of petstore sample there is:
`compile "com.mcxiaoke.volley:library:${volley_version}@aar"`
and the `compileSdkVersion` there has the value `23` but here is `22` in the template.
so which one is right? we need to keep the generated sample in sync at least

Shyri: Hi
Sorry I missed this.

compile `compile "com.mcxiaoke.volley:library:${volley_version}@aar"` should be added to build.mustache, yes. Also     `volley_version = "1.0.19"` in the `ext` section.

I think compileSdkVersion might be updated to `23` because it is the lasestes (Android 6), that's why I used it in the samples.

Should I correct these and send a new PR?

xhh: I see. I can add those 2 lines in my PR of changing android-volley to a sub template of the android client codegen.

I have are two more questions: after adding the above 2 lines:
1. I got this compilation exception when running `gradle build` in the petstore sample:

```
src/main/java/io/swagger/client/api/StoreApi.java:51: error: cannot find symbol
  public void getInventory (final Responses.MapResponse responseListener, final Response.ErrorListener errorListener) {
                                           ^
  symbol:   class MapResponse
  location: class Responses
```
That is, `MapResponse` is not present in the `Responses` class.

2. I also got a warning:

```
WARNING: Dependency org.apache.httpcomponents:httpclient:4.3.3 is ignored for debug as it may be conflicting with the internal version provided by Android.
```

Is apache httpclient and httpmime still being used here? If not, we should remove the dependency, otherwise it would be better to remove the warnings somehow.

wing328:
@xhh's PR has been merged into master.

@Shyri let us know if you need help fixing map (`additionalProperties`) with the Android volley client.

PR #1777, opened 2015-12-29
[Java] Move method StringUtil.toIndentedString to model

### LINKED ISSUE #1701: If I generate models only, I will be missing a class called io.swagger.client.StringUtil
## How to reproduce:

```` shell
java -jar -Dmodels  modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate   -i http://localhost:8080/xxx/swagger.json -l java  -o /somedir  -c  /some.json
````
the some.json is 

````json
{"artifactId":"xxx","artifactVersion":"1.0-SNAPSHOT","groupId":"xxx","library":"jersey2","modelPackage":"xxx.model"}
````

## The problem:

The generated model files will be depending on "io.swagger.client.StringUtil" in their toString() methods, which is not generated .

## Suggestion
We don't need very beautiful output for toString() methods.  I think you can just remove the dependency for StringUtil.

xhh:
Closes #1701


chenjianjx:
There is still one error:  "import {{invokerPackage}}.StringUtil;"  in model classes. 
Please remove the importing too.

xhh:
I'll do it, thanks for reminding.

fehguy:
@xhh I did just update the java model templates, might want to merge those changes at the same time.

xhh:
@fehguy noticed that the StringUtil import has been fixed (removed) by you, I'll submit a PR to remove it from other templates (okhttp-gson and retrofit).

xhh:
@fehguy I see what you meant now. I'll remove the StringUtil import from Java models in JavaClientCodegen.java

fehguy:
OK great--I did update the jaxrs, inflector, and java clients.

dantran:
@xhh 

please check again, my latest generated java client with 2.1.5-SNAPSHOT still have this code in the model

@Override
  public String toString()  {
    sb.append("class Id {\n");
    
    sb.append("    id: ").append(StringUtil.toIndentedString(id)).append("\n");
    sb.append("}");
    return sb.toString();
  }

dantran:
Looks like there are changes just came in few hour ago and  my build uses the snapshots deploy before the changes

PR #1783, opened 2015-12-29
[Akka-Scala] Fix #1781 re specs w/o security defs

### LINKED ISSUE #1781: akka-scala implemention NPE on spec with no securityDefinitions property
Seen in release 2.1.4, and in latest commit on master branch: 564f6e2bb5ff22bdf08672c103c5a14dffdef60a

In the `fromSecurity` method override in `AkkaScalaClientCodegen`, the iterator that looks for an OAuth security definition fails with NPE if the spec contains no `securityDefinitions` property.

The superclass method explicitly returns null in this case, and that null value is used by the akka-scala module, leading to the NPE.

To reproduce this, remove (not just empty out) the `securityDefinitions` property in petstore (and remove all `security` properties, where the removed definitions are referenced).

Possible fixes include:

1. Check for null in `AkkaScalaClientCodegen#fromSecurity`
2. Return an empty list in `DefaultCodegen#fromSecurity` instead of null

Latter seems preferable to me, since an empty list should have the same meaning as omitting the `securityDefinitions` property, and it would prevent others from falling into the same trap in the future.

andylowry:
Per issue recommendation, fix is to DefaultCodegen#fromSecurity, which
now returns an empty list rather than null if no securityDefinitions
property is present in the swagger spec

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -1602,7 +1602,7 @@ public class DefaultCodegen {
      */
     public List<CodegenSecurity> fromSecurity(Map<String, SecuritySchemeDefinition> schemes) {
         if (schemes == null) {
-            return null;
+        	return Collections.emptyList();
         }
 
         List<CodegenSecurity> secs = new ArrayList<CodegenSecurity>(schemes.size());


wing328:
@andylowry thank you for the PR. We'll review and let you know if we've any question.

wing328:
I am able to repeat the issue and confirm the no longer seeing the NPE with the PR.

PR #1785, opened 2015-12-30
added `serverPort`

### LINKED ISSUE #1748: JAXRS code generator does not use the Swagger host in the pom.xml and READ.me files
The JAXRS code generator always uses port 8080 in the generated pom.xml for Jetty instead of what the Swagger host field file defines. This port (8080) shows up both in the generated pom.xml and in the generated READ.me.

fehguy:
fixes #1748

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JaxRSServerCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JaxRSServerCodegen.java
@@ -171,7 +171,7 @@ public class JaxRSServerCodegen extends JavaClientCodegen implements CodegenConf
         String port = "8080";
         if(host != null) {
             String[] parts = host.split(":");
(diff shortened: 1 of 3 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JaxRSServerCodegen.java
```diff
@@ -166,6 +166,16 @@ public void preprocessSwagger(Swagger swagger) {
         if("/".equals(swagger.getBasePath())) {
             swagger.setBasePath("");
         }
+
+        String host = swagger.getHost();
+        String port = "8080";
+        if(host != null) {
+            String[] parts = host.split(":");
+            if(parts.length > 0) {
```
wing328: I think it should be `parts.length > 1` instead (no CI failure after changing to `>1`)

PR #1787, opened 2015-12-30
added BigDecimal java serialization options

fehguy:
This commit adds an option to java clients to serialize BigDecimal values as strings.  It also gives the option to post-process models, for injecting other vendor extensions, imports, etc.


cbornet:
@fehguy Shouldn't it be done also for deserialization ?
Also, since we have {"type":"string", "format":"number"} in the swagger def, what is the point of having bigDecimalAsString ? The def is enough to say how the field should be serialized.

cbornet:
And shouldn't the other languages deal with the "BigDecimal" data type (and map it to string for instance)

PR #1790, opened 2015-12-30
updated templates for pojos, enums

### LINKED ISSUE #1789: update java-based codegens
With the latest commits for enums and more modular files, all the JVM-based clients should be updated.

fehguy:
fixes #1789


naveen237:
Can we update the enum templates to generate java code properly for empty strings ?
If we have a enum with string values like this in my YAML:
Type:
        description: Type of parameter.
        type: string
        enum: ["","String","boolean","Date","integer","decimal"]

The generated classes will have something like this::

public enum TypeEnum {
    (""),
    STRING("String"),
    BOOLEAN("boolean"),
    DATE("Date"),
    INTEGER("integer"),
    DECIMAL("decimal");

which wont compile. I think its a valid usecase and needs to be addressed

fehguy:
Yes you are correct.  Will address that.

naveen237:
Could you tell me whats the targeted release for the change ?

PR #1795, opened 2015-12-31
corrected docs to use config.json

rbpasker:
The docs say you can use -DpackageName=mypackage, but that doesn't work. Changed the docs so they show the proper way using a `config.json. file

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,17 +336,21 @@ TypeScriptAngularClientCodegen.java
 TypeScriptNodeClientCodegen.java
 ```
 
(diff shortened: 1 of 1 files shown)


wing328:
@rbpasker thanks for the PR. Currently, some generators use system properties to obtain configuration. We're in the process to migrate those to use config.json instead.

wing328:
@rbpasker I've updated the wording via https://github.com/swagger-api/swagger-codegen/commit/56951a2df426df368638e380da97db3797529fd3

Issue #1802, opened 2015-12-31
[python] use of Configuration() as default parameter evaluates at definition-time, not runtime


rbpasker:
using this idiom, Configuration().host is evaluated when the method is defined, which is when `from swagger_client import SomeClient` is executed by the interpreter, and **NOT** when the constructor is called.

````
    def __init__(self, host=Configuration().host,
                 header_name=None, header_value=None, cookie=None):

````

https://github.com/Tengah/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/python/api_client.mustache#L69

wing328:
PR merged. Closing the issue.

PR #1803, opened 2015-12-31
Fixed configuration of host in api_client so that it gets done at run-time not load-time

rbpasker:
in Python, a method used as the default value of a formal parameter defines the value when the function is defined (normally at `import` time), not when the method is invoked

So if someone tried 

````
Configuration().host = mySpecialHost
api = MySpecialApi()
````

it would not work because the Configuration().host in api_client was invoked when the module was loaded, not when the constructor is run.

Note: I tested this by editing the generated python code, not the .mustache file

### DIFF
--- a/modules/swagger-codegen/src/main/resources/python/api_client.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/api_client.mustache
@@ -66,8 +66,7 @@ class ApiClient(object):
     :param header_name: a header to pass when making calls to the API.
     :param header_value: a header value to pass when making calls to the API.
     """
-    def __init__(self, host=Configuration().host,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by wing328** [comment]
@rbpasker thanks for the PR. I tested it and the result looks good:
```
OK
______________________________________________ summary ______________________________________________
  py27: commands succeeded
  py34: commands succeeded
  congratulations :)
```
Next time, please update the Petstore sample as mentioned in the contributing guildeline: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

wing328:
FYI. For fixing https://github.com/swagger-api/swagger-codegen/issues/1802

PR #1812, opened 2016-01-05
[Perl] more flexible perl module naming (Allow::This even without WWW::)

captin411:
- No more presuming 'WWW::' is at the beginning
  (default is now "WWW::SwaggerClient" vs "SwaggerClient" in order
   to accomodate)
- Test that module names Like::This write to the filesystem and in
  the mustache templates properly


wing328:
@captin411 thanks for the PR. We'll review and let you know if we've any questions.

wing328:
The change looks good and I also tested locally.

PR #1857, opened 2016-01-08
Java Class Name Clashes with java.lang.*

erikvanzijst:
A swagger file that contains a definition named "object" compiles into a
class named Object. This is technically not a problem as it lives in a
different package than java.lang.Object, but this requires the templates to
refer to Java's Object using its fully qualified name.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/pojo.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/pojo.mustache
@@ -27,7 +27,7 @@ public class {{classname}} {{#parent}}extends {{{parent}}}{{/parent}} {{#seriali
   {{/vars}}
 
   @Override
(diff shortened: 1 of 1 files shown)


erikvanzijst:
This addresses issue #1857

wing328:
@erikvanzijst thanks for the PR. Next time please update the Petstore sample as mentioned in https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md.

Also the Java API client supports different HTTP libraries (Retrofit, Jersey2, etc) and I believe the model templates for these HTTP libraries also need to be updated. Later I'll file a PR to update those.

erikvanzijst:
You're right, I should've added a test and will do so next time. I'm still learning my way around the codebase.

Issue #1858, opened 2016-01-08
The Java equals/hashcode methods ignores state from superclasses


erikvanzijst:
The following object definitions hierarchy produces POJOs who's equals and hashcode methods fail to take into account the member variables inherited from the super classes:

```javascript
{
  ...
  "definitions": {
    "object": {
      "required": [
        "type"
      ],
      "additionalProperties": true,
      "discriminator": "type",
      "type": "object",
      "properties": {
        "type": {
          "type": "string"
        }
      }
    },
    "account": {
      "allOf": [
        {
          "$ref": "#/definitions/object"
        },
        {
          "additionalProperties": true,
          "type": "object",
          "description": "An account object.",
          "properties": {
            "username": {
              "pattern": "^[a-zA-Z0-9_\\-]+$",
              "type": "string"
            },
            "display_name": {
              "type": "string"
            },
            "uuid": {
              "type": "string"
            },
            "created_on": {
              "type": "string",
              "format": "date-time"
            }
          }
        }
      ]
    },
    "team": {
      "allOf": [
        {
          "$ref": "#/definitions/account"
        },
        {
          "additionalProperties": true,
          "type": "object",
          "description": "A team object.",
          "properties": {}
        }
      ]
    }
  }
}
```

```java
package io.swagger.client.model;

import java.util.Objects;
import io.swagger.client.model.Account;

@javax.annotation.Generated(value = "class io.swagger.codegen.languages.JavaClientCodegen", date = "2016-01-07T16:45:23.501-08:00")
public class Team extends Account  {
  @Override
  public boolean equals(java.lang.Object o) {
    if (this == o) {
      return true;
    }
    if (o == null || getClass() != o.getClass()) {
      return false;
    }
    Team team = (Team) o;
    return true;
  }

  @Override
  public int hashCode() {
    return Objects.hash();
  }
```

The `team` type in our schema acts like a marker interface, in that it does not introduce new state and the resulting `equals` method simply always returns `True` for all `Team` objects.

PR #1861, opened 2016-01-08
#1858: The Java equals/hashcode methods ignores state from superclasses

erikvanzijst:
This fixes equals/hashCode for derived classes.

Example of the generated equals/hashCode on a POJO that extends another class:

```java
  @Override
    if (this == o) {
      return true;
    }
      return false;
    }
    Account account = (Account) o;

    return true && Objects.equals(website, account.website) &&
        Objects.equals(createdOn, account.createdOn) &&
        Objects.equals(links, account.links) &&
        Objects.equals(displayName, account.displayName) &&
        Objects.equals(uuid, account.uuid) &&
        Objects.equals(username, account.username)
    && super.equals(o);
  }

  @Override
  public int hashCode() {
    return Objects.hash(website, createdOn, links, displayName, uuid, username, super.hashCode());
  }
```

And one that does not have a parent:

```java
  @Override
    if (this == o) {
      return true;
    }
      return false;
    }
    Object object = (Object) o;

    return true && Objects.equals(type, object.type)
    ;
  }

  @Override
  public int hashCode() {
    return Objects.hash(type);
  }
```

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/pojo.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/pojo.mustache
@@ -34,15 +34,16 @@ public class {{classname}} {{#parent}}extends {{{parent}}}{{/parent}} {{#seriali
     if (o == null || getClass() != o.getClass()) {
       return false;
     }
-    {{classname}} {{classVarName}} = ({{classname}}) o;{{#hasVars}}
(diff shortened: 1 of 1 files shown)


wing328:
@erikvanzijst thanks for the PR. Likely we need similar fix to C# models. I'll create a task for tracking.

PR #1862, opened 2016-01-09
Add auth support to feign library

### LINKED ISSUE #1840: [Feign] Expose Feign builder
The feign builder should be exposed so that we can customize it without touching the generated code (to add basic auth for instance)
Something like
```java
public ApiClient() {
    objectMapper = createObjectMapper();
    feignBuilder = Feign.builder()
            .encoder(new FormAwareEncoder(new JacksonEncoder(objectMapper)))
            .decoder(new JacksonDecoder(objectMapper))
            .logger(new Slf4jLogger());
  }

public Feign.Builder getFeignBuilder() {
    return feignBuilder;
  }

  public void setFeignBuilder(Feign.Builder feignBuilder) {
    this.feignBuilder = feignBuilder;
  }

public <T extends Api> T buildClient(Class<T> clientClass) {
        return feignBuilder.target(clientClass, basePath);
  }
```

cbornet:
Based on the retrofit generator methods.
I will do OAuth later since it is more work.

Fix #1840

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
@@ -223,12 +223,11 @@ public class JavaClientCodegen extends DefaultCodegen implements CodegenConfig {
         final String authFolder = (sourceFolder + '/' + invokerPackage + ".auth").replace(".", "/");
         if ("feign".equals(getLibrary())) {
             supportingFiles.add(new SupportingFile("FormAwareEncoder.mustache", invokerFolder, "FormAwareEncoder.java"));
-        } else {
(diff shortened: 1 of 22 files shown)


cbornet:
Added OAuth support based on OLTU (same as retrofit lib)

cbornet:
Seems that the CI is broken on ruby sample test...

wing328:
I could repeat the issue 10 minutes ago and now the Ruby tests pass locally in my machine. Looks like something has changed/fixed in the server side. I'll restart the CI job to see how it goes.

cbornet:
Just some information on the oauth implementation : since it uses Feign's request interceptor, it cannot renew the token when a 401 error is received (unlike retrofit/okhttp for instance). So the "expires_in" field is used instead and the request might get a 401 if it is done very close to the expiration date. If someone has a better solution ...

cbornet:
You can always use a retryer to mitigate this issue.

PR #1871, opened 2016-01-11
Using logger instead of System.out

rostskadat:


rostskadat:
I tested the build on a sample project with
?> mvn --errors clean install 
?> mvn --quiet clean install 
?> mvn --debug clean install 

Not sure how to test that automatically: it's purely visual while building

rostskadat:
The command "mvn verify -Psamples" exited with 1.

Hum... Who knows Ruby?

wing328:
@rostskadat the issue with Ruby seems to be caused by the server. Now it works locally for me (while I could repeat the issue 20 minutes ago). I'll restart the CI test to see how it goes.

modules/swagger-codegen/src/main/java/config/ConfigParser.java
```diff
@@ -7,11 +7,16 @@
...
+import org.slf4j.LoggerFactory;
+
 public class ConfigParser {
 
+    private static final Logger LOG = LoggerFactory.getLogger(ConfigParser.class);
+
     public static Config read(String location) {
 
-        System.out.println("reading config from " + location);
+    	LOG.debug("reading config from " + location);
```
wing328: @rostskadat thanks for the PR. For consistency, can we use LOGGER instead of LOG?

Also there seems to be some formatting issue in this files and other files, e.g. above line is missing one space.

rostskadat:
Ok. I changed to LOGGER and fixed some indentation problems...

wing328:
@rostskadat thanks!

I did a test and noticed that some `System.out.println` are replaced with either LOGGER.debug or LOGGER.info.

Can we change those to LOGGER.info ? For example current behaviour of `-DdebugModels` is to print out the debugging info for models in the console so it would be nice if we can keep that.

hiveship:
It's a good thing to use LOGGER insteand of System.out. I made comment about it in https://github.com/swagger-api/swagger-codegen/pull/1900

rostskadat:
@wing328 OK.

I changed the LOGGER level in order to allow the following switches to output at the INFO level
debugModels
debugOperations
debugSupportingFiles

rostskadat:
@wing328 github tells me that the branch maven_build_quiet has some conflicts that need to be resolved.
I use Eclipse, and I tried to do a Rebase, a Synchronize with origin/master, but no changes appear except the ones I did...
I'm not really sure how to proceed from there.

wing328:
@rostskadat thanks. Please pull the latest and rebase on the master to resolve the merge conflicts.

wing328:
@rostskadat I'll look into it and merge into master later.

wing328:
#1871 merged into master.

PR #1874, opened 2016-01-11
[Feign]Fix wrong query param key name in feign client

cbornet:
Query params get generated as
```java
@RequestLine("GET /myendpoint?myId={myId}")
```
instead of
```java
@RequestLine("GET /myendpoint?my_id={myId}")
```
when the query param name is "my_id"


wing328:
@cbornet thanks for the fix.

Issue #1848, opened 2016-01-08
Binary file encoding issues and Ruby


jqr:
I'm new to the swagger-codegen project, so apologies if this is the wrong venue or way to suggest changes.

I'm working on making a fix for downloading file data using a Ruby generated swagger client. My primary issue is when downloading a binary file, a file is made on the local filesystem with default encoding (UTF-8) and then binary data is attempted to be written into it, which triggers an attempted encoding transformation. At best, this transformation will fail and you'll get an immediate exception. At worst you'll get a silently corrupted file.

This is happening because the Rails framework [sets default encodings to `UTF-8`](https://github.com/rails/rails/blob/master/railties/lib/rails.rb#L19-L20) and that's used everywhere an encoding isn't specified, like in swagger client file write operations.

I believe the smallest change to fix this is to open the file in binary mode instead of the default. This is just adding a `b` to the flags passed to `File.open`. I provide a better fix below.

```ruby
File.open(path, 'wb') { |f| f.write(response.body) }
```

But the more I look at this code I wonder if we should be doing all this file manipulation in the first place. It's likely similar issue exists in some other language generators too. The Ruby client:

```ruby
# Save response body into a file in (the defined) temporary folder, using the filename
# from the "Content-Disposition" header if provided, otherwise a random filename.
#
# @see Configuration#temp_folder_path
# @return [File] the file downloaded
def download_file(response)
  tmp_file = Tempfile.new '', @config.temp_folder_path
  content_disposition = response.headers['Content-Disposition']
  if content_disposition
    filename = content_disposition[/filename=['"]?([^'"\s]+)['"]?/, 1]
    path = File.join File.dirname(tmp_file), filename
  else
    path = tmp_file.path
  end
  # close and delete temp file
  tmp_file.close!

  File.open(path, 'w') { |file| file.write(response.body) }
  @config.logger.info "File written to #{path}. Please move the file to a proper folder "\
                      "for further processing and delete the temp afterwards"
  File.new(path)
end
```

Summarized:

1. Make a `Tempfile` with a specific `temp_folder_path` solely to get it's path + filename.
2. If a filename was given use that with the path from above
3. Otherwise use `Tempfile` generated filename.
4. Close `Tempfile`
5. Open a new file with the name, write all data to it.
6. Close new file
7. Warn user about the things we're doing
8. Open file once more and leave open

What value does saving to a file in the generated client file provide? It saves the user this bit of Ruby in their application:

```ruby
[...]

PR #1876, opened 2016-01-12
File downloading improvements for Ruby and Java

xhh:
- Java okhttp-gson: sanitize the filename got from response header by removing path delimiters from downloading filename
- Ruby: IO improvements on file downloading
  - Use `File.basename` to sanitize the filename got from response header
  - Write to the `Tempfile` directly and return it to use less IO
  - Set file encoding according to the response body's encoding

See #1848

### DIFF
--- a/modules/swagger-codegen/src/main/resources/ruby/api_client.mustache
+++ b/modules/swagger-codegen/src/main/resources/ruby/api_client.mustache
@@ -170,23 +170,28 @@ module {{moduleName}}
     # from the "Content-Disposition" header if provided, otherwise a random filename.
     #
     # @see Configuration#temp_folder_path
-    # @return [File] the file downloaded
+    # @return [Tempfile] the file downloaded
     def download_file(response)
-      tmp_file = Tempfile.new '', @config.temp_folder_path
       content_disposition = response.headers['Content-Disposition']
       if content_disposition
         filename = content_disposition[/filename=['"]?([^'"\s]+)['"]?/, 1]
-        path = File.join File.dirname(tmp_file), filename
+        prefix = File.basename(filename)
       else
-        path = tmp_file.path
+        prefix = 'download-'
       end
-      # close and delete temp file
-      tmp_file.close!
+      prefix = prefix + '-' unless prefix.end_with?('-')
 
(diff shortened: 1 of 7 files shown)


modules/swagger-codegen/src/main/resources/ruby/api_client.mustache
```diff
@@ -170,23 +170,37 @@ module {{moduleName}}
...
+      tempfile
+    end
+
+    # Sanitize filename by removing path.
+    # e.g. ../../sun.gif becomes sun.gif
+    #
+    # @param [String] filename the filename to be sanitized
+    # @return [String] the sanitized filename
+    def sanitize_filename(filename)
+      filename.gsub /.*[\/\\]/, ''
```
jqr: Why not use `File.basename` here? Ruby regexes are multiline by default so this isn't completely fixing the issue.

modules/swagger-codegen/src/main/resources/ruby/api_client.mustache
```diff
@@ -170,23 +170,37 @@ module {{moduleName}}
...
-      File.open(path, 'w') { |file| file.write(response.body) }
-      @config.logger.info "File written to #{path}. Please move the file to a proper folder "\
-                          "for further processing and delete the temp afterwards"
-      File.new(path)
+      tempfile = nil
+      encoding = response.body.encoding
+      Tempfile.open(prefix, @config.temp_folder_path, encoding: encoding) do |file|
+        file.write(response.body)
+        tempfile = file
+      end
```
jqr: This will work in most cases but will cause random failures when GC happens here. How about we return an opened Tempfile with `close` / `unlink` instructions.

```ruby
```

jqr: Nevermind, I didn't see the assignment to `tempfile`. :+1:

modules/swagger-codegen/src/main/resources/ruby/api_client.mustache
wing328: @wing328 we used File.basename to begin with but it failed some test cases with Windows path.

PR #1864, opened 2016-01-10
[C#] add binary support

wing328:
- added support for binary body parameter and response to C# API client
- added 2 "fake" endpoints in petstore.json to test binary body parameter and response
- added more test cases

PR #1891, opened 2016-01-14
[PHP] addd binary support to response and body parameter

wing328:
- similar to https://github.com/swagger-api/swagger-codegen/pull/1864
- added test cases

PHPUnit 4.7.2 by Sebastian Bergmann and contributors.

....................

Time: 14.33 seconds, Memory: 18.00Mb

OK (20 tests, 3074 assertions)

### DIFF
--- a/modules/swagger-codegen/src/main/resources/php/ObjectSerializer.mustache
+++ b/modules/swagger-codegen/src/main/resources/php/ObjectSerializer.mustache
@@ -241,6 +241,8 @@ class ObjectSerializer
                 $values[] = $this->deserialize($value, $subClass);
             }
             $deserialized = $values;
+        } elseif ($class === 'ByteArray') { // byte array
+            $deserialized = unpack('C*', (string)$data);
         } elseif ($class === '\DateTime') {
             $deserialized = new \DateTime($data);
         } elseif (in_array($class, array({{&primitives}}))) {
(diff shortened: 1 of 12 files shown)


modules/swagger-codegen/src/main/resources/php/ObjectSerializer.mustache
```diff
@@ -241,6 +241,8 @@ class ObjectSerializer
                 $values[] = $this->deserialize($value, $subClass);
             }
             $deserialized = $values;
+        } elseif ($class === 'ByteArray') { // byte array
+            $deserialized = unpack('C*', (string)$data);
```
jqr: I realize the mapped type is a technically a byte array but this seems more natural to return as a string. Even printing this data is complicated as it's an array of integers and PHP's `pack` cannot be called with an `Array`.

```php
$string = "";
foreach($deserialized as $int) {
  $string = $string . pack("C", $int);
}
```

Compare with string returned:

```php
echo $deserialized;
```

File operations also don't support this unpacked format either.

wing328: Here is a 1-liner to restore the byte array: https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/php/SwaggerClient-php/tests/PetApiTest.php#L329

Agree that it's not straightforward but that's how PHP represents byte array.
Ref: http://stackoverflow.com/questions/885597/string-to-byte-array-in-php

PR #1895, opened 2016-01-15
add parentModel to CodegenModel

### LINKED ISSUE #1568: Access parent vars
It would be useful to have access to a parent's vars.
This is for when `supportsInheritance` is set to true in DefaultCodegen

This would allow subclasses to references a super classes properties for doing stuff like the following

``` swift
class Parent{
    var parentVar:String
    var parentVar2:Bool

    init(parentVar:String, parentVar2:Bool){
        self.parentVar = parentVar
        self.parentVar2 = parentVar2
    }
}

class Child:Parent {
    var childVar:String

    //is accessing the parents properties
    init(childVar:String, parentVar:String, parentVar:Bool) {
        self.childVar = childVar
        super.init(parentVar: parentVar, parentVar: parentVar2)
    }
}
```

A parents variables could perhaps be references like this {{#parent}}{{#vars}} ... {{/vars}}{{/parent}}

Without this it's not possible to build the example above which is what I'm trying to achieve in a Swift target

yonaskolb:
This fixes #1568, allowing you to access a parent's model, including it's properties

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenModel.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenModel.java
@@ -9,6 +9,7 @@ public class CodegenModel {
     public String name, classname, description, classVarName, modelJson, dataType;
     public String unescapedDescription;
     public String defaultValue;
+    public CodegenModel parentModel;
     public List<CodegenProperty> vars = new ArrayList<CodegenProperty>();
     public List<String> allowableValues;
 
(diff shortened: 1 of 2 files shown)


fehguy:
I like this functionality but we need to make sure it's safe.  For example, there are some post processing steps for creating models which may be missed from here.

Can you consider some tests for it?

yonaskolb:
Was already implemented https://github.com/swagger-api/swagger-codegen/commit/9511425019a4c160ebe20be7361c55e4c9dfec63

PR #1898, opened 2016-01-15
[Java okhttp-gson] Support "binary" (byte array) for body parameter and response

xhh:
Also updated Java default, jersey2 and okhttp-gson clients to support (constant) query string in the `path` (to work with the 2 "fake" petstore endpoints on binary).


Cloudmersive:
What do you need to do on the Swagger 2.0 API specification side to get this outcome?  Right now when I do type=file, I am getting a File object returned not a byte[]

PR #1899, opened 2016-01-15
Support java code generation for UUIDProperty

hiveship:
A property defined as "type: string & format: uuid" is now generated in
Java using "java.util.UUID" class instead of a classical String. In the
Java, the UUID class do not provide a constructor. You may use the
method UUID.fromString(<uuid_string>) to construct your UUID object.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -651,6 +651,8 @@ public class DefaultCodegen {
             datatype = "map";
         } else if (p instanceof DecimalProperty) {
             datatype = "number";
+        } else if ( p instanceof UUIDProperty) {
+            datatype = "UUID";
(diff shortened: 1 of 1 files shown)


fehguy:
Thank you--I need to confirm this does not break any of the other clients before merging

hiveship:
Ok. I admit that I have only checked Ruby, PHP, Java and JAX-RS generation.

Le vendredi 15 janvier 2016, Tony Tam <notifications@github.com> a écrit :

> Thank you--I need to confirm this does not break any of the other clients
> before merging
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/swagger-api/swagger-codegen/pull/1899#issuecomment-172039937>
> .
>


-- 

*Maëlig NANTEL*
Mobile : +33 (0)6 69 47 39 00
Mail : *maelig.nantel@gmail.com <maelig.nantel@gmail.com>*

hiveship:
@fehguy some news ? Did I need to change anything ?

Issue #1885, opened 2016-01-14
@DefaultValue not added to JavaJaxRS code


fooo:
When generating the Java JaxRS code, the in swagger defined defaultvalue is not mapped to a @DefaultValue specification in the API.

I think it should be possible to change _swagger-codegen/modules/swagger-codegen/src/main/resources/JavaJaxRS/jersey1_18/queryParams.mustache_ to include @DefaultValue like so:

```
{{#isQueryParam}}@ApiParam(value = "{{{description}}}"{{#required}},required=true{{/required}}{{#allowableValues}}, {{> allowableValues }}{{/allowableValues}}{{#defaultValue}}, defaultValue="{{{defaultValue}}}"{{/defaultValue}}) @DefaultValue("{{defaultValue}}") @QueryParam("{{baseName}}") {{{dataType}}} {{paramName}}{{/isQueryParam}}
```

wing328:
@fooo do you mind contributing a fix for the issue as I believe you've identified the right place to make the change?

Ref: guildelines for contributing https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

hiveship:
I submitted a PR https://github.com/swagger-api/swagger-codegen/pull/1902 to handle support the JAX-RS @DefaultValue annotation.

wing328:
PR merged

PR #1902, opened 2016-01-15
Correct issue 1885

hiveship:
Handle the Swagger "default" value for a parameter with the JAX-RS
@DefaultValue annotation. This correction will also be reported in the
new JAX-RS CXF Codegen.
https://github.com/swagger-api/swagger-codegen/issues/1885


fehguy:
Looks good to me!

PR #1922, opened 2016-01-19
[JavaScript] Make JavaScript client work in both Node.js and browser

### LINKED ISSUE #1736: [Javascript] header and form parameters are not used at all
Need to send header and form parameters to `$.ajax`

xhh:
* Replace jQuery with SuperAgent which works in both Node.js and browser
* Use UMD pattern (returnExports.js) to make the module exporting compatible
  with all major systems: AMD, Node.js (CommonJS) and browser
* Implement support of header and form parameters. Closes #1736
* Move HTTP requesting code to `ApiClient` and allow customizing options
  in it, e.g. "basePath"
* Update unit tests accordingly and add some tests for `ApiClient`

Also fixed occasional test failures for Java and Ruby clients.


wing328:

Starting from 2.1.6, JS API client will use [superagent](https://github.com/visionmedia/superagent)

PR #1934, opened 2016-01-21
[Ruby] Support "binary" format for body parameter and response

xhh:
The "binary" format is handled as String in Ruby client.


modules/swagger-codegen/src/main/resources/ruby/api.mustache
```diff
@@ -70,9 +70,10 @@ module {{moduleName}}
 
       # http body (model)
       {{^bodyParam}}post_body = nil
-      {{/bodyParam}}{{#bodyParam}}post_body = @api_client.object_to_http_body({{#required}}{{{paramName}}}{{/required}}{{^required}}opts[:'{{{paramName}}}']{{/required}})
-      {{/bodyParam}}
-
+      {{/bodyParam}}{{#bodyParam}}{{#isBinary}}post_body = {{#required}}{{{paramName}}}{{/required}}{{^required}}opts[:'{{{paramName}}}']{{/required}}
+      post_body = post_body.pack('C*') if post_body
```
jqr: Nearly the same comment from https://github.com/swagger-api/swagger-codegen/pull/1891/files#diff-630a0645a0b3fabb89412db600734cfbR245

A array of 8bit ints is a very non-standard data format in Ruby. As a user, I would prefer `post_body` here without processing.

wing328:
@jqr the PR has been updated to use string instead of array of integer to represent byte array.

Issue #1959, opened 2016-01-22
Java codegen doesn't support multiple security schemes


jasonmartens:
There is a comment in the DefaultGenerator.java:
```
// NOTE: Use only the first security requirement for now.
// See the "security" field of "Swagger Object":
//  https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#swagger-object
//  "there is a logical OR between the security requirements"
if (securities.size() > 1) {
     LOGGER.warn("More than 1 security requirements are found, using only the first one");
}
```
Which is incorrect as of the current version of the spec, which says there is a logical AND, not OR.

xhh:
Looking at the `security` field of [Operation Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#operation-object) the global [Swagger Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#swagger-object), I still see logic OR for the provided security requirements. For example for the Operation Object:

> A declaration of which security schemes are applied for this operation. The list of values describes alternative security schemes that can be used (that is, there is a logical OR between the security requirements). This definition overrides any declared top-level security. To remove a top-level security declaration, an empty array can be used.

jasonmartens:
It appears that there is a contradiction then, since this section:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#security-requirement-object
clearly says that it is an AND. 

It makes sense for the *global* security object to be OR. However, an individual path would need to be AND to support an api_key in addition to oauth for instance.

webron:
There's no contradiction, only a slight confusion.

`security` whether global or at the operation allow you to define a *list* of Security Requirements. Between the items in the list there's a logical OR allowing you to specify different variations.

However, the Security Requirement is an *object* of Security Definitions. Between these there's a logical AND.

wing328:
I suggest we update the spec with examples to show
- API key in header `OR` query
- API key `AND` Client ID in the header

cbornet:
IMO the generator should not set the security on operations but only give a hint on what securities can be used. Then it is the responsibility of the caller to chose which security he wants to use (since there is an OR). This is how it is done in retrofit and feign subgens.

wing328:
@cbornet I left a [comment](https://github.com/swagger-api/swagger-codegen/pull/1961#issuecomment-184597223) in the PR.

[...]

PR #1961, opened 2016-01-22
Allow multiple security requirements with java codegen

jasonmartens:
As reported in #1959, this PR allows multiple security requirements in the generated java client.


wing328:
Thanks for the PR. I've restarted the CI build.

Given that the PR updates the default code generator, which is used by generators other than the Java generator, we'll need more time to evaluate the impact.

wing328:
The CI tests failed with the following error:
```
Tests run: 291, Failures: 2, Errors: 0, Skipped: 0, Time elapsed: 6.621 sec <<< FAILURE! - in TestSuite
testSecurityWithGlobal(io.swagger.codegen.DefaultGeneratorTest)  Time elapsed: 0.195 sec  <<< FAILURE!
java.lang.AssertionError: expected [oauth2] but found [apiKey]
	at io.swagger.codegen.DefaultGeneratorTest.testSecurityWithGlobal(DefaultGeneratorTest.java:96)

testSecurityWithoutGlobal(io.swagger.codegen.DefaultGeneratorTest)  Time elapsed: 0.08 sec  <<< FAILURE!
java.lang.AssertionError: expected [oauth2] but found [apiKey]
	at io.swagger.codegen.DefaultGeneratorTest.testSecurityWithoutGlobal(DefaultGeneratorTest.java:61)


Results :

Failed tests: 
  DefaultGeneratorTest.testSecurityWithGlobal:96 expected [oauth2] but found [apiKey]
  DefaultGeneratorTest.testSecurityWithoutGlobal:61 expected [oauth2] but found [apiKey]

Tests run: 291, Failures: 2, Errors: 0, Skipped: 0
```
Please take a look.

You can run the CI test locally with `mvn verify -P samples`

jasonmartens:
Sorry, I messed up the commits for this. New branch is at:
https://github.com/jasonmartens/swagger-codegen/tree/multiple_security_fix

jasonmartens:
Anything I can do to help get this merged?

fehguy:
@wing328 take a peek when you can

wing328:
@jasonmartens I'll review shortly later today. I thought it's still being discussed in another thread.

I think we also need to prepare spec (with multiple authentication setting) to test the change.

modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
```diff
@@ -1196,9 +1196,21 @@ public CodegenOperation fromOperation(String path, String httpMethod, Operation
...
-            consumes = swagger.getConsumes();
-            LOGGER.debug("No consumes defined in operation. Using global consumes (" + swagger.getConsumes() + ") for " + op.operationId);
+            Boolean hasFormData = false;
+            for (Parameter param : operation.getParameters()) {
+                if (param.getIn().equals("formData")) {
+                    hasFormData = true;
+                }
+            }
+            if (hasFormData) {
+                consumes.add("multipart/form-data");
```
wing328: Technically it can also be `application/x-www-form-urlencoded`.

Instead of adding a default, should the spec owner fix it by adding a proper MIME instead?

wing328:
@jasonmartens Just to manage your expectation, I'll be out for a week so I'll pick this up on Feb 15th. Sorry for the delay.

jasonmartens:
No problem! Have a great vacation.

wing328:
UPDATE: I couldn't pull your work since it's shown as "unknown repository"

I "copied" your change and updated petstore.json with some auth settings for testing purpose as follows:
test_api_client_id AND test_api_client_secret
test_api_key_header OR test_api_key_query

Here's what the change looks like:
https://github.com/swagger-api/swagger-codegen/compare/master...wing328:multi_auth_setting?expand=1

I've updated the PHP sample code and as you can see there's no way to distinguish between AND, OR in the `{{authMethods}}` tag: https://github.com/swagger-api/swagger-codegen/compare/master...wing328:multi_auth_setting?expand=1#diff-286b57ad6967ac98f2f3a6c893dee84fR243

Do you need to do anything to support both AND, OR for auth settings (for Java Petstore API client as an example)?

jasonmartens:
Apologies again for the messed up PR branch, I don't do this a ton and kind of messed it up. 

In the java client, you get something like this:
`    String[] authNames = new String[] { "dkc_auth", "api_key" };
    return apiClient.buildCall(path, "GET", queryParams, postBody, headerParams, formParams, authNames, progressRequestListener);
`

the apiClient.buildCall will then add both of the authentication methods to the call. This is very clearly an AND case (it will apply all of the authentications listed in authNames). If it is a choice between query or header parameters, that is done within an individual auth method. 

I think you are correct that there is no way to represent OR auth methods on a specific endpoint. However, I don't need support for that, but I *definitely* need support for AND (apiKey and OAuth in this case).

wing328:
@jasonmartens I've merged #2155 (based on your work) into master. 

I wonder if you can pull the latest master and see whether you're facing the same issue with `AND` security setting in your case.

wing328:
Closing this as #2155 has merged into master.

Issue #1338, opened 2015-10-05
C#: Need a configuration setting to change the type for arrays


artem-dudarev:
Currently arrays are generated as `List<T>`. I'd like to configure the generator to get real arrays `T[]`.

wing328:
@artem-dudarev do you mind sharing more about the requirement to use `T[]` instead of `List<T>`? Is that a must or more of a personal preference?

Before I started working on the C# generator, it was already using List instead of [] so I just keep it that way (both are fine with me).

artem-dudarev:
This is not a must. I just think the class or method definition is shorter and more readable with arrays then with lists.
Thats why I ask for configuration option.

wing328:
@artem-dudarev thanks for the suggestion. We'll leverage the help from the community to implement this in the future.

jimschubert:
Arrays aren't a direct replacement for `List<T>`. Specifically because initializing an array requires the array size and resizing an array is not cheap. Lists are a lot more efficient in practice.

If anything, I think `IEnumerable<T>` as exposed types and `List<T>` as internal types would be preferable.

wing328:
@jimschubert thanks for the suggestion. May I know if you've cycle to contribute the enhancement?

jimschubert:
I could do this work, but it would be a breaking change. For example,

```
List<Pet> FindPetsByStatus (List<string> status = null);
```

would become

```
IEnumerable<Pet> FindPetsByStatus(IList<string> status = null);
```

This would break code previously generated against `List<Pet>` by:

* Removing the indexer (e.g. pets[1])
* Removing ability to modify the collection
* Requiring explicit type usage to be changed (e.g. `List<Pets> pets = client.FindPetsByStatus()`)

The benefit of using `IEnumerable<T>` would be that you're free to change any internal implementation without breaking consumers in the future.

The [Framework Guidelines](https://msdn.microsoft.com/en-us/library/dn169389(v=vs.110).aspx) specifically say to not return `List<T>` from a public API, which I'd consider generated code. The guidelines suggest using `Collection<T>`, although when implementing APIs I try to stick with `ICollection<T>`. They've somewhat hidden my `IEnumerable<T>` suggestion within that collection suggestion:

> In cases where you are sure that the only scenario you will ever want to support is forward-only iteration, you can simply use `IEnumerable<T>`.

I'd need to look at the code a little more closely to see how much effort it would be.

jimschubert:
[...]

PR #1964, opened 2016-01-23
[csharp] Return ICollection<T> instead of List<T>

jimschubert:
This change is inline with Microsoft's [recommended guidelines for
collections]
(https://msdn.microsoft.com/en-us/library/dn169389(v=vs.110).aspx).

This code is based on the discussion in swagger-api/swagger-codegen#1338

As mentioned in the issue, converting method signatures to return `ICollection<T>` rather than `List<T>` should be considered a **breaking change**.

This change also reduces the specific type of input parameters from `List<T>` to `ICollection<T>`, which is not a breaking change because `List<T>` implements `ICollection<T>`.


wing328:
@jimschubert thanks for the PR. Given that it's a breaking (non-backward compatible) change, I would suggest adding a CLI option (e.g. returnICollection) that defaults to false to begin with.

jimschubert:
@wing328 I've added two options: `useCollection` and `returnICollection`. I rebased my branch and merged master.

* **useCollection**: Deserialize responses into and return `Collection<T>`  
* **returnICollection**: For `List<T>` or `Collection<T>`, return `ICollection<T>`
  instead of the concrete type

wing328:
@jimschubert thanks for adding the option. I tested with useCollection only and got the following error:

screen shot 2016-01-25 at 10 12 49 am

But when I tested witih both useCollection and returnICollection set to true, I didn't get the error.

jimschubert:
@wing328 I don't know why that would happen. Commented on the line above which includes the import.

I noticed that any time I changed the resources, I had to run `mvn clean package` before generating the sample. I didn't get a chance to look into why.

wing328:
Forgot to mention that the error I got is in Pet.cs (model.mustache) and I don't see `System.Collections.ObjectModel` hardcoded in model.mustache.

jimschubert:
Ah. That's weird. I can update the model template tomorrow. I can also rebase.

wing328:
Thank you :beer:

jimschubert:
@wing328 I've added that using directive to model.mustache. Looked like my config was only testing both options. Sorry about that. Couldn't rebase all commits because of the merge from master, but everything on my branch is now up to date.

wing328:
@jimschubert no problem bro. I'll give it another try later and merge into master accordingly.

PR #1965, opened 2016-01-24
[TypeScript] add an option to determine property naming convention

wing328:
Add 4 different naming conventions for Typescript model's property
- original
- camelCase (default)
- PascalCase
- snake_case


fatbone008:
how to use these options? add the specificed type into CLI line? or in the yaml?

wing328:
Yes, something like "--modelPropertyNaming=original" or "modelPropertyNaming" in the config file (JSON). Please refer to project's README on how to customize the output.

abcox:
modelPropertyNaming: {'camelCase', 'PascalCase', 'snake_case' and 'original'}

Matster2:
This doesn't appear to be an option in the generateAPI config. Is it available here or is there some other way to do this.
All my query paramaters start with uppercase and I would like to make them camelcase instead

PR #1970, opened 2016-01-25
Allow maven-codegen-plugin to set additionalProperties even if cliOptions does not define them

evigeant:
The `maven-codegen-plugin` does not allow setting `additionalProperties` when they are not defined in `cliOptions`.

Several generators do not set the cliOptions properly, for example: html and dynamic-html. And in this case, the maven plugin cannot customize the behavior. The `codegen-cli` does not enforce this, so it can customize the behavior.


fehguy:
This solution seems backwards.  How about instead we fix the offending language files and throw a warning when trying to set an option which has not been configured?

evigeant:
Ok, in this PR, I will make the change so that a Warning is emitted when an
undefined option is set, but I will not prevent it as this is limiting for
all languages where cliOptions are not set properly. In another PR, I will
fix the languages which I need.

evigeant:
@fehguy I re-implemented following your suggestion. I think this is a better approach as this will allow generation with Maven to work even for languages which do not have all options defined in cliOptions.

I also added cliOptions for some languages in #1975.

wing328:
@evigeant If it's not too much work, I wonder if you can update indention to use 4-space instead of tab in your change as mentioned in https://google.github.io/styleguide/javaguide.html#s2.3.1-whitespace-characters

Ref: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md#style-guide

evigeant:
@wing328 I fixed the indentation, even some lines which were already there.

wing328:
@fehguy are you happy to merge this PR into master?

PR #1999, opened 2016-01-28
[Java Client] Add Eclipse and IntelliJ IDEA support in build.gradle.mustache

wizardbyron:
I added gradle IDE plugins in `build.gradle.mustache`.

I don't commit the sample changes. I think we should add sample directory in `.gitignore`. 

Let users try codegen and generate samples.


wing328:
The sample directories are used by Travis for CI tests so we can't ignore that.

For more info about the CI test, please run `mvn verify -P samples`.

PR #2011, opened 2016-01-31
Only define @DefaultValue if it's specified in the swagger spec

nickcmaynard:
Specifying that a query param has a default value of "" may be incorrect if the queryParam was completely unprovided.  @DefaultValue("") therefore has meaning that we should be careful to apply.

This pull request alters the query param specification to ensure that @DefaultValue is only specified if a defaultValue is specified in the swagger spec.


wing328:
@nickcmaynard thanks for the PR. There's only one other place (README.md) referencing JaxRSServerCodegen and I'll update that shortly.

nickcmaynard:
@wing328 Thanks!

PR #2019, opened 2016-02-02
Use retrofit2-beta3 and it's dendencies in retrofit2 clients

kungfoo:
Requires some changes to the generated client code
as some package renaming and api changes have occurred.


wing328:
Test result looks good for both `retrofit2` and `retrofit2rx`.

Issue #2015, opened 2016-02-02
Csharp client: Add option(s) to insert "partial" in class definitions 


bluemmc:
@wing328 C# partial classes are a great way to allow clients to customize auto-generated code safely. Happily, they are quite easy to enable - just "partial" keyword in the class definition. 

It would be very nice if  the csharp swagger-codegen client could add an option to enable generation of partial classes instead of the current "closed" classes. In particular for models.

I am not sure if one option should trigger partial for all classes, or there should be separate options for models, supporting files, api files etc. Personally, I just need for models in my particular use case.

wing328:
@bluemmc no concern from me. I remember I saw a PR before about making C# class (model) as partial but the PR didn't get in due to merge conflicts.

May I know if you've cycle to contribute the enhancement by adding an option to add `partial` to all generated classes (e.g. ApiClient, models, API classes, etc)?

bluemmc:
I added you because you looked like to main contributor for the csharp part and I was unable to provide a csharp label myself to issue. Unfortunately, I have yet to get permission from the company I work for to add contributions (working on it) so I can't do it myself at this time.

bluemmc:
BTW: Do you know what the PR is that did not get in due to merge conflicts ?

wing328:
@bluemmc no worry, how urgent do you need the change?

bluemmc:
Not urgently, but I believe I could make good use of the feature next month.

Thinking again about it. I think introducing "partial" always should not be a problem so it does not need to be an explicit option. It can always be on as it has no negative consequences for people not using the feature. 

So what about starting with just making a tiny change for the next release, by just inserting the string "partial" to the model.mustace file's class declaration. I.e. changing "public class {{classname}}" to "public partial class {{classname}}"

wing328:
I researched a bit about partial class in C# and I'll take your suggestion to hardcode it in the model mustache file to begin with.

The only drawback I found (according to some SO answers I read) is that the class definition can be found in multiple files causing it hard to find the full definition but this can be addressed if developers organise the files properly. 

I'll submit a PR shortly. Stay tuned.

wing328:
Submitted #2020 for the change.

wing328:
PR merged into master.

bluemmc:
Great news. Thanks.

PR #2020, opened 2016-02-02
[C#] add partial to C# model

wing328:
For #2015 

All 26 test cases passed without issues.


wing328:

Starting from 2.1.6, C# model will be `partial` so older version of .net/mono framework not supporting `partial` will fail to compile the code.

Issue #1469, opened 2015-10-29
Generate server side asp.net (c#) code?


bagusflyer:
I'm wondering if it's possible to generate server side asp.net code based on the swagger json or yaml file? Thanks.

wing328:
@bagusflyer currently we do have server generator for C# (WebApi). Do you have cycle to help implement it based on other server generator (e.g. JAX-RS, Sinatra, PHP Silex, etc) ?

NahumLitvin:
Unfortunatly I have no time to contribute nowadays but I am joining the request.

wing328:
The beta version of the C# WebApi generator can be as simple as the following to start with:

1) Community implements `getPetById` of [Petstore](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.json#L216) in a WebApi project
2) Swagger core team converts the files in the WebApi project back into mustache template. 

Then we'll implement other methods (POST, PUT, DELETE) one by one.

jimschubert:
Also commenting on this to join the request.

NahumLitvin:
I discovered https://github.com/domaindrivendev/Swashbuckle
it fits my needs.

wing328:
@NahumLitvin I believe Swashbuckle is for generating OpenAPI/Swagger spec and integration with other Swagger  components (e.g. swagger-ui).

The request here is to generate the server stub in C# Web Api given an OpenAPI/Swagger spec.

nin9creative:
So what exactly are you looking for as a start?  A complete ASP. net WebAPI project?  

I may be able to help with that.  I came here looking for the capability as we are moving to a design first approach, and highly leverage WebAPI at the company I'm working at.

wing328:
@nin9creative thanks for offering help. Yes, ideally a ASP.net WebApi that implements the following endpoints to start with:
- [GET] [getPetById](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.json#L303)
- [POST] [addPet](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.json#L66)
- [POST] [updatePetWithForm](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.json#L356)
[...]

PR #2024, opened 2016-02-03
[aspnet] Initial implementation of ASP.NET 5 server

jimschubert:
See Issue #1469

Things to work on:

Gaps:

* Missing swagger definition functionality:
  - defaultResponse
  - examples
  - auth
  - consumes
  - produces
  - nickname
  - externalDocs
  - imports
  - security
  - schema
* Resolve allParams/hasMore issue with headerParams
* Create functional test project stub
* Resolve all issues with value type return values

In this commit:

* Initial cross-platform ASP.NET 5 API server
* Hook up swagger gen via Swashbuckle and xml comment
* Build script (*nix) in project root
* Dockerfile for container deployment


wing328:
@jimschubert thank you :beers: 

I'll review and let you know if I've any questions.

jimschubert:
@wing328 Thanks. It generates a pretty complete server. Nothing snazzy like services or DI. Just pushed a second commit to fix a bad merge of the Haskell Servant example in README.md when I rebased.

I did have to add a hack for operation paths. Something like `/pet/{petId}?testing_byte_array=true` is an invalid identifier for [RouteAttribute](https://msdn.microsoft.com/en-us/library/system.web.http.routeattribute(v=vs.118).aspx). I just replaced the `?` with `/` so the user can build and modify as needed.

modules/swagger-codegen/src/main/resources/aspnet5/model.mustache
```diff
@@ -0,0 +1,134 @@
...
+using Newtonsoft.Json;
+
+{{#models}}
+{{#model}}
+namespace {{packageName}}.Models
+{
+    /// <summary>
+    /// {{description}}
+    /// </summary>
+    public class {{classname}} : {{#parent}}{{{parent}}}, {{/parent}} IEquatable<{{classname}}>
```
wing328: [OPTIONAL] we just added `partial` to the model class. You might want to do the same.

README.md
```diff
@@ -587,6 +588,15 @@ java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate \
   -o samples/server/petstore/haskell-servant
 ```
 
+### ASP.NET 5 Web API
+
+```
+java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate \
+  -i http://petstore.swagger.io/v2/swagger.json \
+  -l aspnet5-webapi \
```
wing328: Should this be `aspnet5` (not `aspnet5-webapi`)?

wing328:
[OPTIONAL] you may also want add `\bin\windows\aspnet5-petstore-server.bat` as there may be more developers using it on the Windows platform. (ref: https://github.com/swagger-api/swagger-codegen/blob/master/bin/windows/csharp-petstore.bat)

README.md
jimschubert: Yeah, thanks for catching this. I had to abort my first rebase in which I changed this, and missed it the second rebase.

jimschubert: I'll leave the header as `ASP.NET 5 Web API`. Technically, MVC and Web API are being merged and this is why I renamed from `aspnet5-webapi` to just `aspnet5`. I think the header will make it clear for now that there aren't any server-side views, and that the generator focuses only on the api.

modules/swagger-codegen/src/main/resources/aspnet5/model.mustache
jimschubert: I can add that. After we recently regenerated the C# client at work, I was thinking about how nice it would be if this was had the partial keyword.

I'll look at the client code, but are there any plans or best practices for partial methods? Or just the class for now?

wing328: To me, adding partial to model classes is just a starting point. If there're demands/use cases for adding partial to methods or other classes, we'll definitely consider it.

jimschubert:
@wing328 I made those changes. Let me know if there's anything else you find.

wing328:
Looks good to me.

PR #2039, opened 2016-02-04
FIX: Fixed FLash Client library

glederrey:
Hi,

I fixed the Flash client library. It had multiple problems.. =) I also added a README.txt that explains how to use it.. (It's a bit more difficult to use than Python..)

- FlashClientCodegen
   - I changed the architecture to make it look like an ActionScript library project.
   - The Number type in as wasn't recognized by Swagger. (You have Double, Float, Integer, etc.) And since it wasn't recognized, Codegen was putting null as default value for Number. It's wrong. Therefore, I changed it to NaN
- ApiInvoker.as
   - A lot of imports were missing. I added them.
   - Bug at line 102/115 that I fixed
   - I commented a part of the function onApiRequestResult because it was trying to parse the response as a XML. And therefore, we couldn't receive anything if it's in JSON, for example. Therefore, having it RAW is better, for the moment. =)
- ApiUrlHelper.as
   - Small bug at line 34.
- api.mustache
   - Here, I simply removed the DefaultValue. It's just because if we put a DefaultValue, then it means for AS that it's an optional parameter. So, if there is other parameters after one parameter with a DefaultValue, AS will trigger an error because it doesn't accept required parameters after optional ones. 
- README.txt
   - Just some explanations.. I hope it will make it more clear how to use it.. I took quite a lot of time to understand how to use it, but I'm not an expert in flash at all.. So, I hope it can help some futur users.. 

That's everything.. So, I don't really improved this library. But at least, this library generated by Swagger will compile directly and will be ready to be used..

PR #2049, opened 2016-02-05
Fix Flash client Library

glederrey:
It's the same commit as #2039.. 

Thank you @wing328 =)


glederrey:
Thanks again!!

wing328:
:beers:

PR #2053, opened 2016-02-06
Use prototypes for model methods

### LINKED ISSUE #2044: JS client: Model methods should be added to prototype
In the generated model classes, each instance contains the methods `constructFromObject`, `toJson`, plus getters and setters for all the fields. Currently, these methods are added as anonymous functions to each instance. Meaning, if I have 100 instances of some class, I have 100 copies of each of the methods.

These should be added to the class's prototype instead. For example, instead of (current version)

```javascript
var Person = function Person(name) {
  var self = this;
  self['name'] = name;
  self['age'] = age;
  self.constructFromObject = function(data) {
      if (!data) {
        return this;
      }
      self['name'] = ApiClient.convertToType(data['name'], 'String');
      return this;
  }
  self.getName = function() {
    return self['name'];
  }
  self.setName = function(name) {
    self['name'] = name;
  }
  self.toJson = function() {
    return JSON.stringify(self);
  }
}
```

it should generate (new version):

```javascript
var Person = function Person(name) {
  this['name'] = name;
}

Person.prototype.constructFromObject = function(data) {
  if (!data) {
    return this;
  }
  this['name'] = ApiClient.convertToType(data['name'], 'String');
  return this;
}

Person.prototype.getName = function() {
  return this['name'];
}
  
Person.prototype.setName = function(name) {
  this['name'] = name;
}

Person.prototype.toJson = function() {
  return JSON.stringify(this);
}
```

Two additional comments:
1. It would be nice to be able to omit the getters and setters entirely, as they cause a lot of bloat that is not desired in many cases.
2. The toJson prototype is the same for every model, and could be moved to a higher level prototype:

```javascript
var ApiModel = function(){}
ApiModel.prototype.toJson = function() { return JSON.stringify(this); }

var Person = function Person(name) {
  ApiModel.call(this);
  this['name'] = name;
}

// all the Person prototype function follow like before, except toJson
```

delenius:
Fixes #2044.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Javascript/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/Javascript/model.mustache
@@ -23,7 +23,6 @@
    * {{description}}
    **/{{/description}}
   var {{classname}} = function {{classname}}({{#mandatory}}{{this}}{{^-last}}, {{/-last}}{{/mandatory}}) { {{#parent}}/* extends {{{parent}}}*/{{/parent}}
-    var self = this;
     {{#vars}}
     /**{{#description}}
      * {{{description}}}{{/description}}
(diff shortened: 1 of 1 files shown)


wing328:
@delenius thanks for the PR. 

I notice that the commit is by "elenius", which means this won't count as your contribution in https://github.com/swagger-api/swagger-codegen/graphs/contributors.

Do you want to fix it by closing this one and submit a new one with the commit correctly linked to @delenius ?

delenius:
Good catch, I will do that.

delenius:
@wing328, I filed PR #2071 to update the petstore example.

PR #2058, opened 2016-02-07
Add model's "additionalProperties" handling in python client

### LINKED ISSUE #2037: additionalProperties issue in python code generation
Assume the following definition in a swagger.yaml file:

```yaml
definitions:
  definition1:
    type: object
    properties:
      prop1:
        type: string
      prop2:
        type: integer
      prop3:
        type: object
        additionalProperties:
          $ref: '#/definitions/definition2'
   
  defintion2:
    type: object
    properties:
       foo:
          type: string
       bar:
          type: string
```

I had issues with the `to_dict` method of `defenition1`,

solved by adding another test in the `to_dict` as follows:
```python
def to_dict(self):
        """
        Returns the model properties as a dict
        """
        result = {}
        for attr, _ in iteritems(self.swagger_types):
            value = getattr(self, attr)
            if isinstance(value, list):
                result[attr] = list(map(
                    lambda x: x.to_dict() if hasattr(x, "to_dict") else x,
                    value
                ))
           # *fix start*
            if isinstance(value, dict):
                result[attr] = dict(map(
                    lambda (k,v): (k, v.to_dict()), value.iteritems()))
           # *fix end*
            elif hasattr(value, "to_dict"):
                result[attr] = value.to_dict()
            else:
                result[attr] = value
```

Not sure how to modify the code generator to create this, or if there is any other way to solve that

mtngld:
Fix #2037 by adding dictionary handling in a model's `to_dict()` method

### DIFF
--- a/modules/swagger-codegen/src/main/resources/python/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/model.mustache
@@ -99,7 +99,7 @@ class {{classname}}(object):
                 result[attr] = dict(map(
                     lambda (k, v): (k, v.to_dict()) if hasattr(v, "to_dict") else (k, v),
                     value.iteritems()
-                    ))
+                ))
             elif hasattr(value, "to_dict"):
                 result[attr] = value.to_dict()
             else:
--- a/modules/swagger-codegen/src/main/resources/python/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/model.mustache
@@ -97,7 +97,7 @@ class {{classname}}(object):
                 ))
             elif isinstance(value, dict):
                 result[attr] = dict(map(
-                    lambda (k, v): (k, v.to_dict()) if hasattr(v, "to_dict") else (k, v),
+                    lambda k, v: (k, v.to_dict()) if hasattr(v, "to_dict") else (k, v),
                     value.iteritems()
                 ))
             elif hasattr(value, "to_dict"):
--- a/modules/swagger-codegen/src/main/resources/python/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/model.mustache
@@ -95,6 +95,11 @@ class {{classname}}(object):
                     lambda x: x.to_dict() if hasattr(x, "to_dict") else x,
                     value
                 ))
+            elif isinstance(value, dict):
+                result[attr] = dict(map(
+                    lambda (k, v): (k, v.to_dict()) if hasattr(v, "to_dict") else (k, v),
+                    value.iteritems()
+                    ))
             elif hasattr(value, "to_dict"):
                 result[attr] = value.to_dict()
             else:
@@ -114,14 +119,14 @@ class {{classname}}(object):
         """
         return self.to_str()
 
-    def __eq__(self, other): 
+    def __eq__(self, other):
         """
         Returns true if both objects are equal
         """
         return self.__dict__ == other.__dict__
 
     def __ne__(self, other):
-        """ 
+        """
         Returns true if both objects are not equal
         """
         return not self == other
--- a/samples/client/petstore/python/swagger_client/models/category.py
+++ b/samples/client/petstore/python/swagger_client/models/category.py
@@ -108,7 +108,7 @@ class Category(object):
                 ))
             elif isinstance(value, dict):
                 result[attr] = dict(map(
-                    lambda (k, v): (k, v.to_dict()) if hasattr(v, "to_dict") else (k, v),
+                    lambda k, v: (k, v.to_dict()) if hasattr(v, "to_dict") else (k, v),
                     value.iteritems()
                 ))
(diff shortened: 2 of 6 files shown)


wing328:
@mtngld thanks for the PR. I notice the commits are not linked to your account, which means this PR won't count as your contribution towards https://github.com/swagger-api/swagger-codegen/graphs/contributors. If you're ok with that, we'll review and merge the PR later.

wing328:
Also not sure how hard it's but it would be nice to add a test case in the Python Petstore sample to cover the bug moving forward.

Ref: https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/python/tests/test_pet_model.py

mtngld:
Hi, hopefully I fixed my account linking issue,

I have no problem adding a testcase, but currently [`petstore.json`](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.json) does not have a model defintion with `additionalProperties`, I can add one, but that might affect more modules...

wing328:
You're right. I'll add a fake model later and would need your help to add a test case if you don't mind.

I think you need to close this PR and open a new one. You can then check the "Commits" tab to confirm.

wing328:
I can see that it's fixed in the commit tab. will review and approve accordingly.

wing328:
@mtngld is the change compatible with python3.x?

I ran the integration test as follows but got errors:
```
cd samples/client/petstore/python
virtualenv venv
source ./venv/bin/activate
./test-all.sh
```
Let me know if you can't repeat the issue locally.

wing328:
and here is part of the error message if it helps:
```
  File "/private/var/tmp/pr/mtngld/swagger-codegen/samples/client/petstore/python/.tox/py34/lib/python3.4/imp.py", line 171, in load_source
    module = methods.load()
  File "<frozen importlib._bootstrap>", line 1220, in load
  File "<frozen importlib._bootstrap>", line 1200, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1129, in _exec
  File "<frozen importlib._bootstrap>", line 1471, in exec_module
  File "<frozen importlib._bootstrap>", line 321, in _call_with_frames_removed
  File "/private/var/tmp/pr/mtngld/swagger-codegen/samples/client/petstore/python/tests/test_store_api.py", line 14, in <module>
    import swagger_client
  File "/private/var/tmp/pr/mtngld/swagger-codegen/samples/client/petstore/python/swagger_client/__init__.py", line 4, in <module>
    from .models.user import User
  File "/private/var/tmp/pr/mtngld/swagger-codegen/samples/client/petstore/python/swagger_client/models/__init__.py", line 4, in <module>
    from .user import User
  File "/private/var/tmp/pr/mtngld/swagger-codegen/samples/client/petstore/python/swagger_client/models/user.py", line 261
    lambda (k, v): (k, v.to_dict()) if hasattr(v, "to_dict") else (k, v),
           ^
SyntaxError: invalid syntax
```

mtngld:
Sorry about that, only tested on 2.7 

Now fixed a small compatibility issue,  `test_all.sh` passed using virtualenv on my machine

wing328:
@mtngld thanks and PR merged.

I saw 
```
test_500_error (tests.test_api_exception.ApiExceptionTests) ... /private/var/tmp/pr/mtngld/swagger-codegen/samples/client/petstore/python/tests/test_api_exception.py:60: DeprecationWarning: Please use assertRaisesRegex instead.
  with self.assertRaisesRegexp(ApiException, "Internal Server Error"):
/private/var/tmp/pr/mtngld/swagger-codegen/samples/client/petstore/python/tests/test_api_exception.py:76: DeprecationWarning: Please use assertRegex instead.
```

I will create a task to track it and see if anyone from the community has cycle to update the code with assertRegex instead.

wing328:
@mtngld btw, one suggestion is to create a new branch instead fixing the issue in the master (I believe creating a new branch for change is part of the Git best practices)

mtngld:
Hi, @wing328, sorry for all the bother I caused, but I found a bug in my PR, I made a new [branch](https://github.com/mtngld/swagger-codegen/tree/issue_2037) that:

1. Fix the bug I found (errrr python2.7 and python3.4 issues...)
2. Performance optimization by switching the order of the `elif`, since I guess `additionalProperties` is probably less used and should be checked last
3. Style optimization by warping lines and staying below 79 chars
4. Change `value.iteritems()` to `value.items()` 

Should I create a new PR? 

The new branch is based on the current master (which includes my previous PR), let me know if you want to simply revert the old PR and I will rebase to a clean and updated master.

Sorry again and thank you for your patience and guidance

We should **definetly** write test case for this scenario since that bug was not caught by `test-all.sh` but was found when using my own swagger file.

wing328:
Please submit a PR for the fix.

William

PR #2062, opened 2016-02-07
[csharp][aspnet5] AbstractCSharpCodegen base

jimschubert:
This moves a bit of logic into a base type to prevent lots of duplicates.

I've only converted csharp and aspnet5 generators, not csharp 2.0.

This base class should provide enough flexibility to support other C# based client and server generators.

Please let me know if you think I should also convert the csharp 2.0 generator. My reason for leaving it as-is was that few people are probably using C# 2.0 anymore and bugs in that generator could have a negative effect in C# 3.0 and higher. I admit my reasoning could be off.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java
@@ -1,0 +1,509 @@
+package io.swagger.codegen.languages;
+
+import io.swagger.codegen.*;
+import io.swagger.models.properties.*;
+import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.util.*;
+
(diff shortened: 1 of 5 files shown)


wing328:
@jimschubert I think `src/Properties/AssemblyInfo.cs` was left in the sample directory after the AssemblyInfo.cs path has changed (as csharp-petstore.sh does not delete files that are no longer used). Please kindly remove it. (and thanks for bringing it up for discussion)

jimschubert:
@wing328 reverted the AssemblyInfo path change, rebased and squashed before I saw your comment about deleting `src/Properties/AssemblyInfo.cs`. I didn't realize GitHub no longer shows comments on squashed.

I had commented that the AssemblyInfo at that location was causing me issues with compiling while the new AssemblyInfo (at the correct path) existed. I just ran the generator again with the correct path and generated csproj and I can't reproduce whatever issue I was having.

I can remove `src/Properties/AssemblyInfo.cs` if you'd like, but I don't think it's actually causing a problem.

wing328:
@jimschubert no worry. I'll take care of it.

I reviewed and tested the change, which works fine as expected. Appreciate your effort to make the C#-related generators better.

PR #2072, opened 2016-02-08
Add usePromise parameter to JS client

### LINKED ISSUE #2014: JS code gen: Using Promises and non-jquery ajax
This is a couple of feature requests for the JavaScript client generator (not the runtime generator, I am generating form a swagger file offline):

1) Is there any way to make the generator produce a JS client  produce a client that uses Promises? It looks like the online generator can do this, but not the offline one, but I could be wrong.

2) Is there any chance we could get rid of jQuery to do the ajax calls? It is unfortunate to have to pull in this heavy dependency just for that, when there are many lightweight ajax libraries out there. Ideally it should work both in a Web client and from Node, so how about something like isomorphic-fetch? This also uses Promises, so it helps with #1 above.

delenius:
This adds a "usePromise" boolean parameter to the JS client generator (disabled by default).

Fixes #2014.
For some discussion on the benefits of using Promises, see http://bluebirdjs.com/docs/why-promises.html

Note: I have not added any unit tests for this feature, because it produces a different client API, so the existing PetStore tests will not work. I could add:

samples/client/javascript/test/api/PetApiPromiseTest.js
samples/client/javascript/test/api/ApiClientPromiseTest.js

and furthermore, modify bin/javascript-petstore.sh to produce a promise version of the test client.

Let me know if you want me to do that, or create tests in any other way.

In the meantime, consider this addition experimental. It does not impact existing functionality since it has to be explicitly enabled using `--additional-properties usePromises=true`


wing328:
@delenius thanks for adding the support for Promises.

Instead of modifying bin/javascript-petstore.sh, I would suggest adding `bin/javascript-promise-petstore.sh` and output the JS client with `promise` in samples under `samples/client/petstore/javascript-promise`. Then you can add test cases in that directory.

If it's not too much work, I would suggest using CLI option (instead of `--additional-properties`) for `usePromises`

delenius:
@wing328, can you point me to how to add a CLI option instead of `--additional-properties`, for a language-specific option? I don't want to pollute the generic set of options.

delenius:
Tests added as discussed above, though some tests are not implemented:
1. Can not check values in the `request` object generated inside `ApiClient`, because it is not returned (a `Promise` is returned instead).
2. Can not check response headers, because they are not returned in the `Promise` version -- only the `data` is returned.

wing328:
would this help #1851 ?

delenius:
That looks like what I did, except that there is no `io/swagger/codegen/options` dir, where that `OptionProvider` goes. How is the parameter in that commit passed on the command line? Maybe the same style already works in my PR.

wing328:
Sorry I missed the lines about adding USE_PROMISES to CLIOption.

wing328:
Btw, I saw commits from others in the commit tabs. Is that due to `git rebase master`?

delenius:
Oh, yeah I was confused by that too. Yes, probably. I was trying to integrate the latest master branch into my feature branch...

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
```diff
@@ -96,6 +82,8 @@ public JavascriptClientCodegen() {
                 "version of the project (Default: using info.version or \"1.0.0\")"));
         cliOptions.add(new CliOption(PROJECT_LICENSE_NAME,
                 "name of the license the project uses (Default: using info.license.name)"));
+        cliOptions.add(new CliOption(USE_PROMISES,
+                "use Promises as return values from the client API, instead of superagent callbacks (Default: false)"));
```
wing328: Instead of putting the default value in the description, you can use `.defaultValue` instead. Here is an example: https://github.com/swagger-api/swagger-codegen/blob/ee8547f3b4e6260c6b9ae3dbfd94652109adcb3e/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PerlClientCodegen.java#L79

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
```diff
@@ -1,35 +1,19 @@
...
-import io.swagger.codegen.CliOption;
-import io.swagger.codegen.CodegenConfig;
-import io.swagger.codegen.CodegenConstants;
-import io.swagger.codegen.CodegenModel;
-import io.swagger.codegen.CodegenOperation;
-import io.swagger.codegen.CodegenProperty;
-import io.swagger.codegen.CodegenType;
-import io.swagger.codegen.DefaultCodegen;
-import io.swagger.codegen.SupportingFile;
+import io.swagger.codegen.*;
```
wing328: We prefer not using wildcard import: https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

Ref: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md#style-guide

wing328:
FYI. I usually rebase my branch (e.g. fix_something) as follows:
```
git checkout master
git pull upstream master 
git checkout fix_something
git rebase master
```
Assuming `upstream` is a remote defined as https://github.com/swagger-api/swagger-codegen.git

delenius:
@wing328, yeah that's how I did my rebase too, except I think I did a git push (to `origin`, i.e. my fork), after `git pull`, which seems to have complicated things.

I fixed the `defaultValue` issue. Note that there are other default values in the same file that use the "manual" style. I didn't want to touch those though.

wing328:
@delenius thanks for the update. We can clean those up later.

I also left you a comment about wildcard import.

delenius:
I for sure did not change those imports (it's because other people's commits are listed with my PR for some reason). I agree on not using wildcards, and I typically use IDE auto-imports which does not generate wildcards. But I can try to fix them anyway.

wing328:
Yes, you're right. @fehguy did that. No worry then.

I'll review your PR and approve shortly.

delenius:
No worries, I already fixed the imports.

wing328:
:+1:

delenius:
@wing328, regarding the other people's commits showing up: Is it because I messed up my rebase somehow? Or does that always happen after a rebase? I'd like to not repeat the same mistake in the future.

wing328:
It never happens to me when I perform a rebase. I've seen others doing that due to a bad rebase.

wing328:
PR merged. Thanks again for the contribution :beers:

Issue #2041, opened 2016-02-04
JS client does not support polymorphic return types


delenius:
In the generated `constructFromObject` functions, the `descriminator` property is ignored, which means that an instance of the abstract superclass is created. For example, if I have this in my swagger.yaml:

```yaml
Vehicle:
  type: "object"
  required:
  - "vehicleType"
  discriminator: "vehicleType"
  properties:
    vehicleType:
      type: "string"

Car:
  allOf:
  - $ref: "#/definitions/Vehicle"
  - type: "object"
    required:
    - "vehicleType"
    properties:
      vehicleType:
        type: "string"

Bicycle:
  allOf:
  - $ref: "#/definitions/Vehicle"
  - type: "object"
    required:
    - "vehicleType"
    properties:
      vehicleType:
        type: "string"
```

then, according to my understanding of how it should work, if the `vehicleType` is `Car`, then a new `Car` instance should be generated, and if `vehicleType` is `Bicycle`, a new `Bicycle` instance should be generated.

Currently, all I get is a new `Vehicle` instance.

delenius:
A workaround would be to have a mode where you don't generate these "models" at all -- just return the raw data. That would also make the client much smaller. With my rather large spec, I get a rather bloated client, and I don't really have any use for all these "model" classes...

demonfiddler:
Agree regarding code-bloat from model classes if these don't actually contribute any useful functionality. Model-driven content assist (auto-complete) is useful functionality but it doesn't necessarily have to be driven from JavaScript - [Tern.js](http://ternjs.net/), for example, has a 'condense' utility whose output is a JSON API definition from which the Tern server can generate completion proposals.

However, from a pure modelling perspective, potential functionality in the model classes could include dynamic enforcement of modelled:
- containment semantics (i.e., containment is intrinsic to JSON but unknown to JavaScript)
- cross-reference semantics (i.e., cross-references are unknown to JSON but intrinsic to JavaScript)
- min, max, length, enum and other constraints
- runtime type safety (i.e., set property accessor rejects invalid parameter)
- multiplicity bound constraints
- uniqueness within multi-valued (array) properties
- round-trip JSON serialisation and deserialisation

Some (but not all) of the above could alternatively be implemented in generated validator classes. But overall I agree that it would be a good idea to make model generation optional.

skydiablo:
hmmm... any fixes ?

triptec:
How would we like to solve this?

[...]

PR #2075, opened 2016-02-08
Move common functionality to new model base class

delenius:
Fixes "additional comment (2)" in #2044.

The `toJson` method that currently exists (identically) in every model class is moved to a common base class, `ApiModel`, that is extended by all model classes (using prototype inheritance).

The new `ApiModel` base class may also be useful for other purposes (i.e. other common functionality among the model classes) in the future.

Note: This also includes the changes to the Petstore example files from https://github.com/swagger-api/swagger-codegen/pull/2071 because it would be impossible not to (since I had to regenerate the example files using the latest code).

A couple of debatable points:
- The name of the common base model class could be changed. It is currently `ApiModel`. Perhaps you prefer `BaseModel`, `ModelBase`, or just `Model`?
- The base model class could be put in the `model` dir. Currently it goes where `ApiClient.js` goes.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
@@ -231,6 +231,7 @@ public class JavascriptClientCodegen extends DefaultCodegen implements CodegenCo
         supportingFiles.add(new SupportingFile("package.mustache", "", "package.json"));
         supportingFiles.add(new SupportingFile("index.mustache", sourceFolder, "index.js"));
         supportingFiles.add(new SupportingFile("ApiClient.mustache", sourceFolder, "ApiClient.js"));
+        supportingFiles.add(new SupportingFile("ApiModel.mustache", sourceFolder, "ApiModel.js"));
     }
 
     @Override
--- a/modules/swagger-codegen/src/main/resources/Javascript/ApiModel.mustache
+++ b/modules/swagger-codegen/src/main/resources/Javascript/ApiModel.mustache
@@ -1,0 +1,33 @@
+(function(root, factory) {
+  if (typeof define === 'function' && define.amd) {
+    // AMD. Register as an anonymous module.
+    define([], factory);
+  } else if (typeof module === 'object' && module.exports) {
+    // CommonJS-like environments that support module.exports, like Node.
+    module.exports = factory();
+  } else {
+    // Browser globals (root is window)
+    if (!root.{{moduleName}}) {
+      root.{{moduleName}} = {};
+    }
+    root.{{moduleName}}.ApiModel = factory();
+  }
+}(this, function() {
+	'use strict';
+
+	/**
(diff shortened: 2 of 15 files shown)


wing328:
We've similar model implementation (refactor model functions into a base class) for Ruby but we later remove that.

The reason is that we want to avoid extra dependency on the model as there are use cases just use the models (not API class) only.

delenius:
I don't understand. This does not depend on an "API class" (though I called the base class, perhaps poorly, `ApiModel`). I put the base class in its own file. And it can even be moved into the `models` dir.

wing328:
Can the model class be used without `ApiModel`? (the fact that I mentioned about API class might have confused you)

In other words, we want to make the model class "self-contained" without dependency on another base class.

delenius:
No, the model classes are not self-contained in that sense.

However, the model classes can already depend on other model classes when you have fields of some user-defined type.

Another potential need for inheritance in the model classes is to support polymorphism properly (see https://github.com/swagger-api/swagger-codegen/issues/2041).

In short, there is nothing wrong with having model classes depend on other model classes, and in fact you already do this.

wing328:
There's definitely nothing wrong and as I said we did it before (with the exact reason you mentioned in this PR). I just wanted to share the reason with you why we rolled back to the original approach..

delenius:
@wing328 since you don't want to do the refactoring above, how do you feel about simply removing the `toJson` method? It is not used anywhere, and it is pretty pointless, since it is always the same one-liner,

```javascript
  {{classname}}.prototype.toJson = function() {
    return JSON.stringify(this);
  }
``` 

It seems more straightforward to just using `JSON.stringify` directly.

wing328:
@delenius once again thanks for this PR. I was really happy when I saw it as you took a huge step further to bring the JS API client to the next level.

I agree that the one-liner (using JSON.stringify) should be pretty straightforward to most JS developers and one should also easily find the solution by doing a google search ([example](http://stackoverflow.com/questions/558518/how-can-i-serialize-an-object-to-json-in-javascript)).

One question I always consider is can a 10-year-old who just learns JS programming for 3 months use the JS API client generated by swagger-codegen? would the `toJson` make his/her life easier in converting the JS models to JSON string? My view is to keep it to make a developer's life easier (which should be the primary goal is an SDK/library).

Of course, another view we also need to consider is whether the `toJson` function would make the JS API client looks less professional.

I've shared my view and if you want to remove `toJson`, please go ahead to file a PR and I'll merge.

delenius:
I'll hold off for a little while to see if we get any tie-breaker opinion on this :)

demonfiddler:
I'm at a loss to understand what the toJson() method is intended to accomplish. The ECMA 5.1 / JavaScript specifications define a toJSON() method that, if present, is called by JSON.stringify(). It strikes me that adding another method with different capitalisation will lead to confusion

Surely standard serialisation via JSON.stringify(model) is perfectly adequate? My vote is to **remove the toJson() method from the template** and in fact my changes for #2258 do exactly that (although I omitted to write that bit up).

PR #2076, opened 2016-02-08
Fix #1487, TypeScript-Angular output path wrong if apiPackage not the…

### LINKED ISSUE #1487: TypeScript and Packages
If I use a config file for TypeScript-Angular and I set modelPackage and apiPackage to different values, I get an error. The generated api File does no longer find the models. Maybe we should use absolut Types? Eg, `services.Bla` instead of just `Bla`?

daveholladay:
The Typescript-Angular codegen added the api.d.mustache files in the constructor & picked up the output patch from the default value for apiPackage (API.Client).

Moving the add to the end of processOpts picks up the overridden value for apiPackage & outputs the file where it should be.

I've made the same change in Typescript-Node to avoid the same bug if the output is changed to fully qualified paths at a later date.


wing328:
@daveholladay thanks for the PR and sorry for taking a bit long to review.

PR #2078, opened 2016-02-09
Add option to omit getters/setters on models

delenius:
Adds an option to the JS client generator, `omitModelMethods`, which causes it to not generate getters and setters for the model classes.

These methods are often not needed or desired, especially when the returned objects are to be further processed and turned into other structures anyway (e.g. [Immutable.js](https://facebook.github.io/immutable-js/) records).

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
@@ -44,6 +44,7 @@ public class JavascriptClientCodegen extends DefaultCodegen implements CodegenCo
     private static final String PROJECT_VERSION = "projectVersion";
     private static final String PROJECT_LICENSE_NAME = "projectLicenseName";
     private static final String USE_PROMISES = "usePromises";
+    private static final String OMIT_MODEL_METHODS = "omitModelMethods";
 
     protected String projectName;
     protected String moduleName;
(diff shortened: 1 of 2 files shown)


wing328:
One way we can leverage the setter method is to support enum values as enum is not natively support in JS.

delenius:
@wing328, how do you mean?

wing328:
Here is an example for Python: https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/python/swagger_client/models/pet.py#L194)

This is a "workaround". For JS, there's no private variable in model classes so developers can bypass the setter.

delenius:
Oh I see, for input validation in the setters. I didn't consider that since the JS generator doesn't actually do that ;) 

Still, my PR adds an *option* that does not break anything...

wing328:
@delenius yes, I'm aware you're adding an option so it's definitely backward-compatible.

Based on your experience, how do JS developers implement enum? It's not a must that we support enum if it's not something supported by JS natively but it would be nice if the JS models can leverage the information defined in enum in the spec to validate the value.

My suggestion is definitely not ideal as you've pointed out that JS developers do not really use setter on models but worth sharing with you how potentially a setter method can be used to "partially" support enum.

delenius:
Well, my sense is that JS developers are used to living without these kinds of safety belts, and tend to use objects in a more loose, free-wheeling way :)

wing328:
@delenius ok. we'll reconsider enum support in JS if there're more demands from JS developers.

demonfiddler:
Why do we need getXxx() and setXxx() **methods** anyway? It's not a JavaScript idiom and in any case it's superfluous since ECMAScript 5.1 allows for get, set property accessor functions so that client code can use direct property reference rvalues and lvalues respectively.

wing328:
@delenius please rebase on the latest master to resolve the merge conflicts.

delenius:
@wing328 - I've fixed it.

wing328:
@delenius thanks for the quick turnaround. PR merged into master

PR #2082, opened 2016-02-09
Add VendorExtensions to root swagger

jhitchcock:


wing328:
@jhitchcock thanks for the PR. Do you think you've cycle to add a simple test case for the root level vendor extensions?

jhitchcock:
Will do

jhitchcock:
Hm actually the global vendorExtensions only appear in the final model sent to moustache, that is only created in the generate() method. 

I can't figure out how to access that via a test as before generate() I don't have it, and after generate() I only have the generator output. 

Any ideas?

fehguy:
@jhitchcock there is a `void preprocessSwagger(Swagger swagger)` function you can override and do any mutation of the swagger definition then.

wing328:
I'll add the test later when I've time. PR merged.

PR #2086, opened 2016-02-09
change the host name as a global

mesteve:
Hi guys,

I just made a modification on the "basePath" under the android folder in order to be able to set it at the beginning of your initialization so you just have to set it once.

Thanks,
mesteve

### DIFF
--- a/modules/swagger-codegen/src/main/resources/android/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/android/api.mustache
@@ -20,7 +20,6 @@ import java.io.File;
 
 {{#operations}}
 public class {{classname}} {
-  String basePath = "{{basePath}}";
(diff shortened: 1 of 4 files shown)


wing328:
@mesteve thanks for the PR. In the commit tab, it shows that your commit is not linked to your github account, which means this PR won't count as your contribution in https://github.com/swagger-api/swagger-codegen/graphs/contributors. You may want to fix that.

I think your PR will make your life easier to set all API class using the same `basePath` while there may be other use cases in which API classes use different `basePath`.

cc @Shyri

mesteve:
I see my account in the PR tab.

wing328:
It's not linked to your github account. You can compare other PRs.

wing328:
@mesteve thanks for the PR but we'll keep the current approach so that the API client can talk to endpoints with different base path.

PR #2092, opened 2016-02-10
ruby: Fix boolean convertion on json to model attribute.

sugi:
Current code will converts boolean true of json to false
because ruby's boolean is not matched with any regex.

Here is single line patch to convert string explicitly before regex check.

```
sugi@tempest:~% irb
irb(main):001:0> true =~ /true/
nil
irb(main):002:0> "true" =~ /true/
0
irb(main):003:0> 
sugi@tempest:~% ruby1.8 -e 'p true =~ /true/'
false
sugi@tempest:~% ruby1.9.1 -e 'p true =~ /true/' 
nil
sugi@tempest:~% ruby1.9.3 -e 'p true =~ /true/'
nil
sugi@tempest:~% ruby2.0 -e 'p true =~ /true/'
nil
sugi@tempest:~% ruby2.1 -e 'p true =~ /true/' 
nil
sugi@tempest:~% ruby2.2 -e 'p true =~ /true/' 
nil
```

### DIFF
--- a/modules/swagger-codegen/src/main/resources/ruby/base_object.mustache
+++ b/modules/swagger-codegen/src/main/resources/ruby/base_object.mustache
@@ -31,7 +31,7 @@
       when :Float
         value.to_f
       when :BOOLEAN
-        if value =~ /^(true|t|yes|y|1)$/i
+        if value.to_s =~ /^(true|t|yes|y|1)$/i
           true
         else
           false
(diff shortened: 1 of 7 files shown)


wing328:
@sugi thanks for the PR and tested with 6 different versions of Ruby to ensure the fix works across different versions of Ruby :beers: 

Do you mind updating the sample code for Ruby Petstore as mention in the contributing guidelines?

 Ref: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

sugi:
Oh, thank you for your comment.

I'll check the document and update this PR in your manner.
Please wait a moment.

sugi:
OK. 
I updated petstore sample and added tests.

I believe that it satisfy your contribute guidelines.
Please notice me if not so.

wing328:
@sugi thanks. Looks good to me and you also updated the test cases as well.

PR #2098, opened 2016-02-10
Fixes typo in template folder for javascript in bin-script

pamuche:
The template folder for javascript starts wit a capital letter. Shell script will fail on linux because of java.io.FileNotFoundException for the module.mustache


wing328:
@pamuche thanks for the PR. You're right and I could repeat the issue in Linux (but not Mac).

PR #2119, opened 2016-02-12
build.gradle missing 2 libs (pom.xml has)

vorburger:
jackson-jaxrs-json-provider & com.brsanthu:migbase64 are missing from the build.gradle mustache template (but the pom.xml one has them)

PS: Why are there no automated tests that detect this?

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/build.gradle.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/build.gradle.mustache
@@ -108,7 +108,9 @@ dependencies {
     compile "com.fasterxml.jackson.core:jackson-core:$jackson_version"
     compile "com.fasterxml.jackson.core:jackson-annotations:$jackson_version"
     compile "com.fasterxml.jackson.core:jackson-databind:$jackson_version"
+    compile "com.fasterxml.jackson.jaxrs:jackson-jaxrs-json-provider:$jackson_version"
     compile "com.fasterxml.jackson.datatype:jackson-datatype-joda:2.1.5"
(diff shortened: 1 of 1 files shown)


vorburger:
FYI @xhh this appears to be an impact of / was forgotten in you https://github.com/swagger-api/swagger-codegen/commit/4f831e032cfb90e37362618c62639a642b550e85

wing328:
Currently, we're using maven for CI tests. I agree with you that ideally any issue with the gradle file should have been uncovered by the CI tests.

Any suggestion to improve the current CI tests so as to catch similar issue moving forward is greatly appreciated.

wing328:
@vorburger I tested it and it works as expected. Thanks for the fix. I wonder if you can do a little bit more:

1) apply the same fix to https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/build.gradle.mustache, which is the build.gradle file for Jersey2 library
2) update Petstore sample by running
```
./bin/java-petstore.sh
```
Ref: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md#testing

For updating the current CI tests to cover the issue moving forward, I've some ideas and will test those after this PR gets merged into master.

wing328:
FYI. I've updated pom.xml for Java API client (okhttp-gson library) to include a test for `gradle check` so that we can catch similar issue with build.gradle moving forward.

https://github.com/wing328/swagger-codegen/commit/516a9413a28e4ac12a3acd03e65d79a35237e5de

wing328:
@vorburger we'll work on item 1 and 2.

PR #2122, opened 2016-02-12
Closes #2118

### LINKED ISSUE #2118: JMeter API ignores basePath in service URL
Spec: https://gist.github.com/foragerr/1daf27b05abab842d197
codegen version: 2.1.5
Language: JMeter

As the title says, basePath should be added to the URL, but is completely dropped by JMeter codegen.

foragerr:
Add basePath without url to jmeter sampler path

### DIFF
--- a/modules/swagger-codegen/src/main/resources/JMeter/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/JMeter/api.mustache
@@ -103,7 +103,7 @@
           <stringProp name="HTTPSampler.response_timeout"></stringProp>
           <stringProp name="HTTPSampler.protocol"></stringProp>
           <stringProp name="HTTPSampler.contentEncoding"></stringProp>
-          <stringProp name="HTTPSampler.path">{{vendorExtensions.x-path}}</stringProp>
+          <stringProp name="HTTPSampler.path">{{basePathWithoutHost}}{{vendorExtensions.x-path}}</stringProp>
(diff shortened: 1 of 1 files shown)


wing328:
@foragerr thanks for the PR. If someone already has `x-path` set, would that result in duplicated base path?

foragerr:
unsure. How does one set `x-path`?
[Looks like JMeter codegen sets `x-path` BTW](https://github.com/swagger-api/swagger-codegen/blob/ee8547f3b4e6260c6b9ae3dbfd94652109adcb3e/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JMeterCodegen.java#L109)

wing328:
Looks like it's adding $ as a prefix to some variables. My guess is that it's for JMeter to recognize that as a variable.

wing328:
My guess is that the author of the JMeter template does to use base path (e..g /v2) in his/her API.

We'll go with your change. Thanks again with the PR.

foragerr:
`${...}` is indeed a variable in JMeter.

wing328:
Ok. Learned something today :smiley:

wing328:
One suggestion if you don't mind. For upcoming PR. I would recommend creating a new branch for the change.

foragerr:
Sure, that makes sense. I was wondering about that when I wanted to push some changes before you merged PR.

Issue #1937, opened 2016-01-21
[Perl] add support for `binary` (byte array)


wing328:
Add support for binary parameter and response type in the Perl API client.

A good starting point is to look at how other languages make similar changes:
Ruby: https://github.com/swagger-api/swagger-codegen/pull/1934 
Java: https://github.com/swagger-api/swagger-codegen/pull/1898
C#: https://github.com/swagger-api/swagger-codegen/pull/1864
PHP: https://github.com/swagger-api/swagger-codegen/pull/1891

If anyone from the community wants to work on this, please reply to let us know.

PR #2125, opened 2016-02-12
[ObjC] Fix #2121, generated method names don't follow coding convention

### LINKED ISSUE #2121: [ObjC] generated method name's `WithCompletionBlock` postfix doesn't follow convention?
API methods for Obj-C always seems to have `WithCompletionBlock` postfix, which doesn't follow the coding convention.  For example, I think the following generated method:
```objc
-(NSNumber*) userConfirmWithCompletionBlock :(NSString*) uid 
     token:(NSString*) token 
     redirect:(NSString*) redirect 
    completionHandler: (void (^)(NSError* error))completionBlock;
```
should have been like the following:
```objc
-(NSNumber*) userConfirmWithUid :(NSString*) uid 
     token:(NSString*) token 
     redirect:(NSString*) redirect 
    completionHandler: (void (^)(NSError* error))completionBlock;
```
FYI, the yaml for this method is:
```yaml
  /Users/confirm:
    get:
      tags:
        - User
      summary: Confirm a user registration with email verification token.
      description: !<tag:yaml.org,2002:js/undefined> ''
      operationId: User.confirm
      parameters:
        - name: uid
          in: query
          description: !<tag:yaml.org,2002:js/undefined> ''
          required: true
          type: string
        - name: token
          in: query
          description: !<tag:yaml.org,2002:js/undefined> ''
          required: true
          type: string
        - name: redirect
          in: query
          description: !<tag:yaml.org,2002:js/undefined> ''
          required: false
          type: string
      responses:
        '204':
          description: Request was successful
          schema: !<tag:yaml.org,2002:js/undefined> ''
      deprecated: false
```
This is as of b54947d (the current latest) (@bajtos wanted to keep informed about this too)

hideya:
Fixes https://github.com/swagger-api/swagger-codegen/issues/2121

Since I needed to reference both the untouched and the camelize names of the first argument in the mustache files,  `CodegenOperation.firstParamAltName` has been introduced.

Now generated method names look like the following, e.g.:
```objc
-(NSNumber*) coffeeShopUpdateAllWithWhere: (NSString*) where
    data: (XXCoffeeShop*) data
    completionHandler: (void (^)(XXInlineResponse2001* output, NSError* error)) handler;
```
instead of:
```objc
-(NSNumber*) coffeeShopUpdateAllWithCompletionBlock :(NSString*) where 
     data:(XXCoffeeShop*) data 
    
    completionHandler: (void (^)(XXInlineResponse2001* output, NSError* error))completionBlock;
```
Note that the argument variable name for `completionHandler` has also been updated from `completionBlock` to `handler`, following the signature of a popular method of `ModalWindowController`:
```objc
    - (void)makeKeyAndOrderFront:(id)sender
                   modalToWindow:(NSWindow*)window
                      sourceRect:(NSRect)rect
               completionHandler:(void (^)(NSInteger result))handler;
```
Also, the name of `-[ApiClient requestWithCompletionBlock:...]` too has been updated to `-[ApiClient requestWithPath:...]` to follow the convention.

@wing328 I have tested the changes against my test cases but haven't made any changes in the test cases and the samples.  Would you please tell me what I need to do for those?

### DIFF
--- a/samples/client/petstore/objc/SwaggerClientTests/Tests/PetApiTest.m
+++ b/samples/client/petstore/objc/SwaggerClientTests/Tests/PetApiTest.m
@@ -35,12 +35,12 @@
     XCTestExpectation *expectation = [self expectationWithDescription:@"testGetPetById"];
     SWGPet* pet = [self createPet];
 
-    [api addPetWithCompletionBlock:pet completionHandler:^(NSError *error) {
+    [api addPetWithBody:pet completionHandler:^(NSError *error) {
         if(error){
             XCTFail(@"got error %@", error);
         }
         NSLog(@"%@", [pet _id]);
-        [api getPetByIdWithCompletionBlock:[pet _id] completionHandler:^(SWGPet *output, NSError *error) {
+        [api getPetByIdWithPetId:[pet _id] completionHandler:^(SWGPet *output, NSError *error) {
             if(error){
                 XCTFail(@"got error %@", error);
             }
@@ -75,12 +75,12 @@
     XCTestExpectation *expectation = [self expectationWithDescription:@"testUpdatePet"];
     SWGPet* pet = [self createPet];
 
-    [api addPetWithCompletionBlock:pet completionHandler:^(NSError *error) {
+    [api addPetWithBody:pet completionHandler:^(NSError *error) {
         if(error) {
             XCTFail(@"got error %@", error);
         }
         else {
-            [api getPetByIdWithCompletionBlock:[NSString stringWithFormat:@"%@",[pet _id]] completionHandler:^(SWGPet *output, NSError *error) {
+            [api getPetByIdWithPetId:[pet _id] completionHandler:^(SWGPet *output, NSError *error) {
                 if(error) {
                     XCTFail(@"got error %@", error);
                 }
@@ -94,12 +94,12 @@
                     [pet setName:@"programmer"];
                     [pet setStatus:@"confused"];
 
-                    [api updatePetWithCompletionBlock:pet
+                    [api updatePetWithBody:pet
                                     completionHandler:^(NSError *error) {
                                         if(error) {
                                             XCTFail(@"got error %@", error);
                                         }
-                                        [api getPetByIdWithCompletionBlock:[pet _id] completionHandler:^(SWGPet *output, NSError *error) {
+                                        [api getPetByIdWithPetId:[pet _id] completionHandler:^(SWGPet *output, NSError *error) {
                                             if(error) {
                                                 XCTFail(@"got error %@", error);
                                             }
@@ -167,13 +167,13 @@ which causes an exception when deserializing the data
     NSLog(@"%@", pet._id);
     pet.tags = [[NSArray alloc] initWithObjects:tag, nil];
 
-    [api addPetWithCompletionBlock:pet completionHandler:^(NSError *error) {
+    [api addPetWithBody:pet completionHandler:^(NSError *error) {
         if(error) {
             XCTFail(@"got error %@", error);
         }
         NSArray* tags = [[NSArray alloc] initWithObjects:@"tony", nil];
 
-        [api findPetsByTagsWithCompletionBlock:tags completionHandler:^(NSArray *output, NSError *error) {
+        [api findPetsByTagsWithTags:tags completionHandler:^(NSArray *output, NSError *error) {
             if(error){
                 XCTFail(@"got error %@", error);
             }
@@ -200,15 +200,15 @@ which causes an exception when deserializing the data
 
     SWGPet* pet = [self createPet];
 
-    [api addPetWithCompletionBlock:pet completionHandler:^(NSError *error) {
+    [api addPetWithBody:pet completionHandler:^(NSError *error) {
         if(error){
             XCTFail(@"got error %@", error);
         }
-        [api deletePetWithCompletionBlock:pet._id apiKey:@"" completionHandler:^(NSError *error) {
+        [api deletePetWithPetId:pet._id apiKey:@"" completionHandler:^(NSError *error) {
             if(error){
                 XCTFail(@"got error %@", error);
             }
-            [api getPetByIdWithCompletionBlock:[pet _id] completionHandler:^(SWGPet *output, NSError *error) {
+            [api getPetByIdWithPetId:[pet _id] completionHandler:^(SWGPet *output, NSError *error) {
                 if(error) {
                     // good
                     [expectation fulfill];
@@ -228,7 +228,7 @@ which causes an exception when deserializing the data
     
     NSURL *fileURL = [self createTempFile];
     
-    [api uploadFileWithCompletionBlock:@1 additionalMetadata:@"special-metadata" file:fileURL completionHandler:^(NSError *error) {
+    [api uploadFileWithPetId:@1 additionalMetadata:@"special-metadata" file:fileURL completionHandler:^(NSError *error) {
         if(error) {
             // good
             XCTFail(@"expected a failure");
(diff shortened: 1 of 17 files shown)


wing328:
Thanks for the PR.

Instead of introducing `CodegenOperation.firstParamAltName`, have you considered adding the `firstParamAltName` in `Map<String, Object> postProcessOperations(Map<String, Object> objs)`? Currently `firstParamAltName` is for ObjC client only and that's why I suggested earlier to add it in `postProcessOperations` for ObjC generator instead.

Please update the ObjC sample as mentioned in https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

To run the integration test, please do
```
cd samples/client/petstore/objc && mvn integration-test -rf :ObjcPetstoreClientTests
```

hideya:
@wing328 Ah, now I see what you meant.  Thank you!
I'm working on `postProcessOperations()` and got a question, if you don't mind me asking...
Could you tell me how the `*****` part in the following code should look like? (I couldn't figure it out...)
```java
    @Override
    public Map<String, Object> postProcessOperations(Map<String, Object> objs) {
        Map<String, Object> operations = (Map<String, Object>) objs.get("operations");
        if (operations != null) {
            List<CodegenOperation> ops = (List<CodegenOperation>) operations.get("operation");
            for (CodegenOperation operation : ops) {
                if (!operation.allParams.isEmpty()) {
                    String firstParamAltName = operation.allParams.get(0).paramName;
                    ******.put("firstParamAltName", camelize(firstParamAltName));
                }
            }
        }
        return objs;
    }
```
I hope the other part of the code looks mostly ok...  Thank you in advance!

hideya:
@wing328 I'm working on updating the Petstore sample and encountering an issue regarding `binary` support.

It looks that the latest `petstore.json` included two new endpoints whose response uses the following schema:
```json
            "schema": {
              "type": "string",
              "format": "binary"
            }
```
For such endpoints, the current Obj-C codegen generates code like the following:
```objc
//
///
/// Fake endpoint to test byte array return by 'Find pet by ID'
/// Returns a pet when ID < 10.  ID > 10 or nonintegers will simulate API error conditions
///
/// @param petId ID of pet that needs to be fetched
/// 
///
/// @return SWGBinary*
-(NSNumber*) getPetByIdWithByteArrayWithPetId: (NSNumber*) petId
    completionHandler: (void (^)(SWGBinary* output, NSError* error)) handler;
```
where `binary` is mapped to a class `SWGBinary`, but there is no `SWGBinary.h/m` file generated, thus compilation of the sample ObjC client fails.  I'm guessing that `binary` is introduced recently and ObjC codegen doesn't support it fully yet? (guessing so from this Issue: https://github.com/swagger-api/swagger-codegen/issues/1937)  Or, I'm possibly doing something wrong, can you tell??

wing328:
@hideya it's my oversight and the issue has been fixed by #2125 (merged into master). Please pull the latest master and rebase.

hideya:
@wing328 thanks!  I've updated the sample and the test files too.

While updating, I noticed an inconvenience of this PR's new method name generation proposal.
It happens when `operationId` includes the name of the first argument in such a way like `getPetById`.
In this particular case, the generated method looks like the following:
```objc
  [api getPetByIdWithPetId:[pet _id] completionHandler:^(SWGPet *output, NSError *error) {
```
FYI, it previously was:
```objc
  [api getPetByIdWithCompletionBlock:[pet _id] completionHandler:^(SWGPet *output, NSError *error) {
```
This may be a limitation of mechanically generating method names from `operationId`s for multiple languages that have different syntax characteristics and conventions.
What would be your take about this?

samples/client/petstore/objc/SwaggerClientTests/Tests/PetApiTest.m
```diff
@@ -75,12 +75,12 @@ - (void) testUpdatePet {
...
     SWGPet* pet = [self createPet];
 
-    [api addPetWithCompletionBlock:pet completionHandler:^(NSError *error) {
+    [api addPetWithBody:pet completionHandler:^(NSError *error) {
         if(error) {
             XCTFail(@"got error %@", error);
         }
         else {
-            [api getPetByIdWithCompletionBlock:[NSString stringWithFormat:@"%@",[pet _id]] completionHandler:^(SWGPet *output, NSError *error) {
+            [api getPetByIdWithPetId:[pet _id] completionHandler:^(SWGPet *output, NSError *error) {
```
hideya: I'm not sure why the conversion to `NSString` was needed in the original code...?

wing328: maybe _id was an integer/long before...

wing328:
@hideya there's no way we can perfectly auto generate the method name to conform's to every language's style.

Personally I'm ok with your current approach. Would it meet the requirement for your use case? I mean do you have `operationId` similar to `getPetById` that's giving you a sub-optimal method name?

hideya:
@wing328 Yeah, I also found similar cases like `getPetById` in our example, but I'm OK with the current approach too since I'd think more development will be done in Swift in future (I don't want to spend too much effort on ObjC, and including any heuristics could be nasty).

wing328:
For perfection, we can introduce a vendor extension tag (e.g. x-objc-operationId) to use the value specified in x-objc-operationId if it's defined. I'll leave it up to you to implement it or not.

hideya:
If you are fine with the current PR, then I'm fine too.  If you'd rather like to have x-objc-operationId implemented, I'm happy to do that.  Please let me know.

wing328:
If you've time, please kindly work on it as this can serve as a workaround to use the "old" method name in Objc API client after upgrading to the latest version that includes your proposed change. In other words, this can make the upgrade backward compatible.

hideya:
Sure thing!  Would you point me an example code of vendor extension tag in the codebase?

wing328:
I think you can use it directly in the mustache template, e.g https://github.com/swagger-api/swagger-codegen/blob/ece56af43c2559c933a2e95bc283078c1ce88f91/modules/swagger-codegen/src/main/resources/Java/libraries/feign/api.mustache

hideya:
I've updated this PR to support `x-objc-operationId` vendor extension tag.
Just in case, please let me summarise my understanding of the usage:

*  the user who creates SwaggerSpec can include optional `x-objc-operationId:` tag per operation to explicitly specify the generated method name for that particular operation.
*  for backward compatibility purpose, for example, the user can put `x-objc-operationId: getPetByIdWithCompletionBlock` to operation 'getPetById'
*  to deal with the issue of unnatural generated method name discussed above, the user can put, for example, `x-objc-operationId: getPetById` which will result in generating method `-(NSNumber*) getPetById: (NSNumber*) petId ...` instead of `-(NSNumber*) getPetByIdWithPetId: (NSNumber*) petId ...`

Does this sound right?  If so, I think this PR is ready for re-review and for possible merge.

fehguy:
Hi, this sounds good, but I would make sure that the java code will set the `operationId` if the `x-objc-operationId` does not exist.  You may already have this in the code but just making sure that conceptually, it will work by default without any special extensions.  They will just be sugar on the generation.

hideya:
@fehguy thanks for the comment.  Yes, the `operationId` gets set regardless of `x-objc-operationId`.  As a matter of fact, Swift code generation for the PetStore sample is working fine, where `x-objc-operationId` is set at all.  Am I answering your question?

wing328:
`x-objc-operationId` is used only in the template: 
```
+-(NSNumber*) {{#vendorExtensions.x-objc-operationId}}{{vendorExtensions.x-objc-operationId}}{{/vendorExtensions.x-objc-operationId}}{{^vendorExtensions.x-objc-operationId}}{{nickname}}{{#hasParams}}With{{vendorExtensions.firstParamAltName}}{{/hasParams}}{{^hasParams}}WithCompletionHandler: {{/hasParams}}{{/vendorExtensions.x-objc-operationId}}{{#allParams}}{{#secondaryParam}}
```
so there should be no impact to other API clients.

wing328:
@hideya  The change looks very good to me. Thanks for the contribution.

PR #2126, opened 2016-02-12
Correct exporting of enum classes

### LINKED ISSUE #2102: [JavaScript] Enums are not exported correctly
In the generated models, when there are enum, an extra segment is introduced, like

```javascript
//export module
if ( typeof define === "function" && define.amd ) {
  define('SomeTypeEnum', [], function() {
    return SomeTypeEnum;
  });
}
```

This has the effect that
- If the module system is AMD, this will just export the enum, and not the main model class
- If the module system is not AMD, this will just export the model class, and not the enum 

The import/export setup also seems a bit inconsistent and strange in general. We are using both returns and setting the "module" parameter. Shouldn't we be using [this UMD template](https://github.com/umdjs/umd/blob/master/templates/commonjsStrict.js)?

I propose that the way we export enums should be as a property of the module class itself. For example, we might have a module `Account` which includes an enum `AccountType`. Then, we should export these (using the style in the template I linked to above) using

```javascript
exports = Account;
exports.AccountType = AccountType;
```

delenius:
Fixes #2102.

This attaches the enum classes to the main model class. This also ensures that the enum is exported along with the model. There is no longer a need for a separate export of the enum.

The enums now looks similar to a local class in Java, which seems appropriate, since they are indeed local to the model class being defined.

Note: This conflicts with https://github.com/swagger-api/swagger-codegen/pull/2115
Take them in any order, and I'll rebase accordingly. We need both to really fix the enums, but the two PRs address two different issues.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Javascript/enumClass.mustache
+++ b/modules/swagger-codegen/src/main/resources/Javascript/enumClass.mustache
@@ -1,17 +1,10 @@
-
-//export module
-if ( typeof define === "function" && define.amd ) {
-	define('{{datatypeWithEnum}}', [], function() {
-        return {{datatypeWithEnum}};
-	 });
-}
-
 var {{datatypeWithEnum}} = {
 {{#allowableValues}}{{#enumVars}}
-	/**
-	 * @const
-	 */
-	{{name}}: "{{value}}"{{^-last}},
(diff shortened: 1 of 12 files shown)


wing328:
@delenius thanks for the PR. #2115 has been merged into master. Please kindly do a rebase on the latest master.

delenius:
@wing328 it should be fine now.

Issue #1924, opened 2016-01-19
Generated Java Client does not support PATCH


sutdisi:
ApiClient throws exception for PATCH: ApiException: unknown method type PATCH

Because getAPIResponse does not know about PATCH.

```java
    if ("GET".equals(method)) {
      response = (ClientResponse) builder.get(ClientResponse.class);
    } else if ("POST".equals(method)) {
      response = builder.type(contentType).post(ClientResponse.class, serialize(body, contentType, formParams));
    } else if ("PUT".equals(method)) {
      response = builder.type(contentType).put(ClientResponse.class, serialize(body, contentType, formParams));
    } else if ("DELETE".equals(method)) {
      response = builder.type(contentType).delete(ClientResponse.class, serialize(body, contentType, formParams));
    } else {
      throw new ApiException(500, "unknown method type " + method);
    }
```

webron:
Looks like just one step before a PR!

sutdisi:
Hey webron, does that mean you're working on this?

webron:
No, it means that it looks like you found the code causing the issue, so it should be easy for *you* to create a PR for it.

wing328:
@sutdisi I believe you're using the default Java client, which uses Jersey 1.x as the HTTP client. 

Jersey1.x only supports PATCH using PostReplaceFilter: http://stackoverflow.com/questions/22355235/patch-request-using-jersey-client

As an alternative, you can use Jersey2.x or Retrofit as the HTTP client for your library. Please refer to https://github.com/swagger-api/swagger-codegen#customizing-the-generator for more information on selecting a different HTTP library in the Java API client.

sutdisi:
Using default client I added the override header and the query parameter in the client, but server still received it as a POST. Instead of PATCH which I was trying to send. I didn't see anything in the links, but is there a server setting I need?

wing328:
@sutdisi if you've use tcpdump or other tools to verify that the HTTP request is good (e.g. PATCH is used instead of POST), then I think it's a server-side error.

wing328:
PR merged into master.

PR #2128, opened 2016-02-13
Adding PATCH header override the default Java client

who:
This "fixes" https://github.com/swagger-api/swagger-codegen/issues/1924 .

The solution was to add the `X-HTTP-Method-Override` to the `POST` call. Using this change, swagger-codegen clients will actually make PATCH requests with POST as the verb, but with a header of `X-HTTP-Method-Override` set to PATCH. It is then the responsibility of the server to recognize the header and swap out the HTTP verb accordingly in a request filtering mechanisgm such as a `com.sun.jersey.spi.container.ContainerRequestFilter`.

Currently, this workaround, and corresponding handling on the server side, is the only way to reliably perform PATCH using Jersey 1.x.
See: http://theza.ch/2009/06/17/override-the-http-request-method-in-jersey/

Note: I have a lot of changes in this diff spit out by the generator; I think someone before me forgot to commit the generated code, perhaps.


wing328:
@who thanks for your contribution as always.

Issue #983, opened 2015-07-15
Request: support for C# .NET 3.5 client


who:
Currently, the C# generator in swagger-codegen conforms to .Net 4.5+.

It would be nice to have a .Net 3.5 compatible generator, so that the resulting DLL could be used in places where .Net 4.5 functionality is not possible.

wing328:
@who thanks for the feedback. Do you mean updating the existing C# client to make it backward compatible with .Net 3.5 framework ?

who:
@wing328 That is one way to do it, yes, but the existing C# client uses many language features that don't translate back to 3.5 very well, so this could be tricky.

wing328:
@who if you can give me a bit more direction on what feature is not available in 3.5, I can start updating the template during weekend.

who:
@wing328 The biggest difference is the lack of `Task`, `await`, and easy async support in 3.5. Also, `dynamic` type is not supported in 3.5. There are some libraries that will be able to support these missing features as a NuGet packages, but this won't be a complete solution. Additionally, RestSharp is compatible with 3.5, but only the 4.0 version contains `RestClient.ExecuteTaskAsync', which the existing C# swagger-codegen client uses heavily. Because of this, adding 3.5 support to the existing template might change it significantly, to the point where creating a separate template may be preferred.

I am going to fork and start working on a PR, but I could use some help if you'd like to collaborate.

who:
@wing328 It looks like this PR broke .Net 3.5 compatibility: https://github.com/swagger-api/swagger-codegen/pull/809

who:
I have filed a question regarding Async capabilities in Resharper when using augmented .Net 3.5 here: https://github.com/restsharp/RestSharp/issues/718

who:
Update from @hallem (RestSharp dev)

> The short answer is no, the Async support in RestSharp requires .Net 4.0+

See: https://github.com/restsharp/RestSharp/issues/718

wing328:
@who thanks for following up with RestSharp dev. In this case, I would agree with your suggestion earlier to add a separate generator for C# .net 3.5.

who:
@wing328 Agreed; a separate generator is the solution. I actually put some work into implementing 3.5 support in the existing templates, but found that it required too many "if .net 3.5" and "if not .net 3.5" conditionals.

wing328:
[...]

PR #2132, opened 2016-02-13
[csharp] Initial settings for v3.5 client compatibility

jimschubert:
* Pass -DtargetFramework=v3.5 to generate .NET 3.5 client
* Run compile.bat or compile-mono.sh to bootstrap RestSharp and Newtonsoft.Json dependencies.

Issue #983

Opening this pull request so others can take a look at the code. Running this with `optionalProjectFile=true`, I'm able to open the project file and compile in Visual Studio 2015 Community. I haven't attempted to compile and run within a .NET 3.5 project.

There were a number of issues that I haven't addressed with this pull request:

1. The SwaggerClientTest project is .NET 4.5. That project isn't compiled along with the client generator, and it would have been a lot more effort to generate that project to target both versions of the framework. The current tests are written to test the async client endpoints only.
2. The client's project structure isn't setup to "just work" with nuget. For example, if you generate the optional project file, it doesn't find the nuget package config and integrate with Visual Studio's nuget tool window.
3. There's no solution generated.

Those issues aren't really blockers. You can quickly generate the code, load the nuget package manager, and install RestSharp and Newtonsoft.Json (which is probably the common workflow anyway).

There were a couple places in the templates where I used the `{{supportsAsync}}` to determine functionality not specifically `async` (see compile.mustache and ApiClient.mustache). In compile.bat, I could have output the string "v3.5" and set the framework installation location conditionally, but checking for `async`, `Task`, and `dynamic` can all be rolled up into a single flag... I just didn't come up with a better name.


wing328:
@jimschubert RE: I just didn't come up with a better name.
I would have chosen a similar name :)

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CSharpClientCodegen.java
```diff
@@ -153,8 +194,6 @@ public void processOpts() {
         supportingFiles.add(new SupportingFile("ApiResponse.mustache",
                 clientPackageDir, "ApiResponse.cs"));
 
-        supportingFiles.add(new SupportingFile("Newtonsoft.Json.dll", "bin", "Newtonsoft.Json.dll"));
-        supportingFiles.add(new SupportingFile("RestSharp.dll", "bin", "RestSharp.dll"));
```
wing328: After removing these lines, the developers need to use NuGet to install these libraries, right ?

It make senses to me as we're trying to support different .net version. I'll make a note in the migration note for 2.1.5 to 2.1.6 upgrade.

jimschubert: Yeah, or run the compile scripts. As I mentioned in the pull request, the project isn't structured to support nuget when loading straight into Visual Studio. It's either run the compile scripts, or I can restructure the output directory to automatically restore.

Both scripts download nuget and grab the right packages based on the .net version selected. They then copy the nuget assemblies to the original bin folder, so I think the extra step is to make sure to execute the appropriate script for the platform. If I remember correctly, that's already mentioned in the readme of the generated source.

I'd personally like to generate the project file and have the nuget packages restored on build. I'd have to pay around with it to see how Xamarin and SharpDevelop handle nuget restore, if you think that'd be a better approach.

wing328: I can recall the reason on why we provided the restsharp.dll. There's a bug with the latest version of restsharp (https://github.com/restsharp/RestSharp/issues/742) which makes the file upload test failed.

wing328: RE: I'd personally like to generate the project file and have the nuget packages restored on build.

I agree that's a better approach.

jimschubert: I can lock restsharp version down to 105.1.0, which is what was provided previously, then it should still work as expected via nuget.

Do you think I should attempt to restructure the project output as part of this pull?

jimschubert: I reverted to RestSharp 105.1.0 and made note of this limitation in the generated Readme.md.

wing328: I'm using mono. According to developers in that thread, .net also has the issue.

I'll merge later today and please try to restructure the project output later with another PR.

wing328:
@jimschubert the change looks very good to me and my test results are ok. 

Thanks for the contribution to bring the C#-related generators to the next level.

PR #2136, opened 2016-02-15
Java okhttp-gson: fix dependencies

xhh:


wing328:
This PR fixes the gradle issue with the Java API client (okhttp-gon)

PR #2140, opened 2016-02-15
Fixed constantize for sub response object

vladshub:
There is an issue with complex response types and Ruby
Response example:
```yaml
responses: 
  200: 
    description: "A code "
    schema: 
      type: "object"
      properties: 
        response: 
          $ref: "#/definitions/LinkResponse"
```
The return type value in all places looks like this:
```ruby
:return_type => 'inline_response_200'
```
and in the api code there is a call to ``` .const_get(return_type).new ``` and it can't find the const.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/ruby/api_client.mustache
+++ b/modules/swagger-codegen/src/main/resources/ruby/api_client.mustache
@@ -168,7 +168,7 @@ module {{moduleName}}
         end
       else
         # models, e.g. Pet
-        {{moduleName}}.const_get(return_type).new.tap do |model|
+        {{moduleName}}.const_get(return_type.split("_").map{|v| v[0].upcase + v[1..-1] }.join).new.tap do |model|
           model.build_from_hash data
         end
       end


wing328:
@vladshub thanks for the PR. What does the response body look like?

inline_response_200 is created because inline scheme is used in your case to define the response. I wonder if your response is actually modelled like this:
``` 
responses: 
  200: 
    description: "A code "
    schema: 
      type: "object"
      $ref: "#/definitions/LinkResponse"
```

vladshub:
Hi @wing328,
The response body looks like this:
```json
{
  "response": {
    "code": "xyz"
  }
}
```
I had to change the swagger json to work since when it looked like this: 
```json
"responses": {
          "200": {
            "description": "A code ",
            "schema": {
              "$ref": "#/definitions/LinkResponse"
              
            }
          }
        }
```
and link response looked like this:
```json
"LinkResponse": {
      "type": "object",
      "properties": {
        "response": {
          "type": "object",
          "properties": {
            "code": {
              "type": "string"
            }    
          }
        }
      }
    }
```

the ruby code that was generated looke like this:
```ruby
module YotpoApi
  class LinkResponse
    attr_accessor :response

    # Attribute mapping from ruby-style variable name to JSON key.
    def self.attribute_map
      {
        
        :'response' => :'response'
        
      }
    end

    # Attribute type mapping.
    def self.swagger_types
      {
        :'response' => :'LinkResponse_response'
        
      }
    end
```

And when it tried to constantize the LinkResponse_response it failed.
Sorry for the invalid description...

xhh:
@vladshub Thank you for reporting the issue and provide the fix. But it would be better to fix it by generating the correct class name reference to the template. See #2169 for it.

wing328:
Closing this one as #2169 has been merged into master.

vladshub:
@wing328 thanks :)

PR #2143, opened 2016-02-16
Do not create empty objects in model constructors

### LINKED ISSUE #2142: [JavaScript] Empty objects in models
When we create instance of the model classes using `new Pet()` or whatever, and we don't supply parameters, we end up with "default" values. For fields that have another model class as their type, this means things like

```javascript
this['category'] = new Category();
```

This creates problems. It means that we can not use the usual javascript idiom

```javascript
if (pet.category) { ... }
```

because the `category` field is now always "truthy" even though it is an empty object. We also waste memory.

I think it would be better to leave all the non-required fields undefined. This is a one-line fix in `model.mustasche` and I will file a PR to that effect.

delenius:
Fixes #2142 

This removes the creation of empty instances
for optional parameters in the model constructors. The problem is that
when the data for these fields is not present, we are left with "empty"
instances, which is not desirable, for several reasons:

1. It means that we cannot check for the presence of a field using the
   usual JS idiom `if (object.field) {...}`.
2. It wastes memory with useless fields and objects.

Note: This completely removes the creation of "default" values
in the constructors. I don't know if there is some other case
where this is a useful thing (are there other kinds of defaults
besides empty instances of model classes?).

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Javascript/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/Javascript/model.mustache
@@ -19,16 +19,23 @@
    * {{description}}
    **/{{/description}}
   var {{classname}} = function {{classname}}({{#mandatory}}{{this}}{{^-last}}, {{/-last}}{{/mandatory}}) { {{#parent}}/* extends {{{parent}}}*/{{/parent}}
-    {{#vars}}
+    {{#vars}}{{#required}}
     /**{{#description}}
      * {{{description}}}{{/description}}
-     * datatype: {{{datatypeWithEnum}}}{{#required}}
-     * required{{/required}}{{#minimum}}
+     * datatype: {{{datatypeWithEnum}}}
+     * required {{#minimum}}
      * minimum: {{minimum}}{{/minimum}}{{#maximum}}
      * maximum: {{maximum}}{{/maximum}}
      **/
-    this['{{baseName}}'] = {{#required}}{{name}}{{/required}}{{^required}}{{{defaultValue}}}{{/required}};
-    {{/vars}}
+    this['{{baseName}}'] = {{name}};{{/required}}{{^required}}{{#defaultValue}}
+    /**{{#description}}
+     * {{{description}}}{{/description}}
+     * datatype: {{{datatypeWithEnum}}}
+     * required {{#minimum}}
+     * minimum: {{minimum}}{{/minimum}}{{#maximum}}
+     * maximum: {{maximum}}{{/maximum}}
+     **/
+    this['{{baseName}}'] = {{{defaultValue}}};{{/defaultValue}}{{/required}}{{/vars}}
   };
 
   {{classname}}.constructFromObject = function(data) {
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
@@ -17,9 +17,17 @@ import io.swagger.models.Model;
 import io.swagger.models.Operation;
 import io.swagger.models.Swagger;
 import io.swagger.models.properties.ArrayProperty;
+import io.swagger.models.properties.BooleanProperty;
+import io.swagger.models.properties.DateProperty;
+import io.swagger.models.properties.DateTimeProperty;
+import io.swagger.models.properties.DoubleProperty;
+import io.swagger.models.properties.FloatProperty;
+import io.swagger.models.properties.IntegerProperty;
+import io.swagger.models.properties.LongProperty;
 import io.swagger.models.properties.MapProperty;
 import io.swagger.models.properties.Property;
 import io.swagger.models.properties.RefProperty;
+import io.swagger.models.properties.StringProperty;
 
 import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
@@ -55,7 +63,7 @@ public class JavascriptClientCodegen extends DefaultCodegen implements CodegenCo
     protected String localVariablePrefix = "";
     protected boolean usePromises = false;
     protected boolean omitModelMethods = false;
-    
+
     public JavascriptClientCodegen() {
         super();
         outputFolder = "generated-code/js";
@@ -64,7 +72,7 @@ public class JavascriptClientCodegen extends DefaultCodegen implements CodegenCo
         templateDir = "Javascript";
         apiPackage = "api";
         modelPackage = "model";
-        
+
         // reference: http://www.w3schools.com/js/js_reserved.asp
         reservedWords = new HashSet<String>(
                 Arrays.asList(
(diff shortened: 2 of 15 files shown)


wing328:
We'd use cases before that developers were expecting default values (as defined in the spec) in the model's properties so I think we need to keep them.

delenius:
@wing328 , actually, swagger-codegen generates these "default" values even though the spec has no defaults specified. Let's look at the Pet class. The yaml has this definition:

```yaml
  Pet: 
    id: Pet
    description: "A pet is a person's best friend"
    required: 
      - id
      - name
    properties: 
      id: 
        type: integer
        format: int64
      tags: 
        type: array
        items: 
          $ref: Tag
      name: 
        type: string
      status: 
        type: string
        description: pet status in the store
        enum: 
          - available
          - pending
          - sold
      category: 
        $ref: Category
      photoUrls: 
        type: array
        items: 
          type: string
```

Here we see that `id` and `name` are requires properties, and `tags`, `status`, `category`, and `photoUrls` are optional. No defaults anywhere. Now let's see what swagger-codegen generates (comments omitted):

```javascript
  var Pet = function Pet(photoUrls, name) { 
    this['id'] = null;
    this['category'] = new Category();
    this['name'] = name;
    this['photoUrls'] = photoUrls;
    this['tags'] = [];
    this['status'] = null;    
  };
```

First, I just noticed a new problem: Swagger somehow treats `photoUrls` as a required field here, but not `id`. That looks like a separate bug.

Getting back to the issue at hand, the "default values" that swagger has generated are:
- `null` for `id` and `status` -- these are harmless (since you can still check for their presence using the falsy `null` value) but pointless (`null` is no more informative than `undefined`).
- An empty array for `tags`. This was not specified as a default by the user. Swagger just decided on it. An empty list is falsy, so you can check for presence of this field using the idiom mentioned above, but an empty list still has a different semantics than a `null` value (I actually use this distinction in my API -- I use `null` for values that have not yet been retrieved from the server; After they have been retrieved, they may turn out to be an empty list).
- En "empty object", `new Category()` for `category`. Again, the user did not request such a default value anywhere, and here we really cannot check whether it is a "real" value, or it's just one of these weird empty objects. Note that *required* fields in the this `Category` object may be missing, so you cannot really use it for anything.

Is there even a way to specify defaults for the model properties? The petstore example only has one `defaultValue`, and it is on a parameter, not a model. If it is possible to put defaults on model parameters, then I agree that it must be supported, but the code generator should not generate these non-specified defaults I mention above.

wing328:
I think model.mustache for other langauges can be a good reference, e.g. https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/csharp/model.mustache#L27 for C#

If my memory serves me well, I need to manually update `Pet` in the petstore sample to test the PR for adding default value support to C# models.

delenius:
The problem is that the spec currently assumes that there is *always* a default value for *every* non-required model property, as can be seen in this line in model.mustasche:

```javascript
this['{{baseName}}'] = {{#required}}{{name}}{{/required}}{{^required}}{{{defaultValue}}}{{/required}};
```

These non-default defaults are produced by `JavascriptClientCodegen.java`,

```java
    @Override
    public String toDefaultValue(Property p) {
        if (p instanceof ArrayProperty) {
            return "[]";
        } else if (p instanceof MapProperty) {
            return "{}";
        } else if (p instanceof RefProperty) {
            return "new " + getTypeDeclaration(p) + "()";
        }

        return super.toDefaultValue(p);
    }
```

Note that it is not checking whether the `Property` has a default specified. Here I also noticed another problem. It looks like [the `Property` interface](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-models/src/main/java/io/swagger/models/properties/Property.java) is missing a getter for `default` -- it only has a setter!

wing328:
That's not within my expectation. I need to step through the code tomorrow to understand how JS code generator processes the default value. Thanks for bringing this up.

delenius:
Also, `setDefault` currently does nothing, in the `Property` classes. So, to summarize,

1. swagger-core doesn't support default model properties yet
2. swagger-codegen makes up its own defaults, like empty lists and empty objects
3. (2) causes weird behavior

When (1) has been fixed, I believe the correct fix to (2) is to change `JavascriptClientCodegen.toDefaultValue()` method to:

```java
    @Override
    public String toDefaultValue(Property p) {
        String defaultValue = p.getDefault();
    }
```   

and in `model.mustache`, check for the presence of a `defaultValue`, rather than just assuming one exists. For now, the safest bet is to return `null` from `toDefaultValue`, and checking whether the default value exists in `model.mustache`. This means no "fake" default values will be produced, and once (1) has been fixed, the template should work as-is, and only trivial changes have to be made in the Java code.

wing328:
For 1, I just did a test by adding `1234` as the default value for Order ID and can find the following code in Ruby and CSharp:
```
      if attributes[:'id']
        self.id = attributes[:'id']
      else
        self.id = 1234
      end
```

```
        /// <summary>
        /// Initializes a new instance of the <see cref="Order" /> class.
        /// </summary>
        public Order()
        {
            this.Id = 1234;

        }
```

delenius:
Interesting. I wonder where they get it from :)
But not in the JS client, right?

wing328:
For the JS client, I got the following (which seems to be correct):
```
  var Order = function Order() {

    /**
     * datatype: Integer
     **/
    this['id'] = 1234;
```

I only tested the default for `integer`. Did you test other type (e.g. string) and found issue?

delenius:
How did you add the default value? I tried this in `pet.yml` and saw no change in the generated output

```javascript
  Pet:
    id: Pet
    description: "A pet is a person's best friend"
    required:
      - id
      - name
    properties:
      id:
        type: integer
        format: int64
        defaultValue: 1234
```

I also tried `default` instead of `defaultValue`.

wing328:
I used `default` instead of `defaultValue`

Ref: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schema-object

delenius:
I tried both :/ Are you committing this along with a test case? Then I can make sure I have the same setup.

wing328:
Here is the `default` I used in `Order`'s `id`:
```
    "Order": {
      "properties": {
        "id": {
          "default": 1234,
          "type": "integer",
          "format": "int64"
        },
        "petId": {
          "type": "integer",
          "format": "int64"
        },
        "quantity": {
          "type": "integer",
          "format": "int32"
        },
        "shipDate": {
          "type": "string",
          "format": "date-time"
        },
        "status": {
          "type": "string",
          "description": "Order Status",
          "enum": [
            "placed",
            "approved",
            "delivered"
          ]
        },
        "complete": {
          "type": "boolean"
        }
      },
      "xml": {
        "name": "Order"
      }
    }
```

modules/swagger-codegen/src/main/resources/Javascript/model.mustache
```diff
@@ -19,15 +19,23 @@
...
      * {{{description}}}{{/description}}
-     * datatype: {{{datatypeWithEnum}}}{{#required}}
-     * required{{/required}}{{#minimum}}
+     * datatype: {{{datatypeWithEnum}}}
+     * required {{#minimum}}
      * minimum: {{minimum}}{{/minimum}}{{#maximum}}
      * maximum: {{maximum}}{{/maximum}}
      **/
-    this['{{baseName}}'] = {{#required}}{{name}}{{/required}}{{^required}}{{{defaultValue}}}{{/required}};
+    this['{{baseName}}'] = {{name}}{{/defaultValue}}
```
xhh: `{{{defaultValue}}}` should be kept here instead of `{{name}}`.
BTW, to generate/support default values here, you can refer to the Ruby client:
https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RubyClientCodegen.java#L268

delenius:
Ah, ok, I was using the `.yml` files in `samples/yaml`. I see the "1234" default being generated now. However, it does not work for strings, booleans, or most other types. The code being used is the following method in `DefaultCodegen.java`. Note how it returns `"null"` in most cases. Only for the number types does it actually retrieve a default property that was set in the spec.

```java
    public String toDefaultValue(Property p) {
        if (p instanceof StringProperty) {
            return "null";
        } else if (p instanceof BooleanProperty) {
            return "null";
        } else if (p instanceof DateProperty) {
            return "null";
        } else if (p instanceof DateTimeProperty) {
            return "null";
        } else if (p instanceof DoubleProperty) {
            DoubleProperty dp = (DoubleProperty) p;
            if (dp.getDefault() != null) {
                return dp.getDefault().toString();
            }
            return "null";
        } else if (p instanceof FloatProperty) {
            FloatProperty dp = (FloatProperty) p;
            if (dp.getDefault() != null) {
                return dp.getDefault().toString();
            }
            return "null";
        } else if (p instanceof IntegerProperty) {
            IntegerProperty dp = (IntegerProperty) p;
            if (dp.getDefault() != null) {
                return dp.getDefault().toString();
            }
            return "null";
        } else if (p instanceof LongProperty) {
            LongProperty dp = (LongProperty) p;
            if (dp.getDefault() != null) {
                return dp.getDefault().toString();
            }
            return "null";
        } else {
            return "null";
        }
    }
```

Anyone know why it doesn't check for defaults in all cases here?

And IMO it would be better to return `null` instead of the string `"null"`. The latter prevents us from checking for it and skipping it in the mustache template.

In short, I am still trying to get to a point where
a) Default values from the spec are used
b) Other default values invented by the code generator are not used

It seems there are several obstacles to both...

wing328:
Don't think we should use the default one. Other languages have better implementation, e.g. https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PhpClientCodegen.java#L409

Please follow those instead.

delenius:
@wing328 yes that looks much better. I will modify this PR along the lines of the PHP client.

delenius:
Ok, I have modified the PR to work along the lines of the PHP client, at least as far as the Java part of the generator goes.

Now, the client has
- Working default values for all the types (only worked for numbers before)
- No defaults where none are specified (previously we had empty lists, empty objects, etc)

Notes:

1. This does remove even the comments for the missing fields. For example, for the `Order` class, the constructor is now completely empty, because there are neither required nor default values. If you like I can put back just the comment blocks for all the fields.
2. If a property is both required and has a default value, the default will not be used. This was also true in the previous version, and it doesn't make sense to provide a default for a req'd prop, so I think it's fine.

wing328:
@delenius For 2, I agree that required property should not have default value as developers must provide a valid value.

For 1, I'm ok with that.

Issue #2117, opened 2016-02-11
Java/pojo.mustache generates incorrect equals when class has an internal field that is the same as the classVarName of the class


vivin:
This is kind of an edge case. A simple example:

Assume we have the following class that has been exposed through the swagger model:
```java
public class Name {

    private String name;

    ...
}
```

The autogenerated code `Name#equals(Object)` ends up having this:

```java
return true && Objects.equals(name, name.name);
```

The fix is quite simple; we just need to prefix the properties with `this`. So the template should be:

```hbs
return true {{#hasVars}}&& {{#vars}}Objects.equals(this.{{name}}, {{classVarName}}.{{name}}){{#hasMore}} &&
        {{/hasMore}}{{/vars}}{{/hasVars}}
```

I can submit a PR for this fix, if that's alright.

wing328:
@vivin the proposed fix looks good to me. Let me know if you need help submitting a PR.

https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

wing328:
FYI. C# is already using the fix you proposed: 

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/csharp/model.mustache#L88

vivin:
Awesome! I will try submitting a PR today or possibly during the weekend.

wing328:
@vivin thanks! Let us know if you need any help.

PR #2150, opened 2016-02-16
Fix for issue #2117

vivin:
@wing328  @fehguy Sorry for being late on this - was with family over the weekend!

This change adds `this` to field names in `equals` within `pojo.mustache` so that an incorrect `equals` method is not generated in cases where a class has an internal field that is the same as the `classVarName` of the class.


wing328:
@vivin no worry. Hope you'd a great weekend with your family.

Next time, please update the Java Petstore sample as mentioned in https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md.

Also I would suggest creating a new branch for your change.

vivin:
Thanks! Will remember the other required changes, next time!

PR #2162, opened 2016-02-17
Handle empty response.body from superagent

delenius:
Superagent does not always produce a `body`. See
http://visionmedia.github.io/superagent/ for details. When it
doesn't, we should deserialize the raw `response.text` rather
than returning `null`. Currently, the JS client always returns
`null` when the return type is String! This commit fixes
that.


delenius:
Currently, the integration tests will fail, but they also fail in master if you regenerate them, so I'll wait for someone else to fix that.

wing328:
@delenius the fix to CI test for JS has been merged into master. Please kindly pull the latest master and perform a rebase.

Sorry for the inconvenience caused by this.

delenius:
Cool, rebase is done, and I have also updated the tests for javascript-promise in this PR.

Issue #1347, opened 2015-10-07
[Java] Support enums in query params


cbornet:
Enums are generated when they are part of a definition object but not when they are a direct method property.
For instance:
```json
"/pet/findByStatus": {
      "get": {
        "tags": [
          "pet"
        ],
        "summary": "Finds Pets by status",
        "description": "Finds Pets by status",
        "operationId": "findPetsByStatus",
        "produces": [
          "application/json",
          "application/xml"
        ],
        "parameters": [
          {
            "name": "status",
            "in": "query",
            "description": "Status value that for filter",
             "type": "string",
             "enum": ["available", "pending", "sold"]
            }
         ],
        "responses": {
          "200": {
            "description": "successful operation",
            "schema": {
              "type": "array",
              "items": {
                "$ref": "#/definitions/Pet"
              }
            }
          }
        }
```
Generates "status" as a String not as an enum.

fehguy:
This can be tricky.  When parameters are passed and mapped directly to enums, any name mismatching will cause big issues on the server side, typically things that you would need to handle on your own unless you're using some Jackson features for deserializing into enum properties

It is valid to expect, and probably worth keeping open and implementing, but it can have some negative side effects.

vivin:
@fehguy This is an issue I am running into as well. It *is* tricky, to be sure! I'm using the Feign library and have a template that generates a client that works with Feign. In Feign you can specify custom "expanders" (basically serializers) to convert an arbitrary object into an representation that is query-string safe. 

Currently it appears that the model built up by swagger-codegen ignores the enum information when building up a CodegenParameter instance. This is especially the case when the query parameter is an array of enums. It appears that all the data exposed by swagger is correctly recognized, but it is not attached to the CodegenParameter instance. For example, I have the following schema that describes a parameter:

```json
{
    "collectionFormat": "multi",
    "enum": [
        "UNKNOWN",
        "CUSTOMER",
        "LEAD",
        "PARTNER",
        "COWORKER",
        "PERSONAL",
        "SYSTEM",
        "VISITOR",
        "ANONYMOUS",
        "VENDOR"
    ],
    "in": "query",
    "items": {
        "enum": [
            "UNKNOWN",
            "CUSTOMER",
            "LEAD",
            "PARTNER",
            "COWORKER",
            "PERSONAL",
            "SYSTEM",
            "VISITOR",
            "ANONYMOUS",
            "VENDOR"
        ],
        "type": "string"
    },
[...]

PR #2163, opened 2016-02-17
Support for enums in query parameters (issue #1347)

vivin:
@fehguy @wing328 

This is a fix to support enums in query parameters. Enum-related information was not being stored on `CodegenParameter` previously; it is now. Test cases have been added to make sure that the enum information is being properly processed from the model.

Let me know if this change is alright. Also, the SDK templates have not been updated to take this change into account. Since I'm most familiar with Java, I think I might be able to update the Java template to support enums, but I think it would be best to make that a separate pull-request (assuming this one is approved).

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenParameter.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenParameter.java
@@ -7,9 +7,9 @@ import java.util.List;
 
 public class CodegenParameter {
     public Boolean isFormParam, isQueryParam, isPathParam, isHeaderParam,
-            isCookieParam, isBodyParam, hasMore, isContainer, 
+            isCookieParam, isBodyParam, hasMore, isContainer,
             secondaryParam, isCollectionFormatMulti;
-    public String baseName, paramName, dataType, collectionFormat, description, baseType, defaultValue;
+    public String baseName, paramName, dataType, datatypeWithEnum, collectionFormat, description, baseType, defaultValue;
     public String jsonSchema;
     public Boolean isString, isInteger, isLong, isFloat, isDouble, isByteArray, isBinary, isBoolean, isDate, isDateTime;
     public Boolean isListContainer, isMapContainer;
@@ -17,6 +17,7 @@ public class CodegenParameter {
     public boolean isEnum;
     public List<String> _enum;
     public Map<String, Object> allowableValues;
+    public CodegenProperty items;
     public Map<String, Object> vendorExtensions;
 
     /**
@@ -81,6 +82,7 @@ public class CodegenParameter {
         output.baseName = this.baseName;
         output.paramName = this.paramName;
         output.dataType = this.dataType;
+        output.datatypeWithEnum = this.datatypeWithEnum;
         output.collectionFormat = this.collectionFormat;
         output.isCollectionFormatMulti = this.isCollectionFormatMulti;
         output.description = this.description;
@@ -111,6 +113,9 @@ public class CodegenParameter {
         }
         if (this.allowableValues != null) {
             output.allowableValues = new HashMap<String, Object>(this.allowableValues);
+        }
+        if (this.items != null) {
+            output.items = this.items;
         }
         output.vendorExtensions = this.vendorExtensions;
         output.isBinary = this.isBinary;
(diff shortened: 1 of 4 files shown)


fehguy:
I think this looks good, thank you!  @wing328 can you please review?

wing328:
OK. I'll take a look and let you know if I've any question

maciejmakowski:
Is this only going to do enum generation for query parameters? Enum generation for path parameters is not going to be solved by this, right?

vivin:
@maciejmakowski Yes, this is just for query parameters.

@wing328 Do you need me to rebase this branch?

maciejmakowski:
@vivin @wing328 - any idea about path parameter support for enumerations? Is there already an open issue? 

I'm basically zero familiar with this project (in terms of what's happening technically under the hood), so I'll probably have a really hard time looking into this myself.

@vivin - Any chance you could maybe add that? I guestimating it could be very similar to adding the query parameter support?

vivin:
@maciejmakowski I will take a look. I have rebase this anyway so that they can merge it, so I will see how easy it will be to add enum support for path parameters. From what I remember from the code, the approach might be pretty similar.

maciejmakowski:
@vivin - That would be awesome! Thanks!

Also, just to ask in advance: How long does it usually take till changes become available in the maven plugin after they've been merged? We're using the maven plugin for code generation, so I'm wondering when we can roughly expect to be able to use those new features (after they were merged of course).

vivin:
@wing328 and @fehguy would probably know :). I'm just a contributor.

maciejmakowski:
@vivin - I have another question regarding implementation: So what would happen if multiple routes would include
a) the same query / path parameter with the same (enum) values, or
b) the same query / path parameter with slightly different (enum) values?

I think your change covers (b), and it would generate two enums in that case, right?

But what about (a)? What if you had one of two "statuses" (with the same enum values) in multiple places? Doesn't this sort of point towards a solution with $refs just like with response objects?

What could also happen namely, is that we want to have the same enum both as a query / path parameter, and in one or more of the response classes / objects.

wing328:
@vivin please rebase on the latest master to resolve the merge conflicts.

vivin:
@maciejmakowski My change only adds the enum metadata to the model that is used by mustache templates; it doesn't change the templates themselves. For the situation you describe, yes, something with `$ref` would be required. I don't remember off the top of my head if you can do that for query and path parameters in the swagger spec. Assuming not, it would be somewhat difficult currently to distinguish between enums. You may have to extend the codegen class yourself to keep track of the enums that have been identified. BTW, can you open a separate issue regarding path params? I think I will limit this change to just query params. It is best not to mix issues.

@wing328 I will rebase this in a bit and update the PR.

wing328:
@vivin please take your time. I'll test tomorrow (Sat).

vivin:
@wing328 Cool, no worries! I just rebased and pushed, but it looks like there is a failing test on master currently (for jdk8).

wing328:
@vivin thanks again for the PR. The change looks good to me and PR merged into master.

If you've time, I wonder if you can submit another PR to update the [test case](https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/java/default/src/test/java/io/swagger/petstore/test/PetApiTest.java#L111)

vivin:
Of course. Will do!

maciejmakowski:
@vivin @wing328 - created separate issue for enum support for path parameters as requested: https://github.com/swagger-api/swagger-codegen/issues/2282

vivin:
@wing328 The templates would have to be updated for that test case to be updated, correct? Although the enum information is there, the templates do not consider it currently. Should I updates the templates first and then add the test case?

wing328:
Yes, please update the mustache template as well.

vivin:
@wing328 Will try to have this done today - was a little busy!

vivin:
@wing328 If you could refresh my memory, is it the `testGetPetByIdInObject` test case that needs to be updated or the `testFindPetsByStatus`?

wing328:
@vivin should be [testFindPetsByStatus](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.json#L160)

vivin:
@wing328 Thank you! One more question. Should the enums be in their own class, or should they be an inner enum of some other class (perhaps PetApi)? I think one problem with making it its own class is that there is no way to tell whether this enum is really the same as another enum with the same name, since the swagger spec doesn't provide a way to use a `$ref` for query parameters. Perhaps one way to do this is to compare the name and the constants. 

Which approach do you think would work best?

Issue #2104, opened 2016-02-10
Code/Architecture needs documentation


ePaul:
As a Java developer trying to contribute some code, I must say that I find it quite hard to orientate myself in the code base.

The [README.md](https://github.com/swagger-api/swagger-codegen/blob/master/README.md) has mainly information about the actual usage, and [CONTRIBUTING.md](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md) has some generic advice.

What I'm missing is a general overview how the code generation (and the configuration) works (the non-language-specific parts). It also doesn't help that most classes and methods don't have any Javadoc, which makes finding out what they are for quite hard.

Could the people who understand how it works please invest some time to write a general overview and Javadocs for the classes in the main package (io.swager.codegen)?

-----

For the current example: I wanted to finally fix #1255 (which I opened myself back in september). I figured out I would add fields `modelNamePrefix` and `modelNameSuffix` in DefaultCodeGen, and use them in `toModelName` (and maybe also `toModelFileName`?). So far, so easy.
The hard part now is figuring out how to get those values into this class from the maven plugin or the CLI.

It looks like some parameters are passed via additionalProperties, others are passed via dedicated setters from outside. To fill additionalProperties from the command line, there is the `--additionalProperties` option defined in io.swagger.codegen.cmd.Generate, but the Maven-Plugin seems to not support that. Also, CLI seems to use the CodegenConfigurator, which then creates a ClientOptInput, which is passed to a DefaultGenerator. The ClientOptInput contains both a ClientOpts object and a CodegenConfig object, and the additional options are passed from the former into the latter. Though in the CodegenMojo, the additional properties are directly set in the CodegenConfig object (but only some special-cased ones and those which are declared by the CodegenConfig implementation as cliOption and passed to the maven-plugin as configOptions).

Sorry, I can't see any real sense in all this. Can someone help me figure this out?
(If someone explains it to me, I can also help to document it, or refactor it so it gets clearer.)

fehguy:
Hi @ePaul you are correct that the README focuses on _usage_ rather than code structure.  There could be a separate development guide for the project which would of course be helpful to a number of people.

The maven plugin is duplicating a lot of work that the codegen proper is doing.  The argument parsing and configuration setting should be the same code, but it's not--hence the confusion it's giving you.  Note from here:

[...]

PR #2168, opened 2016-02-18
[maven plugin] Use configurator for maven plugin

jimschubert:
see swagger-api/swagger-codegen#2104

This reduces code duplication between CLI and maven-plugin.
The maven plugin can now support more options like `language-specific-primitives` and `additional-properties`.

Here's an example configuration to demonstrate passing `additional-properties`:

```xml
<plugin>
      <groupId>io.swagger</groupId>
      <artifactId>swagger-codegen-maven-plugin</artifactId>
      <version>2.1.6-SNAPSHOT</version>
      <executions>
          <execution>
              <goals>
                  <goal>generate</goal>
              </goals>
              <configuration>
                  <inputSpec>/Users/jim/projects/swagger-codegen/modules/swagger-codegen-maven-plugin/examples/swagger.yaml</inputSpec>
                  <language>csharp</language>
                  <configOptions>
                      <additional-properties>optionalProjectFile=true</additional-properties>
                  </configOptions>
              </configuration>
          </execution>
      </executions>
  </plugin>
```


ePaul:
```xml
                  <configOptions>
                      <additional-properties>optionalProjectFile=true</additional-properties>
                  </configOptions>
```
I guess this is out of scope for this pull request, but I think that additional-properties should be of type Map, not string, and thus allow structured entry:

```xml
                  <configOptions>
                      <additional-properties>
                          <optionalProjectFile>true</optionalProjectFile>
                      </additional-properties>
                  </configOptions>
```

I'm not sure how one could do a transition supporting both old and new form, though.

modules/swagger-codegen/src/main/java/io/swagger/codegen/config/CodegenConfiguratorUtils.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.commons.lang3.tuple.Pair;
+
+import java.util.*;
+
+/**
+ * Contains shared logic for applying key-value pairs and CSV strings
+ * to specific settings in CodegenConfigurator.
+ *
+ * <p>
+ *     This class exists to facility in testing. These methods could be applied
```
ePaul: "to facilitate testing"?

jimschubert: Thanks. Must've been muscle memory from the day job.

PR #2170, opened 2016-02-18
Add postProcessAllModels to default generator

wing328:
For processing all models before generating the model files.


hiveship:
I think it's a good thing :)

PR #2184, opened 2016-02-19
Map $ to value for parameter/property name

### LINKED ISSUE #2131: Support `$` property names in JSON Schema
In the case of XML->JSON migrated APIs I often see `$` property in JSON schema (corresponds to the value of the element in XSD).

Right now Swagger Codegen fails on this property, I get code like:
```
  @ApiModelProperty(example = "null", value = "")
  @JsonProperty("$")
  public String get() {
    return ;
  }
  public void set(String ) {
    this. = ;
  }
```

As a simple solution, `$` could be converted to `value`.

wing328:
To fix #2131


wing328:

Starting from 2.1.6, if a variable name is only "$", it will be mapped to "value"

PR #2192, opened 2016-02-20
avoid inner class

### LINKED ISSUE #2191: [PHP] generating wrong class name
I have an API based on [HAL](http://stateless.co/hal_specification.html) and I created the documentation with Swagger, but when I generate the client, some classes was created with dollar `$` in the name, such as: `CampaignPagination$Embedded`.

I take a look and I discovered those [lines](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L2098-L2104)

My configuration:

```yaml
...
paths:
  /campaigns:
    get:
      ...
      responses:
        '200':
          description: An list of campaigns
          schema:
            $ref: '#/definitions/CampaignPagination'
...
definitions:
  CampaignPagination:
    type: object
    properties:
...
      _embedded:
        type: object
        properties:
          items:
            type: array
            items:
              $ref: "#/definitions/Campaign"

  Campaign:
    type: object
    properties:
      id:
        type: string
...
```

tonicospinelli:
it is cased because the property start with underscore and
object is created on th fly

fix #2191

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -2095,14 +2095,6 @@ public class DefaultCodegen {
             word = m.replaceAll(rep);
         }
 
-        // Replace two underscores with $ to support inner classes.
-        p = Pattern.compile("(__)(.)");
-        m = p.matcher(word);
-        while (m.find()) {
-            word = m.replaceFirst("\\$" + m.group(2).toUpperCase());
-            m = p.matcher(word);
-        }
-
         // Remove all underscores
         p = Pattern.compile("(_)(.)");
         m = p.matcher(word);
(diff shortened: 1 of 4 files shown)


wing328:
@tonicospinelli thanks for the PR!

Instead of fixing the bug just for PHP, what about adding the fix in `sanitizeName` instead (so as to fix the same issue for other languages) ?

tonicospinelli:
@wing328 I will do for other languages ASAP :smile:

tonicospinelli:
@wing328 changing the code for `sanitizeName` will not solve the problem because the `camelize` method is called after and it will remove `\` from PHP types as `\DateTime`

Other case, inner class with this syntax `DefaultClass$InnerClass.class` is only for java? If is it, I will change only for java and remove those [lines](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L2098-L2104) from `DefaultCodegen`

What you think about it?

wing328:
Very good suggestion.

I don't think we need inner class support in any languages so I would suggest comment out those [lines](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L2098-L2104) and add a 1-liner comment linking to this discussion.

tonicospinelli:
@wing328 done!

PR #2199, opened 2016-02-21
[C#] Add support for Readonly property

wing328:
- Added support for Readonly property in C#
- Updated constructor to take optional parameters


wing328:
@jimschubert I'll make similar change to modules/swagger-codegen/src/main/resources/aspnet5/model.mustache this week or early next week. Please let me know if you see any potential issue with the change for `aspnet5`

wing328:

Starting from 2.1.6, `readOnly` members (which should not be set with any value) in C# model will not have any public setter.

PR #2202, opened 2016-02-22
Added discriminator support to Objective-C

traviscollins:
The swagger spec includes a discriminator field for Models. This discriminator value is the name of the property on the model that contains the name of the intended subtype for a given object. See discriminator under Fixed Fields in this page

http://swagger.io/specification/#schemaObject

This pull request adds a mustache conditional template to the objective-c model-object template for when the discriminator is defined for a given swagger model. The condition adds a custom initWithDictionary method that uses the value found in the dictionary to determine the appropriate class to return. This causes the discriminator value to be automatically applied per the spec. This implementation fails gracefully when the discriminator is defined but not supplied. The implementation fails hard when the discriminator is defined, and the value defines a class that is not present at runtime. 

Implementing this support required adding the discriminator value from the Swagger Model to CodeGen Model converter. This property can be used to add inheritance support to other languages.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenModel.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenModel.java
@@ -8,6 +8,7 @@ public class CodegenModel {
     public String parent, parentSchema;
     public String name, classname, description, classVarName, modelJson, dataType;
     public String unescapedDescription;
+    public String discriminator;
     public String defaultValue;
     public List<CodegenProperty> vars = new ArrayList<CodegenProperty>();
     public List<String> allowableValues;
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -864,6 +864,10 @@ public class DefaultCodegen {
         m.externalDocs = model.getExternalDocs();
         m.vendorExtensions = model.getVendorExtensions();
 
+        if (model instanceof ModelImpl) {
+            m.discriminator = ((ModelImpl) model).getDiscriminator();
+        }
+
 
         if (model instanceof ArrayModel) {
             ArrayModel am = (ArrayModel) model;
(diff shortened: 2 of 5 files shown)


traviscollins:
A little more clarification on the intention of discriminators and what this pull request does. 

Consider a Swagger definition with three models - Animal, Dog, and Cat. Dog and Cat "inherit" Animal through the "allOf" definition. Animal has a discriminator value of "className". This is how you define class inheritance in the Swagger spec. Now imagine the same Swagger spec defines a method GET /pets that returns a list (array) of Animal objects that were all either Dog or Cat.

Prior to this pull request, the Objective-C swagger client would parse all of the JSON returned down to Animal objects - and strip out all Dog and Cat specific properties. Therefore you could not really use inheritance in returned objects with the swagger Objective-C client. 

After this pull request, client will parse the JSON, using the discriminator field value, and return an array of Dog and Cat objects. The subtype specific properties are preserved, and now inheritance can be used on objects returned from the api.

wing328:
@traviscollins thanks for the PR. Do you mind sharing with me the spec so that I can use it for testing? (you will find my email address in my [Github page](https://github.com/wing328)).

Also I notice that in the Commits tab, the commit is not linked to your Github account (@traviscollins) so the contribution won't count toward https://github.com/swagger-api/swagger-codegen/graphs/contributors. I would recommend you to fix it by probably setting up `user.email` following the instruction [here](https://help.github.com/articles/setting-your-email-in-git/)

traviscollins:
Thanks for that tip about the email setting. Fixed it for next time. 

I included an example discriminator using spec in the pull request.
 
modules/swagger-codegen/src/test/resources/2_0/discriminatorTest.json

I included a JUnit test for the CodeGenModel discriminator property assignment in the pull request. But I wasn't able to determine a viable option for testing the output of the objective-c codegen config against that spec.

wing328:
@traviscollins thanks. We'll review and let you know if we've further questions.

wing328:
@traviscollins thanks again for the PR. To manage your expectation, I'll need a little bit more time to review and will provide you another update this weekend.

traviscollins:
No problem. Thanks for all your work on this project. Very useful.

wing328:
@traviscollins testings are good. PR merged. Thanks for the contribution on this.

PR #2206, opened 2016-02-22
[Java] Do not generate setter method for readonly properties

xhh:
Updated Java clients: default, feign, jersey2, okhttp-gson.


hiveship:
I will also add it to CXF Codegen :)

ePaul:
Note that `readOnly` for a Swagger property just means that the client shouldn't send it to the server, but the server can send it back.
So please make sure it doesn't remove the setter when generating server models.

hiveship:
Ok sorry @ePaul . I won't remove the setter in the CXF Codegen since it only generates a server stub :)

wing328:
FYI. The change applies to `retrofit` and `retrofit2` as well.

cc @cbornet

wing328:

Starting from 2.1.6, there will be no setter for `readOnly` attribute in Java models.

Issue #1255, opened 2015-09-18
Allow modifying the generated model names by prefix or suffix


ePaul:
This is a feature request for a new (optional) generator parameter.

**Summary**

It should be possible to specify a prefix and/or suffix which will be added to all model type names when generating the types.

**Background**
The model names generated from the swagger definition have usually quite natural names ... in Java, I got (in my case) for example Supplier, Organization, BusinessUnit.
I also have same-named domain classes  in the service implementing the API. Now of course I need to write a mapper between those two, and here I can't import both the domain and the API classes. One of them needs to be written out with full package name in the source, and this is ugly:

    private Organization mapOrganization(final 
        de.zalando.einkaufshelden.purchase_order.api.rest.model.Organization organization) {

For the similar problem previously with Soap services we used the suffix `DTO` (for Data Transfer Object) for the objects sent over the wire.

In other projects I've seen a prefix used for all class names to distinguish them from similar ones (think the Swing J* classes).

**Proposed solution**

There should be two configuration parameters (as a command line parameter in swagger-codegen-cli, as a maven configuration parameter in swagger-codegen-maven-plugin) for specifying a prefix and a suffix to be appended to the model names. Default values would be the empty string. (My name proposals would be `modelNamePrefix` and `modelNameSuffix`.)

These could be applied in DefaultCodegen.toModelName before passing the name to the sanitizing transformation.

Alternatively there could be also one single parameter somehow specifying the complete transformation, using a regular expression replacement or similar.

chenjianjx:
+1

jimschubert:
@ePaul you mentioned in #2104 that you intended to do this work. Have you gotten very far?

ePaul:
@jimschubert not much more than mentioned there, but I intend to work on this today.

ePaul:
@jimschubert I didn't get much done today (other urgent stuff interfered), but I pushed a [feature branch in my fork](https://github.com/swagger-api/swagger-codegen/compare/master...ePaul:feature/1255-modify-model-names). Hopefully tomorrow is better.

jimschubert:
@ePaul Hey I've had conjunctivitis for the past couple of days and I've been a lot less productive than normal. I'm going to check out your feature branch today and see what I can contribute to it.

jimschubert:
@ePaul I've implemented these options on swagger codegen cli. Please take a look (commit referenced above).

```
[...]

PR #2211, opened 2016-02-22
Feature/1255 suffix+prefix for model name

ePaul:
This fixes most of #1255 by adding the options `--model-name-prefix` and `--model-name-suffix` to the CLI, with corresponding changes to DefaultCodegen and CodegenConfigurator.

For C# and Java generators we (@jimschubert and me) also added support code as they override `DefaultCodegen.toModelName`.
I guess this needs to be changed for quite some other languages too.

We also need some unit tests for this functionality, these are not yet included. (The existing unit tests still run without changes, as none of them uses these options.)

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConstants.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConstants.java
@@ -82,4 +82,9 @@ public class CodegenConstants {
 
     public static enum MODEL_PROPERTY_NAMING_TYPE {camelCase, PascalCase, snake_case, original}
 
+    public static final String MODEL_NAME_PREFIX = "modelNamePrefix";
+    public static final String MODEL_NAME_PREFIX_DESC = "Prefix that will be prepended to all model names. Default is the empty string.";
+
+    public static final String MODEL_NAME_SUFFIX = "modelNameSuffix";
+    public static final String MODEL_NAME_SUFFIX_DESC = "Suffix that will be appended to all model names. Default is the empty string.";
 }
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
@@ -219,9 +219,6 @@ public class JavaClientCodegen extends DefaultCodegen implements CodegenConfig {
             importMapping.remove("DateTime");
             instantiationTypes.put("array", "java.util.ArrayList");
             instantiationTypes.put("map", "java.util.HashMap");
-        } else {
-            languageSpecificPrimitives.add("List");
-            languageSpecificPrimitives.add("Map");
         }
 
         this.sanitizeConfig();
@@ -490,7 +487,8 @@ public class JavaClientCodegen extends DefaultCodegen implements CodegenConfig {
         String type = null;
         if (typeMapping.containsKey(swaggerType)) {
             type = typeMapping.get(swaggerType);
-            if (languageSpecificPrimitives.contains(type) || type.indexOf(".") >= 0) {
+            if (languageSpecificPrimitives.contains(type) || type.indexOf(".") >= 0 ||
+                type.equals("Map") || type.equals("List")) {
                 return type;
             }
         } else {
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
@@ -219,6 +219,9 @@ public class JavaClientCodegen extends DefaultCodegen implements CodegenConfig {
             importMapping.remove("DateTime");
             instantiationTypes.put("array", "java.util.ArrayList");
             instantiationTypes.put("map", "java.util.HashMap");
+        } else {
+            languageSpecificPrimitives.add("List");
+            languageSpecificPrimitives.add("Map");
         }
 
         this.sanitizeConfig();
@@ -377,6 +380,7 @@ public class JavaClientCodegen extends DefaultCodegen implements CodegenConfig {
 
     @Override
     public String toModelName(String name) {
+        name = super.toModelName(name);
         name = sanitizeName(name); // FIXME: a parameter should not be assigned. Also declare the methods parameters as 'final'.
 
         // camelize the model name
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
@@ -376,22 +376,21 @@ public class JavaClientCodegen extends DefaultCodegen implements CodegenConfig {
     }
 
     @Override
-    public String toModelName(String name) {
-        name = super.toModelName(name);
-        name = sanitizeName(name); // FIXME: a parameter should not be assigned. Also declare the methods parameters as 'final'.
+    public String toModelName(final String name) {
+        final String sanitizedName = sanitizeName(modelNamePrefix + name + modelNameSuffix);
 
         // camelize the model name
         // phone_number => PhoneNumber
-        name = camelize(name);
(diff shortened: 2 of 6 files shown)


wing328:
@ePaul @jimschubert thanks for the PR. I'll review later today and let you guys know if I've any question.

modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConstants.java
```diff
@@ -82,4 +82,9 @@
 
     public static enum MODEL_PROPERTY_NAMING_TYPE {camelCase, PascalCase, snake_case, original}
 
+    public static final String MODEL_NAME_PREFIX = "modelNamePrefix";
+    public static final String MODEL_NAME_PREFIX_DESC = "Prefix that will be prepended to all model names. Default is the empty string.";
```
wing328: I believe you also aware that another way to add the default value is `.defaultValue` in CliOption ([example](https://github.com/swagger-api/swagger-codegen/blob/d05596dad42731356ecc187b7ffec915cd31c524/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PerlClientCodegen.java#L81))

In this case, I'm fine with both approaches as the default is an empty string.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
```diff
@@ -219,6 +219,9 @@ public void processOpts() {
             importMapping.remove("DateTime");
             instantiationTypes.put("array", "java.util.ArrayList");
             instantiationTypes.put("map", "java.util.HashMap");
+        } else {
+            languageSpecificPrimitives.add("List");
+            languageSpecificPrimitives.add("Map");
```
wing328: These 2 lines caused import error in the updated Java client.

To repeat the issue, run
```
./bin/java-petstore-all.sh
mvn verify -Psamples
```
Please commit the updated Java Petstore samples too.

(testing with C# client looks good. so far no issues found)

ePaul: Hmm, looks like having those as primitives removes the imports. I'll have to figure out how this works ...
Likely I'll remove this primitive-definition here and instead add some special-casing code to `getSwaggerType`.

wing328: For other languages, we do add List (array) and Map (dictionary, hash) to the languageSpecificPrimitives (which means no additional import of package/modules is required) but for Java, it requires explicit import of `java.util.List` and `java.util.Map`.

Are you trying to improve the Java import to remove unnecessary import statements from the Java API client?

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
```diff
@@ -377,6 +380,7 @@ public String toParamName(String name) {
 
     @Override
     public String toModelName(String name) {
+        name = super.toModelName(name);
```
wing328: Would there be an issue if we don't call `super.toModelName`, which is defined as follows?
```
    public String toModelName(String name) {
        return initialCaps(name);
    }
```

ePaul: in this super method (DefaultCodegen.toModelName) I added the actual logic of adding suffix and prefix. Of course I could instead duplicate this here instead of calling the super method (like what Jim did in the C# version).

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
ePaul: [Without this addition](https://github.com/swagger-api/swagger-codegen/issues/1255#issuecomment-187125334) I got `ListDto<...>` (and I guess also `MapDto<...>`) generated instead of `List<...>` (and `Map<...>`), because toModelName is called from JavaClientCodegen.getSwaggerType, if the type is not a primitive one (or contains a dot).

wing328: OK, so i think the solution (or workaround) is to check for `List` and `Map` in getSwaggerType (instead of checking against `languageSpecificPrimitives` only), right?

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
wing328: Personally I prefer doing it in JavaCodegen `toModelName` as it would make my life easier reviewing changes. For example if someone changes the default `toModelName`, I've to go through all codegen to see if any one of them calls `super.toModelName` to involve the default one.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
ePaul: Either this, or special-casing them in `toModelName`.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
jimschubert: I agree. I think it's better for generators to opt-in to the prefix/suffix logic than to reimplement the base method just to remove prefix/suffix.

modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConstants.java
jimschubert: That's true for `CliOption`, but I don't think it would work the same for maven/airline cli options, e.g.:

```
 @Option(name = {"--model-name-suffix"}, 
         title = "model name suffix", 
         description = CodegenConstants.MODEL_NAME_SUFFIX_DESC)
```

Airline, for example, only has `allowedValues`.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
ePaul: I think removing the prefix/suffix after the super method added it is not a good idea (if you don't want them, simply don't pass those parameters), but there might be the use case of using a different separator/naming convention (instead of the camelCase) for a programming language, which could be a reason to implement it in the subclass too.

So I would let it stay in the superclass method (which makes it available for all generators which don't override it), but implement it again in the JavaClientCodegen instead of calling `super.toModelName(...)`.

modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConstants.java
ePaul: Maybe better `If the option is not given, nothing will be prepended` (and analogously for the suffix)?

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
ePaul: Done as mentioned above.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
ePaul: I now added the special-casing in `getSwaggerType` and removed the else block here.

ePaul:
I rebased again to solve merge conflicts. @wing328 is it now fine this way or should I still change the parameter description text?

wing328:
@ePaul let me review now

wing328:
@ePaul @jimschubert thanks for the PR.

For other langauges, we'll update them one by one and hopefully it will be ready before the 2.1.6 release.

jimschubert:
@wing328 what's the best way to track languages needing prefix/suffix modifications? Maybe a check list on #1255?

wing328:
updated https://github.com/swagger-api/swagger-codegen/issues/1255 with a
task list.

PR #2218, opened 2016-02-23
[C#] add test case, update c# model constructor

wing328:
- add test cases for readonly property and deserialization
- update constructor to include non-readonly parameters


wing328:
Updated the PR to add required property check to `aspnet5` model constructor.

cc @jimschubert

jimschubert:
The serialization with read-only properties and private setters was my big concern. Looks like `order.Id` in the test covers that and works. Thanks.

wing328:
@jimschubert thanks for reviewing the change. PR merged.

wing328:

Starting from 2.1.6, one must provide all the required parameters when creating a model. Otherwise, an exception will be thrown. In addition, `readOnly` parameters (which should not be set with any value) will be removed from the constructor.
 
```
			Pet p = new Pet(Name: "Csharp test", PhotoUrls: new List<string> { "http://petstore.com/csharp_test" });
```

PR #2225, opened 2016-02-24
Match '.','-','_' in tag sanitization in addition to ' '

jsyi:
This allows for these common characters in tags resulting in more properly cased Api class names.

E.g.,
"order-service" => OrderServiceApi instead of OrderserviceApi


wing328:
@jsyi thanks for the PR. Would you have cycle to add a test case in https://github.com/swagger-api/swagger-codegen/blob/c4d799a4a7e47863437b22ee9b75c7d1828990e5/modules/swagger-codegen/src/test/java/io/swagger/codegen/DefaultGeneratorTest.java to cover the change?

wing328:
@jsyi also do you mind sharing a bit more about your use case? What does the tag look like in your scenario?

wing328:
@jsyi can you provide the swagger spec for reproducing the issue?

wing328:
Closing this as there's no update from you.

Issue #2156, opened 2016-02-17
[Swift] minor: including {{{examples}}} causes unnecessary diffs


hideya:
This is minor but Swift's `api.mustache` includes `{{{examples}}}` to show response example in the comment right above the generated method like the following.  However, if an example includes a property `"type": "string", "format": "date-time"`, it gets expanded to the time codegen ran, like `"2016-02-16T04:07:48.155+0000"`.  This results in unnecessary diffs every time running codegen even when no effective modification is made.  I thought `- examples: {{{examples}}}{{/examples}}` could be eliminated (other languages don't seem to include it and I don't see any strong reason to do so). 
What would you folks think?
```swift
        /**
         
         Place an order for a pet
         
         - POST /store/order
         - 
         - examples: [{example={
  "id" : 123456789,
  "petId" : 123456789,
  "complete" : true,
  "status" : "aeiou",
  "quantity" : 123,
  "shipDate" : "2016-02-16T04:07:48.155+0000"
}, contentType=application/json}, {example=<Order>
  <id>123456</id>
  <petId>123456</petId>
  <quantity>0</quantity>
  <shipDate>2016-02-16T13:07:48.160Z</shipDate>
  <status>string</status>
  <complete>true</complete>
</Order>, contentType=application/xml}]
         - examples: [{example={
  "id" : 123456789,
  "petId" : 123456789,
  "complete" : true,
  "status" : "aeiou",
  "quantity" : 123,
  "shipDate" : "2016-02-16T04:07:48.155+0000"
}, contentType=application/json}, {example=<Order>
  <id>123456</id>
  <petId>123456</petId>
  <quantity>0</quantity>
  <shipDate>2016-02-16T13:07:48.160Z</shipDate>
  <status>string</status>
  <complete>true</complete>
</Order>, contentType=application/xml}]
         
         - parameter body: (body) order placed for purchasing the pet

         - returns: RequestBuilder<Order> 
         */
        public class func placeOrder(body body: Order?) -> RequestBuilder<Order> {
            let path = "/store/order"
            let URLString = PetstoreClientAPI.basePath + path
            
            let parameters = body?.encodeToJSON() as? [String:AnyObject]

            let requestBuilder: RequestBuilder<Order>.Type = PetstoreClientAPI.requestBuilderFactory.getBuilder()

            return requestBuilder.init(method: "POST", URLString: URLString, parameters: parameters, isBody: true)
        }
```

wing328:
[...]

PR #2236, opened 2016-02-25
Fix 2156 - Use fixed date/time in operation.examples to avoid unnecessary diffs

hideya:
Fix https://github.com/swagger-api/swagger-codegen/issues/2156

I initially thought to keep using `Date` class but its automatic timezone handling was a bit annoying and decided to use just a plain text representation.  Hope it is OK.


wing328:
@hideya thanks for the PR. I'm fine with plain text representation. Will merge this PR later today.

Issue #2157, opened 2016-02-17
[Swift] Does <modelName>API class need to be an inner class of <projectName>API ?


hideya:
This is a discussion on request for improvement (not a bug).
For example for the Petstore demo, `PetAPI` class is generated as a inner class of `PetstoreClientAPI`:
```swift
extension PetstoreClientAPI {   
    public class PetAPI: APIBase {
```
Because of that, code to invoke a method of `PetstoreClientAPI` becomes a bit lengthy like the following:
```swift
        let newPet = Pet()
        ...
        PetstoreClientAPI.PetAPI.addPet(body: newPet).execute().then { response -> Void in
```
I'm wondering if this could just be `PetAPI.addPet(body: newPet)` without `PetstoreClientAPI.`?
Note that the model class `Pet` is not an inner class and can be accessed by just saying `Pet`. I don't see any strong reason for `PetAPI` being an inner class.  What would you folks think?

wing328:
Disclaimer: I'm no expert in Swift.

From what I read, Swift's `extensions` allows developers to:
- Add computed properties and computed type properties
- Define instance methods and type methods
- Provide new initializers
- Define subscripts
- Define and use new nested types
- Make an existing type conform to a protocol

I'm not sure if anyone is leveraging these features in the Swift SDK at the moment but personally I don't see drawback providing the flexibility at the moment (to meet the requirement of various use cases)

You do have a point that `PetstoreClientAPI` is a bit lengthy. Personally I also prefer something shorter. Did Apple have any recommended naming convention for `extensions`? (I couldn't find any...)

What about removing `API` from the extension? I believe `PetstoreClient` is also customizable via `{{projectName}}` (e.g. using `Petstore` instead of `PetstoreClient`).

Edubits:
We had a short discussion this earlier as part of #901, where @tkQubo explained he intended it as namespace. In my personal opinion it's not needed to use it, but it shouldn't hurt either (except for the longer method calls). When you currently make it for example a Cocoapods package, there is also a kind of namespace as it will be a separate library.

In Swift 3 a Package Manager will be introduced (see https://swift.org/package-manager/), this will make it officially possible to have the same benefits without the need of the extra verboseness.

hideya:
[...]

PR #2244, opened 2016-02-25
Fix #2120, #2157 - Update Swift API method generation for ease of use

### LINKED ISSUE #2120: [Swift] simplifications for the API method usage?
All the generated APIs return `RequestBuilder<T>` for further process.  I'm wondering what is the motivation for this, which requires extra typing?

I'm wondering why the usage couldn't just be simple like the following (something equivalent to Obj-C version)?
```swift
        CoffeeShopClientAPI.CoffeeShopAPI.coffeeShopCreate(data: coffeeShop) {
            output, error in
            // `output` holds the returned model as an instance of CoffeeShop
            ...
        }
```
instead of:
```swift
        CoffeeShopClientAPI.CoffeeShopAPI.coffeeShopCreate(data: coffeeShop).execute {
            response, error in
            let coffeeShop = response!.body
            ...
        }
```
Could someone share thoughts on this?  (@bajtos wanted to keep informed about this too)

hideya:
Fix https://github.com/swagger-api/swagger-codegen/issues/2120, https://github.com/swagger-api/swagger-codegen/issues/2157  -- Note this is a Breaking Change.

The following illustrates the differences in the usage of generated SDK:

Before:
```swift
        PetstoreClientAPI.PetAPI.getPetById(petId: 1000).execute { response, error -> Void in
                let pet = response?.body
```
After:
```swift
        PetAPI.getPetById(petId: 1000) { pet, error -> Void in
```

PromiseKit ready version of simplified methods are also generated:

Before:
```swift
        PetstoreClientAPI.PetAPI.getPetById(petId: 1000).execute().then { response -> Void in
                let pet = response.body
```
After:
```swift
        PetAPI.getPetById(petId: 1000).then { pet -> Void in
```

The `{{projectName}}API` namespace for API classes (`PetstoreClientAPI` above) now became optional and can be turned on via a CLI option: `-D swiftUseApiNamespace=true` 

In order to allow developers to access the response object when needed, the old methods still exist with `WithRequestBuilder` added to the end of the method name:

Before:
```swift
public class func getPetById(petId petId: Int) -> RequestBuilder<Pet> {
```

After:
```swift
public class func getPetByIdWithRequestBuilder(petId petId: Int) -> RequestBuilder<Pet> {
```

I initially thought to provide an improved abstraction over these `WithRequestBuilder` methods (to make them look closer to C#'s `WithHttpInfo` methods), but it will involve more lines to add (esp. to provide PromiseKit ready version).  I thought it better be done in a separate PR.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
@@ -37,10 +37,12 @@ public class SwiftCodegen extends DefaultCodegen implements CodegenConfig {
   public static final String POD_DESCRIPTION = "podDescription";
   public static final String POD_SCREENSHOTS = "podScreenshots";
   public static final String POD_DOCUMENTATION_URL = "podDocumentationURL";
+  public static final String SWIFT_USE_API_NAMESPACE = "swiftUseApiNamespace";
   protected static final String LIBRARY_PROMISE_KIT = "PromiseKit";
   protected static final String[] RESPONSE_LIBRARIES = { LIBRARY_PROMISE_KIT };
   protected String projectName = "SwaggerClient";
   protected boolean unwrapRequired;
+  protected boolean swiftUseApiNamespace;
   protected String[] responseAs = new String[0];
   protected String sourceFolder = "Classes" + File.separator + "Swaggers";
   private static final Pattern PATH_PARAM_PATTERN = Pattern.compile("\\{[a-zA-Z_]+\\}");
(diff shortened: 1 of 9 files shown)


wing328:
@hideya thanks for the PR. The description is very clear and the changes look good to me.

cc @jaz-ah @Edubits to see if they've any feedback.

hideya:
@wing328 I've updated this PR and included fix for https://github.com/swagger-api/swagger-codegen/issues/2157 (I've also updated the description above accordingly).  Would you please review the changes once again?

wing328:
@hideya looks good to me and that's how I would have done it.

PR #2271, opened 2016-02-29
Fix for clojure API filenames 

timbogit:
When the API name is more than one word (e.g., `UsersAPI` in a `users-service`), then the API namespace name will become `users-service.api.users-api`.

Prior to this fix, the API filename would include a dash, as it is solvely based on the API name (i.e., it currently is `users_service/api/users-api.clj`), which breaks clojure's conventions and leads to compilation errors.

With this fix, the API filename will be correctly converted to have underscores in place of dashes (i.e., the API file will be generated as `users_service/api/users_api.clj`).

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ClojureClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ClojureClientCodegen.java
@@ -161,6 +161,11 @@ public class ClojureClientCodegen extends DefaultCodegen implements CodegenConfi
     }
 
     @Override
+    public String toApiFilename(String name) {
+        return underscore(toApiName(name));
+    }
+
+    @Override
     public String toApiName(String name) {
         return dashize(name);
     }


xhh:
Thanks for the PR.
I tried and it works for `users api` and `UsersApi` by generating the API file named `users_api.clj`.
However, for `users-api` or `users_api`, `usersapi.clj` is generated. It would be better to generate `users_api.clj` as well.

@timbogit do you mind including the improvements in this PR?

timbogit:
Thanks for the comment / suggestion, @xxh!
I was mainly taking the approach to simply fix the "un-requirable" filename for a given API name by 'underscoring' it. I am not familiar enough with the code to actually know where the `tag` name that is fed into the [`apiFilename`](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L2184) originates, and how it can be made to match the `classname` that is used in the [`api.mustache` template](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/clojure/api.mustache#L1), which seems to be needed to implement the suggestion you made above.
I mainly took the 'inspiration' for my fix from the other language codegen classes (like [here](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/DartClientCodegen.java#L218).

If you prefer, I can simply use a standard `replaceAll()` call, like so:
```java
@Override
 public String toApiFilename(String name) {
     return toApiName(name).replaceAll("-", "_");
 }
```
 
Thoughts / pointers welcome!

wing328:
As discussed with @xhh we'll merge this fix and he might add more enhancement and test cases with another PR later.

Thanks @timbogit for the contribution.

xhh:
See #2286 for my PR on further improvements on api names for Clojure client.

timbogit:
Thanks for the merge, and :clap: for the PR that generalizes this for more use cases, and adds test coverage!

PR #2277, opened 2016-02-29
[Android] better local variable naming for Android client

xhh:
And fix the importing of Long and Byte in generated API classes.


wing328:
This also add support for model prefix/suffix (https://github.com/swagger-api/swagger-codegen/issues/1255)

PR #2284, opened 2016-03-01
[Java] Fix invalid imports when model name prefix/suffix is present

xhh:
When generating Java clients with mode name prefix/suffix given, there are invalid imports on `Date` and `File`, e.g. for the Petstore sample (with model name prefix set to "My" and suffix set to "Model"):

    import io.swagger.client.model.MyfileModel;
    import io.swagger.client.model.MyDateModel;

This PR fixes it to:

    import java.io.File;
    import java.util.Date;


wing328:
Related to #2211

Issue #2285, opened 2016-03-01
[Go] Generated client code has several Case / visibility issues


neilotoole:
I'm a Go n00b, but I suspect I've caught several issues with the generated Go code. These are all related to Camelization and the fact that package-visibility in Go is determined by the name being UpperCamelized.

- **Escaping:** For escaping reserved keywords, using `_Name` won't work, because underscore is not considered uppercase by Go (I think), therefore the escaped name will not be visible beyond the package. For example:

````
type Artifact struct { 
        Id  int   ...
        Name  string ...
        CreatedDate  string  ...
        _Type  string... // This is bad
        Links  links  ... // This is also bad (the type name is "Links")
        }
````
* **Lowercase typename:** In the last line above, `links` should be `Links`.
* **Clashing `UpperCamel` param and type names in functions:** There are cases where a function can have a param with the same (UpperCamel) name as its type in the function defintion, e.g. `func (a DefaultApi) UpdateBuild (BuildId int32, Build Build) (error) {` ... this is obviously confusing, but also can cause an error if the generated code tries to refer to the type, e.g. `response := new(Build)`... the compiler will try to pass the object to the `new` function, instead of passing the type. This can be solved by always making functions params `lowerCamel`, e.g. `func (a DefaultApi) UpdateBuild (BuildId int32, build Build) (error) {`

wing328:
@neilotoole thanks for the feedback.  I can help make some changes to improve the reserved keyword handling in Go. Would you have cycle to test the change when it's ready?

neilotoole:
@wing328 I have a PR ready to go in the morning :-)

wing328:
@neilotoole thanks for the contribution :beer:

wing328:
PR merged.

Issue #1948, opened 2016-01-22
[Go] Add unit testing for Go Petstore API client


wing328:
Currently, we only have [test.go](https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/go/test.go) to perform a simple HTTP request.

We want to add more tests with https://golang.org/pkg/testing/ to ensure the Go API client works without issues.

If you would like to contribute this task, please reply to let us know.

wing328:
To begin with, we want to test the following methods as a starting point:
- AddPet
- GetPetById
- UpdatePetWithForm

guohuang:
for these tests, do you want to dynamically generate it or you expect static test?

wing328:
Static test is a good starting point. Later we can add auto-generate test cases similar to PHP, C#, Perl: https://github.com/swagger-api/swagger-codegen/issues/2276

wing328:
Done via #2532 #2544 . Thanks @guohuang

PR #2289, opened 2016-03-01
Several minor fixes in the Go client code generator for Camelization of elements

neilotoole:
See issue #2285 for more details. This fix has been verified locally against a known good swagger file, and the generated client tested against just a handful of operations, so quality is **not** assured. From my quick look around there doesn't appear to be a Go test harness, I think that's something that we many need to work on. Apologies for any violations of style etc, feel free to edit as needed.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/GoClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/GoClientCodegen.java
@@ -44,7 +44,8 @@ public class GoClientCodegen extends DefaultCodegen implements CodegenConfig {
                 "case", "defer", "go", "map", "struct",
                 "chan", "else", "goto", "package", "switch",
                 "const", "fallthrough", "if", "range", "type",
-                "continue", "for", "import", "return", "var")
+                "continue", "for", "import", "return", "var", "error")
+                // Added "error" as it's used so frequently that it may as well be a keyword
         );
 
         defaultIncludes = new HashSet<String>(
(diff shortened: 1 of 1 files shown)


wing328:
@neilotoole yes, thank for the PR. Totally agree that we should have some test cases for Petstore GO API client.

I created a task for tracking here: https://github.com/swagger-api/swagger-codegen/issues/1948

If you've cycle, we would appreciate your contribution on that.

neilotoole:
@wing328 Hopefully I'll have some free cycles coming up in the next few weeks. Do you have a general strategy for dealing with test harness runtimes? (Apologies, I haven't yet had a good look around the rest of the codebase). Given that this test harness would require the installation of the Go runtime, perhaps it should be done in a container? Or any other thoughts?

wing328:
For testing, travis-ci does not support multiple languages so we've to install the lang/package manually if it's required by the test.

As an example, for ObjC/Swift, the tests are not run by the CI so currently I've to manually run the tests when reviewing changes related to ObjC/Swift.

neilotoole:
Hmmmn... well, in this era (so far! so soon!), any CI process worth its salt should allow one to kick off a container to execute tests? Concourse CI for example.

wing328:
I did reach out to other CI provider and hopefully they've better (and free) solution for our special case.

neilotoole:
Best of luck with that! Let me know if I can help. Containers are really the only way to go if you'd like to (ultimately) have full end-to-end test harnesses for each of the supported languages. It's effectively physically impossible to do it otherwise, and that's really what we should be aiming for.

wing328:
Totally agree. I'll share the update once I've more info.

Issue #2319, opened 2016-03-05
[Haskell-Servant] generator creates code that does not compile due to invalid data types


jhedev:
I used the master branch (commit: a4ec9b9) to generate the haskell-servant code for the API defined here: http://data.deutschebahn.com/apis/fahrplan/db-fahrplan-api-specification.yaml

The generated code does not compile due some problems with the generated record data types:

### Invalid names of record fields

For example: 

```yaml
LocationList:
    type: object
    required:
      - StopLocation
    properties:
      StopLocation:
        type: array
        items:
          $ref: '#/definitions/StopLocation'
```
produces code with upper case record field names:

```haskell
data LocationList = LocationList
    { StopLocation :: [StopLocation]
    } deriving (Show, Eq, Generic)
```

which is invalid code.

Another problem arises with the spec (is this actually allowed?):

```yaml
Note:
    type: object
    required:
      - key
      - priority
      - routeIdxFrom
      - routeIdxTo
      - $
    properties:
      key:
        type: string
      priority:
        type: integer
        format: int32
      routeIdxFrom:
        type: integer
        format: int32
      routeIdxTo:
        type: integer
        format: int32
      $:
        type: string
```

which produces

```haskell
data Note = Note
    { key :: String
    , priority :: Integer
    , routeIdxFrom :: Integer
    , routeIdxTo :: Integer
    , $ :: String
    } deriving (Show, Eq, Generic)
```

`$` is not a valid field name.

### Data types without fields

The spec 

```yaml
  LocalTime:
    type: string
  LocalDate:
    type: string
```

will generate empty data types, like this:

```haskell
data LocalTime = LocalTime
    { 
    } deriving (Show, Eq, Generic)
```

I'm not sure if this is really a problem with the code generator or if the specification is the problem. Maybe someone else can help me.

wing328:
cc @algas

algas:
@jhedev '$' causes the trouble. In Haskell, you cannot use '$' in a function or a variable name.

jhedev:
@algas Yeah, I know. You can also not start a field name with an upper case letter :)

What I wanted to point out is that this could maybe be handled during generation, i.e. make sure the field names are valid haskell code. So the generated code can be compiled.

wing328:
@jhedev may I know if you've cycle to contribute a fix?

[...]

PR #2322, opened 2016-03-07
Improve Haskell Code Generation

gibiansky:
I have created a new code generator for Haskell (based on the older `haskell-servant` code generator), and I'd like to propose replacing the `haskell-servant` with this newer `haskell` generator. The reason for the name change is that while the Servant library is used on the backend, the interface to the generated code is almost completely agnostic and requires very little `servant` code.

Main changes:

- Instead of exposing the Servant routes directly and then having to generate clients and servers, the `.API` module exposes a record of functions called `BlahBackend`. Then, generated `runBlahServer` and `createBlahClient` can be used to run either a server (serving the API) or a client (to query the API). This is a cleaner API and requires less knowledge of servant to use.
- Renamed to match Haskell best practices (e.g. `Models` renamed to `Types`, since having `.Types` modules is common and well-understood). Added Haddock documentation in as many places as possible.
- Removed generation of extraneous code. For example, it is inappropriate to have the swagger code generation generate a `LICENSE` file for you; in addition, the generated code is not copyrighted by the author of the code generator. In addition, instead of generating stubs for client and server, demo code is simply put in `README.md`; this means that this Swagger codegen could hypothetically be part of an automatic build process, since it does not generate invalid and incomplete stub code but only a complete library API.
- Replaced usage of types such as `String` and `Integer` with `Text` and `Int`, which are much, much more efficient; `String` should approximately never be used and `Integer` should be used very sparingly (almost never).
- A variety of other small usability improvements.

I would appreciate a few things to hopefully get this merged soon:

1. Examples of Swagger specifications that I can test this on. I would like to make sure that any valid Swagger spec a) generates code that compiles with no changes after `stack build`, b) supports all the features that Swagger does and c) is as intuitive and logical as possible.
2. Guidance on what the right way to merge this is. I do not know whether this should be added in *addition* to `haskell-servant` or as a replacement; from my point of view, this code gen works the way the Haskell code gen *should*, but given that `haskell-servant` is somebody elses work perhaps I need to discuss this further with them. I do not want to step on anyone's toes here, so just let me know what the right procedure here is!


jhedev:
I checked out your branch and tested it as follows:

```
mvn package
java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -i http://data.deutschebahn.com/apis/fahrplan/db-fahrplan-api-specification.yaml -l haskell -o ~/tmp/haskell_new
```

Unfortunately I get a `NullpointerException`:

```
reading from http://data.deutschebahn.com/apis/fahrplan/db-fahrplan-api-specification.yaml
[main] WARN io.swagger.codegen.DefaultCodegen - Empty operationId found for path: get /arrivalBoard. Renamed to auto-generated operationId: arrivalBoardGet
io.swagger.codegen.CodegenParameter@76b1e9b8
Exception in thread "main" java.lang.RuntimeException: Could not process operation:
  Tag: default
  Operation: null
  Resource: get /arrivalBoard
  Definitions: {LocationResponse=io.swagger.models.ModelImpl@65f0ddd9, LocationList=io.swagger.models.ModelImpl@45c6d341, StopLocation=io.swagger.models.ModelImpl@7f868261, DepartureBoardResponse=io.swagger.models.ModelImpl@60f0cabc, DepartureBoard=io.swagger.models.ModelImpl@fafed024, ArrivalBoardResponse=io.swagger.models.ModelImpl@11e2feb1, ArrivalBoard=io.swagger.models.ModelImpl@67db2b07, DepartureOrArrival=io.swagger.models.ModelImpl@eaa9d3d4, JourneyDetailRef=io.swagger.models.ModelImpl@f24ce6f4, JourneyDetailResponse=io.swagger.models.ModelImpl@b14fcd97, JourneyDetail=io.swagger.models.ModelImpl@6ff7abbe, Stops=io.swagger.models.ModelImpl@fce7baa5, Stop=io.swagger.models.ModelImpl@323cad5d, Names=io.swagger.models.ModelImpl@af09dc2c, Name=io.swagger.models.ModelImpl@47a4047f, Types=io.swagger.models.ModelImpl@fd9f7ebd, Type=io.swagger.models.ModelImpl@7288b432, Operators=io.swagger.models.ModelImpl@c7c00637, Operator=io.swagger.models.ModelImpl@48db4816, Notes=io.swagger.models.ModelImpl@b8d74495, Note=io.swagger.models.ModelImpl@769ea71d, LocalTime=io.swagger.models.ModelImpl@87c52206, LocalDate=io.swagger.models.ModelImpl@e1419e65}
  Exception: null
	at io.swagger.codegen.DefaultGenerator.processOperation(DefaultGenerator.java:675)
	at io.swagger.codegen.DefaultGenerator.processPaths(DefaultGenerator.java:567)
	at io.swagger.codegen.DefaultGenerator.generate(DefaultGenerator.java:278)
	at io.swagger.codegen.cmd.Generate.run(Generate.java:195)
	at io.swagger.codegen.SwaggerCodegen.main(SwaggerCodegen.java:36)
Caused by: java.lang.NullPointerException
	at io.swagger.codegen.languages.HaskellServantCodegen.fromOperation(HaskellServantCodegen.java:341)
	at io.swagger.codegen.DefaultGenerator.processOperation(DefaultGenerator.java:623)
	... 4 more
```

With the petstore example I get something similar:
```
  -i http://petstore.swagger.io/v2/swagger.json -l haskell -o ~/tmp/petstore_haskell
reading from http://petstore.swagger.io/v2/swagger.json
io.swagger.codegen.CodegenParameter@647e447
Exception in thread "main" java.lang.RuntimeException: Could not process operation:
  Tag: pet
  Operation: findPetsByStatus
  Resource: get /pet/findByStatus
  Definitions: {Order=io.swagger.models.ModelImpl@df474bd0, User=io.swagger.models.ModelImpl@26e6d282, Category=io.swagger.models.ModelImpl@5d7f7f63, Tag=io.swagger.models.ModelImpl@a3a44263, ApiResponse=io.swagger.models.ModelImpl@e2fa6505, Pet=io.swagger.models.ModelImpl@6dcc67cf}
  Exception: null
	at io.swagger.codegen.DefaultGenerator.processOperation(DefaultGenerator.java:675)
	at io.swagger.codegen.DefaultGenerator.processPaths(DefaultGenerator.java:567)
	at io.swagger.codegen.DefaultGenerator.generate(DefaultGenerator.java:278)
	at io.swagger.codegen.cmd.Generate.run(Generate.java:195)
	at io.swagger.codegen.SwaggerCodegen.main(SwaggerCodegen.java:36)
Caused by: sun.reflect.generics.reflectiveObjects.NotImplementedException
	at io.swagger.codegen.languages.HaskellServantCodegen.makeQueryListType(HaskellServantCodegen.java:399)
	at io.swagger.codegen.languages.HaskellServantCodegen.fromOperation(HaskellServantCodegen.java:342)
	at io.swagger.codegen.DefaultGenerator.processOperation(DefaultGenerator.java:623)
	... 4 more
```

Am I doing something wrong?

wing328:
@algas please kindly review when you've time.

gibiansky:
@jhedev Thanks! I'll test with those two and fix any issues. Will update this thread when they are fixed.

gibiansky:
@algas Since you are the original author of the `haskell-servant` backend, I'd like to know what you think of the changed interface. One thing that has changed is that it no longer includes QuickCheck instances; I think it might be nice to separate those somehow to avoid forcing a dependence on quickcheck for clients or servers that do not want it. Perhaps it could generate a separate `test` directory and use orphans there.

gibiansky:
@jhedev I've pushed a fix to both of those Swagger specs; both now generate a `cabal` package that compiles with `stack build` immediately. Special characters such as `$` get translated into their literal conterparts, e.g. `note$` becomes `noteDollar`. Not clean but best I could think of. Maybe would be better to just translate them all into `'` or `_`?

The issue with the petstore example was handling of form and header parameters, which is also now fixed.

jhedev:
@gibiansky Awesome! Looks good so far. Your ideas of handling the special characters correspond to my thoughts when I tried to fix #2319 . I also used "Dollar" at first, but I would rather prefer something like `_` (That's just my opinion, though).

However, I think there will be a problem with the `FromJSON/ToJSON` instances you generate. Because you use the `genericParseJSON` which just drops the prefix, I think you will have a problem every time you replace a special character like `$`. For example, in case of the `Note` type the parser would expect a JSON object with key `dollar` instead of `$`. 

Correct me if I'm wrong. But this is where I got stuck while trying to fix #2319.

jhedev:
Another issue (?) I noticed:

The given "DBFahrplan" spec contains the following definition:

```yaml
  LocalTime:
    type: string
  LocalDate:
    type: string
```

I'm not sure if this is a valid spec but I would interpret it as `LocalTime` and `LocalDate` being type aliases for `string`. However, the generated code is:

```haskell
data LocalTime = LocalTime
    { 
    } deriving (Show, Eq, Generic)

instance FromJSON LocalTime where
  parseJSON  = genericParseJSON  (removeFieldLabelPrefix "localTime")
instance ToJSON LocalTime where
  toJSON     = genericToJSON     (removeFieldLabelPrefix "localTime")
```

algas:
@gibiansky Thanks. I used QuickCheck in order to generate random values as server response. I worry little about it.

gibiansky:
@jhedev Thanks for the comments. It sounds like I have a little bit more work to do to fix those two issues. I'll try to push a fix soon.

gibiansky:
@jhedev I've pushed two fixes to the issues you raised:

- `$` now gets translated to `'Dollar` instead of just `Dollar`, making it easier to distinguish. Then, `removeFieldLabelPrefix` will either convert `$` to `'Dollar` or `'Dollar` to `$` when working with the JSON. (NOTE: I know this generates the code the way I intend, but I have not tested that it actually works at runtime as I intend.)
- The `type: string` data types will now generate `newtype LocalTime = LocalTime Text` instead of an empty field.

gibiansky:
Also, we can't just translate `$` into `_` since we have to be able to translate it back, which requires something like `Dollar`. The choice of putting `'` in front of it is to avoid mangling people's names when they _intend_ to have something like `amountInDollars`. It is unlikely people would write `amountIn'Dollars` because that is not a valid identifier in pretty much any language except haskell.

jhedev:
@gibiansky I fully agree with your points. Sounds like a good solution.


As you said you did not yet test the JSON encoding/decoding at runtime. A quick manual test showed the following:

```
$ stack ghci
Configuring GHCi with the following packages: db-fahrplan
GHCi, version 7.10.3: http://www.haskell.org/ghc/  :? for help
[1 of 2] Compiling DBFahrplan.Types ( /Users/joel/tmp/haskell_new/lib/DBFahrplan/Types.hs, interpreted )
[2 of 2] Compiling DBFahrplan.API   ( /Users/joel/tmp/haskell_new/lib/DBFahrplan/API.hs, interpreted )

Ok, modules loaded: DBFahrplan.API, DBFahrplan.Types.
λ> import Data.Aeson
λ> :set -XOverloadedStrings
λ> encode $ Note "key" 0 0 0 "dollar"
"{\"RouteIdxTo\":0,\"Priority\":0,\"Dollar\":\"dollar\",\"Key\":\"key\",\"RouteIdxFrom\":0}"
λ>
```

I would expect the output to be, i.e. the `Dollar` field replaced by `$`:

```
"{\"RouteIdxTo\":0,\"Priority\":0,\"$\":\"dollar\",\"Key\":\"key\",\"RouteIdxFrom\":0}"
```

Same goes for the decoding.

jhedev:
Oh it seems this is caused by `Note` still being:

````
data Note = Note
    { noteKey :: Text -- ^
    , notePriority :: Int -- ^
    , noteRouteIdxFrom :: Int -- ^
    , noteRouteIdxTo :: Int -- ^
    , noteDollar :: Text -- ^
    } deriving (Show, Eq, Generic)
```

instead of

```
data Note = Note
    { noteKey :: Text -- ^
    , notePriority :: Int -- ^
    , noteRouteIdxFrom :: Int -- ^
    , noteRouteIdxTo :: Int -- ^
    , note'Dollar :: Text -- ^
    } deriving (Show, Eq, Generic)
```

After I added the `'` it works as expected :)

jhedev:
@gibiansky If you want to do some more testing it might be worth looking at the swagger spec of kubernetes: https://github.com/kubernetes/kubernetes/tree/master/api/swagger-spec I guess that's quite large and comprehensive and should make a good test.

gibiansky:
@jhedev Ah, thanks. I did not fix the generation of the data types. Will do that this evening. Thanks for testing :)

I will try to make a kubernetes API later today and see how it goes.

wing328:
@gibiansky for your next commit, please kindly rebase on the latest master to resolve the merge conflicts.

gibiansky:
Will do! Planning on picking this up this weekend.

wing328:
Thanks. :+1: If you need help with the rebase, let me know.

gibiansky:
Finally picking this back up, much later than anticipated...

Changes:

- Fixed handling of special characters such as '$'
- Rebased onto latest master
- Updated to latest Stackage LTS snapshot (does not yet use Servant 0.5 though)

@jhedev Tried to test on kubernetes but they seem to mostly have Swagger 1.2 specs, not 2.0. Do these generators have to be backwards compatible with older versions of Swagger?

@wing328 Any pointers for next steps?

wing328:
@gibiansky the code generator should work for swagger spec 1.2. 

I'll take a look tomorrow (wed) and let you know if I've any question.

gibiansky:
@wing328 Got it, thanks. I have fixed it to work on the Kubernetes 1.2 API spec and pushed my latest changes. So far, besides just some manual testing, it's been tested using the Kubernetes spec, the Docker spec, and the Fahrplan spec provided by @jhedev

wing328:
@algas I wonder if you can review when you've time. Thanks!

wing328:
@gibiansky I noticed that you've removed the apache license. 

For license on the generated code, please join the discussion at https://groups.google.com/forum/#!topic/swagger-swaggersocket/6JcsR4mWpYk

For the time being, I suggest we keep it.

wing328:
If no further feedback from anyone, I'll merge this PR on coming weekend.

wing328:
PR merged. Thanks @gibiansky for the contribution

PR #2325, opened 2016-03-07
Add scalar converter to retrofit2

### LINKED ISSUE #2177: [Retrofit2] Use multiple converters and remove the custom gson converter
Retrofit 2 can use multiple converters so the custom gson converter is not needed anymore.
The adapter could be built with
```java
adapterBuilder = new Retrofit
               .Builder()
               .baseUrl(baseUrl)
               .client(okClient)
               .addConverterFactory(ScalarsConverterFactory.create())
               .addConverterFactory(GsonConverterFactory.create(gson));
```
This has the advantage of working for both request and response, and also to deal with all primitive types (not only String).

cbornet:
Fix #2177


cbornet:
There seem to be random CI failures with Javascript client test.

wing328:
Thanks for the PR. Let me take a look

Issue #1847, opened 2016-01-08
sort libraries in swagger-generator


fehguy:
currently the list of supported targets feels randomly ordered.  Should sort them alpha

PR #2326, opened 2016-03-07
Keep sort order of APIs

JSurf:
When generating static documentation, the sort order is important. This pull request keeps the sort order from swagger.getPaths()

Might be related to #1847


wing328:
@JSurf thanks for the PR. As shown in the "Commits" tab, the commit is not linked to your account, which means the contribution won't count toward your github account.

JSurf:
Thanks, changed commit author

JSurf:
Actually to keep the original order, linkedhashmap should bei used. treemap will sort alphabetically.What do you prefer?

wing328:
@JSurf I've merged #2327, which sorts both API methods and models. Would that meet your requirement?

Please pull the latest and give it a try.

JSurf:
#2327 should do the trick
Closed because #2327,is merged

PR #2348, opened 2016-03-09
CSharpClientCodegen: Members with same name as the class will not be renamed.

nrekretep:
super.postProcessModels needs to be called. Otherwise renaming of members with same name as the class does not take place. Fixed it.


wing328:
@nrekretep thanks for the PR. 

Can you elaborate a bit on what `DATA_TYPE_WITH_ENUM_EXTENSION` does? Can you share a spec using this vendor extension?

nrekretep:
This vendor extension has already been inside the code as a string literal. I extracted it into a constant.
As far as I have seen the code it represents the name of the enum class without the question mark. The extension is used to make this enum class available inside the enumClass.template and to represent default values of enum variables in a correct way.

Normally you use the "datatypeWithEnum" attribute of the var. But this client code generator defines all enum variables as nullables. Because of this the value of "datatypeWithEnum" contains a trailing question mark.

The following image shows an example where "defaultWithEnum" contains the questionmark and the default value is set correctly because of the extension.

enumvars-default-value

wing328:
@nrekretep thanks for the details.

I'll add a fake model in petstore.json to catch the issue moving forward

PR #2351, opened 2016-03-10
non default java clients wont compile if description is empty in API json.

ieb:
If the API definition contains something like

```
   "integer": {
    "id": "integer",
    "properties": {}
   },
```

with no description property, the generated code for all no default libraries will add a ApiModel annotation but no import. 

pull request appears to fix the compile time issue.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/model.mustache
@@ -13,7 +13,7 @@ import com.google.gson.annotations.SerializedName;
 /**
  * {{description}}
  **/{{/description}}
-@ApiModel(description = "{{{description}}}")
+{{#description}}@ApiModel(description = "{{{description}}}"){{/description}}
 public class {{classname}} {{#parent}}extends {{{parent}}}{{/parent}} {{#serializableModel}}implements Serializable{{/serializableModel}} {
   {{#vars}}{{#isEnum}}
 
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/retrofit/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/retrofit/model.mustache
@@ -13,7 +13,7 @@ import com.google.gson.annotations.SerializedName;
 /**
  * {{description}}
  **/{{/description}}
-@ApiModel(description = "{{{description}}}")
+{{#description}}@ApiModel(description = "{{{description}}}"){{/description}}
 public class {{classname}} {{#parent}}extends {{{parent}}}{{/parent}} {{#serializableModel}}implements Serializable{{/serializableModel}} {
   {{#vars}}{{#isEnum}}
 
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/retrofit2/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/retrofit2/model.mustache
@@ -13,7 +13,7 @@ import com.google.gson.annotations.SerializedName;
 /**
  * {{description}}
  **/{{/description}}
-@ApiModel(description = "{{{description}}}")
+{{#description}}@ApiModel(description = "{{{description}}}"){{/description}}
 public class {{classname}} {{#parent}}extends {{{parent}}}{{/parent}} {{#serializableModel}}implements Serializable{{/serializableModel}} {
   {{#vars}}{{#isEnum}}
 


wing328:
@ieb thanks for the PR. In the current Petstore sample, we do have empty description for ApiModel: https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/java/okhttp-gson/src/main/java/io/swagger/client/model/Pet.java#L16

and both Java1.7 and Java1.8 test it without issue as part of the CI (travis).

If you manually run `mvn test` in the folder `samples/client/petstore/java/okhttp-gson`, do you get any error?

ieb:
Strange, but somehow you also have an ApiModel import, https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/java/okhttp-gson/src/main/java/io/swagger/client/model/Pet.java#L4

and the Pet model appears to have a description.
https://github.com/swagger-api/swagger-codegen/blob/master/samples/yaml/pet.yml#L147



I don't get errors for mvn test in samples/client/petstore/java/okhttp-gson with or without the patch.

Using https://github.com/kubernetes/kubernetes/blob/master/api/swagger-spec/v1.json

generating with 

```
java -jar ../swagger-codegen/modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -l java --library okhttp-gson -i v1.json -o master
```

On master I get
```
x43543:master ieb$ more src/main/java/io/swagger/client/model/V1PersistentVolumeAccessMode.java
package io.swagger.client.model;

import java.util.Objects;

import com.google.gson.annotations.SerializedName;




@ApiModel(description = "")
public class V1PersistentVolumeAccessMode   {

```

With the patch I get 

```
x43543:20160310 ieb$ more src/main/java/io/swagger/client/model/V1PersistentVolumeAccessMode.java
package io.swagger.client.model;

import java.util.Objects;

import com.google.gson.annotations.SerializedName;





public class V1PersistentVolumeAccessMode   {
  

```

The spec location is

https://github.com/kubernetes/kubernetes/blob/master/api/swagger-spec/v1.json#L15882

The behaviour was introduced with https://github.com/swagger-api/swagger-codegen/pull/2342/files#diff-79d6f9708f6ea5ba956fcbc53919c452R528 which is Ok with the Jersey templates but not others.


BTW, I think the Kubernetes spec is going through some changes, so it may not be 100% spec compliant. I have a local copy of v1.1.8 that is patched to make it work with some of the other issues, but this one looked like it was a bug in the templates.

wing328:
The Java Petstore sample does not use petstore.yml. It's generated from https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.json#L1191 instead (which does not have a description for Pet)

I'll try to repeat the issue using kubernets spec and `okhttp-gson` template.

ieb:
Would

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.json#L1230

trigger 

import io.swagger.annotations.ApiModel;

wing328:
@ieb not sure. We'll need to dig into the code to find out.

fehguy:
@ieb yes, this was fixed in https://github.com/swagger-api/swagger-codegen/pull/2342

fehguy:
@wing328 this pr makes sense, should be applied to other java templates

wing328:
@fehguy thanks to the PR from @ieb, other Java model templates are also updated:
```
Java|java_description⚡ ⇒ grep -R @ApiModel *
libraries/okhttp-gson/model.mustache:{{#description}}@ApiModel(description = "{{{description}}}"){{/description}}
libraries/okhttp-gson/model.mustache:  @ApiModelProperty({{#required}}required = {{required}}, {{/required}}value = "{{{description}}}")
libraries/retrofit/model.mustache:{{#description}}@ApiModel(description = "{{{description}}}"){{/description}}
libraries/retrofit/model.mustache:  @ApiModelProperty({{#required}}required = {{required}}, {{/required}}value = "{{{description}}}")
libraries/retrofit2/model.mustache:{{#description}}@ApiModel(description = "{{{description}}}"){{/description}}
libraries/retrofit2/model.mustache:  @ApiModelProperty({{#required}}required = {{required}}, {{/required}}value = "{{{description}}}")
pojo.mustache:{{#description}}@ApiModel(description = "{{{description}}}"){{/description}}
pojo.mustache:  @ApiModelProperty({{#example}}example = "{{example}}", {{/example}}{{#required}}required = {{required}}, {{/required}}value = "{{{description}}}")
```

PR #2366, opened 2016-03-12
Add git_push.sh and .gitignore to client generators

wing328:
Better integration with Github with git_push shell script and .gitignore file


ktunkiewicz:
@wing328 seriously? A bash script to use gihtub? And this is a breaking change for people using their own templates since there is no system property to avoid generation of this file.

We started using swagger-codegen in our company but watching the development process of this library I'll probably freeze the version that we use to 2.1.6 and bugfix things myself... The (breaking) (nonsense) feature added to milestone, PRed and merged, all in one day... Do anybody even discuss changes to that project? Why the hell the bash script to use github? It looks like a toy for beginner programmers who don't know how to use gihtub.

ktunkiewicz:
Sorry for being a bit too agressive with that comment, I just got confused how many things broke down when I updated the swagger-codegen (i'm using customer templates so things usually break if something is changed in templates) and reacted too emotional.
The project is great anyways. Very helpful and working really good. Great job.
But you really should more carefully choose what to change/add to it since if somebody wants to use if for some serious project he will get confused with changes like this one...

wing328:
@ktunkiewicz personally I welcome both positive and negative feedback.

Did you have a chance to read the [release note](https://github.com/swagger-api/swagger-codegen/releases/tag/v2.1.6)? Does it help making the migration smoother?

The git_push is to address a pain point we'd in uploading the auto-generated SDKs to Github for testing (e.g. import/install directly via git, CI integration, etc). I agree that it won't fit into every single workflow but we got to start with "something" to begin with and improve it based on the feedback from the community.

Issue #2367, opened 2016-03-12
[C#] Requests should set User-Agent


jqr:
Revision: 62fbcb9e8bec4a40569e51c7f316af85806628ee

Most of the generated clients set a User-Agent to something like `ruby-swagger-1.2.3`. But the Csharp client does not.

The User-Agent string is really handy for server side debugging of client issues as well as alerting customers to new versions or issues with their deployed version.

You can see how we're currently fixing this in the generated clients in https://github.com/DocRaptor/docraptor-csharp/commit/245cdbc4f068b08405047c670ef93f972727cbdd

At the risk of overloading this issue, I'd love to see a standard way for specifying this version string in the `language-config.son` so that we could have consistent use across languages. Has that been discussed?

wing328:
@jqr agree with you that it's better to make the user agent configurable. I'll spend some time tomorrow (Sunday) to implement it for C# to begin with if it's not very urgent.

jqr:
Not urgent at all. Thanks for your attention!

wing328:
Fixed by https://github.com/swagger-api/swagger-codegen/pull/2376.

Please pull the latest and let us know if you still find issue setting user agent for C# API client.

PR #2371, opened 2016-03-13
[C#] add httpUserAgent option, add configurable user-agent support to C#

wing328:
Tested on both Windows and Mac. All test cases passed.

We'll roll out the feature (configurable user-agent) to other language one by one.

The feature is broken at the moment due to RestSharp: https://github.com/restsharp/RestSharp/issues/787

(for https://github.com/swagger-api/swagger-codegen/issues/2367)


wing328:
I was wrong about setting user agent in RestSharp, which can be done via `RestClient.UserAgent`.

I've submitted https://github.com/swagger-api/swagger-codegen/pull/2376 to fix that.

PR #2376, opened 2016-03-14
Update User-Agent for Ruby, Python, PHP, Java, Android

wing328:
Make the User-Agent configurable via `--http-user-agent` CLI option.

Default to `Swagger-Codegen/{version}/{language}`


jqr:
I'm curious why this was made an option to generate instead of something in the `config.json`. Is there a general move towards options for new things?

wing328:
You can also supply the CLI option in the config, e.g.
```
{
  "sortParamsByRequiredFlag": true,
  "optionalAssemblyInfo": true,
  "httpUserAgent": "XXXXXX"
}
```
I did a test and worked for me. Can you give that a try?

jqr:
Ah, that makes sense. Thanks!

inska:
Hi @wing328,

When can I do like this on the Swift Client?

Thanks,

wing328:
@inska I don't think anyone from the community has started working on it. I'll create a separate task for tracking to implement customisable user-agent in other langauges.

wing328:
Now track via https://github.com/swagger-api/swagger-codegen/issues/2715

PR #2378, opened 2016-03-14
Handle empty date-time gracefully

arnested:
Some API's return an invalid, empty string as a date-time property.

DateTime::__construct() will return the current time for empty input which is probably not what is meant.

The invalid empty string is probably to be interpreted as a missing field/value.

Let's handle this graceful.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/php/ObjectSerializer.mustache
+++ b/modules/swagger-codegen/src/main/resources/php/ObjectSerializer.mustache
@@ -245,7 +245,17 @@ class ObjectSerializer
             settype($data, 'array');
             $deserialized = $data;
         } elseif ($class === '\DateTime') {
-            $deserialized = new \DateTime($data);
+            // Some API's return an invalid, empty string as a
+            // date-time property. DateTime::__construct() will return
+            // the current time for empty input which is probably not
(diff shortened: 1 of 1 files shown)


modules/swagger-codegen/src/main/resources/php/ObjectSerializer.mustache
```diff
@@ -245,7 +245,17 @@ class ObjectSerializer
...
+            // Some API's return an invalid, empty string as a
+            // date-time property. DateTime::__construct() will return
+            // the current time for empty input which is probably not
+            // what is meant. The invalid empty string is probably to
+            // be interpreted as a missing field/value. Let's handle
+            // this graceful.
+            if (!empty($data)) {
+                $deserialized = new \DateTime($data);
+            } else {
+                $deserialized = false;
```
wing328: Instead of setting it to `false`, what about setting it to `NULL` (which means the variable has not been initialised)?

Ref: http://stackoverflow.com/a/138954/677735

arnested: Yes, that would be better! Pushing new version in 2 seconds.

Issue #2258, opened 2016-02-26
[JavaScript] Generator options and template improvements


demonfiddler:
This is to propose the following changes to the JavaScript code generator:
1. RENAME omitModelMethods [=false] to emitModelMethods [=false]. The rationale is two-fold: firstly to suppress their emission by default, since getter and setter methods are not a typical JavaScript programming idiom; their presence bloats code and adds little or no value. _Even_ if it were felt desirable to access properties via a function call, ECMA 5.1 provides for transparent property get set accessor functions. Secondly, the new option name would be consistent with option 2 below.
2. ADD emitJSDoc [=true] so that JSDoc comments are emitted by default but disabled if desired. The JSDoc tags and format to be carefully chosen to conform to the requirements of both JSDoc (for API documentation generation) and Tern.js (to support content assist and validation).
3. ADD useInheritance [=true]. By default generated classes for 'allOf' composed types would use idiomatic JavaScript inheritance via prototype chains for the 'parent' axis and mix in any additional types from the 'interfaces' axis. When set to false, all composed types would be mixed in as at present.
4. REMOVE the toJson() [sic] method from the template, as it bloats code, serves no useful purpose and causes confusion with the toJSON() method recognised by JSON.stringify(). See #2075, #2044.
5. MOVE model methods to the prototype. See #2044.
6. DEFINE prototype properties for those owned properties with default values, rather than adding them to the instance as at present. Adding such properties to the instance causes their values to be serialised to JSON, thus defeating the purpose of default values and furthermore making it impossible to restore the default value without _a priori_ knowledge of the default value! Defining the default property value on the prototype allows the default to be restored by simply deleting the instance property.

I have most of the above implemented and when complete I should like to submit a PR. The changes will include unit and functional tests as appropriate.

Comments invited.

wing328:
@demonfiddler thanks for the suggestion. 

cc @delenius @xhh to see if they've any comment.

xhh:
These suggestions look very good to me :+1:

delenius:
(1) and (2) sound good to me.

I don't know what 'parent' vs 'interface' axis means... Does this take care of https://github.com/swagger-api/swagger-codegen/issues/2041

@demonfiddler, @xhh In the interest of cleanup and de-bloating, please also comment on https://github.com/swagger-api/swagger-codegen/pull/2075

demonfiddler:
@delenius thanks for your response.

[...]

PR #2396, opened 2016-03-16
[JavaScript] Generator options and template improvements

demonfiddler:
Please review and merge the fix and tests for Issue #2258 "[JavaScript] Generator options and template improvements". Many thanks.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenModel.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenModel.java
@@ -6,20 +6,35 @@ import java.util.*;
 
 public class CodegenModel {
     public String parent, parentSchema;
+    public List<String> interfaces;
+
+    // References to parent and interface CodegenModels. Only set when code generator supports inheritance.
+    public CodegenModel parentModel;
+    public List<CodegenModel> interfaceModels;
+
     public String name, classname, description, classVarName, modelJson, dataType;
     public String classFilename; // store the class file name, mainly used for import
     public String unescapedDescription;
     public String discriminator;
     public String defaultValue;
     public List<CodegenProperty> vars = new ArrayList<CodegenProperty>();
+    public List<CodegenProperty> allVars;
     public List<String> allowableValues;
 
-    // list of all required parameters
-    public Set<String> mandatory = new HashSet<String>();
-    
+    // Sorted sets of required parameters.
+    public Set<String> mandatory = new TreeSet<String>();
+    public Set<String> allMandatory;
+
     public Set<String> imports = new TreeSet<String>();
     public Boolean hasVars, emptyVars, hasMoreModels, hasEnums, isEnum;
     public ExternalDocs externalDocs;
 
     public Map<String, Object> vendorExtensions;
+
+    {
+        // By default these are the same collections. Where the code generator supports inheritance, composed models
+        // store the complete closure of owned and inherited properties in allVars and allMandatory.
+        allVars = vars;
+        allMandatory = mandatory;
+    }
 }
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -45,9 +45,12 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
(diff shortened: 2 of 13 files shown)


wing328:
@demonfiddler thanks for the PR. Your 2nd and 3rd commits are not linked to your Github account, which means these contributions won't count toward https://github.com/swagger-api/swagger-codegen/graphs/contributors as your contribution.

wing328:
cc @delenius @xhh

demonfiddler:
@wing328 thanks for the heads-up regarding my 2nd and 3rd commits. I'm still on the Git/GihHub learning curve and hadn't realised the significance of the commit user name WRT attributions. Actually it's only the 3rd commit that contains my changes - the first two were just pulls from remotes. Is there anything could be done to rectify the contributor attribution?

Also, I spent all day yesterday pulling and rebasing to avoid conflicts, and when I raised this PR it could have been merged without conflict. Now it would seem that someone else has made a conflicting commit. I'm not sure how to proceed...

wing328:
@demonfiddler for the rebase, you can try to do it or I can do it for you when I merge later.

About fixing the author for commits, would this help? https://gist.github.com/trey/9588090

demonfiddler:
Thanks for the hints @wing328, I figured out how to do both the rebase and the user name amendment and forced a re-push. Lools like the CI build is now running.

wing328:
:+1:

wing328:
@demonfiddler can you please fix the unit test for JS Petstore API client?

When I run the unit test as following, I got 15 errors:
```
mvn clean && ./bin/javascript-petstore.sh
cd samples/client/petstore/javascript
mvn integration-test
```

After fixing the test cases for `javascript`, please do the same for `javascript-promise` ( samples/client/petstore/javascript-promise)

If you do not have cycle to update the test cases, then we might need to roll back the change for the time being.

(and sorry it was my oversight..)

demonfiddler:
@wing328 - my sincerest apologies, to be honest I had noticed that there were some JavaScript samples but they seemed to contain source-controlled generated API files - not what I'd expected for an integration test, so I left them alone.

Leave this with me, I'll sort it out.

wing328:
@demonfiddler no need to apologize. It's my responsibility to ensure the test cases for Petstore pass before merging and my manual tests didn't cover that (so it's my fault).

Please take your time in updating the test cases.

demonfiddler:
This will take a while, turns out the ITs won't run on Windows. I've had to set up a dev env in an Ubuntu VM but mvn integration-test (actually npm test) fails 'cos it can't find the node executable (it's nodejs in Debian distros). Anyhow, sorted, now making some progress...

delenius:
Please also update and run the "promise" version of the integration tests:

```javascript
mvn clean && ./bin/javascript-promise-petstore.sh
cd samples/client/petstore-promise/javascript
mvn integration-test
```

(this stuff should really be automated)

demonfiddler:
Integration test fixes submitted as Pull Request #2411.

wing328:
@demonfiddler thanks. Tested locally and merged into master.

cbornet:
@demonfiddler I'm doing some fixes and unit tests on inheritance/composition. Can you explain the difference between "vars" and "allVars" ? They seem to do exactly the same thing (except that you need to use one or the other depending on ```supportsInheritance``` which makes a lot of duplicate code in the template (not so much for javascript it seems but for java this is huge))

demonfiddler:
Hello Christophe,

 

vars and allVars are identical if the generator does not support inheritance. If the generator does support inheritance, vars lists those fields owned by the class itself whereas allVars is the complete set of fields both owned and inherited (superclass first). You’ll see that in the JavaScript templates both are used – allVars being used in conjunction with the ‘required’ property to determine the constructor parameters. The JavaScript generator can be configured to support inheritance or not. It implements inheritance via idiomatic JavaScript prototype chains and non-inheritance by declaring one property for each of allVars in the class prototype (but no prototype chain). I’m not that familiar with the Java generator but I recall that it didn’t do a great job of inheritance – I half intended to do some clean-up on it myself but never got the time.

 

I hope this answers your question, please write back if you need more information.

 

Cheers,

 

--A

cbornet:
Oh, I didn't see you were using vars without allVars in other template files !
So that makes sense.
I'm precisely working on inheritance and composition for Java :smile:

cbornet:
Actually, if you also copy the interface properties to vars when supportInheritance is true (vars = own + interfaces but not parent) then it simplifies the model template and it also works well in the other places where you use vars (doc and test).
So I'm still not sure it's useful to separate them...

cbornet:
Reducing vars to the "own" properties would only be useful for languages that support mixins or some kind of multiple inheritance. For them I think we could introduce a supportsMixins options.

PR #2397, opened 2016-03-16
Fix c# client float property default values

mmschettler:
C# decimal literals for floats must be casted to float, because that are
doubles by default


wing328:
@mmschettler thanks for the PR. It looks good to me.

Issue #2315, opened 2016-03-05
Swift code in swagger


saharj:
_From @AnkitaScienceInc on March 4, 2016 15:46_

I can no more generate client code for swift

_Copied from original issue: swagger-api/swagger-editor#839_

wing328:
@AnkitaScienceInc can you please share the spec so that we can reproduce the issue?

wing328:
@AnkitaScienceInc Please also try the latest master as there's been a lot of enhancements to Swift API client.

AnkitaScienceInc:
i used http://editor.swagger.io/#/ where  I pasted JSON to generate a client code. Some days back to was allowing to generate a swift code, now it does not allow.

wing328:
@AnkitaScienceInc can you send me the spec so that I can take a look? if the spec cannot be shared publicly, please you send it to my email account in the Github profile page.

AnkitaScienceInc:
@wing328  i cloned a master branch code and was working great. now i'm facing this issue wherein for a variable with "_" (underscore)  character it generates camel case model variable.

This was not an issue some days back. Now this issue is appearing in a master branch code.

And also when i newly clone the master branch and go to generate an code. It give below issue:
Error: Unable to access jarfile modules/swagger-codegen-cli/target/swagger-codegen-cli.jar
[SwaggerCodeIssue.pdf](https://github.com/swagger-api/swagger-codegen/files/177938/SwaggerCodeIssue.pdf)

wing328:
@AnkitaScienceInc after you cloned the master, did you do a `mvn package` to create the jar?

AnkitaScienceInc:
@wing328 yes i already did. Earlier it was working with the master branch. Last code which i generated was on 3rd March and it had worked.

wing328:
@AnkitaScienceInc just so we're on the same page. The issue you're referring to is the model's **property name** (not the model name).

Again, can you send me the spec so that I can repeat the issue ?

AnkitaScienceInc:
please check Models.swift

// Decoder for [User]
            Decoders.addDecoder(clazz: [User].self) { (source: AnyObject) -> [User] in
                return Decoders.decode(clazz: [User].self, source: source)
            }
			// Decoder for User
            Decoders.addDecoder(clazz: User.self) { (source: AnyObject) -> User in
                let sourceDictionary = source as! [NSObject:AnyObject]
                let instance = User()
[...]

PR #2405, opened 2016-03-18
[Swift] use baseName for swift model property

wing328:
For https://github.com/swagger-api/swagger-codegen/issues/2315


jaz-ah:
+1 this fixed it for all decoding/encoding - thanks!

PR #2406, opened 2016-03-18
SwiftCodegen : change typeMapping for 'long' (Int64 instead of Int).

kenjitayama:
`long` should be mapped to Int64 instead of Int, because `long` will have a range of signed 64 bits.

Following the recommendation of the [Swift Doc](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/TheBasics.html), fixing only the mapping for `long` is necessary and sufficient.

```
Int

In most cases, you don’t need to pick a specific size of integer to use in your code. Swift provides an additional integer type, Int, which has the same size as the current platform’s native word size:

On a 32-bit platform, Int is the same size as Int32.
On a 64-bit platform, Int is the same size as Int64.
Unless you need to work with a specific size of integer, always use Int for integer values in your code. This aids code consistency and interoperability. Even on 32-bit platforms, Int can store any value between -2,147,483,648 and 2,147,483,647, and is large enough for many integer ranges.
```

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
@@ -74,6 +74,7 @@ public class SwiftCodegen extends DefaultCodegen implements CodegenConfig {
         languageSpecificPrimitives = new HashSet<String>(
                 Arrays.asList(
                     "Int",
+                    "Int64",
                     "Float",
                     "Double",
                     "Bool",
@@ -116,7 +117,7 @@ public class SwiftCodegen extends DefaultCodegen implements CodegenConfig {
         typeMapping.put("char", "Character");
         typeMapping.put("short", "Int");
         typeMapping.put("int", "Int");
-        typeMapping.put("long", "Int");
+        typeMapping.put("long", "Int64");
         typeMapping.put("integer", "Int");
         typeMapping.put("Integer", "Int");
         typeMapping.put("float", "Float");


wing328:
@kenjitayama thanks for the PR. Please update the `languageSpecificPrimitives` at https://github.com/kenjitayama/swagger-codegen/blob/fixSwiftCodegenTypeMapping/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java#L76 to include `Int64`

kenjitayama:
@wing328 Didn't notice that part... Added now.

cjolif:
I think this (unfortunately) won't be enough. First as I commented on the issue, I think 32 version is also needed. But also, the code generated now is not functional. Indeed while `Int` can be automatically put in a `AnyObject` this is not the case of `Int64` (see: http://stackoverflow.com/questions/28920232/why-do-integers-not-conform-to-the-anyobject-protocol for why, basically there is some auto-casting Swift is doing to `NSNumber` for `Int` but not bit-exact versions). So in the `encodeToJSON` method this case must be taken into account. I was actually working on a PR myself and faced that when testing the generated code :(

kenjitayama:
@wing328 Thanks for pointing out.
I was using my custom template, and `Int64` was fine for that, and didn't notice that.
Looks like there is more work to do for a complete fix...

kenjitayama:
Sorry, intended to @ mention @cjolif in my previous comment.

kenjitayama:
I'm not sure I can come up with a complete fix, so maybe this PR can be closed for now.

cjolif:
I'll try to find time to propose a PR that does take that into account. But no promise on when.

cjolif:
Looks like I was not as far as I thought. At least a tentative here: #2417

kenjitayama:
@cjolif Awesome! Everyone will have a safer Swift client :)
I see why you needed to use NSNumber, [mentioned in your previous comment](https://github.com/swagger-api/swagger-codegen/pull/2406#issuecomment-198230552).
It seems to me that this is why you said tentative, but looks best for now.
A fix without this kind of workaround would need a bigger fix, with a different approach.

cjolif:
> It seems to me that this is why you said tentative, but looks best for now.

Yes I had the feeling there were possible cleaner solutions so that's why I said "tentative" ;). But as you said yourself:
> A fix without this kind of workaround would need a bigger fix, with a different approach.

I tried a different path, but then indeed I ended up quickly changing a lot of things... so I sticked to the approach above to keep the changes are minimal as possible.

thanks!

PR #2407, opened 2016-03-18
Enhancement: add the ability to generate single file per operation (GET/POST/PUT/etc)

yotammadem:
In order to accomplish my task at work (automatic REST API documentation generator) I needed to be able to generate a different html page per API operation (not per path) Looks like the current swagger-codegen does not allow this. Except for this missing feature, it is a perfect choice. Therefore I'd like to add this feature to the core engine (DefaultGenerator). I've added unit tests to make sure it is working properly, as well as tested that build passes successfully with existing tests


wing328:
@yotammadem thanks for the PR. Your commits (as shown in the Commits tab) are not linked to your github acccount, which means your contribution won't count toward https://github.com/swagger-api/swagger-codegen/graphs/contributors

wing328:
Another suggestion is to create a new branch for the PR (instead of using master), which is part of the git best practice.

yotammadem:
Ok, thanks for those suggestions
I will cancel this PR and start new one
Anyway, I was thinking on a way to do it better

PR #2417, opened 2016-03-19
Generates Int32/Int64 for Swift when those are the required formats. …

### LINKED ISSUE #2133: [Swift] generating Int32/Int64 when those are required in format
Currently whatever is in format, int32 or int64, the code generated Int and not Int32 or Int64 while Swift has an exact mapping.

cjolif:
…Fixes #2133.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
@@ -74,6 +74,8 @@ public class SwiftCodegen extends DefaultCodegen implements CodegenConfig {
         languageSpecificPrimitives = new HashSet<String>(
                 Arrays.asList(
                     "Int",
+                    "Int32",
+                    "Int64",
                     "Float",
                     "Double",
                     "Bool",
@@ -115,10 +117,10 @@ public class SwiftCodegen extends DefaultCodegen implements CodegenConfig {
         typeMapping.put("string", "String");
         typeMapping.put("char", "Character");
         typeMapping.put("short", "Int");
-        typeMapping.put("int", "Int");
-        typeMapping.put("long", "Int");
-        typeMapping.put("integer", "Int");
-        typeMapping.put("Integer", "Int");
+        typeMapping.put("int", "Int32");
+        typeMapping.put("long", "Int64");
+        typeMapping.put("integer", "Int32");
+        typeMapping.put("Integer", "Int32");
         typeMapping.put("float", "Float");
         typeMapping.put("number", "Double");
         typeMapping.put("double", "Double");
--- a/modules/swagger-codegen/src/main/resources/swift/Extensions.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift/Extensions.mustache
@@ -19,6 +19,14 @@ extension Int: JSONEncodable {
     func encodeToJSON() -> AnyObject { return self }
 }
 
+extension Int32: JSONEncodable {
+    func encodeToJSON() -> AnyObject { return NSNumber(int: self) }
+}
+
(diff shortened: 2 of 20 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
```diff
@@ -115,10 +117,10 @@ public SwiftCodegen() {
         typeMapping.put("string", "String");
         typeMapping.put("char", "Character");
         typeMapping.put("short", "Int");
```
cjolif: I have left short untouched because I can't find it in swagger spec (and the test is not covering it)

wing328:
@cjolif thanks for the PR. FYI. I've restarted the CI job

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
kenjitayama: Looks ok because `Int` can handle `short` in any platform.

kenjitayama:
Looks good to me!

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
kenjitayama: Second thought, forgot about this problem:
If a server mistakenly sends back 64bit integer in spite of `short` in the api spec, 64bit platform clients can handle with `Int`, but 32bit platform clients can't. This kind of mistake can go by unnoticed.
So maybe `short` should be mapped to `Int16` ?

I also couldn't find `short` in the swagger spec. Maybe this line should be deleted if `short` really does not (and will not) exist in the specs.
Does anyone know where this `short` comes from?

kenjitayama:
Sorry, second thought, I commented on the line `typeMapping.put("short", "Int");`

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
kenjitayama: I found `typeMapping.put("short", "Short");` in `public DefaultCodegen()`, and it is overridden in `public SwiftCodegen()`
So maybe keep the line but change to `typeMapping.put("short", "Int16");` ?

cjolif: > So maybe keep the line but change to typeMapping.put("short", "Int16"); ?

We can but this won't work I suspect. Indeed we don't have the equivalent of 

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L1165

for shorts... Honestly I think this support for short is not really functional anyway so that's why I preferred not to touch it.

kenjitayama: You're right, looks like it won't work anyways. 
I agree keeping it untouched.
I think we can say the Int to Int16 fix is noncrucial compared to the Int to Int32 fix.

cjolif:
@wing328 do you have rough timeline for 2.2.0? (I see 2.1.6 is on-going).

wing328:
@cjolif v2.1.6 will be released soon.

Usually it takes 2 to 3 months for another stable release.

cjolif:
@wing328 ok thanks for the update

cjolif:
added a commit to deal with a case that is not covered by test.

wing328:
@cjolif @kenjitayama Thanks for the PR and it has been merged into the master.

PR #2421, opened 2016-03-19
Missing multipart (form data) imports.

kkrauth:
Swagger files that declare "in: formData" parameters will result in API class that cannot be compiled due to the usage of "@Multipart" annotations in method signatures.


wing328:
@kkrauth thanks for the fix.

Issue #2422, opened 2016-03-21
Default iOS reachability status


Stuart-campbell:
Hi,

Currently the default reachability status is 'AFNetworkReachabilityStatusNotReachable' this is causing me issues when attempting to check the status as my application starts up.

I suggest that the default status should be 'AFNetworkReachabilityStatusUnknown' as this is true at this point in the application lifecycle.

If this seems like reasonable suggestion happy to create a PR.

Effected line. https://github.com/swagger-api/swagger-codegen/blob/23696c5eb36d77be670e199d91a6818683ec1191/modules/swagger-codegen/src/main/resources/objc/ApiClient-body.mustache#L9

Side note: Thanks for the great work on the library!!

Thanks
Stuart

wing328:
@Stuart-campbell thanks for suggestion and offering help to contribute a PR.

I'm not sure if changing default status from "AFNetworkReachabilityStatusNotReachable" to "AFNetworkReachabilityStatusUnknown" would break any existing applications that use the ObjC SDK. What about adding a CLI option to make the status configurable (and default to AFNetworkReachabilityStatusNotReachable)?

(btw, I used the ObjC SDK a month ago but didn't encountered the issue you mentioned)

Stuart-campbell:
@wing328 

My issue is actually a result of the cache force fallback.

https://github.com/swagger-api/swagger-codegen/blob/23696c5eb36d77be670e199d91a6818683ec1191/modules/swagger-codegen/src/main/resources/objc/ApiClient-body.mustache#L598

I need to always force network as knowing the date the data was returned from the server is important. The library forces cache if offlineState=true but there is no getting for offline state so I check the reachability which sets the offlineState.

On first load offlineState is default to false and reachabilityStatus defaults to AFNetworkReachabilityStatusNotReachable so at this point the two are out of sync meaning your requests will not be forced to use cache but reachabilityStatus is NotReachable.

My suggestion for ReachableUnknown is based on that seeming most logical but I fully understand not wanting to change something it could cause issues for other people. Either a setter reachability status or a getting for offlineState would solve my issue.

I would like to tell it not to use cache if offline but I don't have as good justification for that.

Thanks

wing328:
> Either a setter for reachability status or a getter for offlineState would solve my issue.

Your suggestion sounds good to me. Let us know if you need help implementing the getter/setter.

Ref: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

[...]

PR #2426, opened 2016-03-22
Updates to objc api to address issue #2422

Stuart-campbell:
Added getter for offline state this lets the app check if and ApiClient is going to attempt the network request or just force cache.

Added setter for AFNetworkReachabilityStatus, this is useful for two reason first to set the default before it has been set by the reachability manager. 

Also as the ApiClient uses a shared instance of the reachability manager it is possible for another class to setReachabilityStatusChangeBlock stopping the ApiClient getting reachability change callbacks internally. By adding a setter it would be possible to pass the changes along.

Note: on changes to SWGName, these changes must be because the last person to edit the swagger spec did not regenerate the objc code.


wing328:
@Stuart-campbell thanks for the PR. The change looks good to me and the test cases for ObjC Petstore passed.

PR #2427, opened 2016-03-22
[C#] add UWP support to C# API client

wing328:
Thanks for the help from @jerry1621, we've added UWP support (beta) to C# API client.

The "synchronous" request is still work-in-progress with the following TODO in ApiClient.cs:
```
            // TODO: This is obviously not a valid response, but it prevents compiling-errors
            // 'RestClient.CallApiAsync(..)' is used instead, because 'RestClient.ExecuteAsync(...)' 
            // is available for Windows UWP, while 'RestClient.CallApi(...)' is not
            var response = new object();

            return (Object) response;
``` 

If anyone wants to work on the above and has better suggestion on supporting "synchronous" request, please reply to let us know.

To generate code for UWP, please provide `-D targetFramework=uwp` when generating the C# API client from the command line.


wing328:
As discussed with @jerry1621, he will help implement the sync HTTP request later with another PR.

PR #2433, opened 2016-03-23
[PHP] Parent constructor

arnested:
* Use parent constructor when inheriting (bugfix)
* Set discriminator property in constructor (improvement)


arnested:
@wing328, could you please reconsider adding at least the parent constructor commit (4958f14d1b79a74378810e8272840ae2fb2a54e8) to 2.1.6? Without it my inheritance addition actually breaks some use cases that worked in 2.1.5 (using the constructor with a `$data` argument).

wing328:
@arnested ok let me run some tests now. Will let you know if I've any questions.

wing328:
@arnested do you mind adding https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/discriminatorTest.json#L38-L73 to petstore.json (below "200_response") and add a test case in PHP Petstore to construct the objects with base class?

(my testing so far definitely looks good but I think a test case to cover the issue moving forward would be better)

arnested:
I'm on it now.

arnested:
Ok,

* I added some test cases (and they pass :-)
* Rebased on current master
* Left out the parts about initializing the discriminator property automatically (will push that as another PR for 2.2.0 later -- it is not necessary in the 2.16 release)

Let me know if this was what you had in mind or nor, @wing328.

wing328:
@arnested yes, the new test cases look good. Thanks for the quick turnaround.

PR #2434, opened 2016-03-23
[PHP] Map number without format to float

arnested:
Previously if you specified a property to be of type `number` and didn't
explicitly specify a format Codegen would map it into an (unknown) PHP
class `Number`.

We add a mapping from Swaggers `number` to PHPs `float`.


wing328:
@arnested  Thanks for the PR. We'll include that in the v2.2.0 release as v2.1.6 will be released very soon.

wing328:
@arnested thanks for the contribution. PR merged into master.

PR #2436, opened 2016-03-23
[Dart]: generator used underscored filenames but used non-underscored class names to import the files

ircecho:


wing328:
@ircecho thanks for the PR. We'll include that in the v2.2.0 as v2.1.6 will be released very soon.

PR #2443, opened 2016-03-24
Add withCredentials option to support cookies

### LINKED ISSUE #2442: [JavaScript] Need option for withCredentials
In order to be able to send cookies to the server, we need to call `request.withCredentials()` in `ApiClient.js`. See [the superagent documentation](https://visionmedia.github.io/superagent/#cors). This should be an option, as it may not always be desirable.

In my case, I am calling a server on a different URL, which means I have to use CORS, and I am using cookies for session management (set by Java through the use of `@SessionScoped` beans).

I will submit a PR to add such an option if there are no objections.

delenius:
Fixes #2442


wing328:
@delenius  Thanks for the PR. We'll include that in the v2.2.0 release as v2.1.6 will be released very soon.

fehguy:
See my comment on the original issue, #2442

delenius:
Please consider merging this now that 2.1.6 is out.

delenius:
Closing. I filed a new PR for this issue, which replaces this one.

Issue #2450, opened 2016-03-24
Java and SpringMVC generated server code for "enum" does not work


danjor:
Java and SpringMVC generated server code for "enum" does not work with current master.

Using :

```
swagger: "2.0"
info:
  version: 1.0.0
  title: api
  description: API
  contact:
    name: api team
basePath: /api
schemes:
  - http
consumes:
  - application/json
produces:
  - application/json
paths:
  /customer:
    get:
      operationId: 1
      parameters:
        - name: id
          in: query
          type: string
      responses:
        default:
          description: unexpected error
          schema:
            type: string
definitions:
  ResponseStatus:
    required:
      - status
    properties:
      status:
        type: string
        enum:
          - Ok
          - Error
  Example:
    type: string
    enum:
      - First
      - Second
```

produces the correct class ResponseStatus :

```
@ApiModel(description = "")
@javax.annotation.Generated(value = "class io.swagger.codegen.languages.JavaClientCodegen", date = "2016-03-24T15:35:37.946+01:00")
public class ResponseStatus  {
  
  public enum StatusEnum {
     OK,  Error, 
  };

}
  
  private StatusEnum status = null;

```
  
but the wrong class Example :
            
```
@ApiModel(description = "")
@javax.annotation.Generated(value = "class io.swagger.codegen.languages.JavaClientCodegen", date = "2016-03-24T15:35:37.946+01:00")
public class Example  {

  @Override
  public boolean equals(Object o) {
    if (this == o) {
      return true;
    }
    if (o == null || getClass() != o.getClass()) {
      return false;
    }
    Example example = (Example) o;
    return true;
  }

  @Override
  public int hashCode() {
    return Objects.hash();
  }

  @Override
  public String toString()  {
    StringBuilder sb = new StringBuilder();
    sb.append("class Example {\n");
    
    sb.append("}\n");
    return sb.toString();
  }
}

```
Expected :

```
public enum Example {
    Ok, Error
}

```

wing328:
@danjor can you please test #2460 (submitted by @xhh) to see if that fixes the issue for you?

wing328:
@danjor we've merged https://github.com/swagger-api/swagger-codegen/pull/2508 into master. Please pull the latest master and give it a try.

Closing this for now.

PR #2460, opened 2016-03-28
[Java] Support outer enum class in Java okhttp-gson client

xhh:
This PR fixes the same issue with #2450 for the Java okhttp-gson client.


wing328:
Close via https://github.com/wing328/swagger-codegen/pull/2

Issue #2449, opened 2016-03-24
SubClass annotations are missing from the base class


yaelah:
In order for jackson ObjectMapper to deserialize correctly a base class to its subclasses, we need to include annotations defining all expected subclasses in the parent class.
The codegen currently does not generate those annotations.
For this example YAML:

definitions:
&nbsp;&nbsp;Base:
&nbsp;&nbsp;&nbsp;&nbsp;type: object
&nbsp;&nbsp;&nbsp;&nbsp;discriminator: disc
&nbsp;&nbsp;&nbsp;&nbsp;properties:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;disc:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;type: string
&nbsp;&nbsp;&nbsp;&nbsp;required:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;\- disc
&nbsp;&nbsp;SubClass:
&nbsp;&nbsp;&nbsp;&nbsp;allOf:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;\- $ref: '#/definitions/Base'
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;required:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;\- field
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;properties:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;field:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;type: string

Jackson expects these annotations to be produced, but they are not:
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "disc")
@JsonSubTypes({ @Type(value = SubClass.class, name = "SubClass") })
public class Base {
}

wing328:
@yaelah thanks for the detailed above. Would you have cycle to contribute a fix?

[...]

PR #2467, opened 2016-03-29
Fix issue #2449: SubClass annotations are missing from the base class.

yaelah:
Added a children property to CodegenModel and write out the necessary annotations.
Please give me your feedback on this change and let me know if there is something else I should do.
thanks.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -123,6 +123,19 @@ public class DefaultCodegen {
                     }
                 }
             }
+            // Let parent know about all its children
+            for (String name : allModels.keySet()) {
+                CodegenModel cm = allModels.get(name);
+                if (cm.parent == null) continue;
+                CodegenModel parent = allModels.get(cm.parent);
+                if (parent.children == null) {
+                    parent.children = new ArrayList<CodegenModel>();
+                } else {
+                    parent.children.get(parent.children.size() - 1).hasMoreChildren = Boolean.TRUE; 
+                }
+                parent.children.add(cm);
+                parent.hasChildren = true;
+            }
         }
         return objs;
     }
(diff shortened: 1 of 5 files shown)


wing328:
@yaelah thanks for the PR. As shown in the commit tab, the commit is not linked to your Github account, which means your credit won't count in https://github.com/swagger-api/swagger-codegen/graphs/contributors

I would suggest you to fix it by setting a proper git email address in your local environment.

yaelah:
I am not sure how to get rid of the first 2 commits. Should I just delete this pull request and create a new fork with the correct account?
thanks.

wing328:
@yaelah would this help? https://help.github.com/articles/changing-author-info/

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
```diff
@@ -47,6 +47,7 @@
 
     public JavaClientCodegen() {
         super();
+        supportsInheritance = true;
```
cbornet: This cannot be done at the moment. See : https://github.com/swagger-api/swagger-codegen/pull/1120#issuecomment-219290749
Does it work if you keep supportsInheritance to false ?

modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
```diff
@@ -123,6 +123,19 @@ public void processOpts() {
                     }
                 }
             }
+            // Let parent know about all its children
+            for (String name : allModels.keySet()) {
+                CodegenModel cm = allModels.get(name);
+                if (cm.parent == null) continue;
```
cbornet: ```continue``` is ugly here.
It would be better to put the folllowing code in a ```if (cm.parent != null)``` loop.

yaelah:
This is a matter of taste :)
Open source projects such as WebKit and Chromium prefer early return over indentation.

modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
```diff
@@ -123,6 +123,19 @@ public void processOpts() {
                     }
                 }
             }
+            // Let parent know about all its children
+            for (String name : allModels.keySet()) {
```
cbornet: ```for (CodegenModel cm : allModels.values())``` ?

cbornet:
I'm OK with early return if there is code after the return statement which is not the case here. But you're right, that's opiniated...

PR #2508, opened 2016-04-06
[PHP][C#] better enum support

wing328:
Better enum support for numbers and special characters

PR #2516, opened 2016-04-06
[Swift] make sure to only escape an enum if the actual final variable name is…

jaz-ah:
… going to match the enum name - now that we camelCase variable names this cuts down on the amount of enum escaping we have.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
@@ -332,9 +332,9 @@ public class SwiftCodegen extends DefaultCodegen implements CodegenConfig {
             // the variable name doesn't match the generated enum type or the
             // Swift compiler will generate an error
             if (isReservedWord(codegenProperty.datatypeWithEnum) ||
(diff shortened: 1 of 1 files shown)


wing328:
@jaz-ah when you've time, do you mind giving an example to illustrate the issue?

jaz-ah:
@wing328 so for example before if we had this enum in swagger:
  
"Format": {
    "type": "string",
    "enum": ["plain","html","template"]
}

the code was generating:

public var Format: Format

which would create a swift compiler error.  I made a fix to escape the enum name so we would have:

public var Format: _Format

and _Format would define the enumeration.  But then we started camel casing swift variable names so without my newest fix you'd have:

public var format: _Format

but the escaping is unnecessary in this case - with this change you'll get:

public var format: Format

jaz-ah:
@wing328 let me know if this makes sense ^^

wing328:
@jaz-ah I need more time to review. We've a major PR for enum: https://github.com/swagger-api/swagger-codegen/pull/2508

I need to make sure enum support in other langauges do not have similar issue.

jaz-ah:
@wing328 no prob

wing328:
@jaz-ah PR merged. Sorry for taking this long.

PR #2521, opened 2016-04-07
Fix typo in wget arg

gabrielvv:
The first URL (from the doc) does not work (at least for me) whereas the second does .

- **Previously** wget http://repo1.maven.org/maven2/io/swagger/swagger-codegen-cli/2.1.**5**/...
- **Now** wget http://repo1.maven.org/maven2/io/swagger/swagger-codegen-cli/2.1.**6**/...


wing328:
Thanks for the fix

PR #2524, opened 2016-04-07
Feature/typescript node default base path

bherila:
Possibly need to separate out 91efdd9 but this is a separate ticket... it seems to make sense for the default `basePath` to be defined only once in the file, rather than have the hard-coded default endpoint  multiple times in the .ts file.


bherila:
Can this be merged @wing328  ?

wing328:
Close via #2758

PR #2530, opened 2016-04-08
Hotfix/objectiveCApiClientMustache

matteo-harpoon:
ApiClient moustache files (both header and body) look broken.
```
- (id) deserialize:(id) data class:(NSString *) class
```
Is using the parameter `class` which is the cause of the error.

I’ve changed it from `class` to `className` getting this result:
```
- (id) deserialize:(id) data className:(NSString *) className;
```

### DIFF
--- a/modules/swagger-codegen/src/main/resources/objc/ApiClient-body.mustache
+++ b/modules/swagger-codegen/src/main/resources/objc/ApiClient-body.mustache
@@ -288,25 +288,25 @@ static void (^reachabilityChangeBlock)(int);
 
 #pragma mark - Deserialize methods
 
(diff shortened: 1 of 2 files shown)


wing328:
@matteo-harpoon thanks for the PR. I don't think it's broken as the integration tests for Petstore API pass.

Did the integration fail for you? Here are the steps to run the integration tests:
```
mvn clean && ./bin/objc-petstore.sh
cd samples/client/petstore/objc/SwaggerClientTests/
mvn integration-test
```

wing328:
@matteo-harpoon closing this as there's no update from you.

PR #2532, opened 2016-04-08
issue #1948, added testing code to test go api

guohuang:
#1948 

@wing328 i don't know if this is what you want, it is very similar to the test.go, please take a look
the swagger.io api `http://petstore.swagger.io/v2/pets` does not return anything which causing the tests failed.

here is the test result 
image


wing328:
@guohuang thanks for the PR. I think that's better than what we currently have.

To make it even better, I would suggest using `assert` from https://github.com/stretchr/testify to verify the response.

For other languages (e.g. Ruby), we do check each field from the returned object, e.g. https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/ruby/spec/pet_spec.rb#L73

guohuang:
@wing328 thanks for your suggestion, i have added `assert` function, however, since the endpoints are failing to return anything, i could not test it.

btw, do you expect me to write similar tests for the store and user api as well?

wing328:
@guohuang thanks for the quick turnaround. Let me have a look at the Petstore test later today.

I do not expect you to do so. Later we can add more test cases if needed. Your work is a good reference that other Go contributors can easily follow.

wing328:
@guohuang btw, your latest commit (1f39a4d3a91e0c3892985f2396b2d3c8b85bca3b) is not linked to your Github account somehow ...

wing328:
The issue with the failed test cases is related to the spec and it's addressed in https://github.com/swagger-api/swagger-codegen/pull/2542 (merged into master).

Please pull the latest master when you've a chance.

Issue #2461, opened 2016-03-29
[Python] Add method _with_http_info


wing328:
Similar to PHP, Ruby and other languages, for each method (endpoint) in Python, we wan to add a corresponding method "_with_http_info" (e.g. `add_pet_with_http_info` for `add_pet`) to retrieve the response together with http headers and status code.

Similar PRs: 
[Ruby] https://github.com/swagger-api/swagger-codegen/pull/1677
[PHP] https://github.com/swagger-api/swagger-codegen/pull/1672

If anyone wants to contribute to this task, please reply to let us know.

wing328:
Done via https://github.com/swagger-api/swagger-codegen/pull/3046

Thanks @zhenjun115 for the contribution!

PR #2536, opened 2016-04-09
fix test case test_async_request failed #2461

cobain:
I fixed this test case test_async_request for task Add method _with_http_info #2461.
@wing328 this change is based on your change. hope you can review it kindly. Thanks.


wing328:
@cobain thanks. I'll review and resolve the merge conflicts.

wing328:
@cobain  Above [line 77](https://github.com/cobain/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/python/api.mustache#L77), we need to add a line to set an optional parameter (e.g. kargs['return_data_only'] to be passed down to the HTTP layer (rest.py) so that it will return just the data but not data with http info (header, status).

Please ping me if you need more information on the suggested fix.

wing328:
Closed vis #3046

PR #2537, opened 2016-04-09
[PHP] Set discriminator property in constructor

arnested:
When using inheritance let the constructor set the discriminator property to the model name.

This avoids redundant and error prone stuff like:

```php
  $animal = new \Swagger\Client\Model\Dog();
  $animal->setClassName('Dog');
```

I added a phpunit test case for this and it (and the other test case) pass.


arnested:
This change was previously submitted as part of #2433 but left out to make the other part of that pull request fit for `2.1.6`.

Issue #2543, opened 2016-04-10
[Go] Add SelectHeaderContentType, SelectHeaderAccept to Go API client


wing328:
Similar to other languages, we need logics in Go API client to pick the correct content-type (consumes) and accept (produces).

The goal of this task is to 
 - implement the function `SelectHeaderContentType` (simiar to [SelectHeaderContentType in C#](https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/csharp/SwaggerClientTest/Lib/SwaggerClient/src/main/csharp/IO/Swagger/Client/ApiClient.cs#L315-L331))

 - implement the function `SelectHeaderAccept ` (simiar to [SelectHeaderAccept in C#](https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/csharp/SwaggerClientTest/Lib/SwaggerClient/src/main/csharp/IO/Swagger/Client/ApiClient.cs#L333-L349))

- for the time being, we can put these new functions in Configuration struct or a new struct (e.g. ApiClient, similar to other languages)

- update Go mustache template (api.mustache) to include the logic selecting proper content-type and header: (e.g. [api.mustached](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/csharp/api.mustache#L196-L208) in C#)

If anyone from the community wants to work on this, please reply to let us know.

wing328:
PR merged. Thanks for the contribution from @guohuang

PR #2544, opened 2016-04-10
issue #1948, added missing apiresponse.go, renamed petapi_test.go, fixed compilation error

guohuang:
@wing328 yes, the last PR fixed the api issue, i clean up the code a little bit. however, the tests are still failing because `GetPetById` is not returning any object, do you know what happened? i can't access the api via http call.

image 

#1948

### DIFF
--- a/samples/client/petstore/go/petApi_test.go
+++ b/samples/client/petstore/go/petApi_test.go
@@ -8,7 +8,6 @@ import (
 )
 
 func TestAddPet(t *testing.T) {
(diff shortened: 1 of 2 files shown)


wing328:
The issue with `GetPetById` is that we've not specified the "accept" (produces) in the HTTP request and the endpoint is returning data in XML by default. This task should address it: https://github.com/swagger-api/swagger-codegen/issues/2543

Issue #1946, opened 2016-01-22
[Go] Add a method to include default headers in HTTP request


wing328:
Similar to API clients in other languages, we want to have a method `AddDefaultHeader` to add default headers to HTTP request.

If anyone in the community wants to work on this, please reply to let us know.

A good starting point is https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/go/api.mustache

wing328:
Done via #2546. Thanks @guohuang

Issue #2534, opened 2016-04-09
[Go] Support Default Headers


wing328:
Similar to other langauges, the developers should be able to set default headers for each HTTP request.

To do that, we need:
 - add a map to store the default header in [Configuration.go](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/go/configuration.mustache)
 - in [api.mustache](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/go/api.mustache#L76), we need to obtain the default header map from configuration and then add using `_sling.Set()`

If anyone has questions or wants to contribute to this, please reply to let us know.

wing328:
PR merged. Thanks for the contribution from @guohuang

PR #2546, opened 2016-04-11
issue #1946, added AddDefaultHeader method for GO

guohuang:
@wing328 please review

#1946


wing328:
@guohuang looks good to me.

In https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/go/api.mustache#L52, can we also use  _sling.Set to set the default header provided in the Configuration?

guohuang:
@wing328 I have updated the PR as you requested, initially, i was going to add that on a separate PR because it belongs to #2534, can you close the issues that are already done/merged?

can you show me how to test the generated go code?

PR #2551, opened 2016-04-11
Fix handling optional fields in asyncscala client model

hcwilhelm:
Small fix to generate Option fields correctly in the model of the scalaasync client. 

When displaying the mustache scope with the option -DdebugModels I was not able to 
find the variable "isNotRequired" so I replaced it with the negated version of "required". Now
the model.mustache generates Option fields correctly.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/asyncscala/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/asyncscala/model.mustache
@@ -7,7 +7,7 @@ import java.util.UUID
 
 {{#model}}
 case class {{classname}} (
-  {{#vars}}{{name}}: {{#isNotRequired}}Option[{{/isNotRequired}}{{datatype}}{{#isNotRequired}}]{{/isNotRequired}}{{#hasMore}},{{/hasMore}}{{#description}}  // {{description}}{{/description}}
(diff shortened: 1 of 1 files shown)


wing328:
@hcwilhelm I think `isNotRequired` was used before the 2.x version a while ago (when the generator was written in scala). Thanks for the PR to fix it.

wing328:
@hcwilhelm would you have cycle to file a simlar fir for scalatra: https://github.com/swagger-api/swagger-codegen/blob/4d19da41dd8a3462f88c2be57174aace45101064/modules/swagger-codegen/src/main/resources/scalatra/model.mustache#L10 ?

wing328:
@hcwilhelm thanks for fixing scalatra template.

I found that [{{#requiredParamCount}}](https://github.com/swagger-api/swagger-codegen/blob/90364b3d9c97e39515c6ce655a13e799aa3db8a1/modules/swagger-codegen/src/main/resources/asyncscala/api.mustache#L24-L26) is not set either and we should replace it with something like https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/Java/api.mustache#L51-56

Please let me know if you can help on that also.

hcwilhelm:
I created a new PR that should fix all  {{#requiredParamCount}}  {{/requiredParamCount}} pairs. 

Hope I could help ;)

PR #2556, opened 2016-04-11
Model option scalatra

hcwilhelm:
Small fix for generating optional fields in the Scalatra client model. Also added a new line at EOF


wing328:
@hcwilhelm thanks for the quick fix.

Issue #2561, opened 2016-04-11
In DefaultCodegen.java, CodegenParameter.is<TYPE>Param flags should be set before invoking postProcessParameter


neilotoole:
In [DefaultCodegen#fromOperation](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L1554), we call `fromParameter()` to build the `CodegenParameter` object, and then a few lines [later](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L1581) we check the type of the param, and set the type booleans, e.g. `p.isQueryParam = new Boolean(true)`, etc. These type flags should probably be set inside `fromParameter()`, as that method calls out to `postProcessParameter()`, which is where a language-specific generator can modify the parameter.

TLDR: the `p.is<Type>Param` flags should be set before `postProcessParameter()` is invoked.

wing328:
PR merged into master

PR #2563, opened 2016-04-11
Issue #2561 : set isXYZParam flags before postProcessParameter [updated]

neilotoole:
Addresses #2561 ... populate the p.is<TYPE>Param flags before the call to `postProcessParameter`

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -1578,33 +1578,20 @@ public class DefaultCodegen {
 
 
                 allParams.add(p);
+                // Issue #2561 (neilotoole) : Moved setting of is<Type>Param flags
+                // from here to fromParameter().
                 if (param instanceof QueryParameter) {
-                    p.isQueryParam = new Boolean(true);
                     queryParams.add(p.copy());
                 } else if (param instanceof PathParameter) {
-                    p.required = true;
-                    p.isPathParam = new Boolean(true);
                     pathParams.add(p.copy());
                 } else if (param instanceof HeaderParameter) {
-                    p.isHeaderParam = new Boolean(true);
                     headerParams.add(p.copy());
                 } else if (param instanceof CookieParameter) {
-                    p.isCookieParam = new Boolean(true);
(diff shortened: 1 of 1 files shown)


wing328:
@neilotoole I couldn't find `postProcessParameters`. I believe it's something you plan to implement, right?

Instead of adding it, what about processing the parameters in `postProcessOperations` instead, which can access not only the operations but also the parameters?

neilotoole:
@wing328 `postProcessParameters` is invoked by `fromParameter` [here](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L1935). If my understanding is correct, the purpose of the `postProcessXYZ` methods is to allow the language specific code generators to override them to, well, post-process the base class implementation. That is to say, they are "processing hooks". And there's no downside in not providing as fully-processed an object as possible to the hook. I ran into this situation when I was looking at one possible approach for changing a param name for another [issue](https://github.com/swagger-api/swagger-codegen/issues/2564). Putting the logic into the `GoClientCodegen#postProcessParam` would have been perfect (if I'd ended up using that approach), except that my logic needed to know the param type (e.g. query param, etc), and those fields hadn't been populated yet, but other fields had (e.g. the json). I almost ended up interrogating the json to extract the param type that way (but fortunately I ended up taking another approach). As for using `postProcessOperations` instead... well, I guess we could do that, but in that case, what's the point of having the `postProcessParam` hook at all?

TLDR: There's no reason not to populate objects as much as possible before invoking their post-processing hooks.

neilotoole:
Updated above comment to point out that it's actually `postProcessParameter`, not `postProcessParameters` (with an `s`) ... my apologies @wing328!!!

wing328:
@neilotoole no problem. I just figured that out :)

It all makes sense to me now. I'll review and perform some tests to ensure there's downstream impact.

Thanks for the PR, specially the detailed comments explaining the change in the code.

neilotoole:
@wing328 No problem, apologies for the sloppiness, there's no excuse for that considering that I had my full compliment of 15 cups of tea yesterday... I have no idea what went wrong :-)

wing328:
I ran some tests with C#, PHP, Ruby and more and didn't' find any issue. PR merged.

Issue #2535, opened 2016-04-09
[Go] Support authentication


wing328:
Similar to other langauges, the Go API clients needs to support authentication (HTTP basic, API key, OAuth token)

To do that, we need to add something similar to the following (from [C#](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/csharp/api.mustache#L230-L250)):
```
            {{#authMethods}}// authentication ({{name}}) required
            {{#isApiKey}}{{#isKeyInHeader}}
            if (!String.IsNullOrEmpty(Configuration.GetApiKeyWithPrefix("{{keyParamName}}")))
            {
                localVarHeaderParams["{{keyParamName}}"] = Configuration.GetApiKeyWithPrefix("{{keyParamName}}");
            }{{/isKeyInHeader}}{{#isKeyInQuery}}
            if (!String.IsNullOrEmpty(Configuration.GetApiKeyWithPrefix("{{keyParamName}}")))
            {
                localVarQueryParams["{{keyParamName}}"] = Configuration.GetApiKeyWithPrefix("{{keyParamName}}");
            }{{/isKeyInQuery}}{{/isApiKey}}{{#isBasic}}
            // http basic authentication required
            if (!String.IsNullOrEmpty(Configuration.Username) || !String.IsNullOrEmpty(Configuration.Password))
            {
                localVarHeaderParams["Authorization"] = "Basic " + ApiClient.Base64Encode(Configuration.Username + ":" + Configuration.Password);
            }{{/isBasic}}{{#isOAuth}}
            // oauth required
            if (!String.IsNullOrEmpty(Configuration.AccessToken))
            {
                localVarHeaderParams["Authorization"] = "Bearer " + Configuration.AccessToken;
            }{{/isOAuth}}
            {{/authMethods}}
```
to Go's [api.mustache](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/go/api.mustache#L76) before submitting the HTTP request.

If anyone has questions or wants to contribute to this, please reply to let us know.

wing328:
PR merged. Thanks for the contribution from @guohuang

PR #2567, opened 2016-04-11
issue #2535, added authentication header for GO

guohuang:
link #2535, i am not sure what `GetApiKeyWithPrefix` is, so i did not implement it, let me know if you still want this, and please explain how

@wing328 please review

### DIFF
--- a/modules/swagger-codegen/src/main/resources/go/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/go/api.mustache
@@ -44,6 +44,13 @@ func (a {{classname}}) {{nickname}} ({{#allParams}}{{paramName}} {{{dataType}}}{
     _sling := sling.New().{{httpMethod}}(a.Configuration.BasePath)
 
     {{#authMethods}}// authentication ({{name}}) required
+    {{#isApiKey}}{{#isKeyInHeader}}
+    // set key with prefix in header
+    _sling.Set("{{keyParamName}}", a.Configuration.GetApiKeyWithPrefix("{{keyParamName}}")
+    {{/isKeyInHeader}}{{#isKeyInQuery}}
+    // set key with prefix in querystring
+    _sling.Set("{{keyParamName}}",  a.Configuration.GetApiKeyWithPrefix("{{keyParamName}}")
+    {{/isKeyInQuery}}{{/isApiKey}}    
     {{#isBasic}}
     // http basic authentication required
     if a.Configuration.Username != "" || a.Configuration.Password != ""{
--- a/modules/swagger-codegen/src/main/resources/go/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/go/api.mustache
@@ -43,6 +43,21 @@ func (a {{classname}}) {{nickname}} ({{#allParams}}{{paramName}} {{{dataType}}}{
 
     _sling := sling.New().{{httpMethod}}(a.Configuration.BasePath)
 
+    {{#authMethods}}// authentication ({{name}}) required
+    {{#isBasic}}
+    // http basic authentication required
+    if a.Configuration.Username != "" || a.Configuration.Password != ""{
+        _sling.Set("Authorization",  "Basic " + a.Configuration.GetBasicAuthEncodedString())
+    }
+    {{/isBasic}}
+    {{#isOAuth}}
+    // oauth required
+    if a.Configuration.AccessToken != ""{
+        _sling.Set("Authorization", "Bearer " +  a.Configuration.AccessToken)
+    }
+    {{/isOAuth}}    
+    {{/authMethods}}
+
     // create path and map variables
     path := "{{basePathWithoutHost}}{{path}}"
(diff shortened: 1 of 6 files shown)


wing328:
@guohuang  `GetApiKeyWithPrefix` is for adding a prefix to the authentication API key, e.g.
```
AUTHORIZATION: TOKEN API_KEY_HERE
```
As you can see sometimes the authentication needs a prefix (TOKEN in this case)

so in configuration struct, ApiKey should be a map of string and we need to add ApiKeyPrefix (map of string) as another property in the configuration struct.

modules/swagger-codegen/src/main/resources/go/api.mustache
```diff
@@ -43,6 +43,28 @@ func (a {{classname}}) {{nickname}} ({{#allParams}}{{paramName}} {{{dataType}}}{
 
     _sling := sling.New().{{httpMethod}}(a.Configuration.BasePath)
 
+    {{#authMethods}}// authentication ({{name}}) required
+    {{#isApiKey}}{{#isKeyInHeader}}
+    // set key with prefix in header
+    _sling.Set("{{keyParamName}}", a.Configuration.GetApiKeyWithPrefix("{{keyParamName}}")
+    {{/isKeyInHeader}}{{#isKeyInQuery}}
+    // set key with prefix in querystring
+    _sling.Set("{{keyParamName}}",  a.Configuration.GetApiKeyWithPrefix("{{keyParamName}}")
```
guohuang: @wing328 i haven't figure out how to set queryparam using sling, it seems complicated, i have to generate a struct and pass the parameter there using `QueryStruct`

wing328: Yes, you need to use the QueryParams struct: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/go/api.mustache#L58-L63

modules/swagger-codegen/src/main/resources/go/api.mustache
```diff
@@ -43,6 +43,32 @@ func (a {{classname}}) {{nickname}} ({{#allParams}}{{paramName}} {{{dataType}}}{
 
     _sling := sling.New().{{httpMethod}}(a.Configuration.BasePath)
 
+    {{#authMethods}}// authentication ({{name}}) required
+    {{#isApiKey}}{{#isKeyInHeader}}
+    // set key with prefix in header
+    _sling.Set("{{keyParamName}}", a.Configuration.GetApiKeyWithPrefix("{{keyParamName}}")
+    {{/isKeyInHeader}}{{#isKeyInQuery}}
+    // set key with prefix in querystring
+    {{#hasKeyParamName}}    type KeyQueryParams struct {
```
guohuang: @wing328 code has been updated, but there is no sample that uses query param, so i am not 100% sure if the output will be correct.

wing328: @guohuang thanks. I'll perform a test manually to confirm

modules/swagger-codegen/src/main/resources/go/api.mustache
```diff
@@ -43,6 +43,32 @@ func (a {{classname}}) {{nickname}} ({{#allParams}}{{paramName}} {{{dataType}}}{
 
     _sling := sling.New().{{httpMethod}}(a.Configuration.BasePath)
 
+    {{#authMethods}}// authentication ({{name}}) required
+    {{#isApiKey}}{{#isKeyInHeader}}
+    // set key with prefix in header
+    _sling.Set("{{keyParamName}}", a.Configuration.GetApiKeyWithPrefix("{{keyParamName}}")
```
wing328: it's missing a ")" at the end

guohuang:
@wing328 added the missing ")" thanks for checking

wing328:
@guohuang thanks. I'll take another look and merge into master if the test result is good.

wing328:
@guohuang if the query parameter/apiKey starts with lower case letter, it's pending the PR: https://github.com/swagger-api/swagger-codegen/pull/2568

modules/swagger-codegen/src/main/resources/go/configuration.mustache
```diff
@@ -1,32 +1,53 @@
...
+}
+
+func (c *Configuration) GetApiKeyWithPrefix(apiKeyIdentifier string) string {
+    var returnValue = c.ApiKey[apiKeyIdentifier]
+    var apiKeyPrefix = c.ApiKeyPrefix[apiKeyIdentifier]
+    if apiKeyPrefix != ""{
+        returnValue = apiKeyPrefix + " " + returnValue
+    }
+        
+    return returnValue
```
wing328: @guohuang Suggestion: I think we can skip the local variable `returnValue` and `apiKeyPrefix` to reduce the overhead in declaring additional local variables.

guohuang: ok, i will submit a PR later

Issue #2564, opened 2016-04-11
[Go] [Bug] QueryParam struct field names must be exported (UpperCamelCase)


neilotoole:
Currently, the Go generator can produce code like this:

````go
    _sling = _sling.Path(path)

    type QueryParams struct {
        userId    int32 `url:"user_id,omitempty"`  
}
    _sling = _sling.QueryStruct(&QueryParams{ userId: userId })
````

Note that `userId` is not exported (first letter is lowercase). Therefore sling doesn't see the `userId` value, and doesn't send it in the request.

**Comment by wing328** [comment]
PR merged. Issue closed.

**Comment by nnathan** [comment]
I fetched latest master and found the commit 2f3897f but the changes aren't reflecting. I sense a later merge commit undid the change?

**Comment by wing328** [comment]
@nnathan did you do a `mvn clean package` to build the JAR using the latest master?

**Comment by nnathan** [comment]
Yep. I did a fresh clone and ran `mvn clean package`.

```
$ git rev-parse HEAD
c888434580c601ac9c4734e19d8eefac63518031
$ grep "vendorExtensions.x-exportParamName" modules/swagger-codegen/src/main/resources/go/api.mustache
$ echo $?
1
```

From the above, you can see at HEAD the corresponding changes from commit 2f3897f aren't merged.  Specifically this: https://github.com/swagger-api/swagger-codegen/commit/2f3897f6d5b5df557bf2de526081cb1edcb0d8c2#diff-293df8239a8ee15d445676a48a96a143R87

**Comment by wing328** [comment]
You're right. Looks like the fix has been removed accidentally. I'll take a deeper look later today.

cc @neilotoole @guohuang

**Comment by guohuang** [comment]
@wing328 I don't think we need this commit anymore, as we have already switched to use resty instead of sling, let me know if you still think it is an issue.

**Comment by wing328** [comment]
@guohuang thanks for looking into the issue.

@nnathan what issue did you get when using the latest Go API client? can you show us the code/error to illustrate the issue more?

**Comment by nnathan** [comment]
Our API uses GET query parameters but the generated API (due to the bug mentioned in this ticket) ignores them since the fields in the QueryParams struct are not visible (exported).

**Comment by wing328** [comment]
Can you confirm you are using 2.2.0 or the latest master to generate the Go
API client?

**Comment by wing328** [comment]
I greped the latest Go Petstore API client and confirmedthe `QueryParams` struct has been removed:
```
[...]

PR #2568, opened 2016-04-12
Issue #2564 : Modified Go generator to provide an additional CodegenP…

neilotoole:
…arameter value 'exportParamName' that is always exported (uppercase first letter). See Issue #2564 for more detail.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/GoClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/GoClientCodegen.java
@@ -1,6 +1,7 @@
 package io.swagger.codegen.languages;
 
 import io.swagger.codegen.*;
+import io.swagger.models.parameters.Parameter;
 import io.swagger.models.properties.ArrayProperty;
 import io.swagger.models.properties.MapProperty;
 import io.swagger.models.properties.Property;
@@ -33,6 +34,11 @@ public class GoClientCodegen extends DefaultCodegen implements CodegenConfig {
 
     public GoClientCodegen() {
         super();
+
+        // We use GoCodegenParameter instead of CodegenParameter because we wish
+        // to add an extra field (exportParamName).
+        CodegenModelFactory.setTypeMapping(CodegenModelType.PARAMETER, GoCodegenParameter.class);
+
         outputFolder = "generated-code/go";
         modelTemplateFiles.put("model.mustache", ".go");
         apiTemplateFiles.put("api.mustache", ".go");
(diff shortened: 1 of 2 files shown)


wing328:
To have the param name in Pascal case, have you considered updating `toParamName` in the go generator instead? 

Ref: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/GoClientCodegen.java#L188

neilotoole:
@wing328 The problem is that we need that same param name in different case in different locations. Let's say the base name (in the swagger) of the param (which is a query param) is `user_id`. For the function argument, the Go convention is that this becomes `userId` (well, it should be `userID`, but that's another story), e.g. `myRestCall( userId int32 )`. That is, function parameters are `lowerCamelCase`. However, because we're using sling's QueryStruct mechanism, and we're passing  that struct outside of the package (to sling), the identifier must be `UpperCamel` in order to be exported, e.g. `type QueryParams struct { UserId    int32 ...`. If we were using a template library more sophisticated than mustache I could transform the identifier case just where it's needed for the struct in the template file.. However, I don't immediately see an alternative mechanism whereby we meet the requirements that:

- The identifier is `lowerCamel` for the function argument, per Go conventions
- The identifier is `UpperCamel` for the struct field, as required for export

I'm open to suggestion though!

wing328:
Definitely a valid use case to have another property storing the parameter name in Pascal case.

One approach is to override `postProcessOperations` in the Go generator
similar to [this one](https://github.com/swagger-api/swagger-codegen/blob/d84f848831b66e5941ac26ba0269432959f7ebd4/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaJAXRSServerCodegen.java#L114).

The param name in Pascal case can be stored in `vendorExtensions .x-param-name-pascal-case`.

neilotoole:
@wing328 Interesting, I hadn't considered that approach, and it certainly could be done that way. However, I do sense that as we continue to improve the Go generator (there's still a lot left to meet the full spec, e.g. polymorphic return types) that we'll be needing the "export" / "PascalCase" names in several other situations, and it might be more elegant (from the template's perspective at least) for the exported name to be a first-class citizen. Also, whether we went with this PR's approach or with the vendor extension, I personally would still be inclined to name it `exportParamName` as opposed to the `x-param-name-pascal-case`, as `exportParamName` is more information-bearing... that is, it has semantic value that makes the intended purpose clearer to a reader of the template file.

neilotoole:
@wing328 Let me know your thoughts on my comments above, and whether this PR is good to go, or if you want me to use the vendor extensions mechanism and resubmit.

wing328:
I think both your approach and my suggested approach can meet your requirement. Let me think about it and get back to you by tomorrow.

wing328:
@neilotoole one minor drawback I can think of with `exportParamName` is that other developers who are customizing codegen may mistakenly think that `{{exportParamName}}` is also available in other code generator (PHP, Ruby, etc) but it's not.

Given that `exportParamName` is only available to Go at the moment, I prefer using vendor extension to store it and make it accessible in the template via `{{vendorExtension.x-exportParamName}}`

neilotoole:
@wing328 That's a fair point, I'll go the vendorExtension route.

neilotoole:
@wing328 Created new PR per your suggestion: https://github.com/swagger-api/swagger-codegen/pull/2591

Issue #2221, opened 2016-02-23
Typescript Angular Client validates boolean parameter with !operator


webron:
_From @AndreasM009 on February 23, 2016 7:23_

When creating a Typescript Angular Client which includes methods that has a boolean Parameter, the boolean Parameter is verified using the !Operator. It is not possible to set the boolean Parameter to a false value. Setting the boolean Parameter to false will cause an exception.

Example:
// verify required parameter 'aProductMakerId' is set
            if (!aBooleanParameter) {
                throw new Error('Missing required parameter aBooleanParameter when calling xyz');
            }

_Copied from original issue: swagger-api/swagger-editor#825_

wing328:
@AndreasM009 thanks for the reporting the issue.

In TypeScript, how should one validate the boolean parameter to ensure it's provider by the function caller?

(Disclaimer: not an expert in TypeScript)

wing328:
@AndreasM009 what about using the following instead?
```
if (aBooleanParameter == null) {
```

AndreasM009:
I think you should compare it to 'undefined'.
undefined === aBoolean
________________________________
Von: wing328<mailto:notifications@github.com>
Gesendet: ‎29.‎02.‎2016 06:42
An: swagger-api/swagger-codegen<mailto:swagger-codegen@noreply.github.com>
Cc: AndreasM009<mailto:andreas.mock@live.com>
Betreff: Re: [swagger-codegen] Typescript Angular Client validates boolean parameter with !operator (#2221)

@AndreasM009 what about using the following instead?
```
if (aBooleanParameter == null) {
```

---

wing328:
Here is the discussion to the solution I proposed: http://stackoverflow.com/a/28984306/677735

pixelshaded:
My two cents:

It depends on if we want to support null or not. Right now, the spec itself does not:
https://github.com/OAI/OpenAPI-Specification/issues/229
And this would sway me towards throwing on null params.

But looking in to the future, I think there is a good chance the spec will support nulls. Many APIs give meaning to receiving null data (not my preference personally). 

The spirit of the code is to validate that data was passed. If we get null, data has still been technically passed, but we also want to generate clients that protect the consumer as much as possible. Take the example of an endpoint that expects an array. An empty array might be valid, but a null array might not be.

For the sake of the consumer, I think we should throw on nulls, hence the loose equality check on null.

wing328:
Close via #2574

PR #2574, opened 2016-04-12
[Typescript-Node] Verify required parameter is not undefined instead of falsey.

pixelshaded:
This came up when needing to pass a bool as a param. If the bool was false, the API would throw saying it wasn't supplied, but false is a valid value for the input.

### DIFF
--- a/samples/client/petstore/typescript-node/api.ts
+++ b/samples/client/petstore/typescript-node/api.ts
@@ -355,7 +355,7 @@ export class PetApi {
 
 
         // verify required parameter 'petId' is not null or undefined
-        if (petId == null) {
+        if (petId === null || petId === undefined) {
             throw new Error('Required parameter petId was null or undefined when calling deletePet.');
         }
 
@@ -521,7 +521,7 @@ export class PetApi {
 
 
         // verify required parameter 'petId' is not null or undefined
-        if (petId == null) {
+        if (petId === null || petId === undefined) {
             throw new Error('Required parameter petId was null or undefined when calling getPetById.');
         }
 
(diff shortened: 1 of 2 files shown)


samples/client/petstore/typescript-node/api.ts
```diff
@@ -25,6 +25,20 @@ export class Dog extends Animal {
     "breed": string;
 }
 
+export class FormatTest {
```
pixelshaded: Im not sure why this is here. It is causing a compile error because ByteArray is undefined. I can't figure out what is generating this.

pixelshaded:
I noticed when generating the code that it is now adding a new model FormatTest causing a compile error because of the ByteArray type. FormatTest is not used anywhere in the output file so I'm not even sure why it is being added.

samples/client/petstore/typescript-node/api.ts
wing328: It's for testing different type and format. 

Please update typescript script to use `modules/swagger-codegen/src/test/resources/2_0/petstore.yaml` (the original Petstore spec) for the time being.

wing328: or you can try doing a rebase on the latest master (which should have fixed the issue in the typescript code generator)

wing328:
@pixelshaded Is this related to https://github.com/swagger-api/swagger-codegen/issues/2221 ?

pixelshaded:
Yes this was a fix for that issue.

pixelshaded:
Not sure if we want to check against null. If I remember correctly, the swagger spec does not support null parameters, so this might be a reason to enforce it.

PR #2590, opened 2016-04-13
added import tests for Go

guohuang:
@wing328 please review


wing328:
@guohuang thanks. Looks good as always :)

Issue #2582, opened 2016-04-13
[Go] naming convention for Go files


wing328:
Instead of using Pascal case (e.g. PetApi.go), we want to switch to underscore/snake case (e.g. pet_api.go) following https://github.com/google/go-github/tree/master/github

Here is  good starting point: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/GoClientCodegen.java#L221
(hint: toModelFileName should return snake case while toModelName should camelize the model file name by leveraging toModelFileName)

We also need to override `toApiFilename` e.g. https://github.com/swagger-api/swagger-codegen/blob/c8f4edff68c814fe174397619ba400f09a361230/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RubyClientCodegen.java#L491

This should be a non-breaking change.

wing328:
Done thanks to the contribution from @guohuang

PR #2592, opened 2016-04-13
issue#2582: change go api file name to snake case

guohuang:
For #2582 

@wing328 please review, i am not sure if you want the model name to be lower case or not, let me know if you want to change it.

image


wing328:
@guohuang yes, we need that for model file name as well because stripe-go and go-github uses snake case for model file name:

https://github.com/stripe/stripe-go
https://github.com/google/go-github/tree/master/github

Thanks!

Issue #2579, opened 2016-04-13
[Go] Add configurable user agent to Go API client


wing328:
Similar https://github.com/swagger-api/swagger-codegen/pull/2376, we want to make the user agent in Go API client configurable.

A new property UserAgent should be added to the Configuration struct.

If no user-agent is supplied via CLI or config, it should be default to `Swagger-Codegen/{version}/go`

wing328:
Feature contributed by @guohuang

PR #2593, opened 2016-04-13
isssue#2579: added "UserAgent" in configuration

guohuang:
For #2579 @wing328 I added "UserAgent" to config struct, i tried running the CLI with `--http-user-agent` it replaced the default value as expected.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/go/configuration.mustache
+++ b/modules/swagger-codegen/src/main/resources/go/configuration.mustache
@@ -29,7 +29,7 @@ func NewConfiguration() *Configuration {
         DefaultHeader: make(map[string]string),
         ApiKey: make(map[string]string),
         ApiKeyPrefix: make(map[string]string),
-        UserAgent: "{{#httpUserAgent}}{{{.}}}{{/httpUserAgent}}{{^httpUserAgent}}Swagger-Codegen/{{{version}}}/go{{/httpUserAgent}}",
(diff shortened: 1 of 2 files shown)


modules/swagger-codegen/src/main/resources/go/configuration.mustache
```diff
@@ -27,7 +28,8 @@ func NewConfiguration() *Configuration {
         Debug: false,
         DefaultHeader: make(map[string]string),
         ApiKey: make(map[string]string),
-        ApiKeyPrefix: make(map[string]string) ,
+        ApiKeyPrefix: make(map[string]string),
+        UserAgent: "{{#httpUserAgent}}{{{.}}}{{/httpUserAgent}}{{^httpUserAgent}}Swagger-Codegen/{{{version}}}/go{{/httpUserAgent}}",
```
wing328: @guohuang I think {{version}} (the API spec version) should be replaced by {{packageVersion}} (SDK version)

guohuang: it is inconsistent, in php, it uses `{artifactVersion}}`, anyway, the code has been updated, it worked.

wing328: Yes, different languages have different CLI name for option. For perl, it's moduleVersion. For Ruby, it's gemVersion.

I did it that way because Perl developers would have no idea what artifactVersion is and moduleVersion should be more familiar to them, and similarly for other languages.

wing328:
@guohuang thanks for the contribution as always.

PR #2596, opened 2016-04-14
[Swift] minor spacing cleanup in Models.mustache

jaz-ah:


wing328:
@jaz-ah thanks for the PR. Looks good to me.

Issue #2374, opened 2016-03-13
springmvc codegen not creating @apiresponse annotations with 'response' field when multiple responses


scottball1968:
I have an api that has paths that return multiple possible responses that contain objects.  Unlike issue #1877 these responses all return the same object.  Generation code with springmvc codegen is not creating the @apiresponse annotations with the 'response' field completed for the return object.  It does complete the @apioperation annotation with a response field completed.

Swagger-UI and Springfox then do not create api-docs for response models for the non-default response; it does list all the specified responses.

Generating code with the jaxrs jersey language destination does create the expected annotations with response fields set properly.  

Here is a section of my swagger spec of my api,
```
swagger: '2.0'

info:
  version: "1.1 WIP"
  title: Auth
  description: 'Sapphire Authentication & Authorization API (Adding Patient User Groups)'

schemes:
  - https

securityDefinitions:
  token:
    type:  apiKey
    name: auth_token
    in: header
  
paths:
  /authentication/login:
    post:
      tags:
        - authentication
      operationId: login
      description: Authenticates and establishes a session
      consumes:
        - application/json
      produces:
        - application/json
      parameters:
        - name: Credentials
          in: body
          description: credentials used to authenticate user to sapphire
          required: true
          schema:
            $ref: "#/definitions/Credentials"
      responses:
        500:
          description: unexpected error, response code set 0
          schema:
            $ref: "#/definitions/LoginResponse"
        401:
          description: unauthorized, response code 2(account locked), 3(invalid credentials), 4(password expired)
        200:
          description: Successful Login, response code set 1
          schema:
            $ref: "#/definitions/LoginResponse"
definitions:
  Credentials:
    type: object
    required:
      - username
      - password
      - applicationId
    properties:
      username:
        type: string
        example: sapphireclin
      password:
        type: string
      applicationId:
        type: string
        description: The application to which the user is authenticating
  LoginResponse:
    type: object
    required:
    - responseCode
    - token
    - userPermissions
    - userLogin
    - userFirstName
    - userLastName
    - userEmailAddress
    - userTopOrgId
    - passwordMinLength
    properties:
      responseCode:
        type: integer
        format: int32
        description: 1 = Success, 2 = Invalid Credentials, 3 = Password Expired
      token:
        $ref: "#/definitions/Token"
      userPermissions:
        type: array
        items:
          type: string
      userLogin:
        type: string
      userFirstName:
        type: string
      userLastName:
        type: string
      userEmailAddress:
[...]

PR #2603, opened 2016-04-14
[issue-2374][koly] add return class in @apiresponse and use java lang…

kolyjjj:
* added response in '@ApiResponse' in api.mustache
* remove languageSpecificPrimitives in SpringMVCServerCodegen. It is because the super class of SpringMVCServerCodegen has a languageSpecificPrimitives which has all the value and contains more (the Void type), so it make sense to remove it in the child class, otherwise the Void will not be escaped when generating imports.


wing328:
For #2374

PR #2613, opened 2016-04-15
[swift] Check server response and fail request in case of non-2xx response

tomekc:


tomekc:
To my surprise, in case of error server response (4xx for example) there was no failure.
I was using promise-based api, and promise is always fulfilled, even in case of error response.

To my bigger surprise, it is default Alamofire behavior to remain silent on error. Simply adding validate() to call chain made this straight.

jaz-ah:
+1

wing328:
All test cases passed. PR merged.

PR #2616, opened 2016-04-15
add variable for listContainerType

psev:
Add two variables called returnContainerType and listDatatype.
Also fixed inconsistent indentation near the bottom of _AbstractTypeScriptClientCodegen.java_.

This is required because `import * from "library"` does not work in TypeScript. The required format is to import a library into a namespace, e.g. `import * as lib from "library"`.

The conflict occurs when dealing with containers and custom datatypes. Using the `datatype` and `returnType` variables return something along the lines of `Array<MyClass>` when `Array<lib.MyClass>` is required.

This commit allows TypeScript templates to be created with the proper format.


wing328:
@psev thanks for the PR. The first comment I've is to use 4-space indention for your changes in the Java class files.

Ref: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

psev:
It looks like the Java8 test failed to run at all.

wing328:
@psev do you mind providing a spec for me to easily reproduce the issue? and thanks again for the work to introduce additional boolean flags for different use cases.

wing328:
@psev thanks again for the PR. Closing this as there's no update.

Note: TypeScriptAngularClientCodegen has been renamed to TypeScriptAngularJSClientCodegen

Issue #2611, opened 2016-04-15
[Go] Add pom.xml to automate testing for Go API client (Petstore)


wing328:
Add a pom.xml (similar to Ruby Petstore [pom.xml](https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/ruby/pom.xml)) so that we can later include Go API client as part of the CI.

The pom.xml for Go Petstore client should be in the folder https://github.com/swagger-api/swagger-codegen/tree/master/samples/client/petstore/go

wing328:
PR merged. Thanks @guohuang

PR #2617, opened 2016-04-15
issue#2611: added pom.xml for building Go

guohuang:
For #2611, @wing328, I couldn't find any better way to build the Go code except using ant run. let me know if you have a better way, i am new to maven.


guohuang:
@wing328 updated the pom file as you request, the unit test was code was fine, however, the `GetPetById` call did not return anything,(probably, `AddPet` failed) that is why it is failing, i have commented it out for now, it will be great if we can find the root cause of it.

PR #2634, opened 2016-04-18
Append reserved swift enum type with "Enum"

yonaskolb:
This replaces something like `property:_Type` with `property:TypeEnum`, which is more idiomatic swift


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
```diff
@@ -337,7 +337,7 @@ public CodegenProperty fromProperty(String name, Property p) {
             // Swift compiler will generate an error
             if (isReservedWord(codegenProperty.datatypeWithEnum) ||
                     name.equals(codegenProperty.datatypeWithEnum)) {
-                codegenProperty.datatypeWithEnum = escapeReservedWord(codegenProperty.datatypeWithEnum);
+                codegenProperty.datatypeWithEnum = codegenProperty.datatypeWithEnum + "Enum";
                     }
```
wing328: do you mind indenting the line with 4 additional spaces to make it align with the previous line?

PR #2637, opened 2016-04-18
[Swift] Force required attrs to be defined with unwrapRequired

### LINKED ISSUE #2116: shouldn't the fact a property of a schema object is required reflect in the model object itself?
For example in the Java binding, we do have the annotation that indicates the property is required, but if I'm not mistaken you can still not set it and call the API without the model object warning you you have not provided a required property?

For swift binding the property could be a regular one instead of being a swift optional property to force the user of the API to provide it?

fabdslv:
Fix #2116

When using unwrapRequired option in Swift, we can now force the required attributes to be defined when instantiating a model.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/swift/Models.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift/Models.mustache
@@ -137,9 +137,22 @@ class Decoders {
             // Decoder for {{{classname}}}
             Decoders.addDecoder(clazz: {{{classname}}}.self) { (source: AnyObject) -> {{{classname}}} in
                 let sourceDictionary = source as! [NSObject:AnyObject]
+                {{#unwrapRequired}}
+                let instance = {{classname}}({{#requiredVars}}{{^-first}}, {{/-first}}{{#isEnum}}{{name}}: {{classname}}.{{datatypeWithEnum}}(rawValue: (sourceDictionary["{{baseName}}"] as? String) ?? "")! {{/isEnum}}{{^isEnum}}{{name}}: Decoders.decode(clazz: {{{baseType}}}.self, source: sourceDictionary["{{baseName}}"]!){{/isEnum}}{{/requiredVars}})
+                {{#optionalVars}}
+                {{#isEnum}}
+                instance.{{name}} = {{classname}}.{{datatypeWithEnum}}(rawValue: (sourceDictionary["{{baseName}}"] as? String) ?? "")
+                {{/isEnum}}
+                {{^isEnum}}
+                instance.{{name}} = Decoders.decodeOptional(clazz: {{{baseType}}}.self, source: sourceDictionary["{{baseName}}"])
+                {{/isEnum}}
+                {{/optionalVars}}
+                {{/unwrapRequired}}
+                {{^unwrapRequired}}
                 let instance = {{classname}}(){{#vars}}{{#isEnum}}
-                instance.{{name}} = {{classname}}.{{datatypeWithEnum}}(rawValue: (sourceDictionary["{{baseName}}"] as? String) ?? ""){{#unwrapRequired}}{{#required}}!{{/required}}{{/unwrapRequired}} {{/isEnum}}{{^isEnum}}
-                instance.{{name}} = Decoders.decode{{^unwrapRequired}}Optional{{/unwrapRequired}}{{#unwrapRequired}}{{^required}}Optional{{/required}}{{/unwrapRequired}}(clazz: {{{baseType}}}.self, source: sourceDictionary["{{baseName}}"]{{#unwrapRequired}}{{#required}}!{{/required}}{{/unwrapRequired}}){{/isEnum}}{{/vars}}
+                instance.{{name}} = {{classname}}.{{datatypeWithEnum}}(rawValue: (sourceDictionary["{{baseName}}"] as? String) ?? "") {{/isEnum}}{{^isEnum}}
+                instance.{{name}} = Decoders.decodeOptional(clazz: {{{baseType}}}.self, source: sourceDictionary["{{baseName}}"]){{/isEnum}}{{/vars}}
+                {{/unwrapRequired}}
                 return instance
             }{{/model}}
 			{{/models}}
--- a/modules/swagger-codegen/src/main/resources/swift/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift/model.mustache
@@ -21,15 +21,26 @@ public class {{classname}}: JSONEncodable {
 {{#vars}}
 {{#isEnum}}
     {{#description}}/** {{description}} */
-    {{/description}}public var {{name}}: {{{datatypeWithEnum}}}{{^unwrapRequired}}?{{/unwrapRequired}}{{#unwrapRequired}}{{^required}}?{{/required}}{{#required}}!{{/required}}{{/unwrapRequired}}{{#defaultValue}} = {{{defaultValue}}}{{/defaultValue}}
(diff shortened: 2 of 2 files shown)


cjolif:
I think in model.mustache you should remove the exclamation mark from the required fields. This would  create "real" mandatory fields and not implicitly unwrapped fields. This was not possible before because of the lack of constructor to initialize  them, but now you added it.

yonaskolb:
You don't need the isConstructorFirstArg, you can just use the hasMore field. eg put this at the end `{{#hasMore}}, {{/hasMore}}{{/vars}}`

wing328:
You can also use the following to access the first and last elements in a list:
```
{{#vars}}{{#-first}} this is the first element {{.}} {{/-first}}{{/vars}}
{{#vars}}{{#-last}} this is the last element {{.}} {{/-last}}{{/vars}}
```

fabdslv:
@yonaskolb this doesn't work since I am iterating over `vars` but filtering it with `required`, which means 'hasMore`is not consistant in this case.

@wing328 this does not work either for the same reason. I saw there is a `mandatory` array to iterate over required attributes but these only contain strings and I don't think there is a way to access `vars` dynamically with keys from `mandatory` in Mustache.

yonaskolb:
Yes, sorry you're right. Having said that, you could also add the non required properties to the initializer with default values of nil. That's what I do in my custom templates

wing328:
@fabdslv what about adding "requiredVars" and "optionalVars" instead as I've seen a pattern that more and more langauges require separated access to required and optional attributes/properties?

cjolif:
> @fabdslv what about adding "requiredVars" and "optionalVars" instead as I've seen a pattern that more and more langauges require separated access to required and optional attributes/properties?

:+1:

fabdslv:
@wing328 sure, I think it's the best solution. I didn't want to create redundancy because there's already a `mandatory` array, but I don't think it used somewhere else than in some tests cases. So maybe this array should be replaced with a `requiredVars` array? Could you please check this?

wing328:
@fabdslv `mandatary` is introduced for JavaScript. Can you please use `{{#mandatory}}` to replace `{{#vars}}{{#required}}` (and then use {{^-first}} to replace {{^isConstructorFirstArg}}) to see if it works for you?

You can use `mandatory` for the time being and I'll rename it to `requiredVars` later when introducing `optionalVars`, or you can rename it to `requiredVars` and also update the JS mustache template. I'll leave it up to you to decided.

The mustache tags need some enhancements and clean up and I hope to do it in the upcoming stable release v2.2.0.

fabdslv:
@wing328 As I said before, using `mandatory` is not an option because its list elements type is `String` and not `CodegenProperty` (or maybe it is if we can access `vars` with each `mandatory` value in the template?). Maybe we can leave it this way until you bring `requiredVars` and `optionalVars`?

wing328:
@fabdslv sorry I missed that. How urgent to you need this change to get into master?

I should have some time this weekend to implement `requiredVars` and `optionalVars` unless someone from the community has cycle to implement it before that.

fabdslv:
@wing328 Great, I'll update the PR to use `requiredVars` when it will be available.

fabdslv:
I also removed the forced unwrapping. It is now useless since we force the required attributes to be defined when instantiating the object. Maybe the `unwrapRequired` option should be renamed to `forceRequired` for example?

wing328:
@fabdslv I've filed #2691 to add requiredVars and optionalVars to codegen model. 

**IMPORTANT: please do not use {{#hasMore}} in requiredVars, optionalVars as it's misleading/incorrect. We'll later file a PR to remove {{#hasMore}} and use {{^-last}} instead.**

fabdslv:
@wing328 thanks for your update on this. I updated the PR to use `requiredVars` and `optionalVars`.

modules/swagger-codegen/src/main/resources/swift/Models.mustache
```diff
@@ -137,9 +137,22 @@ class Decoders {
             // Decoder for {{{classname}}}
             Decoders.addDecoder(clazz: {{{classname}}}.self) { (source: AnyObject) -> {{{classname}}} in
                 let sourceDictionary = source as! [NSObject:AnyObject]
+                {{#unwrapRequired}}
+                let instance = {{classname}}({{#requiredVars}}{{^-first}}, {{/-first}}{{#isEnum}}{{name}}: {{classname}}.{{datatypeWithEnum}}(rawValue: (sourceDictionary["{{baseName}}"] as? String) ?? "")! {{/isEnum}}{{^isEnum}}{{name}}: Decoders.decode(clazz: {{{baseType}}}.self, source: sourceDictionary["{{baseName}}"]!){{/isEnum}}{{/requiredVars}})
```
cjolif: It seems only instantiation is different and that the rest is identical between the two cases. Can't we factorize that instead of duplicating the common lines?

fabdslv: I don't think so, because we iterate over different lists (`optionalVars` vs `vars`) and with Mustache it will force us to create two different cases anyway

cjolif: ok right, sorry for having missed that.

wing328:
@fabdslv thanks for the PR. All test cases passed.
```
Test Suite 'UserAPITests' started at 2016-04-27 13:24:04.791
Test Case '-[SwaggerClientTests.UserAPITests test1CreateUser]' started.
Test Case '-[SwaggerClientTests.UserAPITests test1CreateUser]' passed (0.566 seconds).
Test Case '-[SwaggerClientTests.UserAPITests test2GetUser]' started.
Test Case '-[SwaggerClientTests.UserAPITests test2GetUser]' passed (0.491 seconds).
Test Case '-[SwaggerClientTests.UserAPITests test3DeleteUser]' started.
Test Case '-[SwaggerClientTests.UserAPITests test3DeleteUser]' passed (0.490 seconds).
Test Case '-[SwaggerClientTests.UserAPITests testLogin]' started.
Test Case '-[SwaggerClientTests.UserAPITests testLogin]' passed (0.497 seconds).
Test Case '-[SwaggerClientTests.UserAPITests testLogout]' started.
Test Case '-[SwaggerClientTests.UserAPITests testLogout]' passed (0.479 seconds).
Test Suite 'UserAPITests' passed at 2016-04-27 13:24:07.322.
	 Executed 5 tests, with 0 failures (0 unexpected) in 2.524 (2.531) seconds
Test Suite 'SwaggerClientTests.xctest' passed at 2016-04-27 13:24:07.325.
	 Executed 11 tests, with 0 failures (0 unexpected) in 6.481 (6.508) seconds
[INFO] BUILD SUCCESS
[INFO] Total time: 02:38 min
[INFO] Finished at: 2016-04-27T13:24:07+08:00
[INFO] Final Memory: 11M/245M
```

modules/swagger-codegen/src/main/resources/swift/model.mustache
```diff
@@ -21,15 +21,26 @@ public class {{classname}}: JSONEncodable {
...
+    {{/description}}public var {{name}}: {{{datatype}}}{{^unwrapRequired}}?{{/unwrapRequired}}{{#unwrapRequired}}{{^required}}?{{/required}}{{/unwrapRequired}}{{#defaultValue}} = {{{defaultValue}}}{{/defaultValue}}
 {{/isEnum}}
 {{/vars}}
 
+{{^unwrapRequired}}
     public init() {}
+{{/unwrapRequired}}
+{{#unwrapRequired}}
+    public init({{#requiredVars}}{{^-first}}, {{/-first}}{{name}}: {{#isEnum}}{{datatypeWithEnum}}!{{/isEnum}}{{^isEnum}}{{datatype}}!{{/isEnum}}{{/requiredVars}}) {
+        {{#vars}}
```
fabdslv: There's a mistake here : this line should be removed.

fabdslv:
@wing328 I just noticed there's a mistake in the ```init()``` function, I will fix it in the same branch and create a new PR with the fix. Sorry about that.

wing328:
@fabdslv that's ok and thanks for the quick turnaround.

PR #2640, opened 2016-04-18
[Swift] Enum parameters are now handled

### LINKED ISSUE #2531: [Swift] enum parameters do not generate enums
```js         
               "name": "paramName"
               "required": false,
               "type": "string",
               "enum": [
                   "one",
                   "two"
               ]
```
generates:

```swift
    public class func action(paramName: String?)
```

while ideally it should generates:

```swift
   enum ParamName: String {
      case One = "one"
      case Two = "two"
   }
   public class func action(paramName: ParamName?)
```

fabdslv:
Fix #2531

### DIFF
--- a/modules/swagger-codegen/src/main/resources/swift/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift/api.mustache
@@ -16,6 +16,18 @@ extension {{projectName}}API {
 /** {{description}} */{{/description}}
 public class {{classname}}: APIBase {
 {{#operation}}
+    {{#allParams}}
+    {{#isEnum}}
+    /**
+
+    enum for parameter {{paramName}}
+    */
+    public enum {{{datatypeWithEnum}}}_{{operationId}}: String { {{#allowableValues}}{{#values}}
+        case {{enum}} = "{{raw}}"{{/values}}{{/allowableValues}}
+    }
+
+    {{/isEnum}}
+    {{/allParams}}
     /**
      {{#summary}}
      {{{summary}}}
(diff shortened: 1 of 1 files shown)


yonaskolb:
This won't compile if multiple operations have the same enum. Maybe you could prepend the operationId to the enum type?

wing328:
@fabdslv thanks for the PR.

@yonaskolb thanks for reviewing.

I've a PR for C# and PHP (https://github.com/swagger-api/swagger-codegen/pull/2508) and should work for multiple operations with the same enum defined.

I'll try to incorporate your change into my PR.

yonaskolb:
Well it will break in swift if there are multiple operations with the same tag that share any enums with the same name, as they are in the same scope

yonaskolb:
The rawValue of the enum will also need to be extracted when added to the path parameters or the nillableParameters, otherwise they will use the printed escaped enum name which may be different from the actual value

fabdslv:
@yonaskolb Yes, it might be a good solution to add the operational id. I'm not sure I understand your last comment though, what do you mean by "extract"?

yonaskolb:
well for the path params instead of 
`"\({{paramName}})"`
use
`"\({{paramName}}{{#isEnum}}.rawValue{{/isEnum}})"`

and for the query params instead of
`"{{baseName}}": {{paramName}}` 
use 
`"{{baseName}}": {{paramName}}}{{#isEnum}}{{^required}}?{{/required}}.rawValue{{/isEnum}}`

I hope that makes sense. That's what I've done in my custom templates

wing328:
@yonaskolb when you've time, do you mind submitting your custom template change with a PR so that we can review and compare? Thanks!

fabdslv:
I updated the PR to integrate what @yonaskolb was mentioning. Apart from the fact that it fixes the issues, now the Enum name is ```{{{datatypeWithEnum}}}_{{operationId}}``` so that it fixes the conflicts we might encounter.

@wing328 It seems that some changes were made in master that affect my commit and I cannot rebase from master. Could you please take a look at this if you have time? Thanks.

modules/swagger-codegen/src/main/resources/swift/api.mustache
```diff
@@ -16,14 +16,26 @@ extension {{projectName}}API {
 /** {{description}} */{{/description}}
 public class {{classname}}: APIBase {
 {{#operation}}
+    {{#allParams}}
+    {{#isEnum}}
+    /**
+
+    enum for parameter {{paramName}}
+    */
+    public enum {{{datatypeWithEnum}}}_{{operationId}}: String { {{#allowableValues}}{{#values}}
```
cjolif: This is solving the possible clash issue. But for people that don't have a clash, this is created a cumbersome API. Maybe we should add some logic Java-side to know if we have a conflict and do that only in that case?

fabdslv:
@wing328 I managed to rebase, it should be okay now.

modules/swagger-codegen/src/main/resources/swift/api.mustache
wing328: @cjolif I agree with your comment.

@fabdslv I've a pending PR (https://github.com/swagger-api/swagger-codegen/pull/2508) that should be able to handle the situation.

wing328:
@fabdslv thanks! I'll merge this into https://github.com/swagger-api/swagger-codegen/pull/2508 so that we only have javascript and typescript to add better enum support. Hopefully that can be done this weekend...

PR #2629, opened 2016-04-17
Add unescaped Parameter description

yonaskolb:
This allows you to use CodegenParameter.unescapedDescription which will allow for multiline descriptions. This matches the recent changes in Model and Property

PR #2642, opened 2016-04-19
add unescapedNotes to CodegenOperation

yonaskolb:
This matches the recent changes in Model, Property and Parameter https://github.com/swagger-api/swagger-codegen/pull/2629


yonaskolb:
By the way, why is the property called notes instead of description? I would have changed notes to description and unescapedDescription, but that would have been a breaking change

wing328:
@yonaskolb good question but I don't have an answer as it was already using "note" when I started working on this project a year ago.

Given that it will be a breaking change, I suggest we keep it as it's for the time being.

wing328:
@yonaskolb let me know if you gonna do the same for "summary" or you would like someone from the community to work on that instead. Thanks again for the PRs.

Issue #2621, opened 2016-04-16
auto-generated Node.js server code has a error if the API path contains underline(_)


lienhua34:
I have a swagger api spec file, which content is
```
swagger: '2.0'
info:
  version: "0.0.0"
  title: "Hello World API"
paths:
  /hello/{user_id}:
    get:
      description: |
        Returns a greeting to the user
      parameters:
        -
          name: user_id
          in: path
          description: The name of the user to greet
          required: true
          type: string
      responses:
        200:
          description: Return the greeting.
          schema:
            type: string
        400:
          description: Invalid characters in "user" were provided.
```
I edit this file in http://editor.swagger.io, and use “generate server --> Node.js” to generate Node.js server stub code.

After I start the server, I open another terminal and use the following command to call the api,
   $ curl http://localhost:8080/hello/lienhua34

The server responses a error. The error is 
```
lienhua34@nodejs-server-server$ node index.js 
Your server is listening on port 8080 (http://localhost:8080)
Swagger-ui is available on http://localhost:8080/docs
Error: Cannot resolve the configured swagger-router handler: Default_helloUser_idGET
    at swaggerRouter (/home/lienhua34/Downloads/nodejs-server-server/node_modules/swagger-tools/middleware/swagger-router.js:414:18)
    at call (/home/lienhua34/Downloads/nodejs-server-server/node_modules/connect/index.js:239:7)
    at next (/home/lienhua34/Downloads/nodejs-server-server/node_modules/connect/index.js:183:5)
    at /home/lienhua34/Downloads/nodejs-server-server/node_modules/swagger-tools/middleware/swagger-validator.js:350:30
    at /home/lienhua34/Downloads/nodejs-server-server/node_modules/swagger-tools/node_modules/async/lib/async.js:52:16
    at /home/lienhua34/Downloads/nodejs-server-server/node_modules/swagger-tools/node_modules/async/lib/async.js:361:13
    at /home/lienhua34/Downloads/nodejs-server-server/node_modules/swagger-tools/node_modules/async/lib/async.js:52:16
    at async.forEachOf.async.eachOf (/home/lienhua34/Downloads/nodejs-server-server/node_modules/swagger-tools/node_modules/async/lib/async.js:236:30)
    at _asyncMap (/home/lienhua34/Downloads/nodejs-server-server/node_modules/swagger-tools/node_modules/async/lib/async.js:355:9)
[...]

PR #2655, opened 2016-04-20
include underscore when generating nodejs controller and service method

kolyjjj:
This is a fix for [issue-2621](https://github.com/swagger-api/swagger-codegen/issues/2621)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NodeJSServerCodegen.java
```diff
@@ -314,4 +314,10 @@ public void serialize(Double val, JsonGenerator jgen,
         }
         return super.postProcessSupportingFileData(objs);
     }
+
+    @Override
+    public String removeNonNameElementToCamelCase(String name) {
```
wing328: would be nice to add a docstring explaining what the function does, parameter description, etc

Issue #2652, opened 2016-04-20
Load Generator using ServiceLoader to enable alternative generators


alexec:
This would be really useful, allowing you to write your own generators:

```
Alexs-MacBook-Pro:swagger-codegen alex.e.c$ git diff modules/
diff --git a/modules/swagger-codegen-cli/src/main/java/io/swagger/codegen/cmd/Generate.java b/modules/swagger-codegen-cli/src/main/java/io/swagger/codegen/cmd/Generate.java
index 8c7beab..4b6f511 100644
--- a/modules/swagger-codegen-cli/src/main/java/io/swagger/codegen/cmd/Generate.java
+++ b/modules/swagger-codegen-cli/src/main/java/io/swagger/codegen/cmd/Generate.java
@@ -4,11 +4,13 @@ import io.airlift.airline.Command;
 import io.airlift.airline.Option;
 import io.swagger.codegen.ClientOptInput;
 import io.swagger.codegen.CodegenConstants;
-import io.swagger.codegen.DefaultGenerator;
+import io.swagger.codegen.Generator;
 import io.swagger.codegen.config.CodegenConfigurator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;

+import java.util.ServiceLoader;
+
 import static io.swagger.codegen.config.CodegenConfiguratorUtils.*;
 import static org.apache.commons.lang3.StringUtils.isNotEmpty;

@@ -220,6 +222,8 @@ public class Generate implements Runnable {

         final ClientOptInput clientOptInput = configurator.toClientOptInput();

-        new DefaultGenerator().opts(clientOptInput).generate();
+        for (Generator generator : ServiceLoader.load(Generator.class)) {
+            generator.opts(clientOptInput).generate();
+        }
     }
 }
diff --git a/modules/swagger-codegen-cli/src/main/resources/META-INF/services/io.swagger.codegen.Generator b/modules/swagger-codegen-cli/src/main/resources/META-INF/services/io.swagger.codegen.Generator
index e69de29..92623d0 100644
--- a/modules/swagger-codegen-cli/src/main/resources/META-INF/services/io.swagger.codegen.Generator
+++ b/modules/swagger-codegen-cli/src/main/resources/META-INF/services/io.swagger.codegen.Generator
@@ -0,0 +1 @@
+io.swagger.codegen.DefaultGenerator
\ No newline at end of file
```

wing328:
[...]

PR #2664, opened 2016-04-20
Allowed you to provide your own code generator.

alexec:
This is a PR for https://github.com/swagger-api/swagger-codegen/issues/2652 as requested.

Why would I want to do this? The `DefaultGenerator` creates some specific outputs:

* APIs
* Models
* Supporting file

I'd like to be able to create other things, such as per API endpoint files (see below). This allows me to do this, without making any complex changes to the `DefaultGenerator`.

~~~java
public class HmrcGenerator extends DefaultGenerator {

    @Override
    public List<File> generate() {
        List<File> files = super.generate();
        files.addAll(generateEndpoints());
        return files;
    }

    private List<File> generateEndpoints() {
        List<File> files = new ArrayList<File>();
        Map<String, List<CodegenOperation>> paths = processPaths(swagger.getPaths());
        for (List<CodegenOperation> operations : paths.values()) {

            String summary = operations.get(0).summary;
            String filename = String.format("%s/%s/documentation/%s.xml", config.getOutputDir(), swagger.getInfo().getVersion(), summary).replace(" ", "-");
            String templateFile = getFullTemplateFile(config, "endpoint.mustache");
            String template = readTemplate(templateFile);

            Template tmpl = Mustache.compiler()
                    .withLoader(new Mustache.TemplateLoader() {
                        @Override
                        public Reader getTemplate(String name) {
                            return getTemplateReader(getFullTemplateFile(config, name + ".mustache"));
                        }
                    })
                    .defaultValue("")
                    .compile(template);
            try {
                Map<String, List<CodegenOperation>> context = Collections.singletonMap("operations", operations);
                writeToFile(filename, tmpl.execute(context));
            } catch (IOException e) {
                throw new RuntimeException(e);
            }
            files.add(new File(filename));
        }
        return files;
    }
}
~~~

### DIFF
--- a/modules/swagger-codegen-cli/src/main/java/io/swagger/codegen/cmd/Generate.java
+++ b/modules/swagger-codegen-cli/src/main/java/io/swagger/codegen/cmd/Generate.java
@@ -4,7 +4,8 @@ import io.airlift.airline.Command;
 import io.airlift.airline.Option;
 import io.swagger.codegen.ClientOptInput;
 import io.swagger.codegen.CodegenConstants;
-import io.swagger.codegen.DefaultGenerator;
+import io.swagger.codegen.Generator;
+import io.swagger.codegen.GeneratorLoader;
 import io.swagger.codegen.config.CodegenConfigurator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


wing328:
@alexec so the API endpoint files are the XML documentation files, right ?

PR #2676, opened 2016-04-21
Typescript angular2 client can be published to npm

Vrolijkx:
I extended the current typscript angular2 client so the generated code could be published to npm.
To accomplish this i added generation of the next files:
* package.json
* tsconfig.json
* typings.json

I also made the generation of these files optional. They will only be generated when a npmName is passed as option.


Vrolijkx:
This issue will resolve #2677

wing328:
@Vrolijkx thanks for the PR. The change looks good to me. I'll merge it tomorrow (Sat).

Have a nice weekend.

PR #2686, opened 2016-04-22
[Java] Update Java version to 1.7 in generated pom

### LINKED ISSUE #2607: Generated Java Client relies on java.util.Objects (since 1.7), pom.xml specifies Java 1.6
_From @devinjones on April 14, 2016 20:58_

I generated a Java client for a swagger file and imported it as an existing maven project in eclipse.
There were compilation errors related to the `equals()` and `hashCode()` methods, caused by class not found: java.util.Objects

_Copied from original issue: swagger-api/swagger-editor#869_

skirkpatrick:
This fixes #2607 and brings the Java version in the generated pom to match the version in the generated build.gradle.

-

_Copy/pasted from #2607_

The `default.yaml` example spec available in the swagger editor is sufficient for reproducing this.

A minimalist spec would be any spec that has a `definitions` section:
```
swagger: '2.0'
info:
  version: 0.0.0
  title: Simple API
paths:
  /:
    get:
      responses:
        200:
          description: OK
definitions:
  Example:
    type: object
    properties:
      uuid:
        type: string
```

Every generated class under `io.swagger.client.model` has a reference to `java.util.Objects`, which was only added in JDK 1.7. The generated `build.gradle` specifies 1.7 as the target/source version, but the `pom.xml` specifies 1.6.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/pom.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/pom.mustache
@@ -97,8 +97,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>2.3.2</version>
         <configuration>
-          <source>1.6</source>
-          <target>1.6</target>
+          <source>1.7</source>
+          <target>1.7</target>
         </configuration>
       </plugin>
     </plugins>


wing328:
@skirkpatrick thanks for the PR. I've submitted the following PR to do the same for the remaining Java API client using different HTTP libraries (e.g. Jersey2.x, Retrofit1.x, 2.x)

https://github.com/swagger-api/swagger-codegen/pull/2685

PR #2692, opened 2016-04-24
Rewrite Api_Client.go to use resty api

guohuang:
@wing328 please review


modules/swagger-codegen/src/main/resources/go/api_client.mustache
```diff
@@ -2,6 +2,11 @@ package {{packageName}}
...
 import (
     "strings"
+    "github.com/go-resty/resty"
+    "errors"
+    "reflect"
+    "bytes"
+    "path/filepath"
 )
 
 type ApiClient struct {
```
neilotoole: This should be spelled `APIClient`  per the usual [Go conventions](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

modules/swagger-codegen/src/main/resources/go/api_client.mustache
```diff
@@ -38,4 +43,81 @@ func contains(source []string, containvalue string) bool {
         }
     }
     return false
-}
\ No newline at end of file
+}
+
+
+func (c *ApiClient) CallApi(path string, method string,
```
neilotoole: This should be `CallAPI` per Go conventions.

modules/swagger-codegen/src/main/resources/go/api_client.mustache
```diff
@@ -38,4 +43,81 @@ func contains(source []string, containvalue string) bool {
...
+        return response, err
+    case "PATCH":
+        response, err := request.Patch(path)
+        return response, err
+    case "DELETE":
+        response, err := request.Delete(path)
+        return response, err
+    }
+
+    return nil, errors.New("Invalid method " + method)
```
neilotoole: [Idiomatically:](https://github.com/golang/go/wiki/CodeReviewComments#error-strings) `return nil, fmt.Errorf("invalid method %v", method)`

modules/swagger-codegen/src/main/resources/go/api_response.mustache
```diff
@@ -0,0 +1,24 @@
+package swagger
+
+import (
+  "net/http"
+)
+
+
+type ApiResponse struct {
```
neilotoole: Should be `APIResponse` per conventions.

modules/swagger-codegen/src/main/resources/go/configuration.mustache
```diff
@@ -26,7 +26,7 @@ func NewConfiguration() *Configuration {
     return &Configuration{
         BasePath: "{{basePath}}",
         UserName: "",
-        Debug: false,
+        debug: false,
         DefaultHeader: make(map[string]string),
         ApiKey: make(map[string]string),
```
neilotoole: Should be `APIKey` and `APIKeyPrefix` per conventions.

samples/client/petstore/go/go-petstore/pet_api.go
```diff
@@ -35,89 +35,70 @@ func NewPetApiWithBasePath(basePath string) *PetApi{
  * @param body Pet object that needs to be added to the store
  * @return void
  */
-func (a PetApi) AddPet (body Pet) (error) {
-    // verify the required parameter 'body' is set
-    if &body == nil {
-        return errors.New("Missing required parameter 'body' when calling PetApi->AddPet")
-    }
-    _sling := sling.New().Post(a.Configuration.BasePath)
+func (a PetApi) AddPet (body Pet) (ApiResponse, error) {
```
neilotoole: This generated code seems to be a step backwards. It should be  `Pet` returned by this operation surely?

wing328: `AddPet` does not return any `Pet` model for 2xx response: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.json#L29

ApiResponse (should be renamed to APIResponse) contains the status code, http headers and we introduce this similar to the Go Github API client (e.g. https://github.com/google/go-github/blob/master/github/repos.go#L210) just in case the developers need access to the status code and http headers.

guohuang:
@neilotoole thanks for your valuable feedbacks, i have updated all the api names to `API`, except ModelApiResponse which we will take care of that in the next PR.

samples/client/petstore/go/go-petstore/pet_api.go
neilotoole: Ah, ok, I thinking I was looking at a different petstore spec that does indeed return a pet object, makes sense now.

PR #2703, opened 2016-04-25
added condition to validate required field, added more unit tests

guohuang:
@wing328 please review


samples/client/petstore/go/go-petstore/pet_api.go
```diff
@@ -42,8 +42,8 @@ func (a PetApi) AddPet (body Pet) (APIResponse, error) {
   path := a.Configuration.BasePath + "/pet"
 
   // verify the required parameter 'body' is set
-  if &body == nil {
-      return *NewAPIResponseWithError("400 - Bad Request"), errors.New("Missing required parameter 'body' when calling PetApi->AddPet")
+  if reflect.DeepEqual((Pet{}), body) {
```
guohuang: @neilotoole I would like to have your feedback on this regarding to validating if an object parameter is equal or not. `&body == nil` will fail my test, i am running out of idea how to check for nil in Go

modules/swagger-codegen/src/main/resources/go/api.mustache
```diff
@@ -5,7 +5,7 @@ import (
     "strings"
     "fmt"
     "encoding/json"
-    "errors"
+    "reflect"
```
guohuang: @wing328  there is a potential issue with this reference, when there is no model in the parameter(line 65 generate the required parameter validation), if we find a better way to compare null object then we don't need this reference.

samples/client/petstore/go/go-petstore/pet_api.go
neilotoole: @guohuang Unfortunately, with the general model we're using, this is probably the best we can do. I'm investigating an alternative approach as well, stay tuned.

neilotoole: @guohuang Thinking about this even further... just because a parameter is required doesn't mean it can't be the zero value. E.g. with `Pet` above, I think it would be legal for all the fields of `Pet` to be zero values (e.g. empty strings, etc). So this code is problematic.

guohuang: @neilotoole I think I agree with you, validating complex object in Go does not make sense. this is the shortcoming of Go.(not "supporting" nullable object, but on the other hand, Go wouldn't compile if i put `nil` as parameter, which is a good thing)

I have removed this validation in the template as well as the test.

modules/swagger-codegen/src/main/resources/go/api.mustache
```diff
@@ -42,15 +41,27 @@ func (a {{classname}}) {{nickname}} ({{#allParams}}{{paramName}} {{{dataType}}}{
...
 {{/pathParams}}
 
   {{#allParams}}
   {{#required}}
   // verify the required parameter '{{paramName}}' is set
-  if &{{paramName}} == nil {
-      return {{#returnType}}*new({{{returnType}}}), {{/returnType}}*NewAPIResponseWithError("400 - Bad Request"), errors.New("Missing required parameter '{{paramName}}' when calling {{classname}}->{{operationId}}")
+  {{#isListContainer}}
+  if len({{paramName}}) == 0 {
+       return {{#returnType}}*new({{{returnType}}}), {{/returnType}}*NewAPIResponseWithError("400 - Bad Request"), fmt.Errorf("missing required parameter '{{paramName}}' when calling {{classname}}->{{operationId}}")
```
wing328: Does Go have some sort of built-in exception for invalid argument? Similar to [this](http://stackoverflow.com/a/774116/677735) in C#

neilotoole: No, there isn't really a "standard library" of Errors.

wing328: In this case, returning a 400 error may mislead the developers by giving the impression that the server did return something (400) and they use tools like tcpdump, wireshark to capture the request and the response (which do not occur), and they will later figure out that there's no request (which is not a good developer experience we want to deliver)

I'll look into other Go SDKs (e.g. go-github, stripe-go) to see how they handle invalid argument exception.

wing328: I suggest we use `errors.New("missing required parameter '{{paramName}}' when calling {{classname}}->{{operationId}}")` instead


Ref: http://spf13.com/presentation/7-biggest-mistakes-in-go/, starting at slide 19

PR #2708, opened 2016-04-26
Fix ruby model Boolean Attributes not being set

brhou:
If a Boolean attribute is set as false and is passed into the model initializer, the old code will not set the model attribute at all as it only checks for truthy before setting. This fix will check for the presence of the key that is passed to the initializer to determine setting logic.


wing328:
@brhou thanks for the fix and it looks good to me. (this should also fix number/integer attribute set to 0)

Issue #2711, opened 2016-04-26
The classes should implement equals(), hashCode() and toString().


neilotoole:
As a matter of general Java hygiene, all of the modeling classes (e.g. `CodegenModel`) should implement `equals()` and `hashCode()`, especially given that they are stored/accessed via Collections.. subtle bugs can appear when dealing with collections and these methods are not implemented. A friendly `toString()` also makes debugging a lot easier.

wing328:
PR merged. Thanks for the contribution.

PR #2712, opened 2016-04-27
issue #2711 adding equals, hashcode etc to model classes

neilotoole:
See issue #2711


wing328:
@neilotoole thanks for the PR.

Issue #2276, opened 2016-02-29
Add auto-generated test files


wing328:
As part of https://github.com/swagger-api/swagger-codegen/pull/1921, we've add capabilities to generate test files for API and model files. 

If anyone has cycle to add the feature to the following languages, please reply to let us know:

- [ ] Android
- [ ] C++
- [x] C# (#2007)
- [ ] Clojure
- [ ] Dart
- [ ] Flash
- [ ] Javascript
- [ ] Java
- [ ] ObjC
- [x] Perl (#2109)
- [x] PHP (#1921)
- [ ] Python (#2161)
- [ ] Ruby (#2272)
- [ ] TypeScript (#2270)
- [ ] Swift
- [ ] Scala

ePaul:
I don't quite understand: What is the purpose of those generated tests?
The PHP tests seem to be empty, so they are not yet testing anything – It looks as after generation they need to be edited to actually test something. But then the "don't edit this file" comment should go away.

But then what is being tested here? The getters/setters of the model objects? (That kind of test, if deemed necessary, should also be generated. Though I would (as a codegen-user, not as a maintainer of codegen) prefer if the generated code just works and I don't have to test it myself.)

What might be useful would be a test of the actual API methods (which is to be run from a test framework against an actual server) – generating some template as a start for your own written tests.

wing328:
The purpose is to save developer's time in creating those test files from scratch. (I'll remove "don't edit this file" in my next PR)

And yes, ideally we should test getter/setter for each model. The work I've done is just a starting point. Ideally the test fils should leverage all available information provided in the spec to generate all possible test cases.

In my opinion, generating the test code makes the SDK/API client looks more mature. For example, if there are 2 open-source Ruby SDKs for Twitter API - one with test cases and another without any, I would tend to think the one with test cases is more mature. What do you think? 

There are scaffolded test cases for API methods, e.g. https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/csharp/SwaggerClientTest/Lib/SwaggerClient.Test/PetApiTests.cs. Is that what you're looking for?

Thanks again for the feedback. I think what I'll do it o update the wording to elaborate more on the purpose of the scaffolded test cases in the auto-generated test files.

wing328:
@ePaul does it make sense to you?

Later if anyone from the community has cycle, we can update codegen to automatically generate test cases specified by vendor extension in the endpoint definition.

wing328:
[...]

PR #2727, opened 2016-04-27
Add test cases for Python Client

scottrw93:
Generates stubs for the tests cases for Issue #2276


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PythonClientCodegen.java
```diff
@@ -20,19 +20,28 @@
     protected String packageVersion;
     protected String apiDocPath = "docs/";
     protected String modelDocPath = "docs/";
+    
+	private String testFolder;
```
wing328: FYI. I'll make minor change to replace tab indentation with 4-space later.

Issue #2725, opened 2016-04-27
[Go] [Bug] import of 'encoding/json' not always required


neilotoole:
In `api.mustache`, we currently always [import](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/go/api.mustache#L7) `encoding/json`. However, if none of the operations in an API return an object (e.g. if there's only one operation and it's just a DELETE), then this import is unused, and thus we have a Go error.

wing328:
PR merged into master

PR #2728, opened 2016-04-27
Issue #2725: added condition to import json based on return type

guohuang:
@neilotoole can you confirm if this will fix your issue or not?

@wing328 the idea is to check the return type to see if it is primitive, if it is, add the custom import, let me know if you have a better approach.

For #2725


neilotoole:
@guohuang I'll give it a test locally

wing328:
@guohuang @neilotoole I've reviewed and the change looks good to me.

Issue #2722, opened 2016-04-27
[Go] Use "petstore" instead of "swagger" for Go petstore package name


wing328:
For the related discussion, please refer to #2717

guohuang:
isn't the additional property going to replace the default package name?
 https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/GoClientCodegen.java#l121

wing328:
You need to specify it via command line or config file. If not provided, it will use the default `swagger`

wing328:
PR merged. Closing this.

PR #2732, opened 2016-04-28
added packageName in the codegen option

guohuang:
@wing328 please review

For #2722


wing328:
@guohuang looks good to me. Thanks for the contribution as always.

PR #2733, opened 2016-04-28
[PHP] add parameter validation in methord call

abcsun:


modules/swagger-codegen/src/main/resources/php/api.mustache
```diff
@@ -122,7 +122,36 @@ use \{{invokerPackage}}\ObjectSerializer;
...
+        }
+        {{/minimum}}
+        {{#pattern}}
+        if (!preg_match("{{pattern}}", ${{paramName}})) {
+            throw new \InvalidArgumentException('invalid value for "{{paramName}}" when calling {{classname}}.{{operationId}}, must conform to the pattern {{pattern}}.');
+        }
+        {{/pattern}}
+
+        {{/hasValidation}}
+        {{/required}}{{/allParams}}
```
wing328: I would suggest moving {{/required}} to a new line between L125 and L126 to cover all parameters (required and optional)

modules/swagger-codegen/src/main/resources/php/api.mustache
```diff
@@ -172,6 +201,33 @@ use \{{invokerPackage}}\ObjectSerializer;
 
         {{#formParams}}// form params
         if (${{paramName}} !== null) {
+            {{#hasValidation}}
```
wing328: My understanding is that the above check should cover form parameters so there's no need for another set of validation here. Do you agree?

abcsun: Yes, you are right. 
All kind of params(queryParams, formParams.. ) should be validated in the allParams tag. I'll fix it

PR #2745, opened 2016-04-29
escape literal \n with character

mach-kernel:
Due to the `escapeJava()` call, I was getting `\n` escaped in a String which would make its way into Markdown documentation and look ugly. I replaced the "\n" strings with a '\n\ character in order to fix this. My swagger schema is in `yaml` and we have multi-line blocks of documentation. 

This may be useful to people in a similar situation.


wing328:
@mach-kernel thanks for the PR. We should add "unescapedDescription" similar to what we've done for property, parameter and model description: https://github.com/swagger-api/swagger-codegen/search?utf8=%E2%9C%93&q=unescapedDescription

Would you have cycle to add that ?

Issue #2748, opened 2016-04-30
[Go] Generated code should conform to Go fmt


neilotoole:
The Go code should conform to the Go conventions. Currently there are spaces instead of tabs, etc. Essentially, running `go fmt` on the generated code should be a no-op.

wing328:
Totally agree and I believe you're working on it :)

Ref: http://stackoverflow.com/a/19094788/677735

neilotoole:
Well yes, as it turns out, I am working on it :-)

Incoming shortly.

wing328:
@neilotoole  PR merged into master. Thank you for your contribution.

guohuang:
that is strange, i used to love tabs when programming in .net, but then it creates a lot of issues, so i ended up switching to spaces, and love it, i don't know why Go chooses tabs as standard. (and it is width of 8, which is a lot)

wing328:
@guohuang same experience here but given that Go explicitly sets the rule (tab with a width of 8), we don't have a choice.

PR #2750, opened 2016-05-01
Issue #2748 - generated code now conforms more closely to conventions

neilotoole:
Addresses issue #2748 ... The generated code now largely conforms to what `gofmt` would produce. However, note that due to the limitations of mustache, and also some unknown-till-runtime issues (such as what imports may be needed), it's not quite perfect in regard to matching `gofmt`.


samples/client/petstore/go/go-petstore/pet_api.go
```diff
@@ -1,602 +1,566 @@
...
+		"application/json",
+	}
+
+	// set Accept header
+	localVarHttpHeaderAccept := a.Configuration.APIClient.SelectHeaderAccept(localVarHttpHeaderAccepts)
+	if localVarHttpHeaderAccept != "" {
+		headerParams["Accept"] = localVarHttpHeaderAccept
+	}
+
+	formParams["additionalMetadata"] = additionalMetadata	fbs, _ := ioutil.ReadAll(file)
```
wing328: There's a minor issue here. `fbs, _ := ioutil.ReadAll(file)` should be in another line.

neilotoole: Well, that was sloppy. Pushed a fix.

PR #2752, opened 2016-05-02
[csharp] Default folder structure

jimschubert:
This should be considered a breaking change.

Aligns C# project outputs more with community accepted standards and leverages Nuget for package management.

This also moves the generated C# sample code out of the test project's Lib folder. The output structure here was causing some issues with maintainability (e.g. had to update test project with generated code).

(see: https://gist.github.com/davidfowl/ed7564297c61fe9ab814)

Output for a project, `IO.Swagger` will now look like:

```
    .
    ├── IO.Swagger.sln
    ├── README.md
    ├── bin
    ├── build.bat
    ├── build.sh
    ├── docs
    ├── packages
    └── src
        ├── IO.Swagger
        │   └── packages.config
        └── IO.Swagger.Test
            └── packages.config
```

This is a change from the Java-like `src/main/csharp/IO/Swagger/etc` structure and will be a breaking change for some.

@wing328


wing328:
@jimschubert thanks for the PR. I've flagged this as a breaking change (without fallback)

I'll review and let you know if I've any questions.

(no need to rebase for now)

wing328:
PR merged into master.

wing328:

Starting from 2.2.0, we've introduced the following breaking changes for C# API client:

  * `compile-mono.sh` has been renamed to `build.sh`
  * `compile.bat` has been renamed to `build.bat`
  * Solution (\*.sln) and project (\*.csproj) files are generated by default. To skip it, set `optionalProjectFileFlag=false`
  * New folder structure, using "Swagger.IO" as an example:
  ```
    .
    ├── IO.Swagger.sln
    ├── README.md
    ├── bin
    ├── build.bat
    ├── build.sh
    ├── docs
    ├── packages
    └── src
        ├── IO.Swagger
        │   └── packages.config
        └── IO.Swagger.Test
            └── packages.config
            
   ```

jimschubert:
@wing328 minor typo. 

Should be `using "IO.Swagger"`  or the files in the tree should be updated to `Swagger.IO` to avoid confusion. I think I ran `tree` on the output generated from `./bin/csharp-petstore.sh`. Maybe the target project name in that script and the batch script should also be updated to be the more C# standard `Swagger.IO`?

wing328:
@jimschubert what about changing the default namespace to just `Swagger`?

In the coming weekend, I'll also move those test cases for PetStore to the auto-generated test files (which won't be overwritten if the test file already exists).

jimschubert:
I think that's fine. However, if you rename the project after moving your test cases, they will be lost if the API naming convention ever changes (*Api to *Service for example). I think keeping them in a project that doesn't get auto generated is more maintainable because it would just take a search and replace to update it.

Also, we could use the test project to include integration tests for bugs as a regression suite. I don't know that those would belong in the generated tests?

wing328:

Starting from 2.2.0, we've introduced the following breaking changes for C# API client:

  * `compile-mono.sh` has been renamed to `build.sh`
  * `compile.bat` has been renamed to `build.bat`
  * Solution (\*.sln) and project (\*.csproj) files are generated by default. To skip it, set `optionalProjectFileFlag=false`
  * New folder structure, using "IO.Swagger" as an example:
 
screen shot 2016-05-16 at 3 08 19 pm

PR #2753, opened 2016-05-02
Issue 2743

mikolajprzybysz:
This PR adds some order in docs and test, making related files have the same path but different root, as the following example:
```
lib/Api/DefaultApi.php
test/Api/DefaultApi.php
docs/Api/DefaultApi.php
```

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PhpClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PhpClientCodegen.java
@@ -35,6 +35,9 @@ public class PhpClientCodegen extends DefaultCodegen implements CodegenConfig {
     protected String packagePath = "SwaggerClient-php";
     protected String artifactVersion = "1.0.0";
     protected String srcBasePath = "lib";
+    protected String testBasePath = "test";
+    protected String apiDirName = "Api";
+    protected String modelDirName = "Model";
     protected String variableNamingConvention= "snake_case";
     protected String apiDocPath = "docs/";
     protected String modelDocPath = "docs/";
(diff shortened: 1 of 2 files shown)


wing328:
cc @arnested

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PhpClientCodegen.java
```diff
@@ -35,9 +35,13 @@
...
     protected String srcBasePath = "lib";
+    protected String testBasePath = "test";
+    protected String docsBasePath = "docs";
+    protected String apiDirName = "Api";
+    protected String modelDirName = "Model";
     protected String variableNamingConvention= "snake_case";
-    protected String apiDocPath = "docs/";
-    protected String modelDocPath = "docs/";
+    protected String apiDocPath = docsBasePath + "/" + apiDirName;
+    protected String modelDocPath = docsBasePath + "docs/" + modelDirName;
```
arnested: I think there is a `docs` too many
```java
protected String modelDocPath = docsBasePath  + "docs/" + modelDirName
```
should probably just be
```java
protected String modelDocPath = docsBasePath  + "/" + modelDirName
```

modules/swagger-codegen/src/main/resources/php/composer.mustache
```diff
@@ -29,5 +29,8 @@
     },
     "autoload": {
         "psr-4": { "{{escapedInvokerPackage}}\\" : "{{srcBasePath}}/" }
+    },
+    "autoload-dev": {
+        "psr-4": { "{{escapedInvokerPackage}}\\" : "{{srcBasePath}}/" }
```
arnested: Should this have been `{{testBasePath}}`?

arnested:
Would it make sense to expose the `testBasePath` as a CliOption (like `srcBasePath` is)?

Beside the other two comments I think this looks good!

wing328:
@mikolajprzybysz did you have a chance to review the comments by @arnested ?

mikolajprzybysz:
@arnested I would prefer not to add it to CliOptions, otherwise we should also consider adding CliOption for docs also and I think we can agree that it is not needed, I consider test to be the same case. For me it is just important to move them out of the code library to be able to make the code deployable without having to configure exclude path, just to exclude tests. I think if there someone will have real use case for that we can add it, but as long as there is not issue regarding that I would skip it. Regarding all the other feedback, thank can you for the input, I will fix all those points :)

arnested:
@mikolajprzybysz, it make sense not to include CliOptions. Thanks for giving it a thought.

I think this is good to merge. @wing328 when you have some spare time please merge.

wing328:
@mikolajprzybysz thanks for the PR, which has been merged.

@arnested thanks for reviewing the change.

PR #2754, opened 2016-05-02
Add uuid mapping

wing328:
Add uuid mapping for c#, java, ruby, php, TS, JS

This PR also fixes the Typescript template folder issue (lowercase vs PascalCase)


hiveship:
I think uuid generation is also done in Java based codegens ?

wing328:
Not that I'm aware of in https://github.com/wing328/swagger-codegen/blob/add_uuid_mapping/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java#L92

I think you might be referring to various Java server generator.

wing328:
FYI. TS resource folder issues are addressed via the following commits instead: 
0bce28cff226b50bbbea02d627f076b39ea0386d
5c97717fc818950aa956afb2538266dbf9404793

PR #2459, opened 2016-03-28
Rewrite Promise.defer in new style

bherila:
Resolves https://github.com/swagger-api/swagger-codegen/issues/2251

PR #2758, opened 2016-05-03
[Typescript] add support for typescript es6 promises

wing328:
Ref: https://github.com/swagger-api/swagger-codegen/pull/2459

To use the option, pass `-DsupportsES6=true` or do it in config file.

cc @Bherila


wing328:
also cc @Vrolijkx

samples/client/petstore/typescript-angular/tsd.json
```diff
@@ -5,11 +5,20 @@
...
   "bundle": "typings/tsd.d.ts",
   "installed": {
-    "angularjs/angular.d.ts": {
+  	"angularjs/angular.d.ts": {
       "commit": "f6c8ca47193fb67947944a3170912672ac3e908e"
     },
     "jquery/jquery.d.ts": {
       "commit": "f6c8ca47193fb67947944a3170912672ac3e908e"
+    },
+    "request/request.d.ts": {
```
Vrolijkx: Why do we have a dependency on request in the angular client? Isn't request only for node?

Vrolijkx:
Bluebird has all features that ES6 promises have. Can't we change all code to ES6 promises. And only when we need to support ES5  the following code is needed. 
```
{{^supportsES6}}
 import Promise = require('bluebird');
 {{/supportsES6}}
```
This will make the templates simpler because we avoid a lot of conditional blocks.

wing328:
@Vrolijkx so you're suggesting making ES6 as the default? (currently the default is ES5)

Vrolijkx:
@wing328 No is suggest to keep the default ES5 with bluebird. but wright all the code in with te ES6 promise syntax.
We would always write:
```typescript
return new Promise((reject, resolve) =>{
  request(requestOptions, (error, response, body) => {
    if (error) {
        reject(error);
    } else {
       //do some code
       resolve(THE_VALUE_TO_RETURN)
    }
})
```
instead of:
```typescript
let localVarDeferred = promise.defer<ReturnType>();
request(requestOptions, (error, response, body) => {
    if (error) {
        localVarDeferred.reject(error);
    } else {
       //do some code
       localVarDeferred.resolve(THE_VALUE_TO_RETURN)
    }
});
return localVarDeferred.promise
```
The first example also works in bluebird so why support two different code templates for these parts.

wing328:
@Vrolijkx thanks for the explanation. If we switch to the ES6 promise syntax and when developers upgrade the Typescript SDK with the ES6 promise syntax, do you need to make any code change or the new ES6 promise syntax is backward compatible? (this is not to say we cannot make breaking change but I need to understand the impact to the developers/users)

Vrolijkx:
The new syntax is backward compatible. For a consumer of an api no changes are needed.

samples/client/petstore/typescript-angular/tsd.json
Vrolijkx: Maybe the sample generation script updates the wrong values.

bherila:
I seem to recall that Bluebird deprecated the use of Promise.defer anyways. So that's another reason to switch over to the ES6 syntax.

wing328:
@Vrolijkx thanks for the confirmation. I've merged this PR into master for the time being. Please feel free to submit a PR with the suggested change (so that the credit goes back to your Github account). 

If you need help, just let me know.

Issue #2742, opened 2016-04-29
Swager Generator on SpringMVC generates as many times as the method as there are tags


diyfr:
If you use multiple tags on  one rest method, swagger generator with springmvc target generate  as many times as the method as there are tags

Sample 
```
paths:  
  /pets:  
    post:  
      tags:  
        - pet  
        - littlePet  
      summary: Add a new pet to the store  
      description: ""  
      operationId: addPet  
      consumes:  
        - application/json  
        - application/xml  
      produces:  
        - application/json  
        - application/xml  
      parameters:  
        - in: body  
          name: body  
          description: Pet object that needs to be added to the store  
          required: false  
          schema:  
            $ref: "#/definitions/Pet"  
      responses:  
        "405":  
          description: Invalid input  
      security:  
        - petstore_auth:  
          - write_pets  
          - read_pets  
```

Generate

```
  @ApiOperation(value = "Add a new pet to the store", notes = "", response = Void.class, authorizations = {
    @Authorization(value = "petstore_auth", scopes = {
      @AuthorizationScope(scope = "write_pets", description = "modify pets in your account"),
      @AuthorizationScope(scope = "read_pets", description = "read your pets")
      })
  })
  @io.swagger.annotations.ApiResponses(value = { 
    @io.swagger.annotations.ApiResponse(code = 405, message = "Invalid input") })
  @RequestMapping(value = "", 
    produces = { "application/json", "application/xml" }, 
    consumes = { "application/json", "application/xml" },
    method = RequestMethod.POST)
  public ResponseEntity<Void> addPet(

@ApiParam(value = "Pet object that needs to be added to the store"  ) @RequestBody Pet body
)
      throws NotFoundException {
      // do some magic!
      return new ResponseEntity<Void>(HttpStatus.OK);
  }

  

  @ApiOperation(value = "Add a new pet to the store", notes = "", response = Void.class, authorizations = {
    @Authorization(value = "petstore_auth", scopes = {
      @AuthorizationScope(scope = "write_pets", description = "modify pets in your account"),
      @AuthorizationScope(scope = "read_pets", description = "read your pets")
      })
  })
  @io.swagger.annotations.ApiResponses(value = { 
    @io.swagger.annotations.ApiResponse(code = 405, message = "Invalid input") })
  @RequestMapping(value = "", 
    produces = { "application/json", "application/xml" }, 
    consumes = { "application/json", "application/xml" },
    method = RequestMethod.POST)
[...]

PR #2760, opened 2016-05-03
#2742 Issue multiple methods if use multi tags

diyfr:


diyfr:
Duplicate method from JavaResteasyServerCodegen

PR #2762, opened 2016-05-03
[koly] small refactor

kolyjjj:
just a small refactor, please review it carefully, because refactoring has the risk to break the funtionality


wing328:
@kolyjjj thanks for PR. We'll consider doing a refactoring in 2.3.0 or later version.

Closing this for now.

PR #2775, opened 2016-05-04
fixed array return type return as pointer issue

guohuang:
@wing328 the integration test discover a new issue while returning an array as pointer, according to this post, it is not valid http://stackoverflow.com/questions/21278023/golang-cannot-range-over-pointer-to-slice

i have updated the codegen file to check if return type is array or not before adding the pointer.

cc: @neilotoole

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/GoClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/GoClientCodegen.java
@@ -401,13 +401,7 @@ public class GoClientCodegen extends DefaultCodegen implements CodegenConfig {
                 listIterator.add(newImportMap);
             }
         }
-        // add pointer to return type string if it is not array
-        for (CodegenOperation operation : operations) {
-            if((operation.returnContainer == null || operation.returnContainer != "array") 
-                && operation.returnType != null) {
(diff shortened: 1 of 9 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/GoClientCodegen.java
```diff
@@ -401,7 +401,13 @@ public String toOperationId(String operationId) {
...
             }
         }
-
+        // add pointer to return type string if it is not array
+        for (CodegenOperation operation : operations) {
+            if((operation.returnContainer == null || operation.returnContainer != "array") 
+                && operation.returnType != null) {
+                operation.returnType = "*" + operation.returnType;
+            }
+        }
```
wing328: @guohuang  Thanks for the PR. What about using the `isListContainer` mustache tag to check if response type is an array?

e.g. `{{#isListContainer}} ... {{/isListContainer}}`

Ref: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenResponse.java#L17

guohuang: Fixed, I like this solution better, thanks

Issue #2756, opened 2016-05-02
c# - generated api method throws NullReferenceException when using optional FromBody parameter


kbrooks:
I have a api endpoint with an optional body parameter, but the generated code for that endpoint produces a NullReferenceException. 

```
  "put": {
    "tags": ["Entity"],
    "operationId": "EntityUpdate",
    "produces": [],
    "parameters": [
      {
        "name":"schemaName",
        "in":"path",
        "required":true,
        "type":"string"
      },
      {
        "name":"entityName",
        "in":"path",
        "required":true,
        "type":"string"
      },
      {
        "name":"columns",
        "in":"body",
        "required":false,
        "schema":{
          "type":"array",
          "items":{
            "$ref":"#/definitions/Column"
          }
        }
      }
    ] 
...
  },
```

The generated code allows for the parameter to be null, but it throws a NullReferenceException:
```
        public ApiResponse< UpdateEntityResult > EntityUpdateWithHttpInfo (string schemaName, string entityName, List<Column> columns = null)
        {

...
            if (columns.GetType() != typeof(byte[]))
            {
                localVarPostBody = Configuration.ApiClient.Serialize(columns); // http body (model) parameter
            }
            else
            {
                localVarPostBody = columns; // byte array
            }
```
I am using version 2.1.6.

Thanks for your help.

kbrooks:
It looks like all we need to do to fix this is add a null check to [this line](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/csharp/api.mustache#L230). 

I've never used mustache before but I think it should be something like this:
```
if ({{^required}}{{paramName}} != null && {{/required}}{{paramName}}.GetType() != typeof(byte[]))
```

wing328:
@kbrooks the fix by @guohuang has been merged into master. Please pull the latest master and give it a try.

(closing the issue for the time being)

PR #2777, opened 2016-05-04
Issue #2756: add null checking to avoid null exception

guohuang:
For #2756 

@kbrooks we should check for null regardless it is required or not. let me know if this is still an issue.

@wing328 please review


wing328:
@guohuang thanks for the PR. The fix has been merged into master.

Issue #2779, opened 2016-05-05
Unicode Local Text Support issue


catinred2:
checkout here:[escapeText](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L326)

escapeJava would encode the Chinese into unicode like \u8BA1 , so Chinese would not show properly in the generated static html or java code.

wing328:
@catinred2 thanks for reporting the issue. I think this will impact other double byte characters as well.

May I know if you've cycle to contribute a fix?

Ref: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

catinred2:
I'll give it a try tomorrow.

wing328:
@catinred2 thanks. ping me if you need any help.

PR #2780, opened 2016-05-05
Unicode escape issue

catinred2:
#2779


wing328:
@catinred2 thanks for the PR. Have you tried upgrading to http://mvnrepository.com/artifact/org.apache.commons/commons-lang3, which has better support for double byte characters?

Ref: https://dzone.com/articles/commons-lang-3-improved-and-powerful-StringEscapeUtils

catinred2:
@wing328 StringEscapeUtils of either version would not do.

catinred2:
@wing328  the ci fails. I've checked out the detail. it seems that has something about javadoc style. any rules should i follow?

wing328:
@catinred2 I'll take another look later today.

wing328:
@catinred2 I've filed https://github.com/swagger-api/swagger-codegen/pull/2785 to fix the issue. Would you give it a try to see if it works for you as well?

(this is **not** to say your PR is not good but just that we need to upgrade `org.apache.commons.lang` to the latest version `org.apache.commons.lang` anyway so I think I would try to leverage the latest version of the library to address the problem to avoid reinventing the wheel)

catinred2:
@wing328 #2785 works for me. 
thanks for your great work.

wing328:
@catinred2 my PR merged into master. Please pull the latest.

PR #2783, opened 2016-05-05
Replace 'BEARER' with 'Bearer' everywhere for RFC6750 compliance.

GriffinSchneider:
The [OAuth spec](https://tools.ietf.org/html/rfc6750#section-2.1) says that the correct format is `Bearer <token>`, not `BEARER <token>`. My server using [node-oauth2-server](https://github.com/oauthjs/node-oauth2-server) was rejecting my request without this fix.

It looks like the only place `BEARER` was in actual code was `modules/swagger-codegen/src/main/resources/objc/Configuration-body.mustache`. I experienced the issue in ObjC.

This pullreq was created by running
```bash
find . -type f -exec sed -i "" "s/BEARER/Bearer/g" {} \;
```
and then ignoring all the whitespace-only changes that generates.


wing328:
@GriffinSchneider thanks for the PR.

PR #2786, opened 2016-05-06
[Objc] ResponseDeserializer for separating deserialization and service logic

mateuszmackowiak:
Add new class ResponseDeserializer for separating deserialization and service logic.
- Adds support for null JSON values.
- Adds missing errors if invalid type values


wing328:
@mateuszmackowiak thanks for the PR. 

From what I understand, the change is non-breaking, which means developers do not need to make any code change to their application in order to work with the latest auto-generated ObjC SDK. Please correct me if I'm wrong.

mateuszmackowiak:
Yes. No changes are required

wing328:
@mateuszmackowiak thanks for the prompt confirmation.

Issue #2789, opened 2016-05-06
C#: Add support for collectionFormats for query string parameters


sean11:
Currently, all collections, arrays and lists that are sent as a query string parameter get formatted as a comma separated list of strings and do not support any other collectionFormats. See [here](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/csharp/api.mustache#L221).

I've made some preliminary modifications [here](https://github.com/sean11/swagger-codegen/pull/1/commits/c999fcf1850dc9e1742c5d80958ae24629578537) but do not currently have time to make all the changes to unit tests and samples for a proper pull request.

Related Java/Android [Pull Request#893](https://github.com/swagger-api/swagger-codegen/pull/893)
Related Java [Issue#484](https://github.com/swagger-api/swagger-codegen/issues/484)

wing328:
@sean11 Thanks for the contribution. May I know if you've time to submit a PR so that we can review accordingly?

cc @jimschubert

sean11:
Opened PR #2790

jimschubert:
I'll try to also check out the PR some time this weekend.

jimschubert:
@sean11 sorry. That weekend ended up being pretty hectic and I completely forgot about reviewing the pull request. I made a comment, which is more a comment on developers using the existing public method than on your implementation.

PR #2790, opened 2016-05-06
C#: Add support "multi" collectionFormat query string params

sean11:
This will add support to pass collections as a list of keyValuePairs (e.g. "collectionParam=value1&collectionParam=value2&collectionParam=value3) when the collectionFormat="multi" rather than always passing collections as a comma separated list of strings (e.g. "collectionParam=value1,value2,value3").

### DIFF
--- a/modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
@@ -80,7 +80,7 @@ namespace {{packageName}}.Client
 
         // Creates and sets up a RestRequest prior to a call.
         private RestRequest PrepareRequest(
-            String path, RestSharp.Method method, Dictionary<String, String> queryParams, Object postBody,
+            String path, RestSharp.Method method, List<KeyValuePair<String, String>> queryParams, Object postBody,
             Dictionary<String, String> headerParams, Dictionary<String, String> formParams,
             Dictionary<String, FileParameter> fileParams, Dictionary<String, String> pathParams,
             String contentType)
@@ -145,7 +145,7 @@ namespace {{packageName}}.Client
         /// <param name="contentType">Content Type of the request</param>
         /// <returns>Object</returns>
         public Object CallApi(
-            String path, RestSharp.Method method, Dictionary<String, String> queryParams, Object postBody,
+            String path, RestSharp.Method method, List<KeyValuePair<String, String>> queryParams, Object postBody,
             Dictionary<String, String> headerParams, Dictionary<String, String> formParams,
             Dictionary<String, FileParameter> fileParams, Dictionary<String, String> pathParams,
             String contentType)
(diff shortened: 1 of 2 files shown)


sean11:
Related to Issue #2789

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -476,5 +476,39 @@ namespace {{packageName}}.Client
...
+        /// Use collectionFormat to properly format lists and collections.
+        /// </summary>
+        /// <param name="name">Key name.</param>
+        /// <param name="value">Value object.</param>
+        /// <returns>A list of KeyValuePairs</returns>
+        public IEnumerable<KeyValuePair<string, string>> ParameterToKeyValuePairs(string collectionFormat, string name, object value)
+        {
+            var parameters = new List<KeyValuePair<string, string>>();
+
+            if (IsCollection(value) && collectionFormat == "multi")
```
sean11: We can also add support for other collectionFormats as well  (e.g pipe, slash etc).

wing328:
@sean11 thanks for the PR. I'll review and test tomorrow (Tuesday)

wing328:
Btw, your commit (as shown in the Commits tab) is not linked to your Github account, which means this PR won't count as your contribution in https://github.com/swagger-api/swagger-codegen/graphs/contributors.

sean11:
Thanks @wing328. Just amended the commit with the proper author name. I'm going to try and make some time in the coming weeks to add support for other collection formats and update the examples/unit tests.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -145,7 +145,7 @@ namespace {{packageName}}.Client
         /// <param name="contentType">Content Type of the request</param>
         /// <returns>Object</returns>
         public Object CallApi(
-            String path, RestSharp.Method method, Dictionary<String, String> queryParams, Object postBody,
+            String path, RestSharp.Method method, List<KeyValuePair<String, String>> queryParams, Object postBody,
```
jimschubert: I would think this needs to be a method overload. There's no direct conversion between `Dictionary<String, String>` and `List<KeyValuePair<String, String>>`, so newly generated clients would break existing callers.

Also, `CallApiAsync` needs to have the same reflected change.

The type `List<KeyValuePair<String, String>>` will probably also be confusing to some developers. Most people see `KeyValuePair<T, K>` and think "this belongs in a dictionary". I would think `Dictionary<String, List<String>>` would be more appropriate. A `List<Tuple<String,String>>` could also reduce confusion, but it's not exempt from the same issues as `List<KeyValuePair<String, String>>`.

The RFC for URIs (http://tools.ietf.org/html/rfc3986#section-6.2.2.1) claims that scheme and host are case insensitive, and everything else is case sensitive. This means the following won't result in multi values for the same key as some might expect:

```
var list = new List<KeyValuePair<String, String>>();
list.Add(new KeyValuePair("key", "value1"));
list.Add(new KeyValuePair("KEY", "value2"));
```

Obviously, you'd have the same problem with casing in a dictionary. But you'd be reducing the footprint for developer error:

```
var d = new Dictionary<String,List<String>>();
d["key"] = new List<String> { "value1", "value2" };
```

jimschubert: @wing328 What's your take on the method overload comment? My concern is that `CallApi` and `CallApiAsync` are public methods. I don't doubt people are using it outside of the generated code. It could be marked internal, but that would also break those callers.

wing328: @jimschubert definitely valid points. I've heard less than 3 use cases in which the users need to extend the API clients to override the `CallApi` and `CallApiAsync` method but still we should avoid breaking their usage if possible.

sean11: Thanks for the feedback. I agree that we should not change the contracts on the public methods if we don't need to. I think we can just  call a toDictionary() before calling the public ApiClient.CallApi() method. This also means that only one file needs to be modified, api.mustache. I can work on pushing out those changes today. I'll also make sure to update the CallAsync methods as well since I overlooked that in my initial commit (was using targetFramework v3.5).

sean11: On second thought, because every key in a Dictionary must be unique, I don't think it is the appropriate type to be using for query parameters.

markebbers:
When can I expect this change to be merged in a release?

wing328:
@markebbers we wanted to but we're just so buy and lost track of this PR.

Did you have a chance to test the change? Any feedback from you?
```
git checkout -b sean11-csharp-collection-query-param-fix master
git pull https://github.com/sean11/swagger-codegen.git csharp-collection-query-param-fix
```

markebbers:
@wing328 tested with Sean11s branch only because of some GIT merge conflicts. The result is not what I hoped for.

'http://test.com/article/find?order[]=name%2Ccolor' is what I got from a List<String> order and I was hoping for 'http://test.com/article/find?order[]=name&order[]=color'

markebbers:
Made a fork of the master and applied @sean11s work on top of it and now an array is formatted correctly in the query params. (param[]=1&param[]=2&etc...)

wing328:
@markebbers thanks for testing the PR. I'll try to merge over the weekend.

sean11:
I appreciate the effort to get this merged in @wing328 @markebbers. Sorry I couldn't give this more effort after I created the initial PR.

wing328:
Closed via https://github.com/swagger-api/swagger-codegen/pull/5307

PR #2791, opened 2016-05-06
Typescript fetch update

leonyu:
* Add `npmName` and `npmVersion`
 * will take a look at `npmRepository` and `snapshot` since my use case doesn't take advantage of those
* Make sure compilable in both ES6 and ES5
* Fix `enum` implementation (was broken in the PetStore)
 * this implementation uses `string literals` for string enums instead of `Enum` (maybe this would be better served as an option) since TypeScript `Enum` follows C Enum instead of Java Enum that defaults to a number. [see ref](https://stackoverflow.com/questions/15490560/create-an-enum-with-string-values-in-typescript)


wing328:
@leonyu thanks for the contribution. PR merged into master.

PR #2799, opened 2016-05-07
Sanitizer for separating sanitize and service logic

mateuszmackowiak:
Sanitizer for separating sanitize and service logic


wing328:
@mateuszmackowiak thanks for the enhancement.

PR #2811, opened 2016-05-10
Update the injector static variable to contain all injected values.

achew22:
In Angular.js, values are injected into service in one of two ways:

1) Inline (by name).
2) By a static injector variable.

The TypeScript generator uses the 2nd method. This method requires you
to explicitly enumerate all the values you would like to have injected.
If you fail to inject a value the Angular DI system will simply pass you
`undefined`. The constructor is expecting 3 values to be passed (the
final being basePath) but the injector static only defines 2 values.
This results in basePath always being undefined no matter what you
define it to be.

This change updates the injector variable to handle that properly.


wing328:
@achew22 thanks for the PR.

@Vrolijkx please review when you've time. Thanks!

Vrolijkx:
looks also ok for me

PR #2812, opened 2016-05-10
Update TypeScript to do a better check for empty on basePath.

achew22:
In Angular if a value is not defined for injection it is passed as
undefined. That means that most of the time `if (value) {` is a
reasonable test. Unfortunately since `""` (empty string) is also falsey
by nature, an empty string will not trigger the if properly.

Instead you should check `if (value !== undefined) {`.


wing328:
@achew22 thanks for the PR.

@Vrolijkx please review when you've time. Thanks!

Vrolijkx:
Look ok to me.

Issue #2795, opened 2016-05-06
C#: Constructor is malformed if last property is readOnly: true


ggood:
Swagger-codegen 2.1.6. With the OpenAPI spec in this Gist:

https://gist.github.com/ggood/099933cdf7aba3cab4867abd500d1b25

the generated c# constructor won't compile, due to a missing argument:

```
public Foo(string Bar = null, )
        {
            this.Bar = Bar;
            
        }
```

This happens when the last property in a definition is readOnly.

wing328:
@ggood I believe I've addressed this in the latest master. I wonder if you can pull the latest master and give it a try.

jimschubert:
We should probably include a unit test if this is fixed.

I noticed code for required properties last night which I thought might result in this malformed constructor, but I didn't have time to verify my assumption. If this isn't yet fixed, I can grab it early next week.

wing328:
@jimschubert totally agree about adding a test case to cover issues moving forward. I added this before: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml#L807

Let me know that looks good from your perspective (I welcome better and more tests to cover the same issue)

jimschubert:
@wing328 I think it'd be a great idea to add a sample type to `petstore-with-fake-endpoints-models-for-testing.yaml` to verify compilation across all languages. I was thinking of a simple unit test case that compiles the yaml definition from this issue and verifies the string output for the class definition to prevent regression. This would be an additional safety against unintentional changes to `petstore-with-fake-endpoints-models-for-testing.yaml` which would invalidate the test (no longer generating code that would cause the compilation issue).

wing328:
@ggood may I know if you've a chance to try the latest master to see if the issue still persists?

ggood:
I just did a git pull, mvn clean install, and re-ran the generator on my sample. The generated constructor still looks wrong:

```
public Foo(string Bar = null, )
        {


                        this.Bar = Bar;

        }
```

Full constructed class here:

https://gist.github.com/ggood/93a39701368e98e5d59faed56f499857

wing328:
@ggood Thanks for checking. I'll take another look today.

wing328:
[...]

PR #2818, opened 2016-05-10
[C#] fix constructor with readonly property at the end

wing328:
To fix https://github.com/swagger-api/swagger-codegen/issues/2795


ggood:
Pulled from master just now and verified. Looks good to me - thanks.

PR #2827, opened 2016-05-10
[Objc] - Moved Logging to separate Logger file and optimizations

mateuszmackowiak:
- Moved Logging to separate Logger file
- Moved selectHeaderAccept and selectHeaderContentType to sanitizer
- little optimizations

This change will require changes by developers if used logging logic not only by using "debug" property on Configuration shared instance or used directly selectHeaderAccept and selectHeaderContentType.


mateuszmackowiak:
The second commit adds support for variations of application/json type. This was more of a bug.

PR #2840, opened 2016-05-12
Fix regression on swift enum name as well as make sure enum var name …

### LINKED ISSUE #2824: [Swift] enum with colons are generating non compilable swift code
If you have:

```json
 "properties": {
                "prop": {
                    "type": "string",
                    "enum": [
                        "foo:1",
                        "foo:2"
                    ]
                }
}
```

This is valid swagger but generates 

```swift
    public enum Prop: String { 
        case Foo:1 = "foo:1"
        case Foo:2 = "foo:2"
    }
```

which is not legal swift.

### LINKED ISSUE #2835: [Swift] regression in enum names
See https://github.com/swagger-api/swagger-codegen/commit/982a035cc1228d477e8583d9cf070698818082a1#commitcomment-17436465

cjolif:
…with colons ouput compiable swift. Add a test. fixes #2824, fixes #2835


Edubits:
Nice, thanks for the quick PR, looks good to me!

PR #2842, opened 2016-05-12
[PHP] Allow version skip in composer.json

mikolajprzybysz:
It is not possible to skip version because it has default value. It is important to be able to skip version, because it creates a lot of problems.

> Note: Packagist uses VCS repositories, so the statement above is very much true for Packagist as well. Specifying the version yourself will most likely end up creating problems at some point due to human error.
https://getcomposer.org/doc/04-schema.md#version


arnested:
Yes, good catch!

@wing328, please merge.

wing328:
@mikolajprzybysz thanks for the PR. If my memory serves me well, I was the one who sets a default version for the PHP (and other ) SDK. I didn't realise it would cause issue with Packagist.

Is that correct to say that you're able to upload the PHP SDK to Packagist with this fix?

mikolajprzybysz:
@wing328 The problem with setting up the version is that it should match version of your tag, let me show you the issue I had and why I proposed this. We have sdk project X with tag 1.0.0 in github. When I make changes to my schema, I rebuild it with swagger and overwrite my sdk. Afterwards I create tag 1.0.1 and in projects using sdk I update their dependency to use version 1.0.1. Now the problem is when I run composer update he downloads version 1.0.0 and as far I as I understand from this quote I provided it is the expected behavior.

So there are 2 solutions for my problem:
1.  don't forget to update version in my swagger config and always keep it in sync with the tag I am creating
2. remove version

I strongly believe solution number 1 will always create issues in the future because people will simply forget about it.

wing328:
@mikolajprzybysz for your use case, did you specify `artifactVersion` to 1.0.1 (via CLI option or config file) so that composer's version refers to 1.0.1 as well?

mikolajprzybysz:
That was the issue I did not update artifactversion, but with this Project I don't have to maintain it.

PR #2843, opened 2016-05-12
Android volley api client

zhenjun115:
add gitignore.mustache and git_push.mustache for android api client using volley


wing328:
@zhenjun115 the change looks good to me. Thanks for the contribution.

PR #2844, opened 2016-05-12
Support multiple file type parameters

amdonov:
I found that every parameter with a type of file added two method parameters with the name of inputStream and fileDetail. When two or more more file type parameters are used with an operation the names conflict. The provided patch uses the convention of {{paramName}}Stream and {{paramName}}Detail for the variables avoiding the name collision. I only made the fix for jersey2. I assume the same issue occurs with jersey 1, but I don't use those templates.


wing328:
Thanks for the PR but your commit (as shown in the Commits tab) is not linked to your Github account, which means this PR wont' count as your contribution in https://github.com/swagger-api/swagger-codegen/graphs/contributors.

wing328:
@amdonov there has been some changes to jax-rs (jersey2) template. Please kindly do a rebase to the latest master to resolve the merge conflicts.

PR #2861, opened 2016-05-13
[PHP] list_invalid_properties now don't call undefined variables

BeneficialName:


arnested:
Good!

Please merge, @wing328.

wing328:
@BeneficialName thanks for the fix.

@arnested thanks for the review.

PR #2862, opened 2016-05-13
[Objc] Moved the generated files to Api, Core and Model folders

mateuszmackowiak:
Moved the generated files to Api, Core and Model folders
also Added known response codes in comment


wing328:
@mateuszmackowiak thanks for the PR. I would suggest creating a new branch for the change (as per git best practice)

PR #2864, opened 2016-05-13
typescript-angular2 query string fix

revaultch:
This should fix the following problems, I ran into:
1) typescript complaining: "Cannot create property 'xxx' on string '' " (string treated as map)
2) I have a subclass of RequestOptions which will call merge, which behind the scenes calls clone which isn't implemented on string


wing328:
@Vrolijkx would you review the change when you've time? Thanks.

Vrolijkx:
Looks ok for me if the build passes. 
Didn't know URLSearchParams existed but it certainly makes the code look more robust.

wing328:
@revaultch would you please rebase on the latest master to resolve the merge conflicts?

wing328:
@revaultch thanks for your contribution.

wing328:
@revaultch btw, I would suggest make the change in a new branch (as per Git best practices)

Issue #2868, opened 2016-05-14
Use "Title" as model name if provided in swagger definition


dmytrobr:
For inline object definitions, the generator uses "inline_response12323" as a model name which later ends up in a lot of Java classes with meaningless names like InlineResponse12323.

I propose to use "Title" attribute whenever provided. This attribute is part of  Swagger definition (not specific to Swagger by comes from JSON schema) and could be added to any schema object.

Below is JSON example. For such example I would like to have Java class name for parameter be "ModifyCatRequest" (instead of "Data") and for response ModifyCatResponse instead of InlineResponse123. 

I believe it is not specific for Java client, but should be a generic approach to use "title" as model name whenever provided. Other tools (e.g. Swagger UI) are using this approach too

Change is already working at my laptop, so if community supports this change I'll send pull request
```
    "/cat/{id}": {
      "put": {
        "summary": "Modify Cat",
        "operationId": "modifyCat",
        "consumes": [
          "application/json"
        ],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "id",
            "in": "path",
            "type": "string",
            "description": "ECM ID",
            "required": true
          },
          {
            "name": "data",
            "in": "body",
            "schema": {
              "type": "object",
              "title": "ModifyCatRequest",
              "properties": {
                "description": {
                  "type": "string"
                },
                "name": {
                  "type": "string"
                } 
				}
            }
          }
        ],
        "responses": {
          "200": {
            "description": "200 status response",
            "schema": {
              "type": "object",
              "title": "Status",
              "properties": {
                "status": {
                  "type": "object",
                  "title": "StatusItem",
                  "properties": {
                      "reqStatus": {
                      "type": "string",
                      "example": "sample reqStatus"
                    }, 
                  },
                  "required": [
                    "credentials",
                    "reqStatus"
                  ]
                },
                "data": {
                  "type": "object",
                  "title": "ModifyCatResponse",
                  "properties": {
                    "order": {
                      "type": "object",
                      "title": "order",
                      "properties": {                     
                        "id": {
                          "type": "string"
                        }
                      },
                      "required": [
                        "id"
                      ]
                    }
                  }
                }
              }
            }
          }
        }
      }
    },
```

wing328:
PR merged. Issue closed.

PR #2870, opened 2016-05-14
use Title attribute as model name

dmytrobr:
[Use "Title" as model name if provided in swagger definition](https://github.com/swagger-api/swagger-codegen/issues/2868)

this allows to use names defined by user in Response model instead of
"inline_response". Also user could use title to have specific name for
request model instead of parameter name


wing328:

dmytrobr:
Ok, I'll fix the author and committer and create another pull request

PR #2872, opened 2016-05-14
use Title attribute as model name

dmytrobr:
[Use "Title" as model name if provided in swagger definition](https://github.com/swagger-api/swagger-codegen/issues/2868)

this allows to use names defined by user in Response model instead of
"inline_response". Also user could use title to have specific name for
request model instead of parameter name

### DIFF
--- a/modules/swagger-codegen/.gitignore
+++ b/modules/swagger-codegen/.gitignore
@@ -1,1 +1,2 @@
 /.settings/
+/test-output/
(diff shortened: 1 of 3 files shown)


privatwolke:
@wing328 Any chance this could get merged soon? Anything we can help out with?

wing328:
Let me review later today.

wing328:
@dmytrobr @privatwolke One quick question if you don't mind. What if 2 inline models (with different properties) have the same title defined? Would codegen throw exception or just keep the last one (inline model)?

Overall the PR looks good (especially you've written test cases to cover it 👍 ). My question above is more on edge cases that the spec is not entirely correct (which is not something uncommon).

dmytrobr:
@wing328 it is using function to generate unique name. So if 2 inline models have "SameTitle" one of them will get model name "SameTitle_1"

wing328:
@dmytrobr thanks for the explanation. PR merged.

PR #2873, opened 2016-05-14
Fixes instructions for generating PHP Silex code.

webbj74:
The name of this generator was updated from `silex` to `silex-PHP` in @dc7c2a4b, but that change wasn't reflected in README.md.


wing328:
@webbj74 thanks for the PR. We'll soon move that to a community-driven documentation so that the community can more easily contribute to the doc.

PR #2878, opened 2016-05-16
[Objc] - Common protocol for Api 

mateuszmackowiak:
- Added common api protocol 
- NSParameterAssert if missing required param + calling completion block with error rather than throwing an exception

Changes can be breaking if depending on try catch for missing required method argument - Not a common way to solve this problem in objective c.


wing328:
@mateuszmackowiak thanks for the PR. If my memory serves me well, the reason why we used `NSException` for missing required parameters was that we reviewed a few Objc SDKs (hosted on Github) and found that these SDKs used `NSException` to handle validation error. 

If `NSParameterAssert` is more popular and correct, we should make the (breaking) change and keep our ObjC developers informed.

PR #2880, opened 2016-05-16
gradle wrapper for android api client;

zhenjun115:
not tested


cbornet:
Would you also do it for the java clients ?

wing328:
@cbornet I think that's the plan :)

Can you also do a test? (it works for me)

PR #2882, opened 2016-05-16
Adding correct code generation for additional properties in typescript angular2.

### LINKED ISSUE #2751: Generated typescript when using additional properties is incorrect.
The code generated by the Typescript client when having additional properties in the spec file will never compile. The next example is from the typescript-angular2 client. The other Typescript clients have te same problem.
```typescript
export interface User extends models.null<String, string> {
    id?: number;
    userStatus?: number;
}
```
with spec file:
https://gist.github.com/Vrolijkx/b220edc20c6f9379bfc8291ea6e28aec

In other languages they extend from a map to support the additional properties behaviour. But I don't think that's a good solution in Typescript. The support for [Map](https://developer.mozilla.org/nl/docs/Web/JavaScript/Reference/Global_Objects/Map) is very browser dependent. Maps are also not converted correctly to json when using ```JSON.stringify()``` so a custom converter will be needed.

I think the best way to support additional properties in Typescript is to go in the next direction:
```typescript
export interface User {
    [key: string]: string

    id?: number;
    userStatus?: number;
}
```

Vrolijkx:
This branch will partially resolve #2751. Support for other typescript models will come soon.

I also added an integration test. This kind of test makes it easier for me to work test driven. And i also think these kind of test will protect against regression. Feel free to suggest improvements.


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngular2ClientCodegen.java
```diff
@@ -131,7 +139,7 @@ private String addModelPrefix(String swaggerType) {
     @Override
     public void postProcessParameter(CodegenParameter parameter) {
         super.postProcessParameter(parameter);
-        parameter.dataType = addModelPrefix(parameter.dataType);
+         parameter.dataType = addModelPrefix(parameter.dataType);
```
wing328: Minor indentation inconsistency.

PR #2895, opened 2016-05-17
gradle wrapper for java api client;

zhenjun115:
Gradlew look good on window. Please give it more test.


wing328:
Tested on mac and all works as expected. Thanks for the contribution from @zhenjun115

Issue #2105, opened 2016-02-10
add an ignore file for codegen


fehguy:
From #1956, adding a .codegen-ignore file would be helpful for keeping languages from overwriting files unexpectedly.

jimschubert:
I've begun development on this. https://github.com/jimschubert/swagger-codegen/tree/feature/codegen-ignore

wing328:
Thanks for the contribution from @jimschubert to add the support of `.swagger-codegen-ignore` file to skip file overwrite

PR #2899, opened 2016-05-18
Feature/codegen ignore

jimschubert:
Looking for feedback and possibly additional testing on this.

Here's a test scenario:

1. Remove `samples/client/petstore/csharp/SwaggerClient`
2. Create `samples/client/petstore/csharp/SwaggerClient`
3. Add the following to `samples/client/petstore/csharp/SwaggerClient/.swagger-codegen-ignore`
```
# Swagger Codegen Ignore

# This should match build.sh located anywhere.
# Should also work as /build.sh
build.sh

# Exclude all recursively
docs/**

# Explicitly allow files excluded by other rules
!docs/UserApi.md

# Test DirectoryRule
src/**/Api/

# This shouldn't be included because parent directory is excluded by previous rule.
!src/**/PetApiTests.cs

# Exclude a single, nested file
src/IO.Swagger.Test/Model/AnimalFarmTests.cs

```
4. Verify:
  - `build.sh` not created
  - only `docs/UserApi.md` created under docs
  - `src/IO.Swagger.Tests/Api/PetApiTests.cs` not created under `src/IO.Swagger.Tests/Api`
  - `src/IO.Swagger.Test/Model/AnimalFarmTests.cs` not created

see issue #2105


jimschubert:
@wing328 I've rebased against master to get rid of the conflicts (there were a few in `DefaultGenerator.java`).

A main question I have is whether I should add any tests to `DefaultGeneratorTest` to verify whether or not files get generated as intended. I didn't do this because I thought it might be overkill since it would tightly couple those tests to the folder structure created by a single generator.

Also, I didn't do anything with the existing `skipOverwrite` functionality.

PR #2905, opened 2016-05-18
[Typescript-Node] Add Support for useQuerystring Request Option

pixelshaded:
The request node module normally sends arrays in the querystring like so

```
?array[0]=1&array[1]=2
```

However Web API doesn't know how to bind that to the model. Web API supports this method of serialization:

```
?array=1&array=2
```

The request module allows for you to override this default functionality so you can produce the second outcome. This pull request adds support for setting useQuerystring on the instance of the api so that all future calls can use the second method of serialization.


pixelshaded:
Also I'm not seeing how tests are run anymore. The sample output now has a default and npm folder. I went in to the npm folder, ran npm install and npm build and had no errors. This runs the api.ts through tsc but not the client.ts (which is what used to test the client). I'm not seeing anywhere client is actually compiled and run. How are we testing this now?

wing328:
@pixelshaded thanks for the feedback and yes, the integration test for Typescript API client still need some work. I should some time this weekend to make it better.

If you've cycle, feel free to submit a PR for the improvement.

wing328:
@pixelshaded I've filed a PR with better integration tests for Typescript API clients: https://github.com/swagger-api/swagger-codegen/pull/2910.

wing328:
@pixelshaded I've merged #2910 into master to have better test coversage for TS ndoe and angular.

Please rebase on the latest master to leverage these tests for your PR.

wing328:
cc @Vrolijkx

pixelshaded:
I looked through the PR and did a rebase. I'm not seeing anywhere that the tests are being run. Did you want me to get the client script running again?

pixelshaded:
Looks like the rebase blew up this pr.

wing328:
@pixelshaded no worry. I'll cherry-pick your commits later today.

Vrolijkx:
@pixelshaded If we only version the ```.ts``` files the bundeling will become the burdon of the consumer. Also it will become very hard for non typescript angular2 users that like ES5 to use these clients. Maybe we should version both the ```.ts``` and the ```.js``` files.

pixelshaded:
@Vrolijkx 

1. Why is the consumer concerned with bundling server side scripts? And if they are, why should codegen be handling that? Bundling is almost always tightly coupled to the consumer's build system. Why even try to account for something that is not even standardized?
2. How would angular2 use a node client? I wouldn't expect support for running a node library in the browser, even if you can make it happen with things like browserify. I would expect the consumer to be responsible for this overhead, not codegen. Right now, browserify and request (which we use in typescript-node) don't play nicely for multiple reasons: https://github.com/request/request/issues/2090. You want to validate that every dependency we use can be run client side? In which module loaders?
3. The js files in the samples folder aren't generated from codegen. You have to run the ts files through a compiler manually before committing to even update them (just like a consumer would). It really doesn't make sense to version them. They only get created for testing. This would be like versioning your bin folder in a c# library. We care about versioning the source code, not the output. In my eyes, the only reason to version anything in the samples folder is so that in a PR, reviewers can see how the output is changing relative to code changes (the output being ts files, not js) without having to build locally.

Vrolijkx:
@pixelshaded 
When we do the bundeling also non Typescript users can use our generated client with the benefit of TypeDefinitions files. And because we know the client will be used on NodeJs only because we use request we are sure they support the CommonJs import syntax.
You are right about your second point. I got mixed up with the angular2Typescript client i was working on.
You are also correct about the last point it would be better to automatically run this on ```npm prepublish```.
But maybe because we only generate one file with the nodeJS TypesScript client the benefit of bundeling and publishing it automatically wouldn't be that great.

pixelshaded:
Hmmm. I think I must be missing something. Maybe some context? Are we publishing every npm module in samples or something? If so, why? I thought petstore was essentially an example API.

wing328:
@pixelshaded I'll do the cherry-pick over the weekend.

@pixelshaded @Vrolijkx very good discussion. Please keep it going.

Vrolijkx:
At the moment we don't publish the generated clients but it would be nice if we did. It would be like publishing a generated java client to maven repository and then just add the correct dependency to your consumer and your are ready to use it. This makes it very easy for consumers of your client.

We could do the same thing with npm than a consumer doesn't need to copy paste the generated code in his project but just add the correct npm dependency an he's good to go. 
The swagger codegen team could also publish their generated petstore clients to npm this gives interested developers the change to play with the generated code against our working petstore server.
But this are just some of my ideas for the future.

wing328:
Closed via #2989

PR #2922, opened 2016-05-20
add gradle wrapper mustache for android api client;

zhenjun115:
#2880


cbornet:
The gradlew files should be inherited in the volley library from the default generator. They shouldn't be duplicated.

zhenjun115:
Thinks for you notice, i will remove duplicated files.

zhenjun115:
It seem that the shell script do not find any trail spaces in the mustache template, except for newline symbol.
Here is command,
`find . -not \( -name .git -prune -o -name gradle-wrapper.jar -prune \) -type f -print0 | xargs -0 sed -i '' -e "s/[[:space:]]*$//"`

Issue #1939, opened 2016-01-21
Static HTML generation disregards basePath


pniederw:
When adding a `basePath` such as `/v1` in the JSON, this isn't reflected anywhere in the generated static HTML.

`swagger-codegen-cli:2.1.5`

wing328:
That can be fixed with {{basePathWithoutHost}} mustache tag in the static HTML mustache template.

e..g
https://github.com/swagger-api/swagger-codegen/blob/9a54371ca9e0355dfebd7cdfe4c4632684791c5a/modules/swagger-codegen/src/main/resources/flaskConnexion/README.mustache#L25

Would you have cycle to file a PR for that?

pniederw:
Sorry but I'm not able to make code contributions at this time; best I can do is to file accurate issues.

wing328:
Closed via #2928

wing328:
Thanks for the PR from @catinred2

PR #2928, opened 2016-05-21
add basePathWithoutHost

catinred2:
#1939


wing328:
@catinred2 thanks for the PR. Later we'll consolidate `basePathWithoutHost` and `contextPath`.

cbornet:
@wing328 What is the difference between basePathWithoutHost and contextPath ? They seem identical...

wing328:
@cbornet yes, they're almost identical. We'll consolidate them into one.

Which one do you prefer?

PR #2931, opened 2016-05-21
[Swift] Add sample/tests for Swift client without PromiseKit

Edubits:
Added sample/tests for Swift client without PromiseKit. Also moved the tests with PromiseKit to swift-promisekit (as without is the default config).

cc: @jaz-ah


jaz-ah:
+1 looks good

wing328:
Thanks both! PR merged.

Edubits:
Thanks!

jaz-ah:
hey btw one minor thing I noticed - java/android and other client samples are structured this way:
android/
httpclient	volley
java/
default		feign		jersey2		okhttp-gson	retrofit	retrofit2	retrofit2rx

should we move swift to the same structure - i.e.:
swift/default
swift/promisekit

Edubits:
@jaz-ah, I noticed. However some other languages use this format (e.g. JavaScript and TypeScript). I'm fine with both. 
@wing328 would it be an idea to make this consistent for all languages?

jaz-ah:
@wing328 I feel like it would be more straightforward to go w/ samples/client/petstore/language/language_derivation

and move javascript/typescript and I believe scala (akka-scala,async-scala,etc..) into this pattern... up to you I'm fine either way as well.

wing328:
Both are fine. My preference is the suggestion by @jaz-ah 
```
swift/default
swift/promisekit
```

PR #2934, opened 2016-05-22
Mispelling in Objective-C initialization template

alexandern:


wing328:
@alexandern thanks for the PR to fix the typos.

wing328:
@alexandern btw, I would recommend creating a new branch for the change (as per Git best practice)

PR #2947, opened 2016-05-24
[PHP] Check skipping artifiactVersion in other mustache templates, since it is not required anymore

mikolajprzybysz:
I have found some places that still try to print artifactVersion even though it is not required. So I added possibility to skip it.


modules/swagger-codegen/src/main/resources/php/README.mustache
```diff
@@ -6,8 +6,8 @@
 This PHP package is automatically generated by the [Swagger Codegen](https://github.com/swagger-api/swagger-codegen) project:
 
 - API version: {{appVersion}}
-- Package version: {{artifactVersion}}
-- Build date: {{generatedDate}}
+{{#artifactVersion}}- Package version: {{artifactVersion}}
+{{/artifactVersion}}- Build date: {{generatedDate}}
```
arnested: You can just put `{{#artifactVersion}}` and `{{/artifactVersion}}` on lines of their own.

As of recently whitespace will be stripped around them if they are the only thing on a line.

This looks nicer and more readable in the mustache template.

modules/swagger-codegen/src/main/resources/php/configuration.mustache
```diff
@@ -110,7 +110,7 @@ class Configuration
      *
      * @var string
      */
-    protected $userAgent = "{{#httpUserAgent}}{{{.}}}{{/httpUserAgent}}{{^httpUserAgent}}Swagger-Codegen/{{{artifactVersion}}}/php{{/httpUserAgent}}";
+    protected $userAgent = "{{#httpUserAgent}}{{{.}}}{{/httpUserAgent}}{{^httpUserAgent}}Swagger-Codegen{{#artifactVersion}}/{{{.}}}{{/artifactVersion}}/php{{/httpUserAgent}}";
```
arnested: I think we should put in something instead of the `artifactVersion` when it is not set.

I think I read something about the default user agent is intended to follow a format of `Swagger-Codegen/version/language` (but I can't find the reference for that now).

So maybe we should just put in the string "unknown-version" when we don't have a better guess.

wing328: I think we should keep the default version (1.0.0) to make it consistent with other langauges and have the users set the version properly via `artifactVersion`

mikolajprzybysz:
@arnested Regarding #1 Did you mean, something like this:
```
{{#artifactVersion}}- Package version: {{artifactVersion}}{{/artifactVersion}}
- Build date: {{generatedDate}}
```
I checked it and if I do it like this I end up with empty line

arnested:
I meant:
```
{{#artifactVersion}}
- Package version: {{artifactVersion}}
{{/artifactVersion}}
- Build date: {{generatedDate}}
```
With the recent jMustache this should not produce blank lines.

mikolajprzybysz:
@arnested yup, you are correct, it works, thanks for good advice :+1:

modules/swagger-codegen/src/main/resources/php/configuration.mustache
```diff
@@ -513,11 +513,13 @@ class Configuration
      */
     public static function toDebugReport()
     {
-        $report  = "PHP SDK ({{invokerPackage}}) Debug Report:\n";
+        $report  = 'PHP SDK ({{invokerPackage}}) Debug Report:'."\n";
```
wing328: What about replacing "\n" with PHP_EOL? (I thought you've done it via another PR already, maybe you need to rebase on the latest master)

What about remaining lines with double quote? shall we replace it with single quote as well?

PR #2948, opened 2016-05-24
[PHP] Replaced \n with PHP_EOL

mikolajprzybysz:
I would propose to use PHP_EOL instead \n


wing328:
Thanks for the PR. As part of the change, please update the PHP Petstore sample so that Travis can run CI tests on the updated PHP Petstore.
```
./bin/php-petstore.sh
```

PR #2952, opened 2016-05-24
[Swift] change swift directory structure

jaz-ah:
move everything under a swift directory in samples so we have:
swift/default
swift/promisekit
@Edubits @wing328


Edubits:
Looks good! One little todo: edit the path in https://github.com/swagger-api/swagger-codegen/blob/master/pom.xml

PR #2962, opened 2016-05-25
[C#] To fix error when model name is `Client`

### LINKED ISSUE #2943: [csharp] If you have a definition named "Client", generated code wont' compile
##### Description
If you have a class definition named `Client`, the generated code won't compile because of a conflict with the namespace.  `Client` seems like it might be a very common model name (`Api` not so much, maybe `Model`, but probably not as common as `Client`).

##### Swagger-codegen version
master, cloned at the time of posting

N/A

java -jar swagger-codegen-cli.jar generate -i http://localhost:50002/swagger/docs/v1 -l csharp -c swagger-codegen-csharp-config.json --import-mappings Object=System.Object -o ./

##### Steps to reproduce
Just have a definition named Client in your spec.

##### Related issues
Couldn't find any related issues.

##### Suggest a Fix
Change the generated namespace from Client to ApiRestClient (worked well for my in my template).

wing328:
To fix #2943


modules/swagger-codegen/src/main/resources/csharp/api.mustache
```diff
@@ -81,10 +81,10 @@ namespace {{packageName}}.Api
         /// <returns></returns>
         public {{classname}}(String basePath)
         {
-            this.Configuration = new Configuration(new ApiClient(basePath));
+            this.Configuration = new {{packageName}}.Client.Configuration(new ApiClient(basePath));
```
jimschubert: I believe you should still need to remove `using {{packageName}}.Client;` from line 7 in this file and make all `ApiClient` references fully qualified. I'll pull your branch and the examples in Windows.

jimschubert: I get compilation errors on the SwaggerClient solution in Visual Studio:

image

modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml
```diff
@@ -990,6 +990,16 @@ definitions:
         type: array
         items:
           $ref: '#/definitions/ReadOnlyFirst'
+  Configuration:
+    type: object
+    properties:
+      configuration:
+        type: string
+  Client:
```
jimschubert: I think this model needs to have an associated endpoint, since the clashing issue is in the *Api types.

modules/swagger-codegen/src/main/resources/csharp/api.mustache
jimschubert: I made those changes and opened a PR to your branch. wing328/swagger-codegen#5

I didn't update the yaml (see my other comment).

modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml
wing328: I'll add it tomorrow as I've another test case that I need to cover.

wing328:
The PR would need more work as the documentation will look poor if using fully-qualified name for model names, etc. Closing this for now.

PR #2965, opened 2016-05-25
[Objc] Basic Core data support [Optional]

mateuszmackowiak:
Basic CoreData support: 
- Generates entities NSManagedObject
- Generates basic xcdatamodel
- Generates converter class from NSManagedObject to basic object and back

Change is non breaking


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ObjcClientCodegen.java
```diff
@@ -154,6 +152,7 @@ public ObjcClientCodegen() {
         instantiationTypes.put("map", "NSMutableDictionary");
 
         cliOptions.clear();
+        cliOptions.add(new CliOption(CORE_DATA, "Should generate core data models").defaultValue("n"));
```
wing328: for consistency with other cli options, shall we use "true", "false" instead?

wing328:
I'll cherry-pick the commits related to CoreData

wing328:
Close via #2988

Issue #2963, opened 2016-05-25
Add Apache v2 license to auto-generated files


wing328:
We will update the auto-generated files for all languages (both API clients and server stubs).
#2960 for C# API client is a good reference.

API clients
- [ ] Android
- [x] C++
- [x] C# (#2960)
- [ ] Clojure
- [ ] Dart
- [ ] Flash
- [x] Go
- [x] Javascript
- [ ] Java
- [x] ObjC
- [x] Perl
- [X] PHP (#2971)
- [X] Python (#2972)
- [x] Ruby (#2946 #2951 #2953)
- [ ] Swift
- [ ] Scala
- [x] TypeScript

Server stubs
- [ ] C# ASP.NET5
- [ ] Haskell Servant
- [ ] Java Spring Boot
- [ ] JAX-RS RestEas
- [ ] JAX-RS CXF
- [ ] PHP Lumen
- [ ] PHP Slim
- [x] Ruby on Rails 5
- [x] Scalatra (#3122)

ePaul:
Is this needed for some legal reason? Also, is this legally sound?

One could say that the generated code is derived from both the template files (which certainly have the same license as the rest of this project) and the API definition – the latter one is provided by the user, and could have conflicting licensing requirements (e.g. a copyleft license as GPL). In that case I guess the generated code would need to be GPL, too (or the user would not be allowed to use the generator).

wing328:
@ePaul thanks for your questions and please refer to the following discussion for more information.

https://groups.google.com/forum/#!searchin/swagger-swaggersocket/license/swagger-swaggersocket/6JcsR4mWpYk/pd3__T0mHwAJ

https://twitter.com/TimHaines/status/729717172889489408

Disclaimer: I'm no expert in software licensing

ePaul:
@wing328 Thanks. Just for the future, it would be nice if (in cases there has been some discussion) this discussion is linked right when the issue is opened, so people know why stuff is done.

wing328:
Yup, definitely valid point. I've been trying my best to keep the community informed but sometimes I did miss it. Feel free to ask whenever you've questions for us.

xtoddx:
tl;dr: You should license the template files Apache 2, but not assume it applies it to the resulting generated code.

[...]

PR #2966, opened 2016-05-25
[Objc] Add Apache v2 license to auto-generated files #2963

mateuszmackowiak:


modules/swagger-codegen/src/main/java/io/swagger/codegen/SupportingFile.java
```diff
@@ -44,4 +48,17 @@ public int hashCode() {
         result = 31 * result + (destinationFilename != null ? destinationFilename.hashCode() : 0);
         return result;
     }
-}
\ No newline at end of file
+
+    static public class GlobalSupportingFile extends SupportingFile {
```
jimschubert: Is there a specific reason this class is nested?

modules/swagger-codegen/src/main/resources/_common/LICENSE.mustache
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
jimschubert: If a template file doesn't need to be processed through mustache, you can change the `.mustache` extension to something else to skip running it through the processor. [See DefaultCodegen](https://github.com/mateuszmackowiak/swagger-codegen/blob/d82c77da34a29cd9bba71dfde57c9479ae39ad7f/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java#L550).

This obviously won't be a significant performance gain, but it's useful to know in case we find some text that breaks the mustache template processor.

jimschubert:
If our goal is to include a LICENSE file for all generated client and server code, you could do like I did for the `.swagger-codegen-ignore` and write it at the end if it doesn't exist. [See this part of DefaultGenerator](https://github.com/mateuszmackowiak/swagger-codegen/blob/d82c77da34a29cd9bba71dfde57c9479ae39ad7f/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java#L596).

We'd probably need to discuss a common convention of naming the file `LICENSE`. This wouldn't address license headers in code files, but it would ensure the license exists in all generated code.

modules/swagger-codegen/src/main/java/io/swagger/codegen/SupportingFile.java
mateuszmackowiak: I treat this as a integral part of the SupportingFile but I can move it to separate place.

wing328:
@mateuszmackowiak thanks for the PR. Later when you've time, can you please add documentation to `modules/swagger-codegen/src/main/java/io/swagger/codegen/GlobalSupportingFile.java` and its methods?

wing328:
@jimschubert I've reused your code to add a default license for all generators (as we cover too many languages/frameworks) and filed #2993.

jimschubert:
@wing328 Looks good!

PR #2969, opened 2016-05-25
[Go] add apache v2 LICENSE

guohuang:
@wing328 please review, should we share the partial header and the license file so that we don't have to copy it on every language?


wing328:
@guohuang good question. I tried before by putting the file one level up but encountered cross-platform issue. If you find a way to do it, please file a PR.

PR #2971, opened 2016-05-25
[PHP] Add apache v2 license to PHP client

arnested:
See #2963.

The PHP client already had some license headers. I have changed it to include the same license header and approach as the C# implementation (#2960).


wing328:
@arnested thanks for the PR, including minor enhancement to the doc.

PR #2973, opened 2016-05-26
update package.json and fix webpack.html

leonyu:
* Update `tsify`
* Fix webpack.html file


leonyu:
The version of `tsify` appeared to be the problem with CI. This PR updates it to latest. 

It might be beneficial to pin the NPM package versions. There are a couple of methods, I will need to investigate which one is the best for codegen.

wing328:
@leonyu thanks for the fix. I'll review and merge later today.

wing328:
@leonyu that works. I've included TS Fetch API clients in the CI via https://github.com/swagger-api/swagger-codegen/pull/2976.

PR #2977, opened 2016-05-26
Remove unnecessary 2nd call to SwaggerParser.

pablograna:
Both CodeGenMojo and CodegenConfigurator where parsin the swagger input.
The result in CodeGenMojo was discarded. I simply removed the line in
CodeGenMojo.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/HaskellServantCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/HaskellServantCodegen.java
@@ -7,7 +7,6 @@ import io.swagger.models.properties.*;
 import io.swagger.models.Model;
 import io.swagger.models.Operation;
 import io.swagger.models.Swagger;
-import sun.reflect.generics.reflectiveObjects.NotImplementedException;
 
 import java.util.*;
 import java.io.File;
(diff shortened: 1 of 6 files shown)


wing328:

pablograna:
I think I've fixed it, now the page shows the commits.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/HaskellServantCodegen.java
```diff
@@ -432,11 +431,11 @@ private String makeQueryListType(String type, String collectionFormat) {
             case "pipes": return "(QueryList 'PipeSeparated (" + type + "))";
             case "multi": return "(QueryList 'MultiParamArray (" + type + "))";
             default:
-                throw new NotImplementedException();
+                throw new UnsupportedOperationException();
```
wing328: I suggest we keep `NotImplementedException` as existing programs may be a dependency on this. If we switch it to `UnsupportedOperationException`, we'll need to flag this change as breaking change, which seems not necessary.

wing328:
Yes, I think you've fixed it as confirmed in the commit tabs.

Btw, I would suggest you to create a new branch for the change as per Git best practice.

pablograna:
Here is the 'clean' PR, just the line removed in a new branch:

https://github.com/swagger-api/swagger-codegen/pull/3005

Should I close this PR?

Issue #2970, opened 2016-05-25
Request: Go server codegen


peterbourgon:
First, forgive me if this is the wrong way to make this request. Thank you for the Go client generator. I would love to bring Swagger into my project (https://gokit.io) but to do that I'd need a Go server generator. I want to file the issue to register the interest and offer my help to review any PR someone could make — I don't have any Java chops, so I unfortunately won't be much help in making the PR :)

wing328:
@peterbourgon I think there are many Go server-side framework out there. Why one do you prefer?

I reviewed before and would suggest https://github.com/gin-gonic/gin

peterbourgon:
Most Go web "frameworks" are nonidiomatic and should probably be avoided for codegen purposes, including Gin. I would recommend plain `net/http` handlers + selected pieces from https://github.com/gorilla as necessary; probably just https://github.com/gorilla/mux for routing is enough.

wing328:
If anyone from the community wants to implement the Go server generator based on @peterbourgon suggestion or using some Go sever-side framework, please reply to let us know.

guohuang:
i could give it a shot tomorrow

fehguy:
I would like to see go-swagger in the server generator.  @casualjim can you help?

ysqi:
@peterbougon ,Beego (a golang web framework) will be useful to you, You can write the API by Beego and use Beego command tool to generate document automatically for Swagger , more doc [here](http://beego.me/docs/advantage/docs.md#generate-document-automatically)
code such as like this:
```go

// @Title Get Product list
// @Description Get Product list by some info
// @Success 200 {object} models.ZDTProduct.ProductList
// @Param   category_id     query   int false       "category id"
// @Param   brand_id    query   int false       "brand id"
// @Param   query   query   string  false       "query of search"
// @Param   segment query   string  false       "segment"
// @Param   sort    query   string  false       "sort option"
// @Param   dir     query   string  false       "direction asc or desc"
// @Param   offset  query   int     false       "offset"
// @Param   limit   query   int     false       "count limit"
// @Param   price           query   float       false       "price"
// @Param   special_price   query   bool        false       "whether this is special price"
// @Param   size            query   string      false       "size filter"
// @Param   color           query   string      false       "color filter"
// @Param   format          query   bool        false       "choose return format"
// @Failure 400 no enough input
// @Failure 500 get products common error
// @router /products [get]
func (c *CMSController) Product() {
[...]

PR #2979, opened 2016-05-27
issue#2970, [Go] add go server codegen template

guohuang:
For #2970

@wing328 this is the initial commit for the go server codegen, there is no implementation for the endpoint, all returning 200 now. welcome for any feedback.

cc: @peterbourgon see if this works for you.


modules/swagger-codegen/src/main/resources/go-server/README.mustache
```diff
@@ -0,0 +1,21 @@
+# Swagger generated server
+
+## Overview
+This server was generated by the [swagger-codegen](https://github.com/swagger-api/swagger-codegen) project.  By using the [OpenAPI-Spec](https://github.com/OAI/OpenAPI-Specification) from a remote server, you can easily generate a server stub.  This is an example of building a node.js server.
```
jimschubert: README wasn't updated. This line references a node.js server. Also, instructions for running the server are missing.

modules/swagger-codegen/src/main/resources/go-server/logger.mustache
```diff
@@ -0,0 +1,23 @@
...
+
+        inner.ServeHTTP(w, r)
+
+        log.Printf(
+            "%s\t%s\t%s\t%s",
+            r.Method,
+            r.RequestURI,
+            name,
+            time.Since(start),
+        )
```
peterbourgon: Is this meant to be an example that you'd expect users to change completely, or do you expect users to "embrace and extend" from this foundation?

modules/swagger-codegen/src/main/resources/go-server/main.mustache
```diff
@@ -0,0 +1,15 @@
+package main
+
+import (
+	sw "./{{apiPath}}"
```
peterbourgon: Relative imports (import strings that begin with `.`) aren't kosher. This needs to be a fully-qualified path.

modules/swagger-codegen/src/main/resources/go-server/routers.mustache
```diff
@@ -0,0 +1,55 @@
...
+	Name        string
+	Method      string
+	Pattern     string
+	HandlerFunc http.HandlerFunc
+}
+
+type Routes []Route
+
+func NewRouter() *mux.Router {
+
```
peterbourgon: Extraneous newline

modules/swagger-codegen/src/main/resources/go-server/logger.mustache
guohuang: @peterbourgon I expect user will extend this as needed, do you have any suggestions to include some features there?

modules/swagger-codegen/src/main/resources/go-server/main.mustache
guohuang: all the generated code (except main.go) will be in that folder, so this is better than fully qualified path, (I expect people copy the generated code to their own project)

peterbourgon: Hm, I don't think you understand :) Relative import paths compile _in some circumstances_ to aid in very specific testing scenarios. But "Go programs cannot use relative import paths within a work space" so this breaks as soon as the user puts this code into their project. In fact, the "feature" is so often mis-used, that it is being discussed to remove it from the language entirely. See https://golang.org/cmd/go/#hdr-Relative_import_paths. 

I don't know much about how swagger-codegen works. But, I would expect to run some CLI tool from my project directory, which would be in my $GOPATH. And I'd expect to provide the fully-qualified import path prefix. So something like this

```
$ cd $GOPATH/src/github.com/peterbourgon/myproject
$ swagger-codegen --import-prefix=github.com/peterbourgon/myproject --output-dir=myapi resources/swagger.spec
reading resources/swagger.spec
putting generated code in myapi/
using import path "github.com/peterbourgon/myproject/myapi" for generated code
writing main.go
writing logger.go
writing controller.go
done
$
```

`--import-prefix` could be deduced from the current context (reading GOPATH) and `--output-dir` could have some sensible default.

If this isn't possible, the next-best thing would be to put a big WARNING or DISCLAIMER around that import statement.

```go
import (
    // WARNING!
    // Change this to a fully-qualified import path
    // once you place this file into your project.
    // For example,
    //    sw "github.com/myname/myrepo/{{apiPath}}"
    sw "./{{apiPath}}"
)
```

modules/swagger-codegen/src/main/resources/go-server/logger.mustache
peterbourgon: OK. In that case I'd just suggest to change the tab to a space.

guohuang: according to `gofmt` https://golang.org/cmd/gofmt/#hdr-Examples, it has to be tab and width=8  

yes, it is ridiculous, I love spaces too, but it's their standard.

peterbourgon: Oh, hehe, no, I mean the `\t` literal in the log line string :)

modules/swagger-codegen/src/main/resources/go-server/routers.mustache
guohuang: fixed

modules/swagger-codegen/src/main/resources/go-server/README.mustache
guohuang: good catch, I have added instruction for running the server. this is still the initial PR, we need to extend it later.

modules/swagger-codegen/src/main/resources/go-server/logger.mustache
guohuang: sorry, I misunderstood it. it's fixed

modules/swagger-codegen/src/main/resources/go-server/main.mustache
guohuang: ah, thanks for explaining, since everyone will have a different fully-qualified path, I will take the warning approach.(code has been updated.)

ysqi:
@guohuang ,  Can use Swagger to test API in [Beego](https://github.com/astaxie/beego) , and doc see [here](http://beego.me/docs/advantage/docs.md#generate-document-automatically)  , You can Write the API use Beego and use beego command tool to generate document automatically for Swagger

wing328:
@guohuang thanks for the PR. I'll merge it tomorrow (Tuesday)

wing328:
@ysqi `Beego` looks promising. if you've cycle to contribute a Beego server stub generator, please let me know and I'll show you some good starting point.

Issue #2967, opened 2016-05-25
[C#] Map `number` to `Decimal`


wing328:
Similar to Java which maps `number` to `BigDecimal`,  we should do the same for C# to make it consistent across different langauges by mapping `number` to `Decimal` in C#.

If anyone has any concern or suggestion, please kindly voice out.

jimschubert:
Would this be the same as #2813 , or would that issue then require an option to generate as `Double` instead? I think we had discussed this previously.

wing328:
My understanding is that we need the following mapping:

float	(type:number, format: float) => float	
double (type: number, format: double) => double
number (type: number, no format) => decimal

(currently number is mapped to double as well)

wing328:
#2813 has been closed due to no response from the user.

Issue #2813, opened 2016-05-10
how to create decimal in yaml for C# generate code


saharj:
_From @deepakbhise on May 9, 2016 15:24_

_Copied from original issue: swagger-api/swagger-editor#921_

wing328:
@deepakbhise the current type mapping can be found in https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java#L112

For your use case, is it correct to say that you must use decimal?

wing328:
@deepakbhise closing the issue as there's no update from you.

PR #2990, opened 2016-05-28
[csharp] map number/no format to Decimal

jimschubert:
Changes the default primitive for `number` with no specified format from `double` to `decimal` in C#.

Other number/integer mappings remain unchanged:

| type | format | C# type |
|-------|----------|------------|
| number | float | System.Float |
| number | double | System.Double |
| number | ------ | System.Decimal |
| integer | int64 | System.Int64 | 
| integer | int32 | System.Int32 | 
| integer | ------ | System.Int32 | 

See #2967 and #2813.

@wing328 this seems like an extremely simple fix, and I feel like I've missed something. I generated the csharp client and verified that types looked fine and tests passed.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java
@@ -77,6 +77,7 @@ public abstract class AbstractCSharpCodegen extends DefaultCodegen implements Co
                         "string",
                         "bool?",
                         "double?",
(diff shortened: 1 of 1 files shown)


wing328:
@jimschubert I agree with you that this should be a simple fix. I'll mark this as a breaking change with fallback as `number` will be mapped to `decimal` instead of `double`.

wing328:

Starting from 2.2.0, `number` (without format) will be mapped to `decimal?` for C#-related generator to make it more consistent with other languages (e.g. Java).

To fallback, please specify the `format` of the property/parameter as `double`

PR #3001, opened 2016-05-30
Feign: disable FAIL_ON_UNKNOWN_PROPERTIES by default

reda-alaoui:
By default, the feign jackson decoder doesn't fail on unknown properties.
I think the generated feign generated code should follow the same behaviour.


reda-alaoui:
Sorry, I didn't create the PR from a branch.
I will recreate it.

PR #3002, opened 2016-05-30
add java8-offset dateLibrary

ebelanger:
Adds support for Java8 OffsetDateTime instead of LocalDateTime.


cbornet:
Is it really useful to add yet another configuration option to the generator ?
You can already change the mappings with ```--type-mappings DateTime=OffsetDateTime --import-mappings OffsetDateTime=java.time.OffsetDateTime```

ebelanger:
and I didn't know that :)  Thank you so much for this :)

PR #3015, opened 2016-05-31
Use OffsetDateTime instead of LocalDateTime

### LINKED ISSUE #2138: Don't use LocalDateTime for java8 datetime mapping
LocalDateTime doesn't have any time zone so it's error prone for systems that don't communicate in the same timezone.
Instant, ZonedDateTime or OffsetDateTime types should be considered instead.

cbornet:
Fix #2138


wing328:
@cbornet thanks for the PR. If I understand it correctly, existing apps would break and the developers will need to replace `LocalDateTime` with `Instance` to fix the issue after upgrade.

Would it be hard to make it backward compatible by introducing an option `LocalDateTime` so that developers can still use `LocalDateTime` if needed (and java8 will default to `Instant`)?

cbornet:
I need to check but maybe it's possible to keep LocalDateTime with --typeMappings/--importMappings options. I will check if it works or if dateLibrary takes precedence.

wing328:
@cbornet thank you. I think there may be still Java7 developers using LocalDateTime so keeping LocalDateTime will provide them a better upgrade experience to the upcoming version.

cbornet:
This PR only affects java8 users so java7 users will not see any difference. Note that we could do the same kind of improvement for java7 by replacing joda's LocalDateTime by DateTime.

cbornet:
Actually joda already uses DateTime so it's fine there.
I tried --typeMappings/--importMappings to set LocalDateTime but the dateLibrary overwrites it :disappointed: 
So I will do a java8-localdatetime option for backward compatibility

wing328:
@cbornet the library option for `jaxrs` seems not correct
```
swagger-codegen|cbornet-java_instant⚡ ⇒ java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar config-help -l jaxrs
...
	library
	    library template (sub-template) to use (Default: jersey1)
	        <default> - HTTP client: Jersey client 1.19.1. JSON processing: Jackson 2.7.0
	        feign - HTTP client: Netflix Feign 8.16.0. JSON processing: Jackson 2.7.0
	        jersey2 - Jersey core 2.x (default)
	        okhttp-gson - HTTP client: OkHttp 2.7.5. JSON processing: Gson 2.6.2
	        retrofit - HTTP client: OkHttp 2.7.5. JSON processing: Gson 2.3.1 (Retrofit 1.9.0)
	        retrofit2 - HTTP client: OkHttp 3.2.0. JSON processing: Gson 2.6.1 (Retrofit 2.0.2). Enable the RxJava adapter using '-DuseRxJava=true'. (RxJava 1.1.3)
	        jersey1 - Jersey core 1.x
...
```
It should only show jersey1 and jersey2.

This is not your fault but if you've cycle I wonder if you can fix it in this PR. (we probably need to review all available options later as most seems to be carried over from Java client generator and may not apply in the Java server generator)

cbornet:
I am working on a refactoring of the Java codegens introducing an AbstractJavaCodegen. This should solve this issue.

wing328:
OK. I believe you will submit a separate PR for the refactoring. 

I'll merge this PR tomorrow (Thur)

cbornet:
Yes this will be another PR

PR #3030, opened 2016-06-02
3006/sbt for java scala

clasnake:
Add sbt support for jersey2, fix dependency issue for okhttp-gson.
Only include junit is not enough to get the test cases to run when typing sbt test.


wing328:
@clasnake the change looks good to me. Thanks for the contribution.

clasnake:
@wing328 Great. I'll add sbt support for other java clients and the scala client later.

PR #3031, opened 2016-06-02
[ObjC] Fix various bugs in generated code.

GriffinSchneider:
Fix a bug that's been around for a while where ApiClient.timeoutInterval didn't do anything. Every time `requestWithPath:...` is called, a new `requestSerializer` is set, which is why the timeout update needs to happen in `setRequestSerializer:`. `60` is the default timeout interval for `NSURLRequest`. I couldn't find a way to get that value programmatically.

Also, when I updated my clone of `swagger-codegen`, my generated code failed to compile for a number of reasons. I had to fix these bugs to get the compile working again. All of these were introduced in the past month or so, it seems like I'm just cleaning up after someone else's mess here.
- Change `#import <AFNetworking.h>` back to `#import <AFNetworking/AFNetworking.h>`. XCode can't locate the header with a normal CocoaPod dependency when it's just `<AFNetworking.h>`.
- Fix usages of `SWG` instead of `{{classPrefix}}` which make the compile break whenever a `classPrefix` is configured.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/objc/ApiClient-body.mustache
+++ b/modules/swagger-codegen/src/main/resources/objc/ApiClient-body.mustache
@@ -50,6 +50,7 @@ static NSString * {{classPrefix}}__fileNameForResponse(NSURLResponse *response) 
 - (instancetype)initWithBaseURL:(NSURL *)url {
     self = [super initWithBaseURL:url];
     if (self) {
+        self.timeoutInterval = 60;
         self.requestSerializer = [AFJSONRequestSerializer serializer];
         self.responseSerializer = [AFJSONResponseSerializer serializer];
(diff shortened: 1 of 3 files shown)


wing328:
@GriffinSchneider thanks for the PR.

cc @mateuszmackowiak

wing328:
@GriffinSchneider thanks for the fix on `SWG`.

I also did a grep on SWG under the ObjC resource template directory and confirmed there's no more reference to SWG

mateuszmackowiak:
@wing328 +1

wing328:
@mateuszmackowiak thanks for the review. PR merged.

@GriffinSchneider thanks for the PR. I would recommend creating a new branch for your future PR as per Git best practices

Issue #3029, opened 2016-06-02
[Python] Model's constructor should provide argument to initialize properties


taxpon:
##### Description
- Currently, model's constructor do not provide arguments to initialize properties
- If there are arguments to initialize propeties, developer can initialize instance with less typings in some cases.

##### Swagger-codegen version
2.2.0-SNAPSHOT

##### Suggest a Fix
 Generated code example (`animal.py`)
###### before
```python
    def __init__(self):
        self._class_name = None
        self._color = 'red'
```

###### after
```python
    def __init__(self, class_name=None, color='red'):
        self._class_name = class_name
        self._color = color
```

taxpon:
Merged #3034. Issue closed.

PR #3034, opened 2016-06-02
Update python constrouctor to provide arguments for instance initialization

taxpon:
PR for Issue #3029


wing328:
@taxpon thanks for the PR.

cc @scottrw93

wing328:
@taxpon here are some feedbacks I would like to share with you:

1) currently your implementation will list *all* properties in the constructor arguments. There are read-only properties that we should skip.

Please refer how C# provides the constructor: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/csharp/modelGeneric.mustache#L42

(readWriteVars contains all properties except readonly)

Ref: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schema-object

2) Later we may also want to add validation to make sure required properties has a value provided by the user. One caveat is that the deserialisation process relies on creating an model using the default constructor: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/python/api_client.mustache#L570

If validation is added to required property, that line may throw exception.

scottrw93:
@wing328 On point 2, I don't see this as an issue, assuming  the constructor adds some condition to only check the validity of a value when it is not `None` when validation is implemented. The deserialisation  code with assign all values as `None` anyway.

wing328:
@scottrw93 OK. I believe we also have test cases to cover deserializations (e.g. get_pet_by_id) so the test cases will complain if the validation introduces any issue.

taxpon:
@wing328 I fixed regarding your mention 1). I think we should create another issue for 2) if it's required. By the way, I think generator should not provide `@peoperty.setter` method for readonly variables. What do you think? There are setters for readonly variables in current generated code.

An example in `read_only_first.py`, there is a method like this:
```python
    @bar.setter
    def bar(self, bar):
        """
        Sets the bar of this ReadOnlyFirst.


        :param bar: The bar of this ReadOnlyFirst.
        :type: str
        """

        self._bar = bar
```

wing328:
@taxpon yes, you're right that read only attribute should not have setter and we've implemented that in some of the languages (C#, PHP, Ruby, maybe a few more). It would be great if you can help implement that. If you need some starting points, please feel free to ping me.

taxpon:
@wing328 I've omitted property.setter methods for readonly properties in python client (2ea3720).

wing328:
@taxpon thanks for adding readonly properties support to the python client. I ran the tests but failed with the following error messages:
```
======================================================================
ERROR: Failure: ValueError ('/Users/williamcheng/Code/swagger-api/swagger-codegen/samples/client/petstore/python/swagger_client/__init__.py/__init__.pyc' is not a package)
----------------------------------------------------------------------
  File "/Users/williamcheng/Code/swagger-api/swagger-codegen/samples/client/petstore/python/.tox/py34/lib/python3.4/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/Users/williamcheng/Code/swagger-api/swagger-codegen/samples/client/petstore/python/.tox/py34/lib/python3.4/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/Users/williamcheng/Code/swagger-api/swagger-codegen/samples/client/petstore/python/.tox/py34/lib/python3.4/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/Users/williamcheng/Code/swagger-api/swagger-codegen/samples/client/petstore/python/.tox/py34/lib/python3.4/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/Users/williamcheng/Code/swagger-api/swagger-codegen/samples/client/petstore/python/.tox/py34/lib/python3.4/imp.py", line 245, in load_module
    return load_package(name, filename)
  File "/Users/williamcheng/Code/swagger-api/swagger-codegen/samples/client/petstore/python/.tox/py34/lib/python3.4/imp.py", line 210, in load_package
    raise ValueError('{!r} is not a package'.format(path))
ValueError: '/Users/williamcheng/Code/swagger-api/swagger-codegen/samples/client/petstore/python/swagger_client/__init__.py/__init__.pyc' is not a package
```
May I know if the tests passed for you in your local machine?

taxpon:
I ran tests in my local machine and passed successfully. I'll check again.

wing328:
@taxpon Let me clone your repo in a new folder to see if I get the same issue...

wing328:
@taxpon sorry the problem probably due to local issue in my machine. I gave it try in another machine and it works:
```
______________________________________________________ summary _______________________________________________________
  py27: commands succeeded
  py34: commands succeeded
  congratulations :)
```

wing328:
FYI. I've added CI for Python API client via https://github.com/swagger-api/swagger-codegen/pull/3114

dthorpe:
If a property labeled read-only has no setter generated for it, then how is deserialization supposed to work?  As in, returning an object in an http result which contains properties marked in swagger as read-only.

I'm seeing an issue where we get an AttributeError: attribute cannot be set error in api_client.py::__deserialize_model() line 589, the call to  `setattr(instance, attr, self.__deserialize(value, attr_type))`

The setattr call is failing because the instance (whose type is generated by swagger-codegen) doesn't have a property setter!

wing328:
@dthorpe thanks for reporting the issue with the deserialization. I think what we'll do it to remove the read-only attribute handling in the Python model template so that all properties will have a setter: https://github.com/swagger-api/swagger-codegen/issues/3210

wing328:
cc @scottrw93

wing328:
@dthorpe the PR (fix) by @scottrw93 has been merged into master. Please pull the latest master to give it a try.

PR #3039, opened 2016-06-03
[Spring] Add an option to only generate stubs of the API and no server files

### LINKED ISSUE #3025: [Spring] Spring generators
I have started working on a [spring-cloud-feign](http://projects.spring.io/spring-cloud/spring-cloud.html#spring-cloud-feign) client generator and I see that the most flexible way to do it would be to generate the models and interfaces for the APIs as described [here](http://projects.spring.io/spring-cloud/spring-cloud.html#spring-cloud-feign-inheritance) and let the caller configure its @FeignClient beans the way he wants (w/wo eureka, w/wo hystrix, etc...).

Most of the files and options would be common with spring-mvc generator. The spring-mvc and springboot generators also share most of their files and options. So I think we should regroup all spring generators either with a common abtract codegen (as is done for jaxrs/c#/typescript) or by library support (as is done for java/android). Which solution is better in your opinion ? Or do we keep the code duplicated ?
cc @fehguy @wing328

Note : spring-mvc and springboot are server generators, spring-cloud-feign would be a client generator

cbornet:
This allows to:
* easily upgrade your existing Spring-MVC / Spring-Boot server with new endpoints from the swagger file (kind of like the j8-async library but without java8's "default" keyword)
* easily create @FeignClient clients in a Spring-Cloud micro-service application

Note: although the option is on Spring-Boot codegen, the generated code also works in a Spring-MVC server.

Fix #3025


cbornet:
I've also added an option to use the x-accept x-contentType vendor extensions that are used in the java-feign generator because @FeignClient beans need a single produces/consumes content type.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SpringBootServerCodegen.java
```diff
@@ -119,9 +131,6 @@ public void addOperationToGroup(String tag, String resourcePath, Operation opera
         if (basePath == "") {
             basePath = "default";
         } else {
-            if (co.path.startsWith("/" + basePath)) {
-                co.path = co.path.substring(("/" + basePath).length());
-            }
```
ePaul: That looks like a change unrelated to the change in the PR description. What is its purpose?

ePaul:
Generally a good idea. It goes in the direction of allowing to generate which each build instead of only once. (I'm doing this with [the templates in swagger-codegen-tooling](https://github.com/zalando-stups/swagger-codegen-tooling/tree/master/swagger-codegen-templates).)

Though I'm not sure if this will work easily, as the whole maven project is generated, so you still need several maven calls if this generation is done from maven.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SpringBootServerCodegen.java
cbornet: It's because if you put a RequestMapping annotation on the interface class then Spring-MVC will make a controller of any implementation bean (even if it's not a @Controller) and this is undesirable in use cases such as spring-cloud-feign. This commit removes the RequesMapping on the interface and puts the path segment into the methods. In my opinion this is also better for controllers since you can use RequestMapping on the controller to configure the basePath (eg. "v2" in petstore).

cbornet: And this is exactly what you do in your interface implementation :smile: (except that you put back in the template the path segment that was removed by the codegen) !

cbornet:
You can generate only api and models without supporting files using -D options if you want.

thebignet:
LGTM

thebignet:
Probably commit 556f529 should have been kept apart for this PR.

cbornet:
I have a lot of pending PRs that I want to do for the Spring gen. I try to do it with small increments but if I do them too small, it will take forever...

thebignet:
That's understandable @cbornet, though the longer the PR, the longer it takes to be merged.

And I too want this PR merged !

ePaul:
:+1:

Issue #2800, opened 2016-05-08
when language is jaxrs the defined template is ignored


gpekofsky:
If I set
 -l jaxrs -t /path/to/template
The template is always ignored.

This seems to be because in JavaJerseyServerCodegen.java
embeddedTemplateDir = templateDir = JAXRS_TEMPLATE_DIRECTORY_NAME + File.separator + "jersey1_18";

The specified template path is ignored altogether.

wing328:
@gpekofsky thanks for reporting the issue. Did you have a chance to try other server stub generator (e.g. springmvc) to see if the same issue persists? That would help us to easily plot the fix from other generator if the issue is jaxrs-specified. Thanks.

narayanan:
Generating server for jersey2 also does not work. Looks like these two are related. Code generator always generates jersey1_18 even if the value for `--library` option is `jersey2`. Here is the command I used to generate jersey2 server.

`java -jar swagger-codegen-cli.jar generate -i ./my-api-swagger.yaml -l jaxrs -o my-api --library jersey2`

narayanan:
Update: I don't see the "jersey2 server" not getting generated issue when I built the CLI jar from master and tried the same command. Looks like the issue is fixed recently.

Looking forward for 2.2.0 release to leverage the fix

wing328:
@gpekofsky would you please pull the latest master and give it another try?

aboretum:
It depends on where it works as intended, if the intended use for `-t` is for us to supply the entire template (`mustache`) files for server stub, then it is working on my machine for jersey-1.18. If it is used to overwrite only an individual file, the `-t` option seem to be ignored completely.

wing328:
@gpekofsky @narayanan please pull the latest and give it another try. (I've merged a PR that should have fixed the issue)

qtstc:
I am seeing the issue with version 2.16 and the current master build. Tried both cli tool and maven plugin, the template directory is not used. I don't think it is specific to jaxrs though, tried ruby, also seems to have the same problem. The template was used in 2.14 and 2.15.

wing328:
@qtstc thanks for doing a check with the latest master to confirm the issue. I'll look more into it and hopefully file a PR for the fix by this weekend.

wing328:
I did some tests and the `-t` (template) option works for me for both `ruby` and `jaxrs`.

Please share the full commands you used for repeating the issue.

aboretum:
[...]

PR #3043, opened 2016-06-03
Stop Reading Custom Template Files from subdirectory

qtstc:
#2800


wing328:
@qtstc thanks for the PR. I tested with the following:
- ruby without template [ok]
- ruby with template [ok]
- jaxrs (default: jersey2) without template [ok]
- jaxrs (library=jersey1) without template [ok]
- jaxrs (default: jersey2) with template [ok]
- jaxrs (library=jersey1) with template [ok]

PR #3046, opened 2016-06-04
enable python api client to return just data without http header info when need

zhenjun115:
fix bug #2536


wing328:
@zhenjun115 thanks for the contribution. PR merged into master.

PR #3056, opened 2016-06-06
[Objc] Support for x-unique-id-key vendorExtensions

mateuszmackowiak:


modules/swagger-codegen/src/test/resources/2_0/petstore.json
```diff
@@ -880,7 +880,8 @@
       "properties": {
         "id": {
           "type": "integer",
-          "format": "int64"
+          "format": "int64",
+          "x-unique-id-key"  : true
```
wing328: Let's keep the petstore.json/yaml as it's and add additional test-related model properties in `modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml` instead (you will need to update `./bin/objc-petstore.sh` to use petstore testing yaml file)

modules/swagger-codegen/src/main/resources/objc/model-body.mustache
```diff
@@ -13,7 +13,11 @@
   }
   return self;
 }
-
+{{#vars}}{{^vendorExtensions.x-unique-id-key}}{{/vendorExtensions.x-unique-id-key}}{{#vendorExtensions.x-unique-id-key}}
+-(NSString*)indexPropertyName {
+    return @"{{name}}";
+}
+{{/vendorExtensions.x-unique-id-key}}{{/vars}}
```
wing328: If this vendor extension is objc-specified, shall we name it as `x-objc-unique-id-key` so that it's immediately clear to the others this vendor extension is used by objc client generator?

mateuszmackowiak: Hi,
So I was actually thinking that it would be a good idea to add a unique id for all clients. It’s just an idea but it’s very common for any object to have this kind of property..
But of course I can change it and add the x-objc prefix.

wing328: Can you please elaborate a bit more on its usage? From the example above, it adds an additional property "indexPropertyName", right? Is that used by database/ORM to indicate whether it's key (in DB) or not?

wing328: For Ruby on Rails, we need something similar to ensure a model property is unique, e.g.
```
validates_uniqueness_of :admin_user_id, :message => "This user has yet a competitors list"
```
Ref: http://stackoverflow.com/a/12764552/677735 

cc @zlx

zlx: Good idea! Let's add model support and validate the uniqueness validates.

mateuszmackowiak: indexPropertyName it a JSONModel framework way to support unique ids. It overrides the getter.
I'v also added it the core data model.

wing328: @mateuszmackowiak thanks for the explanation. Given that it takes a boolean value, what about naming it as "x-is-unique" instead, which means the value must be unique if set to true?

Issue #2194, opened 2016-02-21
[Swift] Add ObjectMapper to Swift


wing328:
We want to leverage [ObjectMapper](https://github.com/Hearst-DD/ObjectMapper) for easier conversion between models and JSON.

If anyone from the community wants to work on this task, please reply to let us know.

Edubits:
Would be awesome! Note there is [AlamofireObjectMapper](https://github.com/tristanhimmelman/AlamofireObjectMapper) as well, which makes using ObjectMapper in combination with Alamofire even easier.

wing328:
For those who are interested in adding the mapping for ObjectMapper, e.g.
```
    // Mappable
    func mapping(map: Map) {
        username    <- map["username"]
        age         <- map["age"]
        weight      <- map["weight"]
        array       <- map["arr"]
        dictionary  <- map["dict"]
        bestFriend  <- map["best_friend"]
        friends     <- map["friends"]
        birthday    <- (map["birthday"], DateTransform())
    }
```

Here is the logic you can borrow from PHP model template (which needs something very similar for mapping attribute names):
```
    static $attributeMap = array(
        {{#vars}}'{{name}}' => '{{baseName}}'{{#hasMore}},
        {{/hasMore}}{{/vars}}
    );
```
Ref: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/php/model.mustache#L64-L67

romk1n:
Would be great to move to this model instead of current one.

wing328:
@romk1n may I know if you would be interested in making the contribution?

romk1n:
@wing328 unfortunately don't have time to get it done, sorry :(

cbornet:
👍

PR #3072, opened 2016-06-07
Alternative swift code generator

mabedan:
I've decided to create a separate code generator for swift, given that the approach taken is very different than the existing swift code generator.

Main differentiation points:
- Handling for different response data type, per status code
- Usage of swift enums with associated values, for cleaner response callback parameters

Here are a few known issues:
- handling of NSDate in query parameters needs to be improved
- Inclusion of query parameters, in case there're body parameters as well
- Lack of unit tests
- possibility of class name collision, in case of generating multiple spec files. Could be solved by `pod`ification

Here's an example of spec file + generated swift files. [api](https://gist.github.com/mabedan/36549b524870b064beeb6785dba1f432#file-raygunapi-swift), [model](https://gist.github.com/mabedan/36549b524870b064beeb6785dba1f432#file-errordatamodel-swift), [spec](https://gist.github.com/mabedan/36549b524870b064beeb6785dba1f432#file-raygun-yaml)

### DIFF
--- a/modules/swagger-codegen/src/main/resources/swift2/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift2/api.mustache
@@ -11,8 +11,9 @@ import AlamofireObjectMapper
 
 {{#description}}
 /** {{description}} */{{/description}}
-class {{classname}} {
+public class {{classname}} {
     public static var alternativeHost:String? = nil
+    public static var useLocalHost:Bool = false
 {{#operation}}
 
     enum {{operationId}}Response {
@@ -32,7 +33,10 @@ class {{classname}} {
         {{^pathParams}}let{{/pathParams}}{{#pathParams}}{{^secondaryParam}}var{{/secondaryParam}}{{/pathParams}} path = "{{path}}"{{#pathParams}}
         path = path.stringByReplacingOccurrencesOfString("{{=<% %>=}}{<%paramName%>}<%={{ }}=%>", withString: "\({{paramName}})", options: .LiteralSearch, range: nil){{/pathParams}}
         let URLString:String
-        if let alternativeHost = {{classname}}.alternativeHost {
+
+        if {{classname}}.useLocalHost {
+            URLString = "{{termsOfService}}\(path)"
+        } else if let alternativeHost = {{classname}}.alternativeHost {
             URLString = alternativeHost + "{{basePathWithoutHost}}" + path
         } else {
             URLString = "{{basePath}}" + path


wing328:

wing328:
cc @Edubits @jaz-ah

wing328:
@mabedan if the current swift generator can achieve the following, do you agree we should just keep one Swift generator?
> Handling for different response data type, per status code
> Usage of swift enums with associated values, for cleaner response callback parameters

For `Handling for different response data type, per status code`, do you mean the response data type is different based on the status code? Can you share your Swagger spec with us?

For  `Usage of swift enums with associated values`, can you also share the spec for this and how that's represented in swift?

mabedan:
> if the current swift generator can achieve the following, do you agree we should just keep one Swift generator?

The initial reason I started rewriting the swift layer, was the crashes I was experiencing inside ```Decoder``` methods (I assume they're already fixed), and I wasn't able to read/understand what was happening in the code.
[ObjectMapper](https://github.com/Hearst-DD/ObjectMapper) library is widely used in the swift community, and I think it makes sense to make use of it, instead of reimplementing it. Another advantage is that it supports both serlialization/deserialization out of the box.

So to answer your question, for my own project, moving away from ObjectMapper would be a dealbreaker.

One more point which I forgot to mention, is that body parameters are ignored in the original swift implementation. This one supports body parameters, but still lacks form parameters.

> For Handling for different response data type, per status code, do you mean the response data type is different based on the status code? Can you share your Swagger spec with us?

By definition, ```schema```s are defined per status code. Here's a code snippet from [swagger spec](http://swagger.io/specification/)

    responses:
        '200':
          description: pet response
          schema:
            type: array
            items:
              $ref: '#/definitions/Pet'
        default:
          description: error payload
          schema:
            $ref: '#/definitions/ErrorModel'

Under responses array you can keep adding different status codes, and define a schema for each.

> For Usage of swift enums with associated values, can you also share the spec for this and how that's represented in swift?
 
    enum PostEntryResponse {
        case Res200(String, NSData)//default
        case Res202(DetailsDataModel, NSData)
        case Res400(ErrorDataModel, NSData)
        case Failure(Int?, NSData?)
    }
    
    ....
    
    Raygun.requestEntries() {response in 
        swift response {
        case .Res200(let string, _):
            print(string)
            break
        case .Res202(les detailsDataModel, _):
            self.presentDetailsViewController(detailsDataModel)
            break
        case .Res400(les errorDataModel, _):
            self.reportError(errorDataModel)
            break
        case .Failure:
            fatalError("don't do this at home")
        }
    }

The purpose is to have elegant and simple return/callback values which can support multiple types, and can be used within a switch statement.

Edubits:
Hi @mabedan, thanks for your contribution. I would prefer to see if we can use your changes to improve the default generator. Using ObjectMapper was something suggested earlier by @wing328 in https://github.com/swagger-api/swagger-codegen/issues/2194 and would be a great improvement.

mabedan:
That works for me too.
Unfortunately I wont have time to actively invest in changing the default generator given that I'm working on this one as part of my job, but would be happy to help other contributors.

Erdemus:
Many thanks, @mabedan, your generator is awesome. Current one even couldn't pass parameters in headers.

wing328:
Hi all, it's worth sharing with you that `swift3` generator has been added to the latest master.

Please pull the latest to give it a try.

yonaskolb:
In case it's useful for anyone, I took my own approach here to as well https://github.com/yonaskolb/SwagGen

wing328:
@mabedan thanks for the work. Please kindly try the Swift3 or Swift4 generator to see if it meets your requirement.

PR #3073, opened 2016-06-07
fix date format in example generator

bitcloud:
As stated on http://swagger.io/specification/#dataTypes the date-time format defined by RFC3339 requires a colon in the time-offset.

I think #2236 introduced that one.

The z-schema and the javascript swagger-tools validators even don't validate the date format without the colon. e.g. If you enable the output validation in the nodejs sample app, the request to http://localhost:8080/v2/store/order/1 failes with an `Error: Response validation failed: failed schema validation`


cbornet:
i would rather use ```org.joda.time.Datetime.now()```

cbornet:
If the swagger ObjectMapper can be configured to register the JodaModule of course....

wing328:
@bitcloud thanks for the PR.

PR #3075, opened 2016-06-07
[Feign] Support joda dates in feign client and use it in petstore sample

cbornet:


wing328:

Starting from 2.2.0, Java Feign API client will use `org.joda.time.LocalDate` (date), `org.joda.time.DateTim` (date-time) instead of `java.util.Date`

cbornet:
@wing328 Actually, java.util.Date is still the default. You have to set dateLibrary=joda to use joda.

wing328:
Ah I miss that. Thanks for pointing it out. Deleted the upgrade note.

cbornet:
My idea is to first add support for joda in all java clients then do the default date lib swap after that.

PR #3080, opened 2016-06-08
[JAVA] Generate server code using only JAX-RS 2.0 specification

arturdzm:
Create a server code using only annotations and imports from JAX-RS 2.0 specification.


cbornet:
Can we do it as a jax-rs library instead ?

arturdzm:
Did you mean to change language name?

wing328:
@arturdzm thanks for the PR. I think @cbornet meant instead of creating a new generator `jaxrs-spec`, what about reusing the existing `jaxrs` generator and use CLI option (e.g. `onlyJAXRS20Spec`) to generate JAX-RS server stub using only JAX-RS 2.0 spec?

For C# API client, we also support `UMP` (Universal Windows Platform) using the same C# client generator using similar technique (CLI option: `supportUWP`)

The benefit is easier maintenance and also new feature can be more easily added to both.

arthurdm:
I vote for jaxrs-spec as a language, rather than adding an option to the existing jaxrs language, for the following reasons:

- it will be easier for users to see that they can choose an spec-only output if it's a language, because we don't document each set of options for each language on the main wikis (ie: https://github.com/swagger-api/swagger-codegen/tree/master#java-jax-rs-jersey-v118)

- we already have jersey vs resteasy vs cxf as languages, so it feels natural to have the spec-only option as a language.  Otherwise, one could think:  "can I reuse my -l jaxrs-cxf with a spec-only option too? "

- in this PR, Artur not only cleaned up the imports, but also where the POM grabs the needed APIs from, so there are no 3rd party references that the jersey lang has...which makes it quite different for it just be an option.

- the current jersey-based language (-l jaxrs) already has jersey1 and jersey2 as options.  It will be confusing to define the combination of options, since one would expect "jersey1 + spec-only" would give jaxrs 1.1, while "jersey2 + spec-only" would give jaxrs2.0, which is not true in this case.

cbornet:
I had a look at the jax-rs generators and as of today there is indeed one generator per language with a lot of duplicated template files. So the jaxrs-spec must obviously be done as a new language for now.

But in the (near) future, I think we should remove this code duplication using java inheritance for the codegen classes and library inheritance for the templates.

As for the visibility of libraries, I agree something must be done in the documentation and in the editor or else contributors will be reluctant to use them to factorize the templates.

wing328:
@arturdzm thanks for the contribution. The PR has been merged into master.

For your upcoming PR, I would suggest creating a new branch for the change as per git best practice.

arturdzm:
@wing328 Thank you for merging. Will do new branch next time. 
From what I understand REST API app for  generator.swagger.io automatically picks up new languages.
I wonder how often it's updated with new builds from GitHub?

wing328:
@arturdzm there were discussion about updating generator.swagger.io regularly (e.g. weekly, daily) to the latest master but nothing has been finalised yet.

Are you aware that you can run your own generator server locally in your environment?

PR #3083, opened 2016-06-08
Add joda support to retrofit2 clients and use it in samples

cbornet:
also adds back the petstore tests


cbornet:
Note: the PR has gson typeAdapters for joda that can be copied to other gson based libs.
@xhh

PR #3099, opened 2016-06-09
Use joda in jersey1/jersey2 client sample

cbornet:


samples/client/petstore/java/jersey2/src/test/java/io/swagger/client/api/PetApiTest.java
```diff
@@ -1,155 +1,288 @@
...
+            assertEquals(404, e.getCode());
+        }
+    }
+
+    @Test
+    public void testUploadFile() throws Exception {
+        Pet pet = createRandomPet();
+        api.addPet(pet);
+
+        File file = new File("hello.txt");
```
wing328: I think we need to keep hello.txt as required by the test above.

But given that the test case passed, it seems to me something is not right as the test is supposed to fail due to hello.txt not present.

bin/java-petstore-jersey2.sh
```diff
@@ -26,6 +26,8 @@ fi
 
 # if you've executed sbt assembly previously it will use that instead.
 export JAVA_OPTS="${JAVA_OPTS} -XX:MaxPermSize=256M -Xmx1024M -DloggerPath=conf/log4j.properties"
-ags="$@ generate -i modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml -l java -c bin/java-petstore-jersey2.json -o samples/client/petstore/java/jersey2"
+ags="$@ generate -i modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml -l java -c bin/java-petstore-jersey2.json -o samples/client/petstore/java/jersey2 -DdateLibrary=joda,hideGenerationTimestamp=true"
 
+rm -rf samples/client/petstore/java/jersey2/src/main
```
wing328: I would suggest adding a line to echo a message, e.g. "Removing files and directories under samples/client/petstore/java/jersey2/src/main" so that the user/developer is aware that the script will purge the existing Java files.

samples/client/petstore/java/jersey2/src/test/java/io/swagger/client/api/PetApiTest.java
wing328: Please ignore. I read the code and understand what you did.

bin/java-petstore-jersey2.sh
wing328: FYI. I've added the message

PR #3107, opened 2016-06-10
Set joda as default dateLibrary for java codegens that support it

### LINKED ISSUE #3087: Swagger spec date format not supported by Java client
##### Description

The generated Java client does not correctly parse the Swagger date format from API responses.

According to the[ swagger spec](http://swagger.io/specification/) a response field of type string, format date must return a full-date defined by [RFC 3339](http://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14) which specifies the format as 

>    full-date       = date-fullyear "-" date-month "-" date-mday

This is not supported by the default Java client, which generates using a single date format, used for both date and date-time which is

>     DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSXXX");

This means simple date fields are not supported and cause the client to throw an InvalidFormatException during the parsing of a response model which contains Date fields..

##### Swagger-codegen version

2.1.6

Source swagger file is https://github.com/amadeus-travel-innovation-sandbox/sandbox-content/blob/master/swagger.yml

java -jar modules\swagger-codegen-cli\target\swagger-codegen-cli.jar generate -i https://raw.githubusercontent.com/amadeus-travel-innovation-sandbox/sandbox-content/master/swagger.yml -l java -o /java-client

##### Steps to reproduce
1. Build java client using above command
2. Register on sandbox.amadeus.com and create a new app to get an API key
2. Create a new unit test using the java client to invoke our hotel API to demonstrate the bug, like so:
```

	private static final String APIKEY = "YOUR_API_KEY_HERE";
	private static final DefaultApi API = new DefaultApi();
	private static final DateTimeFormatter DATE = ISODateTimeFormat.date();

	@Test
	public void testHotelAirportSearch() throws ApiException {
		LocalDate checkIn = new LocalDate().plusWeeks(6);
		LocalDate checkOut = checkIn.plusDays(1);
		HotelSearchResponse hotelSearchRS = API.hotelAirportSearch(APIKEY, "BOS", DATE.print(checkIn), DATE.print(checkOut), 25, "EN", "USD", null, null, 3, false, false, null);
		assertEquals("US", hotelSearchRS.getResults().get(0).getAddress().getCountry()); 
		assertEquals(checkIn, hotelSearchRS.getResults().get(0).getRooms().get(0).getRates().get(0).getStartDate());
	}
```

##### Related issues
No noted related issues

##### Suggest a Fix
There are two good approaches to this.
Either define a date and a dateTime formatter, and use them appropriately, or define a looser date parser for all API responses that can handle any valid ISO Date/Time

cbornet:
Fix #3087


wing328:
**Upgrade Note from 2.1.6 to 2.2.0**

Starting from v2.2.0, Java API client will use [Joda](http://www.joda.org/joda-time/) instead of `java.util.Date`

To fall back, please provide the option `-DdateLibrary=legacy` when generating the API client through command line.

cbornet:
Also applies to spring-boot

Issue #2849, opened 2016-05-12
java/jersey2 should expose builder or setter for ApiClient.httpClient


jerrylawson:
The httpClient value in https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache#L58 is private without a setter method. This does not allow for an application to provide its own Client instance to be used.

It would be good to have ApiClient either provide a protected or public setHttpClient(Client) method, or provide a protected getHttpClient() and use that within ApiClient, rather than accessing httpClient directly.

wing328:
@jerrylawson thanks for the suggestion. May I know if you've cycle to contribute the enhancement?

Ref: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

jerrylawson:
Yes, I can contribute this change

wing328:
@jerrylawson  PR merged. Please pull the latest to give it a try.

PR #3108, opened 2016-06-10
Adding the ability to get and set the httpClient for jersey2

JasonQuinn:
Fix for #2849

### DIFF
--- a/samples/client/petstore/java/jersey2/src/main/java/io/swagger/client/ApiClient.java
+++ b/samples/client/petstore/java/jersey2/src/main/java/io/swagger/client/ApiClient.java
@@ -102,6 +102,15 @@ public class ApiClient {
     return json;
   }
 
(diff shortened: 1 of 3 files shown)


wing328:

wing328:
@JasonQuinn if you don't want to fix it, we'll merge this PR on coming Wed.

samples/client/petstore/java/jersey2/src/main/java/io/swagger/client/ApiClient.java
```diff
@@ -102,6 +102,15 @@ public JSON getJSON() {
     return json;
   }
 
+  public Client getHttpClient() {
```
jerrylawson: Jason, I originally wrote the issue #2849 . It would be useful in your PR to change the buildHttpClient() method to protected instead of private, allowing a subclass to override it.

PR #3113, opened 2016-06-10
it's the future

taf2:
html5 is all the rage


ePaul:
Thanks for the pull request, but could you please write some more sentences about why this is done? (You can simply edit the initial comment of the pull request.)

wing328:
@taf2  I agree with @epaul that a little bit more info would be good.

Also, please run `./bin/html-petstore.sh` to update the Petstore API documentation (as mentioned in the contributing guidelines)

PR #3117, opened 2016-06-12
Unserialuse a Timestamp date passed from Server to PHP client

corleonis:
In some cases the server can return a date in timestamp format instead of ISO8601.

That is a valid date format that can be either 10 or 14 digits respectfully for
seconds or milliseconds time stamp.

Currently the code throws a runtime error as we try to instanciate a DateTime object
that can't make use of a timestamp format.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/php/ObjectSerializer.mustache
+++ b/modules/swagger-codegen/src/main/resources/php/ObjectSerializer.mustache
@@ -239,6 +239,13 @@ class ObjectSerializer
             // be interpreted as a missing field/value. Let's handle
             // this graceful.
             if (!empty($data)) {
+                // In some cases the server can send us a valid data
+                // represented as a timestamp either in seconds, 10 digits,
+                // or in milliseconds, 14 digits. We need to transform it in
+                // in valid format before passing it back as DateTime object.
+                if (1 === preg_match('/\d{10,14}/', $data)) {
+                    $data = date(DATE_ISO8601, $data);
+                }
                 return new \DateTime($data);
             } else {
                 return null;


cbornet:
Although it cannot harm to accept deserialization of timestamps, note that this is not in the [swagger spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types). The "date-time" format in the spec is [rfc3339](http://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14). So if the server outputs timestamps for date, it should declare those fields as "int" in its swagger def. Note that you can also specialize with a [custom format](http://json-schema.org/latest/json-schema-validation.html#anchor106) such as ```"type": "integer", "format":"timestamp"``` ("Implementations MAY add custom format attributes. Save for agreement between parties, schema authors SHALL NOT expect a peer implementation to support this keyword and/or custom format attributes.")

wing328:
cc @arnested

arnested:
I tend to agree with @cbornet.

We _could_ add support for deserializing timestamp although it would not compliant with the Swagger specification. On the other hand there are countless ways to get a `date-time` almost right / slightly wrong so this would end up making our code quite complex and obscure.

@corleonis can you elaborate a bit on your use case? Would it be a solution for you to use an `int` as suggested by @cbornet?

Or are you trying to integrate with a producer providing the specification (as `date-time`) and a faulty implementation (actually returning a timestamp) of it?

For the latter case (which could be quite common because a lot of faulty implementation exists) we could maybe add some way to implement a "data cleaner function" to hook into the deserializer.

But please provide some more details on the use case.

corleonis:
The problem we're trying to solve is that we have a Java application backed with Swagger and Spring Boot (Jackson). The Jackson serialiser by default converts all dates to timestamps and we have a fair bit of data in ES, the application has been live live for a while. 

So the DTO that is used to store the data and retrieved the data is both used for the front facing API as a response and as a data object when storing to ES thus we can't easily change the type to `int` and actually makes good sense to be `Date`. 

Now we want to integrate that service into a PHP application for which we used the Swagger Codegen library. The date is correctly mapped by the annotations and the matching DTO on PHP looks all OK but there is mismatching in the data types, e.g. we get timestamp instead of ISO date and the PHP implementation of the serialiser doesn't like/support it though to me it looks like a valid example, even though ISO date is better than timestamp, e.g. could carry more data. 

I really hope this change could be useful for other people too, though agree that our situation is a bit of an edge case.

wing328:
To address the issue, what about adding a new property/field in PHP configuration class: `DateTimeFormat` - which allows developers to specify the datetime format (e.g. `Y-m-d\TH:i:sO` (ISO8601)?

It should default to `Y-m-d\TH:i:sP` (RFC3339) and the ObjectSerializer will use it for serialisation/deserialization of datetime property.

Ref: http://php.net/manual/en/class.datetime.php

(understood that the spec only supports RFC3339 for datetime but there are many Restful backend out there not using RFC3339 for datetime at the moment, the suggestion above aims to add flexibility to handle datetime in other formats)

corleonis:
Sounds like a good solution, would allow us to decide how to handle dates. The only problem I could see if people have mixture of let's say `RFC3339` and another format could be timestamp for example this would have same problem if it's a strict rule. Thinking about it adding a list of valid date format could be the solution though I feel I am overcomplicating it :)

wing328:
For mixture of datetime format within the same JSON payload, we can ask the developers to document the datetime attribute as string instead (as a workaround).

Having a mixture of datetime format in the same payload is not a good design in my opinion.

corleonis:
Yes completely agree!

ePaul:
I guess the original problem was the wrong implementation on the server side. I think we've had a different ticket for that? (I can't find it right now.) Fixing this on the client side seems a bit wrong.

wing328:
@ePaul yes, that's one way to look at the problem. I think there may be restful backends in use that were created before swagger spec was created/invented. Of course we can ask them to change their backend to fit into swagger spec. My view is that providing a configuration option would not be a bad thing to ease the adoption.

cbornet:
About Spring boot with Jackson, it supports rfc3339 serialization of dates. It's just a config of Jackson and our generator does it for example. If it's not possible to change the config due to backward compatibility issues, then the swagger spec must be changed so that Date is mapped to integer. If you are using springfox, that should be possible by springfox config.

cbornet:
Also note that it is incorrect to output an int timestamp and have the swagger declare it as "string" even with a date-time format. It will most probably cause issues and not only with the codegen but with other swagger tools (eg. Swagger-ui)

arnested:
I have been trying to think of ways to solve this problem in a general fashion for problems where a server produces data in formats contrary to what the specification says.

One way to that (and it's possible to do today with 2.1.6) is to implement your own ApiClient class to catch and manipulate the reponse before it reaches the object serializer:
```php
<?php

namespace MyNamespace;
use Swagger\Client\ApiClient;

class MyApiClient extends ApiClient {
  public function callApi($resourcePath, $method, $queryParams, $postData, $headerParams, $responseType = NULL) {
    $response = parent::callApi($resourcePath, $method, $queryParams, $postData, $headerParams, $responseType);
    // Manipulate $response here to turn timestamps into RFC3339 date strings.
    return $response;
  }
}
```
And then use your own ApiClient:
```php
$config = new \Swagger\Client\Configuration();
$api_client = new \MyNamespace\MyApiClient($config);
$api = new \Swagger\Client\Api\DefaultApi($api_client);
```

arnested:
If implementing ones own version of `callApi` it would be useful to know the path of the endpoint (without path parameters expanded - i.e. `/pet/{petId}`).

Pull request #3256 adds that information to the `callApi` method.

PR #3122, opened 2016-06-13
Add Apache v2 license to auto-generated scalatra files.

clasnake:
The formatting in the sample model breaks after regenerating according to the current template. I think we may need another pr to fix the formatting.


wing328:
Yes, we can fix the code format/style later with another PR.

wing328:
For https://github.com/swagger-api/swagger-codegen/issues/2963

PR #3124, opened 2016-06-13
[Objc] - CoreData support for ArrayModels

mateuszmackowiak:
- Also added support for uuid and password types


modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenModel.java
```diff
@@ -35,7 +36,7 @@
     public Set<String> allMandatory;
 
     public Set<String> imports = new TreeSet<String>();
-    public Boolean hasVars, emptyVars, hasMoreModels, hasEnums, isEnum, hasRequired;
+    public Boolean hasVars, emptyVars, hasMoreModels, hasEnums, isEnum, hasRequired, isArrayModel;
```
wing328: In the next version (stable release after 2.2.0), we may want to revise the mustache tag e..g

`{{#isArrayModel}}` will become `{{#isListContainer}}{{#isModel}}`

PR #3125, opened 2016-06-13
Added missing c# primitives and mappings for non-existent primitive wrappers

carlin-q-scott:
Someone made the assumption that a primitive declaration of "int?" would match "int" as well; It doesn't. Also, there are no box classes for int and float so I mapped those to nullable int and float, respectively. There were also some primitives that just weren't defined as primitives but they were defined as reserved words.


wing328:

wing328:
Please also run `./bin/csharp-petstore.sh` to update the C# Petstore sample and run the tests locally.

carlin-q-scott:
Oh huh, I assumed it would link it up to my account using my key. I wonder how of my commits are like this :(.

I'll make the recommended changes and open another PR.

carlin-q-scott:
The build failure seems like an intermittent failure unrelated to the PR since the JDK8 build passed and the failure is for TypeScript, not C#.

jimschubert:
I think this PR is required for the C# codebase, however I think there's an issue with the concept of a "primitive" here. Nullable types aren't primitives (`typeof(int?).IsPrimitive // returns false`). Neither are streams, byte arrays, object, DateTimeOffset, etc.

In [DefaultCodegen](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java), everything in this hash set will cause the property to be marked primitive on the `CodegenProperty`. While I believe it's probably required to avoid these types being added to imports, I think `DeafultCodegen` makes some assumptions about code that become more and more incorrect about C# and .NET in general.

carlin-q-scott:
Yeah, there are a bunch of things wrong with how it's currently implemented in regards to primitives and keywords but I don't have the time to fix all of that. The code currently uses nullable primitives in place of real primitives so that they can have default values when passed into functions which is bad but to avoid that it would have to use default(primitive type) which has its own problems.

PR #3126, opened 2016-06-13
[PHP] Handle empty array response correct

arnested:
An empty array can be a perfectly valid response and should be returned as such. Unfortunately the response is converted to `null` in `api.mustache`.

The problem is the use of this code in `api.mustache`:
```php
if (!$response) {
    return array(null, $statusCode, $httpHeader);
}
```

If `$response` is an empty array we will enter this and return `null`. But this will also produce false response for other values as well, i.e. an endpoint returning `false`.

If we look into the possible values of `$response` there is actually no use for this condition.

A test case for testing correct handling of empty arrays has been added (would fail on existing code base, succeeds with this change). [Existing testcases all succeeds as well](https://circleci.com/gh/arnested/swagger-codegen/85).


wing328:
Thanks for the PR. We need similar test for other clients as well to ensure they can handle empty array response.

Issue #3129, opened 2016-06-14
[typescript-angular2] Handle 204 response without json conversion


taxpon:
##### Description
- Currently angular2 clients parse received response as json in any time
- According to [HTTP Status Code Definition](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html), 204 response **MUST NOT** include a message-body, so when I use generated clients for 204 response, I get an error `SyntaxError: Unexpected end of JSON input`

##### Swagger-codegen version
2.2.0-SNAPSHOT

##### Suggest a Fix
Update following part in `typescript/api.mustache` so that it can handle response with status code 204 correctly. I'll create a PR. :
```typescript
        return this.http.request(path, requestOptions)
            .map((response: Response) => response.json());
```

taxpon:
Merged #3130

PR #3130, opened 2016-06-14
Handle 204 response correctly in typescirpt-angualr2

taxpon:
PR for Issue #3129


wing328:
@taxpon thanks for the PR

cc @Vrolijkx

wing328:
@taxpon the change looks good to me. Later we might need to revise the code to support status code defined in the spec.

wing328:
and thanks for fixing the typo as well. Good catch!

PR #3136, opened 2016-06-14
Added support for the latex language

rodrigoabella:
- added latex language


wing328:
@rodrigoabella thanks for the PR to add latex documentation. Do you need any help adding the latex doc generator?

Issue #3138, opened 2016-06-14
c# aspnet 5 server codegen - would like classes to have virtual method for inheritance


stunney:
##### Description

Would like to have public methods be marked as virtual so that they can be sub-classed.

##### Swagger-codegen version

Latest as of 2016-06-14 @ 16:40 EST

##### Command line used for generation

C# aspnet server codegen

##### Suggest a Fix

Update the aspnet moustache file for the class to add " virtual" after "public" for the method signature.

Will be creating a PR shortly for this issue.

stunney:
Changes merged.  Closing issue.

PR #3139, opened 2016-06-14
Issue #3138

stunney:
https://github.com/swagger-api/swagger-codegen/issues/3138

Adds " virtual" to the controller moustache file so that the controllers
can be inheritted from.


stunney:
@wing328 would it be possible to get this PR in sooner rather than later? :)

wing328:
@stunney please run `./bin/aspnet5-petstore-server.sh` (mac/linux) or `bin/windows/aspnet5-petstore-server.bat` (windows) to update the Petstore sample

cc @jimschubert

jimschubert:
I don't see a problem with this change, but also can't think of a use case for this inheritance. Can you provide an example?

stunney:
Well the HttpPostAttribute and RouteAttribute have the Inherited value of true, so any subclass that overrides the methods will get the decorators implicitly.

What we are aiming to do is generate code from the main which will generate a versions NuGet package.  That package will be included in our actually impl service to separate the code gen cleanly.

TBH generating a pure abstract class would be "better" but this is the least intrusive and non breaking change I could think of.

stunney:
I've noticed that the Swashbuckle.Swagger.Annotations.SwaggerOperationAttribute is not marked with the inheritted flag.  This can easily be fixed though.

stunney:
@wing328, @jimschubert 
I have executed bin/windows/aspnet5-petstore-server.bat (windows).  We shall see what Travis-ci has to say about it :)

stunney:
And actually, with regard to the SwaggerOperationAttribute (and the others), the default value for inherited is true, provided one actually marks an attribute class with the AttributeUsageAttribute attribute :)

stunney:
@wing328, @jimschubert,
Any change I could get this PR accepted sooner than later?

I have more to make and wanted this one off my plate, pretty please with sugar on top :)

jimschubert:
@stunney

I don't have merge rights, I only do reviews. I think @wing328 was probably waiting for a response from you, since your third previous comment sounded like you had some travis-ci validations of your own running.

stunney:
Nope, all is done on my end.  @wing328 , can you help a coder out?

Thanks for the feedback, @jimschubert, I will attempt to be less ambiguous in the future.

wing328:
@stunney I've merged the PR. Thanks for the contribution.

stunney:
Thanks @wing328 !! Glad I could contribute.

Issue #3006, opened 2016-05-31
[Scala][Java] Add "sbt" build config to Java/Scala SDK


wing328:
Currently, Java supports maven and gradle to build the library. We want to add "sbt" as well: http://www.scala-sbt.org/

This will make it easier for Scala developers to build the Scala/Java SDK to use it in their Scala program.

If anyone wants to work on this, please reply to let us know.

jimschubert:
I think this duplicates #241.

I can take a look at creating the build file.

wing328:
#241 is for scala while for this one, I want to cover Java API client as well ( so that it can be more easily used in scala with build.sbt)

I think @clasnake may have done it already.

wing328:
Just did a check. Scala Petstore client still misses the bulid.sbt file. If you've cycle, please help create the build file for Scala API client.

clasnake:
@wing328 @jimschubert I've added sbt for feign, retrofit2, retrofit2rx and jersey2 for the java client. The scala client and default client for java haven't been done yet. Please help create that sbt if you've cycle.

jimschubert:
@wing328 @clasnake I've added sbt to the scala client. I looked at some of the other implementations and they all use `build.sbt` which I've had issues with in the past. I didn't know if sbt vs scala build files should be an option, but I always use the .scala files. Also, others drop sbt/gradle/pom to the output directory and my PR does not do this. I didn't know if this was a standard or what.

wing328:
@jimschubert for the issue you encountered in the past, is there a SO link (or other URls) with more details that I can look into?

jimschubert:
I've just looked through my bookmarks and don't see anything relevant. It may be that I've used sbt for multiproject builds with lots of customizations and plugins in the past (3 yrs ago). I've only recently picked sbt back up in the past few months.

Do you think build.sbt would be better and only offer the other format of requested by users?

wing328:
I think we can go with build.sbt first for day 1 requirement and if users complain, we can then look into the issue.

Given that the issue you encountered was a few years ago, I think the issue may have been addressed already.

Given that your PR has generated ProjectBuild.scala already, we can go with both (build.sbt and ProjectBuild.scala and include the instruction in the auto-generated README)

jimschubert:
[...]

Issue #241, opened 2014-08-08
Use build.sbt for generated Scala files, instead of pom.xml


ngocdaothanh:
Currently, the generated Scala files are used with a pom.xml file. It should be build.sbt by default.

wing328:
@ngocdaothanh agree that build.sbt is preferable to pom.xml for scala.

Do you mind coming up with a sample build.sbt for https://github.com/swagger-api/swagger-codegen/tree/master/samples/client/petstore/scala ? Then we'll use it to come up with a build.mustache template.

Thanks in advance.

wing328:
Closed via #3149

PR #3149, opened 2016-06-16
[scala] Add SBT buid support for scala client

jimschubert:
See #241 and #3006

Adds build.sbt to scala client. Removed previous option-based functionality.

~~This adds a configuration option to support sbt (`-DbuildTool=sbt`) or gradle (`-DbuildTool=gradle`) to the scala client.~~

~~I've made the option value generic enough not to cause clutter. Unlike some of the java implementations like `JavaClientCodegen`, this will generate files for the selected build tool and not all files for both.~~

~~Also, unlike other sbt support which creates `build.sbt` this creates a `ProjectBuild.scala`. It's my understanding that some plugins don't work correctly when scalaVersion with binary version 2.11 is defined in `build.sbt` and a plugin requires 2.10. If this is incorrect, let me know and I'll change to build.sbt. However, both the .sbt and .scala build files can live side by side.~~

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ScalaClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ScalaClientCodegen.java
@@ -87,6 +87,8 @@ public class ScalaClientCodegen extends DefaultCodegen implements CodegenConfig 
         supportingFiles.add(new SupportingFile( "gradle-wrapper.jar",
                 gradleWrapperPackage.replace( ".", File.separator ), "gradle-wrapper.jar") );
 
+        supportingFiles.add(new SupportingFile("build.sbt.mustache", "", "build.sbt"));
+
         importMapping.remove("List");
         importMapping.remove("Set");
         importMapping.remove("Map");
--- a/modules/swagger-codegen/src/main/resources/scala/build.sbt.mustache
+++ b/modules/swagger-codegen/src/main/resources/scala/build.sbt.mustache
@@ -1,0 +1,33 @@
+lazy val root = (project in file(".")).
+    settings(
+        version       := "{{artifactVersion}}",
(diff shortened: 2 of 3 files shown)


cbornet:
Why not generate the files for all the build tools (without option) and then the user can choose which tool to use on the generated project. That's how it's done in java clients.

jimschubert:
Because an option is a better user experience than expecting the user to
delete 10-12 other files they don't need. Since I've added support for the
`.swagger-codegen-ignore`, users could regenerate and not need to worry
about deleting additional files again... but that's only relevant after an
initial cleanup.

cbornet:
Yes I didn't read your PR description well enough. That's why I deleted my comment.

jimschubert:
Sorry, I responded by email. I guess that's one downside to the email reply feature.

wing328:
About whether using an option or not, my view is a bit different. I thought it would be a good idea to generate pom.xml, build.gradle and sbt.build as part of the Java API client/server stub generation because it should be up to the developers (consumers of the API client/server stub) to decide how they want to build the source code.

jimschubert:
@wing328 that approach assumes that all build related files have unique filenames, which in this case they do. I can remove the option, but as a user I would find the lack of an option annoying.

I've had too many situations where a developer comes to me and says "I've tried to update a dependency but it doesn't seem to work". This would leave more than one place to update dependencies, and people in larger teams could unknowingly be maintaining code according to different build tools.

jimschubert:
I've force-pushed an update which only uses build.sbt, and doesn't include an option.

wing328:
> I've had too many situations where a developer comes to me and says "I've tried to update a dependency but it doesn't seem to work"

Definitely valid point. I hope swagger codegen can address this particular problem as developers only need to report to us to update the dependencies instead of having the developers (consumers of the SDKs) to update it locally (or the owner updates it themselves).

PR #3151, opened 2016-06-16
[Objc] Moved default petstore demo sample to default folder and created a new with core data sample

mateuszmackowiak:


modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java
```diff
@@ -82,12 +82,14 @@ public Generator opts(ClientOptInput opts) {
         }
         if(System.getProperty("modelDocs") != null) {
             generateModelDocumentation = Boolean.valueOf(System.getProperty("modelDocs"));
+            LOGGER.error("AAAAAAAAAAAAAAAA generateModelDocumentation "+generateModelDocumentation);
```
cbornet: This should probably not be committed...

modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java
```diff
@@ -82,12 +82,14 @@ public Generator opts(ClientOptInput opts) {
...
         }
         if(System.getProperty("apiTests") != null) {
             generateApiTests = Boolean.valueOf(System.getProperty("apiTests"));
         }
         if(System.getProperty("apiDocs") != null) {
             generateApiDocumentation = Boolean.valueOf(System.getProperty("apiDocs"));
+            LOGGER.error("AAAAAAAAAAAAAAAA generateApiDocumentation "+generateApiDocumentation);
```
cbornet: This should probably not be committed...

PR #3152, opened 2016-06-16
Fix Java CXF on parameter names containing '_' or '-'

salemXpresso:
[java-cxf] 	Fix Java CXF on parameter names containing '_' or '-'. The template uses only the parameter name which result on misnaming parameters.


wing328:

wing328:
@salemXpresso I believe your fix is for XML only. Can you also fix the issue for JSON as well?

salemXpresso:
It's not a problem if I am not counted among contributors. Thanks anyway for mentioning it.

The annotations @XmlElement is not dedicated to XML, it's also used for JSON as well. For my part, I am impacted with JSON elements and this annotation solves the problem. Would you like me to send you an example ?

wing328:
@salemXpresso thanks for the details. Please also run `./bin/jaxrs-cxf-petstore-server.sh` to keep the Petstore sample up-to-date.

salemXpresso:
@wing328 The petstore is now up to date.

PR #3156, opened 2016-06-16
Fix result not defined error in javascript ApiClient

### LINKED ISSUE #3155: [JAVASCRIPT] Uncaught ReferenceError in constructFromObject
##### Description

I'm running into an error in `constructFromObject`:

```
Uncaught ReferenceError: result is not defined
```

##### Swagger-codegen version

2.1.6

I can try to supply a simplified declaration if needed. I think the issue is just a typo, so I don't really want to spend the time testing this further.

```
java -jar bin/swagger-codegen-cli.jar generate -l javascript -i swagger.yaml -c conf/javascript.json
```

The conf that is pointed to has an empty json object.

##### Related issues

I couldn't find any...it seems like this is caused by a typo.

##### Suggest a Fix

The fix is to rename `result[k]` to `obj[k]`. The docs say that `obj` is the target. `result` is not declared anywhere in the function, so I'm assuming it was just a typo.

ferristseng:
I was getting a `Uncaught ReferenceError: result is not defined` in `constructFromObject` while using the Javascript client.

It looks like `result` was never being initialized. I think `obj` should be modified like in the block above.

Closes #3155


wing328:

ferristseng:
Thanks for letting me know. I think it should be fixed now.

alexstanbury:
+1 this, had me debugging for hours, plus the fact that someone used the 'length' property in the api which also caused mayhem when setting it on the obj... 👍

wing328:
@ferristseng yes, that's fixed as shown in the commit tab.

Can you please also run the following to update the Petstore sample so that the change will be verified by the CI (travis)?
```
./bin/javascript-petstore.sh
./bin/javascript-promise-petstore.sh
```

wing328:
@alexstanbury thanks for reviewing the change.

alexstanbury:
@wing328 will this change propagate to the swagger editor immediately? I am using that to generate my client. Thanks.

wing328:
@alexstanbury I don't think so. There were some discussion about setting a weekly/daily to deploy the latest master of Swagger Codegen to generator.swagger.io (which is used by editor.swagger.io to generate API clients or server stubs) but from what I know generator.swagger.io is still using stable release v2.1.6.

To use the latest master to generate API clients or server stubs, https://github.com/swagger-api/swagger-codegen/#getting-started is a good starting point.

ferristseng:
Ok, updated the Petstore sample, and committed it.

wing328:
@ferristseng thanks for the PR, which has been merged into the master.

@alexstanbury please pull the latest master and let us know if you still see the same issue.

Issue #3096, opened 2016-06-09
Angular 1: Required parameter check causes unexpected behaviour


moanrose:
The required parameter check in the api.mustache causes unexpected behaviour if the actual parameter value evaluates to false e.g "false", 0, -1, false

swagger-codegen/modules/swagger-codegen/src/main/resources/typescript-angular/api.mustache

The code

if (!{{paramName}}) {
	throw new Error('Missing required parameter {{paramName}} when calling {{nickname}}');
}

Needs to be changed to:

if ({{paramName}} === undefined || {{paramName}} === null) {
	throw new Error('Missing required parameter {{paramName}} when calling {{nickname}}');
}

I can make a pull request myself, but i do not have the necessary permissions to create a branch - should i make the change on a separate fork?

wing328:
@moanrose thanks for the suggestion.

Yes, agree that we should reuse the same check in `typescript-node` to validate required parameter: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/typescript-node/api.mustache#L200-L205

May I know if you've cycle to contribute a fix?

Ref: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

wing328:
`typescript-angular2` needs similar fix: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache#L44-L51

wing328:
@moanrose thanks for the PR, which improves both typescript-angular1 and typescript-angular2

PR #3158, opened 2016-06-17
Changed check for required parameter to check for null or undefined

moanrose:
The required parameter check in the api.mustache caused unexpected behaviour if the actual parameter value evaluates to false e.g "false", 0, -1, false

The code was changed to check for undefined or null instead

See issue #3096


wing328:

wing328:
@moanrose it's fine that you don' want to fix the commit's author. I'll merge this on coming Tue.

PR #3163, opened 2016-06-17
Improve error message for connection failures

DanGe42:
Previous ApiException message would simply print out the result of the
`curl_getinfo($curl)` call, which might be useful only if the developer
actually wanted very low-level information from curl about why a call
failed. The new message should print out a higher-level but more
informative, human-readable message.


wing328:
@DanGe42 thanks for the PR. Please also run `./bin/php-petstore.sh` to update the Pestore sample (which will be tested by travis-ci).

cc @arnested

wing328:
@DanGe42 can you keep `serialize($response_info)` in the error message as the HTTP body may sometime contains useful information such as error message or special error code, e.g.
```
{
    "error": {
        "status": 400,
        "message": "invalid id"
    }
}
```

arnested:
Thank you for a good and nice improvement, @DanGe42!

Keeping the data from `curl_get_info()` sure would be nice but serializing it into the error message doesn't really improve readability I think.

How about [`setResponseObject($reponse_info)`](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/php/ApiException.mustache#L99) on the `ApiException` object?

modules/swagger-codegen/src/main/resources/php/ApiClient.mustache
```diff
@@ -223,7 +223,16 @@ class ApiClient
 
         // Handle the response
         if ($response_info['http_code'] == 0) {
-            throw new ApiException("API call to $url timed out: ".serialize($response_info), 0, null, null);
+            $curl_error_message = curl_error($curl);
+
+            // curl_exec can sometimes fail but still return a blank message from curl_error().
+            if (!empty($curl_error_message)) {
+                $error_message = "API call to $url failed: ".curl_error($curl);
```
arnested: Using `$curl_error_message` instead of another call to `curl_error($curl)` would improve readability slightly I think.

DanGe42:
Thanks for looking at my pull request!

@wing328 @arnested: The context for this change was that a developer using our PHP SDK reported that he kept getting this timeout exception before disabling SSL verification (eek!) to resolve the issue. I'm fairly confident that the `curl_error` will surface this error, thus why I made this change.

In this case, this branch is only taken when `$response_info['http_code'] == 0`. From what I remember, and from a cursory google search (which brought up this [StackOverflow link](http://stackoverflow.com/questions/10227879/php-curl-http-code-return-0)), it looks like the result of `curl_getinfo` doesn't really contain any useful information if the `'http_code'` field is 0 because this only occurs if there was a problem making the request (and thus no HTTP body comes back).

I can still add a `setResponseObject($response_info)` if either of you think it's worth doing so.

modules/swagger-codegen/src/main/resources/php/ApiClient.mustache
DanGe42: That was an oversight. Fixed.

DanGe42:
Hmm the CI build failed. My gut instinct tells me this looks transient, but I don't know how to confirm:

```
Concurrency config is parallel='methods', perCoreThreadCount=true, threadCount=2, useUnlimitedThreads=false
Running io.swagger.client.api.PetApiTest
Tests run: 9, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.209 sec <<< FAILURE!
(snip)
Results :
Tests in error: 
  testFindPetsByStatus(io.swagger.client.api.PetApiTest): {"code":500,"type":"unknown","message":"something bad happened"}
Tests run: 36, Failures: 0, Errors: 1, Skipped: 1
```

arnested:
Thank you, @DanGe42 (and sorry for not noticing your comments until now).

I think this looks good and should be merged.

What do you think, @wing328?

DanGe42:
FYI, I noticed my branch was conflicting, so I squashed and rebased on top of master.

Rerunning bin/php-petstore.sh changed a lot of the sample files, so I ran it first without my changes and committed those into a separate commit. The test failures from CI are coming from the Java (??) client. I tried finding running the PHP tests separately (which I guess I can do with `php samples/client/petstore/php/test.php`), but it just exits with `Caught exception: [404] Error connecting to the API (http://petstore.swagger.io/v2/pet/10005)`.

wing328:
@DanGe42 don't worry about the CI test failure. We'll restart it if the job fails and let you know if the job consistently fails.

arnested:
@DanGe42, I usually navigate to the folder `samples/client/petstore/php/SwaggerClient-php` and run `mvn integration-test` for running the test cases.

arnested:
@DanGe42, could you please rebase this against current master and regenerate the samples (there is now both a `bin/php-petstore.sh` and a `bin/security/php-petstore.sh`.

We are trying to get the ready for an upcoming release soon.

DanGe42:
@arnested done

modules/swagger-codegen/src/main/resources/php/ApiClient.mustache
```diff
@@ -224,7 +224,19 @@ class ApiClient
...
+            // curl_exec can sometimes fail but still return a blank message from curl_error().
+            if (!empty($curl_error_message)) {
+                $error_message = "API call to $url failed: $curl_error_message";
+            } else {
+                $error_message = "API call to $url failed, but for an unknown reason. " .
+                    "This could happen if you are disconnected from the network.";
+            }
+
+            $exception = new ApiException($error_message, 0, null, null);
+            $exception->setResponseObject($response_info);
```
wing328: One suggestion for enhancement in the future. We may want to add a constructor to initalize the Response Object as well, e.g.
```
            $exception = new ApiException($error_message, 0, null, null, $response_info);
```
so that the above 2 lines can be done into one without declaring a temporary variable (`$exception`)

DanGe42:
Awesome. Thanks for merging!

wing328:
@DanGe42 thanks for your contribution!

arnested:
Yeah. Awesome! Thank you for improving Swagger Codegen, @DanGe42!

PR #3165, opened 2016-06-18
[PHP Lumen] sort the endpoints in ascending to avoid the route priority issure

### LINKED ISSUE #2643: Does the generator support to generate API endpoints with match priority?
I am try to create a lumen generator, but apis generated from the mustache file may not work well because the match priority question.

this is my [route.mustache](https://github.com/abcsun/swagger-codegen/blob/lumen/output/LumenServer/src/main/resources/LumenServerCodegen/app/Http/routes.mustache) file. 

and these routes was produced from the [petstore](http://petstore.swagger.io/v2/swagger.json)
```
/**
 * GET getUserByName
 * Summary: Get user by user name
 * Notes: 
 * Output-Formats: [application/xml, application/json]
 */
$app->GET('/user/{username}', function($username, $null = null) use ($app) {
     
    
    

    return response('How about implementing getUserByName as a GET method ?');
    });


/**
 * GET loginUser
 * Summary: Logs user into the system
 * Notes: 
 * Output-Formats: [application/xml, application/json]
 */
$app->GET('/user/login', function($null = null) use ($app) {
     
    $username = Request::input('username');    
    $password = Request::input('password');    
    
    

    return response('How about implementing loginUser as a GET method ?');
    });
```

the question is /user/login was shadowed by previously defined /user/{username} api.
 
Any suggests? Thanks.

abcsun:
fix #2643

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/LumenServerCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/LumenServerCodegen.java
@@ -88,12 +88,12 @@ public class LumenServerCodegen extends DefaultCodegen implements CodegenConfig 
         /**
          * Api Package.  Optional, if needed, this can be used in templates
          */
-        apiPackage = "io.swagger.client.api";
+        apiPackage = "app.Http.Controllers";
 
         /**
          * Model Package.  Optional, if needed, this can be used in templates
          */
-        modelPackage = "io.swagger.client.model";
+        modelPackage = "models";
 
         /**
          * Reserved words.  Override this with reserved words specific to your language
@@ -154,7 +154,7 @@ public class LumenServerCodegen extends DefaultCodegen implements CodegenConfig 
      * instantiated
      */
     public String modelFileFolder() {
-        return outputFolder + "/" + sourceFolder + "/" + modelPackage().replace('.', File.separatorChar);
+        return outputFolder + "/" + modelPackage().replace('.', File.separatorChar);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/LumenServerCodegen.java
```diff
@@ -171,9 +164,27 @@ public String modelFileFolder() {
     @Override
     public String apiFileFolder() {
         return outputFolder + "/app/Http/controllers";
```
wing328: Please replace "/" with file separator (OS specified)

https://github.com/swagger-api/swagger-codegen/blob/dbe8ef9df89556965d52c2a680c21730efd40665/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClosureAngularClientCodegen.java#L98

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/LumenServerCodegen.java
```diff
@@ -154,7 +154,7 @@ public String escapeReservedWord(String name) {
      * instantiated
      */
     public String modelFileFolder() {
-        return outputFolder + "/" + sourceFolder + "/" + modelPackage().replace('.', File.separatorChar);
+        return outputFolder + "/" + modelPackage().replace('.', File.separatorChar);
```
wing328: I think you need to replace `/` with `File.separatorChar` for `(outputFolder + "/" + modelPackage())`

wing328:
@abcsun you can refer to https://github.com/swagger-api/swagger-codegen/pull/3251 on how I fix the folder path for Windows platform.

Issue #3168, opened 2016-06-18
[csharp] Allow additional-properties to set Assembly Info


RowanJKWalker:
##### Description

These lines of code, https://github.com/swagger-api/swagger-codegen/blob/v2.1.6/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CSharpClientCodegen.java#L142-L146 , sets the values of:
- packageTitle
- packageProductName
- packageDescription
- packageCompany
- packageCopyright

This means they have the same values for every CSharp client generated.

We want to be able to set these values via additional-properties on command line generation.

##### Swagger-codegen version

https://github.com/swagger-api/swagger-codegen/tree/v2.1.6

##### Related issues
https://github.com/swagger-api/swagger-codegen/issues/1797

##### Suggest a Fix
For each of the following:
- packageTitle
- packageProductName
- packageDescription
- packageCompany
- packageCopyright

Add a check to see if a value has been specified in the additional-properties, e.g.:
```
if (!additionalProperties.containsKey(CodegenConstants.PACKAGE_NAME)) {
            additionalProperties.put("packageTitle", this.packageTitle);
}
```

jimschubert:
@RowanJKWalker Thanks for providing this feedback.

One clarification:

`CodegenConstants.PACKAGE_NAME` is the namespace while `packageTitle` is the assembly's title which would display in DLL properties, these aren't the same thing.  From [AssemblyTitleAttribute Class](https://msdn.microsoft.com/en-us/library/system.reflection.assemblytitleattribute(v=vs.110).aspx):

> Specifies a description for an assembly.

While there's no steadfast rule that these _can't_ be the same, it would be restrictive to make them the same.

This should be a relatively simply fix, so I'll start on it today.

jimschubert:
@RowanJKWalker actually, since this is a pretty simple fix, would you like to provide a PR?

My recommendation is to add the constants to `CodegenConstants`, then add the properties and their defaults to `AbstractCSharpCodegen` and pull/set values in `processOpts` like others:

```
// {{packageVersion}}
if (additionalProperties.containsKey(CodegenConstants.PACKAGE_VERSION)) {
    setPackageVersion((String) additionalProperties.get(CodegenConstants.PACKAGE_VERSION));
} else {
    additionalProperties.put(CodegenConstants.PACKAGE_VERSION, packageVersion);
}
```

RowanJKWalker:
Cool, will do. Thanks!

[...]

PR #3169, opened 2016-06-19
Issue-3168: [csharp] Allow additional-properties to set Assembly Info #3168

RowanJKWalker:
[csharp] Enabling Assembly Info to be set by the following command line's additional-properties:
- packageTitle
- packageProductName
- packageDescription
- packageCompany
- packageCopyright

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConstants.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConstants.java
@@ -54,11 +54,17 @@ public class CodegenConstants {
 
     public static final String PACKAGE_NAME = "packageName";
     public static final String PACKAGE_VERSION = "packageVersion";
+    
     public static final String PACKAGE_TITLE = "packageTitle";
+    public static final String PACKAGE_TITLE_DESC = "Specifies an AssemblyTitle for the .NET Framework global assembly attributes stored in the AssemblyInfo file.";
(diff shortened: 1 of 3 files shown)


RowanJKWalker:
@wing328 @jimschubert here is my Pull Request for https://github.com/swagger-api/swagger-codegen/issues/3168

jimschubert:
I'll test this later today or tomorrow. One thing I'd add is constant descriptions for title and description to differentiate from PACKAGE_NAME. Iirc, title is just a non-code representation of package name, while description can be a short summary of the package.  

We could probably use the explanations from MSDN Assembly Info types (AssemblyTitleAttribute for example).

RowanJKWalker:
Good idea, will do it now.

RowanJKWalker:
I have added a DESC for each of the new CodegenConstants. I went with the approach of saying exactly which Assembly Attribute would be set by specifying a value for the additional property.

jimschubert:
Looks good. Tested the changes and the AssemblyInfo gets updated as expected.

I think the *_DESC constants' wording is fine for now. These aren't automatically used anywhere, but they'll remove the guesswork for anyone looking at the file. If anyone wants to reuse the constants, we can make the descriptions more generic. I'd also like to eventually expose a list of supported additional properties and descriptions, but I don't think that's a high priority.

RowanJKWalker:
Cool, looking forward to the merge.

wing328:
@RowanJKWalker thanks for the PR.

PR #3065, opened 2016-06-06
[Java] Add support for JSR310 in Jersey2 Clients

ant3:
If the dateLibrary=java8 option is set then jersey2 client code will be generated using JSR310 classes rather than Joda time.  Because JSR310 requires java8 the java target version is also changed in the build files.

If dateLibrary=joda (or undefined) then the support files use Joda time classes as before.

Previously the model classes used JSR310 classes but the support files such as JSON.java still referenced Joda time classes.  I'm not sure this would ever have worked properly and even if it did it was messy.

This option (-DdateLibrary=java8) should probably be documented somewhere but I'm not sure where a good place is?

PR #3174, opened 2016-06-19
[Java] Add support for JSR310 in Jersey Clients

ant3:
This is further (belated) work on #3065 

The Jackson module for JSR310 was incorrectly named and generated code did not compile.  This PR fixes that as well as improving the formatting in the mustache templates (as suggested in previous PR).


wing328:
I tested both jersey1 and jersey2 Java API client and both compiled without issue

wing328:
@ant3 for your upcoming PRs, please run the following to update the Petstore sample for Java (jersey1, jersey2) Petstore clients:
```
./bin/java-petstore.sh
```
so that the changes will be tested by travis-ci

ant3:
Thanks for merging this so quickly.  The 2 scripts you mentioned don't specify dateLibrary=java8 so wouldn't have picked up the issue.  I could add a java8-petstore-jersey2.sh to test this if you want.

wing328:
Yes, we probably need another script to test dateLibrary option. 

It's still a good practice to generate the Java Petstore for default dateLibrary option so as to ensure the default option is not impacted by the change.

modules/swagger-codegen/src/main/resources/Java/ApiClient.mustache
```diff
@@ -66,7 +66,7 @@ public class ApiClient {
     objectMapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
     objectMapper.enable(SerializationFeature.WRITE_ENUMS_USING_TO_STRING);
     objectMapper.enable(DeserializationFeature.READ_ENUMS_USING_TO_STRING);{{#java8}}
-    objectMapper.registerModule(new JavaTimeModule());{{/java8}}{{^java8}}
+    objectMapper.registerModule(new JSR310Module());{{/java8}}{{^java8}}
```
cbornet: Why did you use JSR310Module ? It's deprecated and the one to use is JavaTimeModule !

ant3: The generated pom used a slightly older version of the jackson JSR310 library - which didn't have the newer JavaTimeModule.  However, I take your point and have updated the dependency and the code in my latest PR.

ant3:
I have created PR #3186 to address issues raised in comments above.

Thanks for bearing with me.

PR #3175, opened 2016-06-19
Fix the test dir location

ant3:
Previously only a main source directory was added to the build, but this means that the generated tests don't compile properly as they have test scope dependencies.  Adding a main source dir and a test source dir fixes this.

This change means that instead of adding the top-level generated source directory as a source root, the plugin tries to add the correct main and test roots.  It has to do a little gueswork and may not work for all languages, but should be OK for many (most?) languages.  It is likely that the languages where it doesn't work would not be built with maven.  If there is an issue the plugin can be instructed not to add any source roots by setting addCompileSourceRoot to false.


wing328:
@ant3 thanks for the PR.

cc @ePaul

wing328:
@ant3 sorry for taking this long to review. Please submit a new one based on the latest master if you still want this enhancement to be merged into master.

Issue #3140, opened 2016-06-14
[Jaxrs] Missing Import in Generated Model for Map<String, List<T>>


qtstc:
##### Description

I was trying to add this property
```
        "projections": {
          "type": "string",
          "additionalProperties": {
            "type": "array",
            "items": {
              "type": "number"
            }
          }
        }
```
In the generated model class, the `import java.math.BigDecimal;` is missing.
##### Swagger-codegen version
Happens on 2.16 and current master build.
##### Command line used for generation
`java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -i test.json -l jaxrs`

wing328:
PR merged. Issue closed.

Thanks again for the PR from @qtstc

PR #3176, opened 2016-06-20
Add Import for Inner CodegenProperties

qtstc:
https://github.com/swagger-api/swagger-codegen/issues/3140


wing328:
@qtstc thanks for the PR. I've added some test models for `number` and the test results look good.

For your upcoming PR, I would recommend creating a new branch instead of using master.

PR #3179, opened 2016-06-20
Dart - Java compatibility

areiter:
JSON representation of dart and java generated code is not the same.
I added the @Property annotation to the dart model.mustache 

before a property `some-name` was translated to `someName` in JSON but should be `some-name`


wing328:
@areiter thanks for the PR. Please run `./bin/dart-petstore.sh` to keep the Dart Petstore sample up-to-date.

areiter:
@wing328 done

Issue #3194, opened 2016-06-22
Model imports don't support nested Arrays and Maps.


faresd:
Model imports don't support nested Arrays and Maps.
##### Description

The problem can be reproduced using Swagger property the contains a nested Arrays such as:

``` yaml
"values" : {
          "type" : "array",
          "items" : {
            "type" : "array",
            "items" : {
              "$ref" : "#/definitions/ApiNameValue"
            }
          }
        }
```

In the model generated (in my case JS) don't include the object "ApiNameValue" in the imports.
##### Swagger-codegen version

2.1.6

##### Swagger declaration file content or url

https://swagger-ui-1243.appspot.com/apis/search/v1

wing328:
It should be fixed in the latest stable version v2.2.3. Please give it a try.

PR #3195, opened 2016-06-22
A quick fix to support 1 sub array like [[Object]]

faresd:
A pull request following ticket #3194, Imports don't support nested Arrays and Maps.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -1501,30 +1501,38 @@ public class DefaultCodegen {
         }
 
         property.baseType = getSwaggerType(p);
-
+        // A quick fix to support 1 sub array like [[Object]]
       	if (p instanceof ArrayProperty) {
-        		property.isContainer = true;
-        		property.isListContainer = true;
-        		property.containerType = "array";
-        		ArrayProperty ap = (ArrayProperty) p;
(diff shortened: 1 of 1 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
```diff
@@ -1501,30 +1501,38 @@ public CodegenProperty fromProperty(String name, Property p) {
         }
 
         property.baseType = getSwaggerType(p);
-
+        // A quick fix to support 1 sub array like [[Object]]
```
jimschubert: I think this comment, if necessary, should be above your ~7 line addition and not above the entire block for `ArrayProperty`.

wing328:
@faresd thanks for the PR. I believe your fix only handle 1D and 2D arrays. Ideally we are looking for a recursive way to support arrays of any dimensions.

wing328:
@faresd I've filed https://github.com/swagger-api/swagger-codegen/pull/3199 to support array/map of enum for any dimensions. Here is the function with the recursive logic:

https://github.com/swagger-api/swagger-codegen/pull/3199/files#diff-0c1b6d04d7c2c41f87f7c710a5610d88R1634

Would you have cycle to revise your PR to recursively identify the inner type of array/map for import?

faresd:
@wing328 I don't think I will have the time implement the recursion.

wing328:
@faresd thats ok. We appreciate your contribution and will merge this PR after 2.2.0 issue if no question from anyone.

PR #3207, opened 2016-06-23
Add jsdoc type annotations to model constructors

### LINKED ISSUE #3164: [JavaScript] Missing model annotations in generated JSDoc?
When creating a JS client with the `emitJSDoc=true` option, the parameters of the model classes are lacking type annotations. For example, I have a generated `model/Document.js` class with this constructor:

```javascript
  /**
   * Constructs a new <code>Document</code>.
   * @alias module:model/Document
   * @class
   * @param documentId
   * @param modified
   */
  var exports = function(documentId, modified) {
    ...
```

The generated HTML page for this model class has a table, which is also missing this type information, as seen here:
screenshot 2016-06-17 16 05 08

Somehow the listings below the table *does* have the information. Not sure where it got it.

Also, how can I create annotations in Java that will generate entries in the "Description" fields in the table?

The generated markdown version looks like this:

```markdown
# Sunflower.Document

## Properties
**documentId** | **String** |  | 
**modified** | **Boolean** |  | [default to false]
```

So, I'm not sure where the problem needs to be addressed. But shouldn't the JavaScript look like this:

```javascript
  /**
   * Constructs a new <code>Document</code>.
   * @alias module:model/Document
   * @class
   * @param documentId {String}
   * @param modified {Boolean}
   */
  var exports = function(documentId, modified) {
    ...
```

delenius:
Fixes #3164


wing328:
@delenius thank you for the PR. I've commented out the test cases which break the JS API client.

Please rebase on the latest master and generate the Petstore sample again.

delenius:
Now I see this test failure:

```
  ApiClient
    defaults
      1) "before each" hook for "should have correct default values with the default API client"


  0 passing (12ms)
  1 failing

  1)  "before each" hook for "should have correct default values with the default API client":
     ReferenceError: _number is not defined
      at new exports (src/model/FormatTest.js:66:23)
      at Context.<anonymous> (test/model/FormatTest.spec.js:18:16)
```

delenius:
Oh, that might be my fault. Looking into it.

PR #3224, opened 2016-06-27
[PHP] Better code injection handling for PHP API client

wing328:
For https://github.com/swagger-api/swagger-codegen/pull/3201


arnested:
LGTM 👍 

One thing, though, is the addition of the injection attempt on the `host`, `basepath`, and `scheme` properties in `petstore-with-fake-endpoints-models-for-testing.yaml`. This make all the test cases fail...

wing328:
@arnested thanks for reviewing. We will probably need to create another spec to cover different code injection scenarios.

wing328:
A couple of update:
- create a new swagger spec for security tests
- add the same security fix to PHP server generator: Lumen, Silex, Slim but the code won't compile and we will address those one by one.

PR #3228, opened 2016-06-28
[Ruby] Add params_encoding configuration to be passed to api_client's request options

cliffano:
params_encoding configuration allows the client to determine how to construct array parameter with multi collection format.

For example:

If I have the following spec
```
- name: myparam
  in: query
  required: false
  type: array
  collectionFormat: multi
```
and then I pass multiple value in the ruby code
```
{ :myparam => ['myvalueA', 'myvalueB', 'myvalueC' ] }
```
it will generate request parameters
```
myparam[0]=myvalueA&myparam[1]=myvalueB&myparam[2]=myvalueC
```

However, if I want to generate unindexed parameter names, I have to pass params_encoding option to Typhoeus (which then passes it to Ethon https://github.com/typhoeus/ethon/blob/master/lib/ethon/easy/queryable.rb#L103).

This patch allows the client to configure params_encoding option like
```
    SwaggerClient.configure { |conf| [
      conf.params_encoding = :multi
    ]}
```
which will generate request parameters
```
myparam=myvalueA&myparam=myvalueB&myparam=myvalueC
```


wing328:
@cliffano thanks for the PR. The change, together with test cases, docstring, etc, looks very good to me.

cc @zlx

zlx:
Seems good to me @wing328

wing328:
@zlx thanks for reviewing.

PR #3236, opened 2016-06-28
Disable path HTML-escaping in Ruby api template

cliffano:
Prior to this PR, the default generator HTML-escapes the variables passed to Mustache, which breaks specification with a path that contains a HTML entity.

In my use case, the path contains an equal sign (e.g. /foo=bar), which gets escaped to &#x3D; by Mustache's Escaper (into /foo&#x3D;bar).
This obviously breaks the API client because it sends a request to /foo&#x3D;bar instead of /foo=bar .

This PR ensures that the path stays AS-IS in the generated client code.

### DIFF
--- a/modules/swagger-codegen/src/test/java/io/swagger/codegen/DefaultGeneratorTest.java
+++ b/modules/swagger-codegen/src/test/java/io/swagger/codegen/DefaultGeneratorTest.java
@@ -1,6 +1,7 @@
 package io.swagger.codegen;
 
 import io.swagger.codegen.languages.JavaClientCodegen;
+import io.swagger.codegen.languages.RubyClientCodegen;
 import io.swagger.models.Swagger;
 import io.swagger.parser.SwaggerParser;
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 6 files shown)


wing328:
@cliffano what about using {{& tag_name }} (or {{{tag_name}}}) to avoid the HTML escape instead?

Your change in the default generator will impact other code generator (e.g. C#) which requires the HTML escaped mustache tag.

cliffano:
@wing328 Sorry, I didn't realise that some languages require escaped paths.

I have moved the unescaped path handling from the DefaultGenerator to Ruby's api template.
Agreed with your recommendation. I guess because Mustache escaping is applied to all or none, the only way out is to opt-out at language template level.

cliffano:
Updated the title to reflect the latest change.

cliffano:
Added a minor update to Ruby's API client template to handle file downloads which Content-Disposition header doesn't include any filename attribute.

modules/swagger-codegen/src/test/java/io/swagger/codegen/DefaultGeneratorTest.java
```diff
@@ -221,6 +222,32 @@ public void testGenerateUniqueOperationIds() {
...
+            apiFileGenerated = true;
+            // Ruby client should set the path unescaped in the api file
+            assertTrue(FileUtils.readFileToString(file, StandardCharsets.UTF_8).contains("local_var_path = \"/foo=bar\""));
+          }
+        }
+        if (!apiFileGenerated) {
+          fail("Default api file is not generated!");
+        }
+    }
+
```
wing328: @cliffano thanks for writing the test case to cover the change. If it's not too much work, I would suggest we put the test under a new Java class in the [ruby folder](https://github.com/cliffano/swagger-codegen/tree/8bacbfb6916cf2a3cb8b9e69d62d9e49fda37fa2/modules/swagger-codegen/src/test/java/io/swagger/codegen/ruby). 

Ref: [Java test cases](https://github.com/cliffano/swagger-codegen/tree/8bacbfb6916cf2a3cb8b9e69d62d9e49fda37fa2/modules/swagger-codegen/src/test/java/io/swagger/codegen/java)

cliffano:
@wing328 Test for generated ruby client has been moved to java.io.swagger.codegen.ruby package as per suggestion https://github.com/swagger-api/swagger-codegen/pull/3236/files/8bacbfb6916cf2a3cb8b9e69d62d9e49fda37fa2#r69063609 .

wing328:
@cliffano thanks for the contribution and I've merged the PR into the master.

PR #3238, opened 2016-06-29
regenerate ruby petstore files

chadrosen:
This diff was created by simply running the bin/ruby-petstore.sh script from the top level directory. I assume that this generates a diff because the ruby sample code has drifted from some changes that weren't updated.


wing328:

chadrosen:
@wing328 I followed the commands in that stack overflow article. Let me know if I did this correctly.

chadrosen:
I think I messed this up with a commit. I'm going to throw it away and do it again.

PR #3248, opened 2016-06-29
Use okttp builder instead of instance in retrofit2

### LINKED ISSUE #3188: [Java] Bug in client code generation with Retrofit2 fails to add new Auth Interceptor to OkHttpClient used
##### Description

When trying to use Java swagger codgen auto-generated client with basic Auth it fails.

##### Swagger-codegen version
2.1.6
As far as I know this is not a regression.
2.1.6 is the first version I am using.
This does not look like something added lately.

generate
                    -l java
                    --artifact-version 1.0.0
                    --group-id com.foo.bar
                    --artifact-id bar-sdk
                    -i swagger.json
                    -o ./output
                    --invoker-package com.foo.bar.invoke
                    --model-package com.foo.bar.model
                    --api-package com.foo.bar.api
                    --library retrofit2

##### Steps to reproduce

Creating a new ApiClient with basicAuth fails:
ApiClient apiClient = new ApiClient("basicAuth", "user@bar.com", "password");
..with the following Exception:
<<<<
Exception in thread "main" java.lang.UnsupportedOperationException
	at java.util.Collections$UnmodifiableCollection.add(Collections.java:1055)
	at com.foo.bar.sdk.invoke.ApiClient.addAuthorization(ApiClient.java:252)
	at com.foo.bar.sdk.invoke.ApiClient.<init>(ApiClient.java:55)
	at com.foo.bar.sdk.invoke.ApiClient.<init>(ApiClient.java:64)
	at com.foo.bar.sdk.invoke.ApiClient.<init>(ApiClient.java:84)
	at com.foo.bar.sdk.invoke...
>>>>

In the swagger.json file I have the following section defined:
<<<<
...
  "securityDefinitions": {
        "basicAuth": {
            "type": "basic",
            "description": "HTTP Basic Authentication. Works over `HTTP` and `HTTPS`"
        }
    },
...
>>>>

##### Related issues

n/a

##### Suggest a Fix

The reason for the failure is that OkHttpClient swagger codegen auto-generated ApiClient tries to add Interceptor (in our case BasicAuthHttp) to OkHttpClient's interceptors via Map.add() instead of reconstructing it with OkHttpClient.Builder().

Details:
In generated ApiClient the method addAuthorization looks like the following:
<<<<
    public void addAuthorization(String authName, Interceptor authorization) {
        if (apiAuthorizations.containsKey(authName)) {
            throw new RuntimeException("auth name \"" + authName + "\" already in api authorizations");
        }
        apiAuthorizations.put(authName, authorization);
        okClient.interceptors().add(authorization);
    }
>>>>

.. in the last line "okClient.interceptors().add(authorization);" is the one that is failing with java.lang.UnsupportedOperationException.

The reason is that okClient.interceptors() which is initiated by the OkHttpClient Builder to Collections.unmodifiableList which does not allow adding items of course. This is a snippet from the constructing code:
<<<<
private OkHttpClient(Builder builder) {
    this.dispatcher = builder.dispatcher;
    this.proxy = builder.proxy;
    this.protocols = builder.protocols;
    this.connectionSpecs = builder.connectionSpecs;
    this.interceptors = Util.immutableList(builder.interceptors);
...
>>>>

I would like to suggest replacing the "okClient.interceptors().add(authorization);" with building a new OkHttpClient and adding the authorization via it's Builder.

cbornet:
Fix #3188


wing328:
## Upgrade Note ##

This change removes the following methods from Java Retrofit2 API client:
```
    public OkHttpClient getOkClient()
    public void addAuthsToOkClient(OkHttpClient okClient)
```
and add the following instead:
```
    public OkHttpClient.Builder getOkBuilder()
    public void addAuthsToOkBuilder(OkHttpClient.Builder okBuilder)
```

PR #3249, opened 2016-06-29
[C#] better code injection handling for C# API client

wing328:
For #3201


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java
```diff
@@ -656,4 +656,16 @@ public String toEnumName(CodegenProperty property) {
     public String testPackageName() {
         return this.packageName + ".Test";
     }
+
+    @Override
+    public String escapeQuotationMark(String input) {
```
jimschubert: I think this function and `escapeUnsafeCharacters` need to be heavily commented in all language overrides.

Consider 1 or 2 years down the road:

1. This method is removing quotes, not escaping. Why?
2. Comment on line 662 could be read as avoiding SQL code injection in generated code. Maybe clarify the comment?
3. From a usage perspective, it's not clear whether this is single or double quotes.
4. It's not clear the intended context of this method (Mustache templates vs. in generated code).

Issue #3250, opened 2016-06-29
getHasExamples() is missing in CodegenOperation class


alexralko:
##### Description
getHasExamples() is not implemented for CodegenOperation class
So it is not possible to use {{hasExamples}} tag in template (e.g. in modules\swagger-codegen\src\main\resources\htmlDocs\index.mustache)

##### Swagger-codegen version
2.2.0-SNAPSHOT

##### Command line used for generation

java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -i http://petstore.swagger.io/v2/swagger.json -l html

java version "1.7.0_79"
Apache Maven 3.0.5

##### Suggest a Fix

Implement getHasExamples() method for CodegenOperation class

wing328:
PR merged.

PR #3252, opened 2016-06-29
add getHasExamples() method to support {{hasExamples}} tag in templates

alexralko:
add getHasExamples() method to support {{hasExamples}} tag in templates #3250


wing328:
@alexralko thanks for the PR, which has been merged into the master.

PR #3256, opened 2016-06-29
[PHP] Add path without expanded path parameters to callApi

arnested:
In continuation of #3117 it could be useful to know the path of an endpoint (without path parameters expanded) in the `callApi` method of `ApiClient`.

This is for use cases where you would create a derived class from `ApiClient` for manipulating responses from the server before further processing (#3117) or add extended logging of the API calls (a use case I currently have where I want to track timing information for each API call).


wing328:
## Upgrade Note ##

This change will update the method signature of `callApi` with an additional optional parameter `endpointPath` as follows:
```
public function callApi($resourcePath, $method, $queryParams, $postData, $headerParams, $responseType = null, $endpointPath = null)
```

If your class overrides `callApi`, please make sure you update the method signature accordingly.

PR #3273, opened 2016-07-01
use baseName instead of paramName

qtstc:
Use baseName instead of paramName in @FormParam(). 

baseName is the one specified in the "name" field in the schema, and is normally of the form "user_id". I believe paramName is converted from baseName to be used as Java parameter name, and it is of the form "userId".

Currently, for the schema
```
parameters: [
    {
            "name": "user_id",
            "in": "formData",
            "type": "string"
    }
]
```
the generated code would be ```@FormParam("userId") String userId```. The Jersey framework would then expect the name of the form parameter to be "userId" instead of user_id.

We are using the baseName instead of paramName in pathParams.mustache and queryParams.mustache.


wing328:
@qtstc thanks for the PR. Please do a rebase on the latest master to resolve the merge conflicts.

gmousset:
There is the same issue with SpringMVC and path parameters.

wing328:
@gmousset please pull the latest master and give it another try as the spring generator has changed a lot (e.g. default to generate spring boot server stub instead).

If it's still an issue, please help submit a PR so that we can review the fix.

cc @cbornet

qtstc:
@wing328 merged conflict resolved

wing328:
@qtstc  PR merged. Thanks for the contribution.

PR #3278, opened 2016-07-02
Better handling of multi-line comments for all generators

wing328:
- to cover all remaining generators
- to handle block comment starter (e.g. /*, =begin)
- to add underscore instead of removing the block comment starter and terminator


wing328:
For #3201

edstover:
Hi,

I think this change has the side-effect of changing media type declarations from "*/*" to "*_/_*".  I've noticed this in generated API specifications when the following yaml is used:

consumes:
  - application/json
  - text/plain
  - "*/*"
produces:
  - application/json

The resulting Java API when generated using "spring" for the language looks like this:

        produces = { "application/json" }, 
        consumes = { "application/json", "text/plain", "*_/_*" },
        method = RequestMethod.GET)

Can the substitution with underscores be suppressed for media types?

wing328:
@edstover "*/*" is better handled via https://github.com/swagger-api/swagger-codegen/pull/4155

Please pull the latest master to give it a try.

Issue #3285, opened 2016-07-04
[JAVA] Java-Types (BigDecimal, LocalDate, ...) are also effected by modelNamePrefixes and Suffixes


d0x:
##### Description

When swagger is using Java-Types (BigDecimal, LocalDate, ...) as fields, those are also renamed.

In case `modelNameSuffix` is set to `Dto`, this will into Types like `BigDecimalDto` which is wrong.

This snipped is taken from the `DefaultCodegen`:

```
    /**
     * Output the proper model name (capitalized)
     *
     * @param name the name of the model
     * @return capitalized model name
     */
    public String toModelName(final String name) {
        return initialCaps(modelNamePrefix + name + modelNameSuffix);
    }
```

and this from `AbstractJavaCodegen`:

```
    @Override
    public String toModelName(final String name) {
        final String sanitizedName = sanitizeName(modelNamePrefix + name + modelNameSuffix);

        // camelize the model name
        // phone_number => PhoneNumber
        final String camelizedName = camelize(sanitizedName);

        // model name cannot use reserved keyword, e.g. return
        if (isReservedWord(camelizedName)) {
            final String modelName = "Model" + camelizedName;
            LOGGER.warn(camelizedName + " (reserved word) cannot be used as model name. Renamed to " + modelName);
            return modelName;
        }

        // model name starts with number
        if (name.matches("^\\d.*")) {
            final String modelName = "Model" + camelizedName; // e.g. 200Response => Model200Response (after camelize)
            LOGGER.warn(name + " (model name starts with number) cannot be used as model name. Renamed to " + modelName);
            return modelName;
        }

        return camelizedName;
    }
```
##### Swagger-codegen version

2.2.0-SNAPSHOT
##### Suggest a Fix

Use the `typeMapper.values` and ignore those.

I'll open a PR for this.

ePaul:
Good catch, thanks.

ePaul:
Do you have a test case where this happens (i.e. an OpenAPI definition + parameters to Codegen)?

d0x:
You can reproduce it like this:

```
java -jar swagger-codegen-cli.jar generate --library feign --model-name-suffix Dto -i dtos.yaml
```

Where `dtos.yaml` should look like this:

```
swagger: '2.0'
info:
  title: Suffix on Types
  description: Shows that this doesn't work yet.
  version: "1.0.0"
paths:
  /example:
    get:
      summary: Product Types
      responses:
        200:
[...]

PR #3288, opened 2016-07-04
[JAVA] #3285 ModelNamePre- and Suffixes should not be applied to Types.

d0x:
As described in #3285 .

In case you know a better way then `typeMapping.values().contains(name)` just tell it.

Best Regards,
Christian.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java
@@ -343,11 +343,6 @@ public abstract class AbstractJavaCodegen extends DefaultCodegen implements Code
 
     @Override
     public String toModelName(final String name) {
-        // Don't do any kind of sanitizing to Java types (e.g. BigDecimal)
-        if(typeMapping.values().contains(name)) {
-            return name;
-        }
-
         final String sanitizedName = sanitizeName(modelNamePrefix + name + modelNameSuffix);
 
         // camelize the model name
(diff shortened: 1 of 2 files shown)


wing328:
@d0x thanks for the PR.

cc @ePaul

d0x:
I removed the check from both changed `toModelName()`-Methods and instead fixed it inside `getSwaggerType()`.

In the fix I remove the checks `languageSpecificPrimitives.contains(type) || type.indexOf(".") >= 0 || type.equals("Map") || type.equals("List") || type.equals("File") || type.equals("Date")` after getting the type from the `typeMapper`. I don't understand or see any reason to change any known `typeMapper` type? If I understood it correct, those Types are already adjusted for the target language. Or do you have a reason? If yes, I could rollback the last commit.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java
```diff
@@ -520,21 +520,16 @@ public void setParameterExampleValue(CodegenParameter p) {
...
         if (typeMapping.containsKey(swaggerType)) {
-            type = typeMapping.get(swaggerType);
-            if (languageSpecificPrimitives.contains(type) || type.indexOf(".") >= 0 ||
-                type.equals("Map") || type.equals("List") ||
-                type.equals("File") || type.equals("Date")) {
-                return type;
-            }
-        } else {
-            type = swaggerType;
+            return typeMapping.get(swaggerType);
```
ePaul: This looks sensible for me, though I fear that it changes the meaning of `typeMapping`, when comparing with most other languages, which still apply `toModelName` after applying the typeMapping, other then when it is in `languageSpecificPrimitives`.  I'm not quite sure what the actual intended purpose of type mapping is.

From the documentation (in README.md):

        --type-mappings <type mappings>
            sets mappings between swagger spec types and generated code types in
            the format of swaggerType=generatedType,swaggerType=generatedType.
            For example: array=List,map=Map,string=String

This is still not totally clear. I guess what we want is "if something is the result of a type mapping, use it as-is and don't transform it further" (so no escaping, sanitizing, prefixing, ...). (This is how it is implemented here.)

@fehguy @wing328 is this a sensible approach for type mapping which we should document in general, and also implement for all other languages this way?

wing328: > "if something is the result of a type mapping, use it as-is and don't transform it further"

I agree with your interpretation but I might have missed an edge case I've encountered before.

I'll try to merge this so that we've a potential fix for Java (#4430)

wing328:
@d0x thanks for the PR, which has been merged into master.

My apologies for taking this long to merge this PR as I've missed it.

joergi77:
@wing328 
are your sure it's merged? there is still:
 ```
    if (typeMapping.containsKey(swaggerType)) {
            type = typeMapping.get(swaggerType);
            if (languageSpecificPrimitives.contains(type) || type.indexOf(".") >= 0 ||
                type.equals("Map") || type.equals("List") ||
                type.equals("File") || type.equals("Date")) {
                return type;
            }
        } else {
            type = swaggerType;
}
```
in the actual master file....

PR #3292, opened 2016-07-05
Fix typo in README.md

RowanJKWalker:
On line 379 -DapiTest should be -DapiTests and -DmodelTest should be -DmodelTests.


wing328:
@RowanJKWalker thanks for the fix.

PR #3303, opened 2016-07-05
Fix missing json import in go client template

yannrouillard:
I used the last source code of swagger-codegen to generate proper go client code but compilation failed with the following error:

    undefined: json in json.Unmarshal

This commit intents to fix this.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/go/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/go/api.mustache
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"errors"
 	"net/url"
+	"encoding/json"
 	{{#imports}}"{{import}}"
 {{/imports}}
 )


wing328:

wing328:
I think it would be a problem to hardcode it in the api.musatche if the swagger spec does not contain any endpoint that deal with model (json).

Can you provide a spec that we can reproduce the issue?

wing328:
cc @guohuang @neilotoole

guohuang:
@wing328 it will be a problem for other users who run the code gen, i need to know the spec which needs this reference, then i will programmatically import the reference instead.

wing328:
@guohuang yes, I agree.

guohuang:
@yannrouillard what is the return type of the function that is missing the reference? we only add that reference if the return type is not primitive.

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/GoClientCodegen.java#L380

wing328:
@yannrouillard thanks for the PR again but we've to close this as there's no update from you.

Issue #3300, opened 2016-07-05
[JAX-RS] Codegen Creates form-data variables with duplicate names


jencodingatwork:
##### Description

The JAX-RS Code generator creates variables with duplicate names when more than one file is included in the API call which results in a compilation error.

##### Swagger-codegen version
2.1.6

##### Swagger declaration file content or url

```
paths:
  /Plugin:
    post:
      tags:
        - Plugin
      summary: Upload Plugin
      description: This can only be done by the logged in user.
      operationId: createPlugin
      consumes:
      - "multipart/form-data"
      produces:
        - application/json
        - application/xml
      parameters:
        - in: formData
          name: json
          description: Uploaded Plugin
          required: true
          type: string
        - in: formData
          name: jarFile
          description: Jar File to upload
          required: true
          type: file
        - in: formData
          name: configFile
          description: config File to upload
          required: true
          type: file
      responses:
        200:
          description: successful operation
```

##### Command line used for generation

java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -i C:/PluginAPI/plugin.yaml -l jaxrs -o C:/Temp/JAX_RStest java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -i C:/PluginAPI/plugin.yaml -l jaxrs -o C:/Temp/JAX_RStest

##### Steps to reproduce

1. Execute command
2. Look at Plugin API code, where you should see the following method. Notice that there are two variables named inputStream, two variables named fileDetail, and all params are set to “file”.

```
@POST 
    @Produces({ "application/json", "application/xml"})
    @Consumes({"multipart/form-data"})
    @io.swagger.annotations.ApiOperation(value = "Upload Plugin", notes = "This can only be done by the logged in user.", response = void.class, tags={ "Plugin", })
    @io.swagger.annotations.ApiResponses(value = { 
        @io.swagger.annotations.ApiResponse(code = 200, message = "successful operation", response = void.class) })
    public Response createAnalytic(@ApiParam(value = "Uploaded Plugin", required=true)@FormParam("json")  String json,
            @FormDataParam("file") InputStream inputStream,
            @FormDataParam("file") FormDataContentDisposition fileDetail,
            @FormDataParam("file") InputStream inputStream,
            @FormDataParam("file") FormDataContentDisposition fileDetail,@Context SecurityContext securityContext)
    throws NotFoundException {
[...]

PR #3304, opened 2016-07-05
Issue #3300 Modified JavaJaxrs resources to fix duplicated variable names

jencodingatwork:
Fixed issue where the JAX-RS code generator creates variables with duplicate names when more than one file is included in the API call which results in a compilation error by modifying the api.mustache file, formParams.mustache file, and serviceFormParams.mustache file.


wing328:
@jencodingatwork thanks for the PR, which has been merged into master.

PR #3312, opened 2016-07-06
[Java] Set Java default library to "okhttp-gson"

wing328:
We'll switch the default of Java API client use `okhttp-gson` HTTP library. To use the default (Jersey 1.x), please set library to `jersey1`.


cbornet:
It seems the jersey1 petstore sample has lost its tests in the process...

wing328:
@cbornet oh my bad. I'll take a look this weekend.

wing328:
Fixed via https://github.com/swagger-api/swagger-codegen/pull/3728

PR #3322, opened 2016-07-07
Raise Exception Instead of String in Python Test

qtstc:
```
======================================================================
ERROR: test_delete_pet (tests.test_pet_api.PetApiTests)
----------------------------------------------------------------------
  File "/home/travis/build/swagger-api/swagger-codegen/samples/client/petstore/python/tests/test_pet_api.py", line 203, in test_delete_pet
    raise "expected an error"
TypeError: exceptions must be old-style classes or derived from BaseException, not str
```
in https://travis-ci.org/swagger-api/swagger-codegen/jobs/143133869.

The test itself seems flaky. And that's not fixed in this PR.


wing328:
@qtstc I could not repeat the error but your improvement on how the exception is thrown looks good to me.

PR #3324, opened 2016-07-08
Fix Objective-C client .podspec description field

tar500:


wing328:
@tar500 can you share more what you're trying to fix? Any spec or example description to repeat the issue?

The current approach using multi-line string in Ruby looks fine to me.

wing328:
cc @mateuszmackowiak

tar500:
@wing328 I'm tyring to fix a cocoapods 1.0.0+ error:
"- ERROR | description: The description is empty."

Currently in our project we have next generated lines in a podspec:
```
 s.description      = <<-DESC
                         
                         DESC
```

I believe this is invalid, so I removed that. Just googled that those are ruby operators for multiline strings.

I'm sorry I grabbed your attention this way, without providing any comments.
Hope you can help to figure out how to avoid passing raw ruby operators to the .podspec output.
I have no access to the original .json for a generated swagger client right now, but I can provide one later if needed.

Thanks!

wing328:
Fix via #3404. Closing this one.

PR #3327, opened 2016-07-08
Improve jsdoc for API return values

### LINKED ISSUE #3340: [JS] jsdoc for API return values
There are a couple of problems in the generated jsdoc, for the javascript client:

- Return value are not documented when using usePromise option (instead, there is an incorrect text added to the last parameter)
- When not using usePromise option, data type in API callbacks is not a hyperlink to the relevant model class doc

I filed PR #3327 which fixes these.

delenius:
Fixes #3340

To accomplish this, I had to remove the curly brackets out of the jsdoc strings in `vendorExtensions.x-jsdoc-type`. This was so that I could do `{@link <&vendorExtensions.x-jsdoc-type>}`. This meant I had to manually add back the curly brackets in the places where they were needed.


delenius:
Hold off on this, I found a problem I need to fix first.

delenius:
Ok problem fixed, please consider this PR.

wing328:
@delenius thanks for the PR, which has been merged into master.

PR #3351, opened 2016-07-12
fix[ruby]: Problem with List of Enum

### LINKED ISSUE #3323: [RUBY]: Problem with List of Enum
##### Description

Ruby client cannot deserialize the response when the model contains a list of enum.

##### Swagger-codegen version

master 89741c75f7ab289596b3d1d386946f0fbb45f859

##### Detailed description

In our model we have a list of enums:

``` json
"userActions": {
    "type": "array",
    "items": {
        "type": "string",
        "enum": [
            "AcceptOrder",
            "...",
            "UploadBinaryFile"
        ]
    }
}
```

This leads to the following check being created in the model:
``` ruby
    # Custom attribute writer method checking allowed values (enum).
    # @param [Object] user_actions Object to be assigned
    def user_actions=(user_actions)
      validator = EnumAttributeValidator.new('Array<String>', [])
      unless validator.valid?(user_actions)
        fail ArgumentError, "invalid value for 'user_actions', must be one of #{validator.allowable_values}."
      end
      @user_actions = user_actions
    end
```
The relevant part of the file (swagger-codegen/modules/swagger-codegen/src/main/resources/ruby/partial_model_generic.mustache) is:
``` mustache
    {{#isEnum}}
    # Custom attribute writer method checking allowed values (enum).
    # @param [Object] {{{name}}} Object to be assigned
    def {{{name}}}=({{{name}}})
      validator = EnumAttributeValidator.new('{{{datatype}}}', [{{#allowableValues}}{{#values}}"{{{this}}}"{{^-last}}, {{/-last}}{{/values}}{{/allowableValues}}])
      unless validator.valid?({{{name}}})
        fail ArgumentError, "invalid value for '{{{name}}}', must be one of #{validator.allowable_values}."
      end
      @{{{name}}} = {{{name}}}
    end

    {{/isEnum}}
```
The data sent to mustache (generated with -DdebugModels) is the following:
``` json
{

    "baseName": "userActions",
    "getter": "getUserActions",
    "setter": "setUserActions",
    "datatype": "List<String>",
    "datatypeWithEnum": "List<UserActionsEnum>",
    "name": "userActions",
    "defaultValue": "[]",
    "defaultValueWithParam": " = data.userActions;",
    "baseType": "List",
    "containerType": "array",
    "example": "null",
    "jsonSchema": "{\n  \"type\" : \"array\",\n  \"items\" : {\n    \"type\" : \"string\",\n    \"enum\" : [ \"AcceptOrder\", \"...\", \"UploadBinaryFile\" ]\n  }\n}",
    "isPrimitiveType": true,
    "isContainer": true,
    "isListContainer": true,
    "isEnum": true,
    "items": {
        "baseName": "userActions",
        "getter": "getUserActions",
        "setter": "setUserActions",
        "datatype": "String",
        "datatypeWithEnum": "UserActionsEnum",
        "name": "userActions",
        "defaultValue": "null",
        "defaultValueWithParam": " = data.userActions;",
        "baseType": "String",
        "example": "null",
        "jsonSchema": "{\n  \"type\" : \"string\",\n  \"enum\" : [ \"AcceptOrder\", \"...\", \"UploadBinaryFile\" ]\n}",
        "isPrimitiveType": true,
        "isNotContainer": true,
        "isString": true,
        "isEnum": true,
        "_enum": [
            "AcceptOrder",
            "...",
            "UploadBinaryFile"
        ],
        "allowableValues": {
            "values": [
                "AcceptOrder",
                "...",
                "UploadBinaryFile"
            ]
        },
        "vendorExtensions": { },
        "nameInCamelCase": "UserActions"
    },
    "vendorExtensions": { },
    "nameInCamelCase": "UserActions"

}
```
As you can see, isEnum is true for the List of Enum, which leads to the problem, because the ruby code tries to check for the values of the enum even though it is a List.

##### Possible solutions

Either isEnum is erroneously true for Lists of Enum and this needs to be fixed, or the Ruby client needs to be expanded so it has special checks if an isEnum is true and the value is in fact a List of enum values.

ircecho:
Fixes #3323

### DIFF
--- a/modules/swagger-codegen/src/main/resources/ruby/partial_model_generic.mustache
+++ b/modules/swagger-codegen/src/main/resources/ruby/partial_model_generic.mustache
@@ -124,8 +124,10 @@
       return false if @{{{name}}}.nil?
       {{/required}}
       {{#isEnum}}
+      {{^isContainer}}
       {{{name}}}_validator = EnumAttributeValidator.new('{{{datatype}}}', [{{#allowableValues}}{{#values}}"{{{this}}}"{{^-last}}, {{/-last}}{{/values}}{{/allowableValues}}])
       return false unless {{{name}}}_validator.valid?(@{{{name}}})
+      {{/isContainer}}
       {{/isEnum}}
       {{#hasValidation}}
       {{#minLength}}
@@ -150,6 +152,7 @@
 
     {{#vars}}
     {{#isEnum}}
+    {{^isContainer}}
     # Custom attribute writer method checking allowed values (enum).
     # @param [Object] {{{name}}} Object to be assigned
     def {{{name}}}=({{{name}}})
@@ -160,6 +163,7 @@
       @{{{name}}} = {{{name}}}
     end
 
+    {{/isContainer}}
     {{/isEnum}}
     {{^isEnum}}
     {{#hasValidation}}


wing328:
@ircecho thanks for the PR.

Next time, please update the Ruby Petstore sample by running `./bin/ruby-petstore.sh` so that the CI can test the change.

PR #3377, opened 2016-07-15
[Travis] Cache maven directory

jgavris:
Cache maven directory for faster builds!


wing328:
@jgavris thanks for the PR to make the build faster.

cbornet:
Nice one! The buid time was 15-16mn, is now 8-9mn !
It's so fast that it broke the CI...

jgavris:
🎈

PR #3378, opened 2016-07-15
[Python] Fix flaky test with id collisions

jgavris:


cbornet:
:+1: This is really annoying...

PR #3004, opened 2016-05-30
Fix inheritance issues due to duplicated properties in java

cbornet:
* If a composed model (allOf) doesn't have any parent and one of its interface has a discriminator field, then set this interface as parent.
* Upgrade to v1.0.21 of the parser (see https://github.com/swagger-api/swagger-parser/pull/246)
* Configure supportsInheritance=true for java based gens.

Fix #2096 
Fix #3147 
Fix #1556 
Fix #1348

PR #3388, opened 2016-07-15
updated versions for release

fehguy:


cbornet:
See #3004 for the parser upgrade

fehguy:
@cbornet that is a huge PR.  I can try merging into this branch

fehguy:
@cbornet as @wing328 has tagged this to 2.3.0, I suggest we hold off.  Objections?

cbornet:
No problem.
Then just pick https://github.com/swagger-api/swagger-codegen/pull/3004/commits/fc516115236136beb17ee8c296317375d268d354 to solve the CI issue.

cbornet:
On the other hand, I think that upgrading the parser without merging my PR can lead to surprises as allOf will no longer set the parent and this is an issue when you have a discriminator (see Dog/Cat/Animal). So maybe wait my PR to upgrade the parser.

fehguy:
@wing328 what do you think?  I'll update this PR to include the discriminator fixes from #3004 but not merge until we discuss.

wing328:
@fehguy yes, as discussed with @cbornet we plan to include #3004 in v2.3.0
(and as @fehguy have pointed out, it's a huge PR and I don't think we should include it in the v2.2.0 release. Personally I didn't have a chance to test #3004, which is not to say the PR is not ready)

fehguy:
OK I will have to adjust the PR then.

wing328:
UPDATE: I'm drafting the release note, which is 50% done.

cbornet:
Maybe just pick https://github.com/swagger-api/swagger-codegen/pull/3004/commits/2ce9a11b7665b11f1903e936226244eddc185b2c which sets the interface that has a discriminator as parent. And check that the generated samples don't change (if not included, Cat & Dog will no longer extend Animal).

PR #3399, opened 2016-07-18
Make jersey tests wait that jetty is closed before terminating

cbornet:
This fixes the CI.


jgavris:
👍  Hit this a few times recently

PR #3401, opened 2016-07-19
Stop Shortening Underscore in Enum Value

qtstc:
"__" is a valid value for enum, but it is shortened to "\_" in AbstractJavaCodegen. In fact, "\_" shouldn't be used because it may not be supported in the future according to this post http://stackoverflow.com/questions/34521690/why-is-a-single-underscore-character-an-illegal-name-for-a-lambda-parameter.

For the following field:
```
"state": {
  "type": "string",
  "enum": [
    "GOOD",
    "BAD",
    "__"
  ]
}
```
The generated enum is
```
  public enum StateEnum {
    GOOD("GOOD"),
    BAD("BAD"),
    _("__");

    private String value;

    StateEnum(String value) {
      this.value = value;
    }

    @Override
    @JsonValue
    public String toString() {
      return value;
    }
  }
```

I am also getting the following compiler warning when using Java 8:
```
[WARNING]   (use of '_' as an identifier might not be supported in releases after Java SE 8)
```


wing328:
@qtstc what about mapping `__` to `DOUBLE_UNDERSCORE` instead?

There's similar request to map symbol (e.g. #, @, &, etc) to enum value.

qtstc:
@wing328 I think given that "__" is a valid identifier in Java, just like "o1" or "user", we shouldn't modify it just for readability in a one-off case. 
I agree that we need to deal with other symbols though. And I think what the code does right now (e.g.  replacing those symbols with "_") is simple and reasonable.

wing328:
@qtstc Thanks for writing the test case to cover the change. Can you put it under `modules/swagger-codegen/src/test/java/io/swagger/codegen/java` instead? I think that's folder where developers expect to find Java-related test cases.

qtstc:
@wing328 moved the file to ```modules/swagger-codegen/src/test/java/io/swagger/codegen/java```

Issue #2938, opened 2016-05-23
[typescript-angular2] APIKey setting unavailable


inska:
##### Description

Thank you @Vrolijkx, and generated code have no error, and build successfully. However, there are some missing Swagger spec on the client. We need to setup API Key for communicating with API Server. In other client such as Swift or Python, I am updating api_key in singleton configuration class to set API Key.. but that seems missing in typescript-angular2 client. 

Please consider above case and put them in.
##### Swagger-codegen version

swagger-api version is 2.1.6, and master branch with codegen
##### Swagger declaration file content or url

petstore

tanaka-takayoshi:
I'm afraid the same issue is existing in typescript-node. Should I create a new issue for typescript-node or don't need to do so?

wing328:
@tanaka-takayoshi I prefer creating a separate issue for tracking.

May I know if you've cycle to contribute the fix?

timadamson:
https://github.com/swagger-api/swagger-codegen/pull/3781

timadamson:
I've create a pull request: https://github.com/swagger-api/swagger-codegen/pull/3781

wing328:
Thanks for the PR from @timadamson 

@inska please pull the latest master to give it a try.

PR #3403, opened 2016-07-19
Make typescript-angular2 generated client code customizable

conet:
This will enable clients to customize API parameters like basePath, headers, maybe response processing.

I'm not sure if this is the best approach but we found it difficult to implement a header based auth mechanism and to customize the base path use by the API.

Base path customization is necessary to run the same code on different endpoints (instances) while header based auth is also requested in #2938.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngular2ClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngular2ClientCodegen.java
@@ -67,6 +67,7 @@ public class TypeScriptAngular2ClientCodegen extends AbstractTypeScriptClientCod
         super.processOpts();
         supportingFiles.add(new SupportingFile("models.mustache", modelPackage().replace('.', File.separatorChar), "models.ts"));
         supportingFiles.add(new SupportingFile("apis.mustache", apiPackage().replace('.', File.separatorChar), "api.ts"));
+        supportingFiles.add(new SupportingFile("apiservice.mustache", apiPackage().replace('.', File.separatorChar), "ApiService.ts"));
         supportingFiles.add(new SupportingFile("index.mustache", getIndexDirectory(), "index.ts"));
         supportingFiles.add(new SupportingFile("gitignore", "", ".gitignore"));
         supportingFiles.add(new SupportingFile("git_push.sh.mustache", "", "git_push.sh"));
--- a/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
@@ -1,8 +1,9 @@
 {{>licenseInfo}}
-import {Http, Headers, RequestOptionsArgs, Response, URLSearchParams} from '@angular/http';
-import {Injectable, Optional} from '@angular/core';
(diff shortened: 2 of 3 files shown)


wing328:
cc @Vrolijkx

wing328:
@conet please run `./bin/typescript-angular2-petstore-all.sh` to update the TS Angular2 Petstore client.

Vrolijkx:
Look ok for me.

wing328:
@Vrolijkx thanks for reviewing. I believe this PR is a breaking change (non-backward compatible) so we'll merge it into `2.3.0` branch instead.

Vrolijkx:
Yes it's a breaking change so better to merge it in 2.3.0

timadamson:
We should note that it's already possible to alter the basePath by injecting the BASE_PATH OpaqueToken or by registering a factory. 

Authentication methods have been added in https://github.com/swagger-api/swagger-codegen/pull/3781

sebastianhaas:
I agree with @timadamson, setting the base path is already possible using an Opaque Token. #3781 added a configuration object, so this would be the third approach on controlling the API.

I suggest we combine forces somehow?

JohannesHoppe:
I highly suggest to switch from [`Http`](https://angular.io/api/http/Http) to [`HttpClient`](https://angular.io/guide/http) instead. This requires fewer changes in the mustache templates and brings us super power. 😄 

`HttpClient` is an evolution of the existing Angular HTTP API, which exists alongside of it in a separate package, `@angular/common/http`. It requires Angular 4.3.0-rc.0 and above, so this is a breaking change.

We could then use injectors to change the behavior. Here is an example from [netbasal.com](https://netbasal.com/a-taste-from-the-new-angular-http-client-38fcdc6b359b):

```ts
import {
  HttpRequest,
  HttpHandler,
  HttpEvent
} from '@angular/common/http';

@Injectable()
class JWTInterceptor implements HttpInterceptor {
  
  constructor(private userService: UserService) {}
  
  intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {

    const JWT = `Bearer ${this.userService.getToken()}`;
    req = req.clone({
      setHeaders: {
        Authorization: JWT
      }
    });
    return next.handle(req);
  }
}
```

JohannesHoppe:
Nevermind, I just realized that newest templates already utilize `HttpClient`... 👍

wing328:
@JohannesHoppe yup, please check it out and let us know if you've any feedback.

@conet please pull the latest master to give it a try when you've time.

JohannesHoppe:
`HttpClient` is all I need! Readme should mention that you now have to include `HttpClientModule` on your own OR the import could be directly done in the generated `ÀpiModule`, too. (see also https://github.com/swagger-api/swagger-codegen/issues/6722)
I don't think that this PR is still necessary. @conet do you agree?

wing328:
> Readme should mention that you now have to include HttpClientModule on your own OR the import could be directly done in the generated ApiModule, too.

@JohannesHoppe I wonder if you can file a PR to improve the auto-generated README when you've time: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/typescript-angular/README.mustache

JohannesHoppe:
See also #6727, I realized it's not about the README, but about a bogus import.

Thanks for the great work to maintain swagger-codegen. I really appreciate this tool and recommend it to every client.

PR #3411, opened 2016-07-19
Run petstore.swagger.io locally

wing328:
Run petstore.swagger.io locally to
- improve performance
- avoid issues with 2 or more jobs sharing the same backend (http://petstore.swagger.io)
- avoid bad data in public petstore server causing test failures


jgavris:
👍  Awesome

Issue #3418, opened 2016-07-19
version 2 basePath missing with slim server


jkt628:
##### Description

when a version 2 IDL file uses basePath, the php client generator includes basePath as part of the URL, e.g., http://localhost/basePath/foo.  unfortunately, the routing for the slim server lacks this value, processing http://localhost/foo.  a simple client built according to the README will not just communicate with the default slim server.

looking at samples/server/petstore i suspect this is a problem for:
* aspnet5
* go-api-server
* jaxrs-cxf
* lumen
* scalatra
* silex
* slim

##### Swagger-codegen version

2.1.6, 2.2.0 and master

##### Swagger declaration file content or url

```yaml
swagger: '2.0'
info:
  version: 0.0.1
  title: basePath
host: localhost
schemes:
  - http
basePath: '/basePath'
paths:
  /foo:
    get:
      operationId: getFoo
      responses:
        '200':
          description: retrieved foo
```

##### Command line used for generation

'''bash
java -jar ../swagger-codegen-cli.jar generate -l php --config swagger.json -o gen -i basePath.yaml
java -jar ../swagger-codegen-cli.jar generate -l slim -o gen -i basePath.yaml
'''

##### Steps to reproduce

extract [basePath.zip](https://github.com/swagger-api/swagger-codegen/files/372381/basePath.zip).

on one tty:
'''bash
make server
'''

on another tty:
'''bash
make client
'''

##### Related issues

#1425

##### Suggest a Fix

with #1439 put {{basePathWithoutHost}} in appropriate places.

wing328:
@jkt628 thanks for fixing the issue, which should be fixed by #3418. Please pull the latest master to give it a try.

wing328:
@jkt628 and thanks for checking other server stub generator. We'll fix them one by one.

wing328:
@abcsun when you've time, can you help fix **PHP Lumen generator** with the context path (e.g. /v2/) ? #3418 is a good reference.

wing328:
@guohuang when you've time, can you help fix **Go API server generator** with the context path (e.g. /v2/) ? #3418 is a good reference.

jkt628:
@wing328 @abcsun thanks for the quick fixes

wing328:
@jkt628 for jaxrs-cxf, it does not come with `src/main/webapp/WEB-INF/web.xml` so I think the developers need to set it in their web server's configuration.

wing328:
[...]

PR #3422, opened 2016-07-20
[PHP][Slim] Add missing context path

wing328:
Add missing context path (e.g. /v2/) to index.php in PHP Slim


wing328:
Linked to #3418

PR #3438, opened 2016-07-22
Fix ExampleGenerator

vfpfafrf:
Fixes several issues in ExampleGenerator:

-  Incorrect float & double generation (values  taken from DecimalProperty)

Definition:
```
 "Test" : {
      "type" : "object",
      "properties" : {
        "pFloat" : {
          "type" : "number",
          "format" : "float"
        },
        "pDouble" : {
          "type" : "number",
          "format" : "double"
        }
      }
    }
```
Result:
````
examples['application/json'] = {
  "pFloat" : 1.3579000000000001069366817318950779736042022705078125,
  "pDouble" : 1.3579000000000001069366817318950779736042022705078125
};
```

-  Incorrect object example generation as string, not an object:

Definition:
```
 "Test" : {
      "type" : "object",
      "properties" : {
        "obj" : {
          "type" : "object"
        }
      }
    }
```

Result:
```
examples['application/json'] = {
  "obj" : "{}"
};
```

-  ComposedModel generates empty example result

Definition:
```
    "Test" : {
      "allOf" : [
        {
          "$ref" : "#/definitions/TypeOne"
        },
        {
          "$ref" : "#/definitions/TypeTwo"
        }
      ]
    },
    "TypeOne" : {
      "type" : "object",
      ...
    },
    "TypeTwo" : {
      "type" : "object",
      ...
    }
```

Result:
```
  examples['application/json'] = "";
```

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/examples/ExampleGenerator.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/examples/ExampleGenerator.java
@@ -98,7 +98,7 @@ public class ExampleGenerator {
             }
             return mp;
         } else if (property instanceof ObjectProperty) {
-            return "{}";
+            return Collections.EMPTY_MAP;
         } else if (property instanceof RefProperty) {
             String simpleName = ((RefProperty) property).getSimpleRef();
             Model model = examples.get(simpleName);
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/examples/ExampleGenerator.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/examples/ExampleGenerator.java
@@ -2,22 +2,7 @@ package io.swagger.codegen.examples;
 
 import io.swagger.models.Model;
 import io.swagger.models.ModelImpl;
-import io.swagger.models.properties.ArrayProperty;
-import io.swagger.models.properties.BooleanProperty;
-import io.swagger.models.properties.DateProperty;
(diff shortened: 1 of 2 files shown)


wing328:
@vfpfafrf thanks for the PR. I noticed that you've changed some methods from public/protected to private. Can we keep those as public/protected as there are use cases in which the developers customize the generator?

vfpfafrf:
@wing328 I rollback method modifies to public/protected. May be it should be interface with abstract  and default generator implementations ?

wing328:
> May be it should be interface with abstract and default generator implementations ?

Yes, I think that's better as the same Example Generator cannot meet the syntax requirement of **all** languages (e.g. C# float value needs to end with "f", e.g. `1.37f`)

vfpfafrf:
> C# float value needs to end with "f", e.g. 1.37f

I dont think this is a problem  - example objects use deserialization from JSON-string -  from swagger-codegen/modules/swagger-codegen/src/main/resources/aspnet5/mapReturn.mustache (or from listReturn.mustache)

```
var example = exampleJson != null
            ? JsonConvert.DeserializeObject<Dictionary<{{#returnType}}{{{returnType}}}{{/returnType}}>>(exampleJson)
            : new Dictionary<{{#returnType}}{{{returnType}}}{{/returnType}}>();
```

But the problem is in  swagger-codegen/modules/swagger-codegen/src/main/resources/aspnet5/controller.mustache where exampleJson declared as null-object, so result is always empty:

```
string exampleJson = null;
```

I don't have experience in C#, so i can't fix it in right way

PR #3440, opened 2016-07-22
dart codegen improvements

### LINKED ISSUE #3334: dart codegen: add a vm bash script and proper argument for dart vm creation
see pull request #3325

I intend to implement this.  Just thought a reference would be good.

close2:
generated library has better exceptions, type conversion is better (/fixed), dart strong mode is enabled, petstore test is refactored and now looks like a real project (generated lib in a different directory) (this more or less closes #3334)

I have not split the test into 2 projects for browser and vm because dart will support creating libraries which support browser and vm at the same time soon.


wing328:
@close2 thanks for the PR ( 👍 to comments in the code and explanation of the change)

Issue #3441, opened 2016-07-24
[csharp] RootNamespace and AssemblyName are incorrect for Project.mustache


jimschubert:
##### Description

Client generated assembly uses assembly name and default namespace of `Swagger Library` rather than `IO.Swagger` (as the project is defined). This breaks from common C# convention and causes problems when adding new files (tested in Xamarin, VS2015, Rider EAP).

##### Swagger-codegen version

2.2.0/master

##### Command line used for generation

n/a, sample code can be used for evaluation

##### Steps to reproduce

1. Open `samples/petstore/csharp/SwaggerClient/IO.Swagger.sln`
2. Build the solution
3. Inspect output artifacts
    - Notice: `bin/Debug/Swagger Library.dll`
    - Expected: `bin/Debug/IO.Swagger.dll`
4. Add a new file to the `IO.Swagger` project in your IDE of choice. 
    - Notice: `namespace Swagger Library {`
    - Expected: `namespace IO.Swagger {`

##### Related issues

n/a

##### Suggest a Fix

Change `packageTitle` in `Project.mustache` to `packageName`. The `packageTitle` property is intended for assembly metadata, not root namespace and assembly filename.

jimschubert:
This is a simple fix to a problem I just encountered. Since I've already fixed it locally, I'll build and open a PR.

wing328:
PR merged. Issue closed.

PR #3442, opened 2016-07-24
[C#] RootNamespace/AssemblyName use packageName

jimschubert:
Uses packageName instead of packageTitle for `RootNamespace` and
`AssemblyName` in the C# generate client project. packageTitle is intended
to add additional information to AssemblyInfo.cs, and changing
packageTitle to non-FQN text breaks adding new files to the project.

packageTitle, per CodegenConstants description:

> Specifies an AssemblyTitle for the .NET Framework global assembly attributes stored in the AssemblyInfo file.

see #3441


wing328:
@jimschubert thanks for the PR, which has been merged into master.

Shall we do the same for NancyFx? https://github.com/swagger-api/swagger-codegen/blob/61015488d2911bfaba9aa9277072112d1c48e81f/modules/swagger-codegen/src/main/resources/nancyfx/Project.mustache#L10

cc @mstefaniuk

jimschubert:
Yeah, I think it would make more sense to follow the practice of naming assemblies consistently with the namespace rather than some freeform text. For some reason, I thought NancyFX was written for .NET Core, so I didn't think to check its project file.

wing328:
OK. I'll fix it in my work-in-progress C# PR.

wing328:
For NancyFX, I've submitted https://github.com/swagger-api/swagger-codegen/pull/3449

Issue #3459, opened 2016-07-26
[Go] [Bug] Check for required parameters is invalid


neilotoole:
##### Description

The generated code for API calls with required parameters is largely invalid. There's a few (bad) different things going on. Let's take a look. Here's a generated snippet:

```go
func (a ProjectApi) CreateProject(project Project) (*Project, *APIResponse, error) {

	var httpMethod = "Post"
	// create path and map variables
	path := a.Configuration.BasePath + "/projects"

	// verify the required parameter 'project' is set
	if &project == nil {
		return new(Project), nil, errors.New("Missing required parameter 'project' when calling ProjectApi->CreateProject")
	}
```
Note that structs are passed by value, not reference (otherwise it would be `CreateProject(project *Project)`. Therefore the check `if &project == nil {` is meaningless... the supplied `project` parameter will, at a minimum, be the zero value for that struct type, therefore `&project` can never evaluate to `nil`.

Let's take another example:

```go
func (a ArtifactApi) GetArtifacts(executionId int32) ([]Artifact, *APIResponse, error) {

	var httpMethod = "Get"
	// create path and map variables
	path := a.Configuration.BasePath + "/artifacts"

	// verify the required parameter 'executionId' is set
	if &executionId == nil {
		return new([]Artifact), nil, errors.New("Missing required parameter 'executionId' when calling ArtifactApi->GetArtifacts")
	}
```
In this case, this code is invalid Go: `if &executionId == nil {`.

Additionally, this is invalid: `return new([]Artifact) ...`, as it returns a pointer to a slice (`*[]Artifact`), instead of a slice .

##### Swagger-codegen version

`master`, but I suspect it's been around for a while.

##### Swagger declaration file content or url

Unfortunately I can't share the specific swagger, but at a minimum, this should happen with any swagger operation that takes a required parameter that is an `int`, or a `struct`, or returns an array.

##### Suggest a Fix

Validating required parameters is non-trivial in Go, due to the nature of the language. The basic outline of a solution is to use pointers everywhere, or to use interfaces. Both of these would require significant rework and any such changes should be planned and designed carefully. 

Currently the generator is completely broken: we need to remove the section of code that checks for required parameters, and revisit the whole topic. I create a PR shortly.

Also, we need to have automated testing of generated Go clients. The PRs that introduced these breaking changes should have been caught (just by the Go compiler alone).

wing328:
@neilotoole thanks for the details and the PR for the fix.

[...]

PR #3460, opened 2016-07-26
Issue #3459 removed check for required params

neilotoole:
This addresses #3459 by removing the check for required params.


neilotoole:
I haven't looked in detail, but it's possible the TravisCI failure was not due to the PR itself.

```bash
$ docker pull swaggerapi/petstore
Using default tag: latest
latest: Pulling from swaggerapi/petstore
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

wing328:
@neilotoole Thanks for the PR. I've restarted the build and all tests passed. (I think the issue is with Docker Hub)

What would be the right way, if any, to check required parameters in Go?

PR #3471, opened 2016-07-27
[JAVA] Add fluent methods to add items to list and put items to maps

cbornet:
So you can do
```java
Pet myPet = new Pet()
    .addPhotoUrlsItem(firstUrl)
    .addPhotoUrlsItem(secondUrl);
```


cbornet:
@wing328 I only generated jersey1 because that should be enough for CI check. Do you want me to also generate the other samples ?

PR #3475, opened 2016-07-28
support for hierarchies with more than one level, fixes #3474

### LINKED ISSUE #3474: Child in hierarchy of more than one level doesn't inherit from its parent
##### Description

When you create a hierarchy with more than one level the generated java code of the child doesn't inherit (extends) from its parent

##### Swagger-codegen version
2.2.0

##### Swagger declaration file example
```
swagger: '2.0'
info:
  title: Enum Test
  description: Test enum inside array
  version: "1.0.0"
host: localhost:8080
schemes:
  - http
  - https
consumes:
  - "application/json"
produces:
  - "application/json"

paths: {}
  
definitions:

  example:
    type: object
    discriminator: example_type
    properties:
      id:
        type: string
      name:
        type: string
      example_type:
        type: string
        enum:
        - subexample
        - subsubexample
  subexample:
    type: object
    allOf:
    - $ref: '#/definitions/example'
    - type: object
  subsubexample:
    type: object
    allOf:
    - $ref: '#/definitions/subexample'
    - type: object
```

##### Hierarchy generated classes 

```
public class ExampleModel {
    // ...
}
```
```
public class SubexampleModel extends ExampleModel {
    // ...
}
```
```
public class SubsubexampleModel {
    // ...
}
```

##### Error detected
SubsubexampleModel should extends from SubexampleModel

##### Suggested fix

First of all the parent property in a ComposedModel isn't set anymore in the parser. That's because of this issue and commit:
https://github.com/swagger-api/swagger-parser/issues/245
https://github.com/swagger-api/swagger-parser/pull/246/commits/be6f0687645829c78e8126b64b3f20e576e0f36f

Now you define the parent in swagger-codegen inside **DefaultCodegen.fromModel** method.
You set as the parent the first "interface" that has a discriminator. But in the case of a hierarchy with more than one level the discriminator would be in the root of the hierarchy and this condition would be false.
I change the code to first search for a parent with discriminator and if it does not exist, then I set as the parent the first of the ComposedModels inside the "interfaces".
I will add a pull request with my solution.

Thanks!
Regards,

Adrian

fiore-adrian:


wing328:
@fiore-adrian thanks for the PR. Can you also share a spec for reproducing the issue?

fiore-adrian:
@wing328 I have opened an issue for this PR. https://github.com/swagger-api/swagger-codegen/issues/3474

wing328:
@fiore-adrian sorry I missed that. Will look into that tomorrow.

wing328:
cc @cbornet

arnested:
This fixes related problems in the PHP client. If @cbornet doesn't see any problems in this I'll suggest we merge it.

cbornet:
The problem is that this PR would treat composition as inheritance which is wrong. Also in the issue definition it's not clear if subsubmodel composes submodel or inherits from it. I think it's an issue with the swagger spec itself.
Maybe @fehguy can give its opinion ?

cbornet:
Imagine the following : B1 has A1 with a discriminator in its allOf. B2 has A2 with a discriminator in its allOf. C has B1 and B2 in its allOf. So shall C extend B1, B2 or neither of them ?

cbornet:
The solution here would be to look recursively into the interfaces tree for the discriminator field and set the parent accordingly if any.

Issue #3410, opened 2016-07-19
[Qt5/C++] Arrays of primitive types not working


philicious:
##### Description

this swagger

```
  Product:
    type: object
    properties:
      categories:
        type: array
        description: Category Ids
        items:
          type: integer
```

results in 

```
//.h
QList<qint32>* categories;  // ok !
//.cpp
obj->insert("categories", QJsonValue(categories));  // bad
```

instead of obj->insert.. it should use `toJsonArray` from SWGHelpers

Its working fine if the arrays items are SWG\* objects
##### Swagger-codegen version

latest from git
##### Command line used for generation

`java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -t modules/swagger-codegen/src/main/resources/qt5cpp -i ../retailers.yaml -l qt5cpp -o ../generated-retailer`
##### Steps to reproduce

generate an SDK based on a spec with primitive type in array
##### Suggest a Fix

model-body.mustache template needs to be adjusted. I lack swagger-codegen internals knowledge

wing328:
> instead of obj->insert.. it should use toJsonArray from SWGHelpers

@philicious If you can provide what the exact code looks like, I can create a PR based on that.

philicious:
@wing328 unfortunately I cant tell you the exact code, or I would have PR'ed myself. afair it will require editing https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/qt5cpp/model-body.mustache#L71-L92  which is where I was lacking understanding of all these `{{^isContainer}}{{#complexType}}` etc and how they relate/derive to/from the yaml

wing328:
{{#isContainer}} .. {{/isContainer}} means the parameter or property is either an array or a map.

{{#complexType}} .. {{/complexType}} means it's a model (non-primitive type).

`^` is for inverted section.

philicious:
@wing328 ok I see. However I've meanwhile moved on to the next project and probably wont have the chance developing a fix. However I could test changes if you (or anyone) will pick this up

wing328:
@philicious that's ok. We'll see if anyone from the community is interested in contributing the fix. Thanks again for reporting the issue.

marache:
Hi,
I can confirm the issue, it's not in the mustache file that the issue is located. An array of simple types don't get "labelled" as ComplexType thus failing to generate the correct code path. I can have a look and see if I can find out why...

[...]

PR #3476, opened 2016-07-28
Qt5: Added missing datatype handlers for QByteArray for swagger-type "byte"

philicious:
Now you can have sth like this in your swagger and it will work with Qt5 SDK

```
  ProductObject:
    type: object
    properties:
      image:
        type: string
        format: byte
```

the property will be base64 encoded/decoded to/from JSON


wing328:
@philicious thanks for the PR to improve the Qt5 C++ generator.

I noticed that you described the image as `type: string, format: byte` (base64-encoded string). Should it be better described as `type: string, format:binary` instead?

philicious:
@wing328 actually no. Binary would be an octet stream. And I think you cannot wrap any octet / byte in JSON as that's a string. So you need encoding

wing328:
@philicious thanks for the swift response and PR merged.

wing328:
@philicious for your next PR, please also run `./bin/qt5-petstore.sh` to update the Qt5 C++ Petstore sample.

philicious:
oh right @wing328 , will do. I forgot about it since my last PR :innocent: 

btw I think its only #3410 missing now for fully implementing swagger spec datatypes in Qt5 generator.

PR #3479, opened 2016-07-29
Confluence Wiki Documentation

jhitchcock:
Addition of a generator to create confluence wiki markup which can then be pasted into confluence to generate docs as per image

cwiki


wing328:
@jhitchcock thanks for the new generator, which has been merged into master.

PR #3480, opened 2016-07-29
[PHP] Client properties validation #3457

### LINKED ISSUE #3457: [PHP] [BUG] Client properties validation
##### Description

When you generate a PHP Client and you try to get data from a resource that has a property with a minLength it throws an exception if the value is null. The problem is that this behaviour is right if the property is required, but if it is not required the check is the same so it cannot be null and it becomes required when it isn't.

##### Swagger-codegen version

api version: 2.2.0-SNAPSHOT

    https://gist.github.com/joskfg/96633d30a8788bde414c1b77d696fdaa

Used online generator.
```json
{
    "swaggerUrl": "http://localhost/swagger.json",
    "options": {
        "modelPackage": "Company\\Package\\Model",
        "apiPackage": "Company\\Package\\Api",
        "invokerPackage": "Company\\Package",
        "packagePath": "Package",
        "srcBasePath": "src",
        "composerVendorName": "company",
        "composerProjectName": "package",
        "artifactVersion": "1.0.0"
    }
}
```

##### Steps to reproduce

Create a PHP Client for a given resource with non-required properties with minLength and maxLength. When you'll retrieve data from the resource if the non-required parameter is null you'll get an error telling that it needs to have al least N chars (the defined in the swagger), but it is not required so it should allow null values.

##### Related issues

Not found

##### Suggest a Fix

If the properties are not required the validations should be applied if there is some value in the property, if it is null, the property shouldn't be checked.

arnested:
Fixes #3457.


arnested:
Could use a test case testing this though.

wing328:
linked to https://github.com/swagger-api/swagger-codegen/issues/2663

akira28:
Sorry to inform this is bugged :(
generated models have syntax errors like these:
```
if (!is_null(${{$this->container['id_compliance_scan']}}) && (strlen($this->container['id_compliance_scan']) < 36)) {
    $invalid_properties[] = "invalid value for 'id_compliance_scan', the character length must be bigger than or equal to 36.";
}
```

arnested:
@akira28, you are right. Don't know how I managed to do that. And don't know how I managed not to catch that.

Fix ready in a few minutes.

arnested:
See #3493 for the fix. Thanks for catching this, @akira28.

wing328:
@arnested thanks for the quick fix.

@akira28 please pull the latest master to give it another try.

akira28:
I'll try it and let you know. Thanks for the quick fix!

akira28:
@arnested @wing328 seems to work correctly now. Thank you very much

arnested:
Thank you very much for reporting back, @akira28!

PR #3487, opened 2016-07-31
fixing bug of rendering an extra spaces into @FeignClient annotation

szantopeter:
FeignClient annotations are redered incorrectly with an extra space. This causes a problem that spring properties are not loaded. E.g: the below annotation was generated

```@FeignClient(name="${ swaggerPetstoreSimple.name:swaggerPetstoreSimple}", url="${ swaggerPetstoreSimple.url:https://localhost:8080/api}", configuration = ClientConfiguration.class)```

and overwriting 

swaggerPetstoreSimple.name:swaggerPetstoreSimple

was not possible because of the extra spaces.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/JavaSpring/libraries/spring-cloud/apiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/JavaSpring/libraries/spring-cloud/apiClient.mustache
@@ -3,6 +3,8 @@ package {{package}};
 import org.springframework.cloud.netflix.feign.FeignClient;
 import {{configPackage}}.ClientConfiguration;
 
-@FeignClient(name="${ {{{title}}}.name:{{{title}}}}", url="${ {{{title}}}.url:{{{basePath}}}}", configuration = ClientConfiguration.class)
+{{=<% %>=}}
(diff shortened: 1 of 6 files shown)


cbornet:
Can you also regenerate the spring-cloud sample ?

wing328:

szantopeter:
@wing328 thanks for spotting, I fixed my profile so commits should be linked to it
@cbornet I updated the spring/feign examples and added instructions how to generate them (it took me a while to find the shell scripts)

wing328:
@szantopeter thanks for the feedback. We've documented Petstore sample update procedure here: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md#testing but seems like not many people notice it.

We'll probably create a PR template to draw people's attention about updating Petstore sample so that CIs can verify the change.

szantopeter:
@wing328 I think most people (users of SWAGGER) will not be as much interested in how to generate the samples, rather what options they need to specify to generate a client from their SWAGGER files using the same technologies. In some cases this is not so obvious because it is not just the language has to be set but also additional options, libraries, etc. Maybe it would worth adding this information to all samples (I only added it to two of them) By the way I read the how to contribute page, but it seems I missed the relevant part at the bottom... ;)

PR #3488, opened 2016-07-31
[android] Make all models implement 'Serializable'

dylanmckay:
This allows the models to be passed to `Activity` objects when creating them,

### DIFF
--- a/modules/swagger-codegen/src/main/resources/android/libraries/volley/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/android/libraries/volley/model.mustache
@@ -13,7 +13,7 @@ import com.google.gson.annotations.SerializedName;
  * {{description}}
  **/{{/description}}
 @ApiModel(description = "{{{description}}}")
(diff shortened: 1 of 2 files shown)


cbornet:
In java generators this is an option.

dylanmckay:
It doesn't look like this is supported by Android, is that correct?

If not, I can add it.

wing328:
@dylanmckay yes please. Here is the related-code: https://github.com/swagger-api/swagger-codegen/search?l=java&q=serializableModel&utf8=%E2%9C%93

and then you can use the `{{#serializableModel}} ... {{/serializableModel}}` tag in the template :https://github.com/swagger-api/swagger-codegen/blob/48bdae8fe798742c82f2016650ebdef22889d4b4/modules/swagger-codegen/src/main/resources/Java/model.mustache#L10

wing328:
@dylanmckay may I know if you need any help adding the option?

wing328:
Add ed via https://github.com/swagger-api/swagger-codegen/pull/3933 instead.

PR #3494, opened 2016-08-01
NodeJS generator - Harken to the the warnings ...

osher:
2 types of warnings are emitted in bulk whenever using the nodeJS generator:

```
[main] WARN io.swagger.codegen.DefaultCodegen - escapeUnsafeCharacters should be overriden in the code generator with proper logic to escape unsafe characters
```
and
```
[main] WARN io.swagger.codegen.DefaultCodegen - escapeQuotationMark should be overriden in the code generator with proper logic to escape single/double quote
```

I did some guesswork: since JS is a c-type language like C# and Java - I looked there for reference, and followed their example.


osher:
uhhh....
how do I label it as `Server: Nodejs` ?

wing328:
@osher the change looks good to me. Thanks for the contribution.

PR #3391, opened 2016-07-18
Feature/objc tasks

wberger:
I propose a change to the objc ApiClient. Instead of internally created and maintained request IDs the API implementation should return NSURLSessionTasks. This change has two advantages:
- Less internal state
- These task object provide real cancellation as compared to the request ID based implementation (even if the task is currently running)

Furthermore I changed the configuration object to be protocol based in order to allow for more flexible integration into projects and more complex configuration setups. A default configuration implementation is provided and can be passed in, if no custom implementation is required.

Finally I renamed the ApiClient to ApiSessionManager to go with the AFNetworking wording.

PR #3510, opened 2016-08-03
Feature/objc tasks

wberger:
As agreed in PR #3391 I created a new PR for branch 2.3.0.

I have updated the integration tests and documentation templates to conform to the breaking changes.

Further changes compared to PR #3391:
- Rename ApiSessionManager back to ApiClient. This is also a better naming in the Swagger context
- Create a default initializer for the Api classes


wing328:
@wberger thanks for the PR

Did you try to do a rebase? There seems to be an issue as it contains 70+ commits.

cc @mateuszmackowiak

wberger:
Something is wrong with the rebase, some commits need extra merging. Thus I created a new branch from 2.3.0 and build a completely new git history. I will create new PR.

wberger:
I have created a new, clean PR: #3522

This should be the final one ;)

PR #3514, opened 2016-08-03
inject base_path opaque token for typescript angular2

### LINKED ISSUE #3513: [typescript-angular2] baseUrl injectable
##### Description

The constructors for the Api classes in typescript-angular2 currently have the definition:
```typescript
constructor(protected http: Http, @Optional() basePath: string)
```
To be able to provide a value and have DI use this value, would it be better to set this up as:
```typescript
constructor(protected http: Http,  @Optional()@Inject(BASE_PATH) basePath: string)
```
And define a variables file with:
```typescript
export const BASE_PATH = new OpaqueToken('basePath');
```
And then this could be utilised on app bootstrap by:
```typescript
Import { BASE_PATH } from './service/variables';
bootstrap(AppComponent, [
    { provide: BASE_PATH, useValue: 'http://myservice:5000' }
]);
```

##### Swagger-codegen version
2.2.0

I'd be happy to make the contribution via a PR, just checking that there is no prior agreed way. And that this would be an appropriate way to handle this case?

damienpontifex:
Enable setup of base path for apis in angular2 bootstrap by providing the value to resolve #3513 

Need verifying I haven't missed anything as first code contribution to swagger-codegen and building the project from source

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngular2ClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngular2ClientCodegen.java
@@ -68,6 +68,7 @@ public class TypeScriptAngular2ClientCodegen extends AbstractTypeScriptClientCod
         supportingFiles.add(new SupportingFile("models.mustache", modelPackage().replace('.', File.separatorChar), "models.ts"));
         supportingFiles.add(new SupportingFile("apis.mustache", apiPackage().replace('.', File.separatorChar), "api.ts"));
         supportingFiles.add(new SupportingFile("index.mustache", getIndexDirectory(), "index.ts"));
+        supportingFiles.add(new SupportingFile("variables.mustache", getIndexDirectory(), "variables.ts"));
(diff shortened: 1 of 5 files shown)


wing328:
@damienpontifex thanks for the PR.

cc @Vrolijkx

wing328:
@damienpontifex is it correct to say that this is a breaking change (non-backward compatible)?

If yes, please re-submit the PR for `2.1.3` branch instead.

damienpontifex:
@wing328 I don't believe it to be a breaking change as it just provides another avenue for passing in the configuration value. 
I would be happy to have more feedback from others who are using it if it is actually causing a breaking change.

Vrolijkx:
Nice addition to the angular2 api. Doesn't seem a breaking change to me. should be completely backward compatible.

damienpontifex:
@wing328 @Vrolijkx just quickly added a note in the readme of how to set the custom base path. Let me know if there is any other sort of documentation we would normally add to these.

martinmcwhorter:
I just ran into this missing OpaqueToken issue today. Im glad to see there is a pull request and hopefully this will be be in a release soon enough.

wing328:
@damienpontifex thanks for the PR, which has been merged into master.

PR #3516, opened 2016-08-04
Add missing spring-cloud client to Swagger Editor

szantopeter:
Swagger Editor is not listing spring cloud among the available clients. This is because Swagger editor reads the available clients from the backend. Code generators are are categorized as either Server or Client. The spring code generator is categorized as Server therefore it won't be listed as a client. I created a new code generator only so it can be registered as Client. This is clearly a workaround, but a proper solution would require a bigger refactor and I am not sure it worth the effort. It also worth mentioning that even though the list of available languages are coming from the backend, the human readable name is coded into Swagger Editor's

```
$scope.capitalizeGeneratorName
```

method. This requires changes in two modules when a new code generator is created. This pull request solves the issue for spring-cloud client, but it wont work for java/feign or spring/stubs because it requires additional parameters, not just language. I think a proper solution would be that every code generator returns a list of available server and client implementations along with all options, then the two apis

http://generator.swagger.io/api/gen/clients
http://generator.swagger.io/api/gen/servers

should return something like

[{
    "displayName": "Java spring cloud client",
    "options": {
        "l": "spring"
    }
}, {
    "displayName": "Java spring stubs",
    "options": {
        "l": "spring",
        "D": "DinterfaceOnly=true"
    }
}]

This is either a breaking change or a new endpoint, or versioning, or an extra parameter needs to be introduced.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SpringCloudCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SpringCloudCodegen.java
@@ -1,0 +1,27 @@
+package io.swagger.codegen.languages;
+
+import io.swagger.codegen.*;
+
+public class SpringCloudCodegen extends SpringCodegen {
+
+    public SpringCloudCodegen() {
+        super();
+        setLibrary(SPRING_CLOUD_LIBRARY);
+    }
+
+    @Override
+    public CodegenType getTag() {
+        return CodegenType.CLIENT;
+    }
+
+    @Override
+    public String getName() {
+        return "spring-cloud";
+    }
+
+    @Override
+    public String getHelp() {
+        return supportedLibraries.get(SPRING_CLOUD_LIBRARY);
+    }
+
+}
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SpringCloudCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SpringCloudCodegen.java
@@ -2,6 +2,13 @@ package io.swagger.codegen.languages;
 
 import io.swagger.codegen.*;
 
+/**
+ * This is a temporary workaround, so spring cloud can be listed by
+ *
+ * http://generator.swagger.io/api/gen/clients
+ *
+ * as a client and eventually Spring cloud can be displayed by the swagger editor' generate client menu.
+ */
 public class SpringCloudCodegen extends SpringCodegen {
 
     public SpringCloudCodegen() {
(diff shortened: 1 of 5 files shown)


szantopeter:
@wing328 @cbornet any thought on this?

cbornet:
See https://github.com/swagger-api/swagger-editor/issues/713

wing328:
Agreed with @cbornet that the feature should be added in the editor instead of codegen.

szantopeter:
If I read swagger-api/swagger-editor#713 correctly then the proper solution is already on it's way, so should this PR be closed? @wing328 I am slightly confused what should be added to the editor?

wing328:
@szantopeter currently, there's no way you can customize the API client/server stub generation in editor.swagger.io. I believe swagger-api/swagger-editor#713 refers to a dialog window to allow users passing various options via JSON payload in the requests to generator.swagger.io.

szantopeter:
@wing328 the solution I provided in this PR will display a Spring cloud option under the client menu in swagger editor, should it be useful until the redesign is implemented?

wing328:
@szantopeter first of all, the community and I truly appreciate your contribution but even though we merge this, the swagger editor team still needs to make changes in their side in order to pick up new generators in the front end and they also need to update generator.swagger.io to use the latest stable version (not latest master) of swagger codegen so we still need to wait for a while for changes in the latest master to be visible in editor.swagger.io and generator.swagger.io.

I believe you're aware that you can generate API clients or server stubs online as described here: https://github.com/swagger-api/swagger-codegen#online-generators

szantopeter:
Hi @wing328 

Swagger editor calls http://generator.swagger.io/api/gen/clients to generate the list of available clients. My PR adds spring-cloud to the list. 

spring-cloud-swagger-api

Swagger editor will add it to the "generate client" menu as "Spring cloud"

spring-cloud-swagger

Screenshots are generated with my modified swagger-codegen and unmodified swagger editor, so I belive there is no change required in swagger editor. (I generated a number of spring cloud clients from swagger editor and they all work fine) As you pointed out my PR will take effect only if swagger-codegen is released and deplyoed. If swagger-api/swagger-editor#713 will be included in the next release then it is a better solution. If that won't be included then my PR may have benefits, but I am perfectly fine either way. Personally I am not blocked without this PR because I generate clients with the maven plugin.

wing328:
@szantopeter thanks for the info. Seems like swagger editor has a lot of enhancements to it since I last checked it a few months ago.

wing328:
Closing this as it should be done in Swagger Editor. Please continue the discussion in https://github.com/swagger-api/swagger-editor/issues/713

Issue #2487, opened 2016-04-01
Enumerations with non-alphanumeric characters


degroves:
[JSON spec and generated Java code](https://gist.github.com/degroves/4e1af1a72779c59b60ef42c731c61cbd)

I generated java code using 2.1.5:

java -jar swagger-codegen-cli.jar generate -i enum_example.json -l java -o target

Unless I've misinterpreted the JSON schema specification (json-schema.org/latest/json-schema-validation.html), an enumeration of type string can have any Unicode character in the enumeration name.  While using mathematical symbols is unusual, they are Unicode characters and so the swagger code generator should generate Java code which can compile for this enumeration.

wing328:
Welcome to GitHub and thanks for reporting the issue. For your case, what's the expected `MyEnum.java` look like from your perspective?

sbelan:
@wing328 .. similarly, if ENUMS are case sensitive... Swagger always creating enum with UPPERCASE
e.g:
In my case server is expecting value as Success and Fail, so I mentioned the same in swagger specification but during client creation swagger always created it in uppercase like SUCCESS and FAIL.
I did quick fix for this but let me know your advice on this.

wing328:
@sbelan please open a separate issue for tracking and provide more information such as spec to reproduce the issue and which language(s) you're referring

degroves:
@wing328: Something along these lines:



    public enum MyEnum
    {
        LESS_THAN("<"), NOT_EQUALS("!="),  EQUALS("=");

        private final String description;

        MyEnum(final String description)
        {
            this.description = description;
        }

        public static void MyEnum getByDescription(final String description)
        {
            // Some efficient mapper here
        }
    
    }


This might also solve the problem @sbelan described.  You could have SUCCESS map to "Success" and FAIL map to "Fail".

The problem is how to come up with a generalized mapping.  I don't have a solution for that.  There's a lot of characters in the Unicode character set.  You might need some user input to come up with a good mapping.

sbelan:
@degroves .. we resolved this issue by overriding postProcessModels() from JavaClientCodegen and changed logic for enum during inserting into enum map.

degroves:
@sbelan, thanks for the pointer, I might do the same.

wing328:
@degroves #3519 (for Java-related generators) merged into master to provide the mapping between symbol and symbol name. 

Please pull the latest to give it a try.

wing328:
[...]

PR #3519, opened 2016-08-04
[Java] better enum naming for symbol

wing328:
For https://github.com/swagger-api/swagger-codegen/issues/2487

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -106,7 +106,7 @@ public class DefaultCodegen {
     // How to encode special characters like $
     // They are translated to words like "Dollar" and prefixed with '
     // Then translated back during JSON encoding and decoding
-    protected Map<Character, String> specialCharReplacements = new HashMap<Character, String>();
+    protected Map<String, String> specialCharReplacements = new HashMap<String, String>();
 
     public List<CliOption> cliOptions() {
         return cliOptions;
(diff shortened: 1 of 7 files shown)


sebciok:
Looking at the changelist there may be a copy/paste error:

       specialCharReplacements.put(">=", "Greater_Than_Or_Equal_To");
       specialCharReplacements.put("!=", "Greater_Than_Or_Equal_To");

Also, would it be possible to do it for . (dot/period) as well?

That's the case I ran into (we have an enumeration with both . (dot) and _ (underscore) and they are both being translated to _ (Java codegen). It would be really nice if underscore on its own would be translated to UNDERSCORE. 

Is there any guess when the new version with this fix would come out?

wing328:
@sebciok fixed via https://github.com/swagger-api/swagger-codegen/pull/3531

sebciok:
@wing328 Fantastic, Thank you.

wing328:
@sebciok thanks again for reporting the issue. Please pull the latest to help confirm it addresses the issue for you as there're quite a few enum-related enhancements and bug fixes. Although we've test cases but still we our users to do a test if possible.

sebciok:
I got latest, built codegen and was able to successfully generate my enumeration. Thank you, works great. 

I run into an unrelated issue where generated code could not resolve JsonCreator annotation, but it could be something on my side. 

If we want to handle more special characters (I do not need them, but just for completeness), I found that those ones are still missing:

QUESTION_MARK ?
COMMA ,
QUOTE '
DOUBLE_QUOTE "
SLASH /
BACKSLASH \
LEFT_PAREN (
RIGHT_PAREN )
LEFT_BRACE {
RIGHT_BRACE }
LEFT_BRACKET [
RIGHT_BRACKET ]
TILDE ~
BACKTICK `

wing328:
> I ran into an unrelated issue where generated code could not resolve JsonCreator annotation, but it could be something on my side.

Which HTTP library did you use? The default one is `okhttp-gson`, which does not use JsonCreator annotation.

For the remaining special characters, we'll add those later.

sebciok:
I just used a default generated maven project without any changes, so it looks like I was using com.squareup.okhttp. I will try generation, on another, larger swagger definition.

wing328:
Thanks for the info.

FYI. Other symbols mapped via https://github.com/swagger-api/swagger-codegen/pull/3546

wing328:
@sebciok thanks for the info. I'm able to repeat the issue and filed a fix: https://github.com/swagger-api/swagger-codegen/pull/3551

Issue #2766, opened 2016-05-03
Underscore prefix dropped from property names in generated Node Typescript classes


webron:
_From @saikatguha on April 29, 2016 22:9_

```
"Foo": {
  "type": "object",
  "properties": {
    "_links": { "$ref": "#/definitions/Bar" }
  }
}
```

becomes 

```
class Foo {
  "links": Bar;
}
```

_Copied from original issue: swagger-api/swagger-editor#905_

wing328:
@saikatguha to resolve it, you need to set `modelPropertyNaming` to `original`:
```
swagger-codegen|master ⇒ java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar config-help -l typescript-node


CONFIG OPTIONS
	sortParamsByRequiredFlag
	    Sort method arguments to place required parameters before optional parameters. (Default: true)

	ensureUniqueParams
	    Whether to ensure parameter names are unique in an operation (rename parameters that are not). (Default: true)

	modelPropertyNaming
	    Naming convention for the property: 'camelCase', 'PascalCase', 'snake_case' and 'original', which keeps the original name (Default: camelCase)

	npmName
	    The name under which you want to publish generated npm package

	npmVersion
	    The version of your npm package

	npmRepository
	    Use this property to set an url your private npmRepo in the package.json

	snapshot
	    When setting this property to true the version will be suffixed with -SNAPSHOT.yyyyMMddHHmm (Default: false)
```

saikatguha:
I am using [editor.swagger.io](editor.swagger.io) and don't have the option to change the codegen settings used there. If they are using the default options, `original` naming should probably be the default. 

Also changing property names without also using a custom serializer/deserializer to map the changed names back to the original schema names in API operations breaks the client.

wing328:
Please refer to https://github.com/swagger-api/swagger-codegen#online-generators on how to customize the SDK generated by generator.swagger.io, which is used by editor.swagger.io to generate SDKs.

I agree with you that ideally there should be a mapping between JSON key and object attributes but I'm not aware of any JS/TS lib to do that.

For PHP, Ruby, Python API clients, we end up implementing our own object serializer/deserializer to handle the mapping between JSON keys and object attributes. 

May I know if you've cycle to contribute the enhancement (mapping between JSON keys and object attributes)?

saikatguha:
[...]

PR #3521, opened 2016-08-04
Convert snake case models naming to cameCase naming

iCEAGE:
Simply based on the JavaScript's code style guide naming should be camelCase.
`baseName` might be snake_case or some other variations but when the `name` is used it is camelCase.


wing328:
Thanks for the PR but we need to use the base name (original name) since the base name can contain special characters, e.g. -, $

iCEAGE:
@wing328 Thanks for your reply.

To be consistent in all the clients, I think `name` should be used. I've seen in the other clients e.g. Go client, special characters (`-`) are replaced with `_`. I think that happens in the [toVarName()](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/GoClientCodegen.java#L185) method.

I'm wondering what [sanitizeName()](https://github.com/swagger-api/swagger-codegen/blob/5b4eb96ba1b2e49c16db75963a091e259dacacc4/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L3047) method does? Don't you think should we replace the special characters with style guide's way?

iCEAGE:
````
-{{/interfaceModels}}{{/useInheritance}}{{#vars}}{{#required}}    _this['{{baseName}}'] = {{name}};{{/required}}
````
I'm not sure what this part does but in my case it just generates some empty lines. You can find the spec [here](https://newsapi.aylien.com/swagger/spec/news-api.json).

In the spec in one of the model there are some attributes which are snake_case. For example there is a Stories model which has an attribute called `next_page_cursor`. I expect the following:
```
console.log('API called successfully. next page cursor is: ' + data.nextPageCursor);
````
but I get:
````
console.log('API called successfully. next page cursor is: ' + data.next_page_cursor);
````
Let me know if you have any questions. Thanks

wing328:
@iCEAGE basically we need a way in JS API client to maintain a mapping between JS model property name and JSON key name. I'm not aware of any node module to do that.

Here is a similar discussion but for typescript: https://github.com/swagger-api/swagger-codegen/issues/2766

wing328:
Sorry we can't merge this as this will break others. Please use the `-t` option with the customized templates to meet your requirement.

Issue #1981, opened 2016-01-26
[Objective-C] Feature Request: Use AFHttpSessionManager instead of AFHTTPRequestOperationManager


wberger:
Currently the client's inherit from `AFHTTPRequestOperationManager`, which is no longer supported in AFNetworking 3.

Regarding future updates it would be great to have it use `AFHttpSessionManager` instead. This way the generated code should be compatible with both AFNetworking 2 and 3.

wing328:
@wberger thanks for pointing it out. Would have cycle to help make the change to use `AFHttpSessionManager ` instead?

wberger:
Hi,

I have started with a generator that is based on `AFHttpSessionManager`. For now it is implemented in a custom module and does not support all features of the standard `AFHTTPRequestOperationManager` version, as I currently don't have the time to test the currently missing features.

Maybe someone could pick that up? Here is the link to the current implementation, which I think is not yet ready for a pull request:

[objc-SessionManager Module](https://github.com/wberger/swagger-codegen/tree/feature/objc-sessionmanager/output/objcSessionManager)

wberger:
@wing328 Do you have any feedback on this? I think the suggested approach should work quite well, but I need some input from someone who has the bigger picture in mind in order to cover all intended use cases.

wing328:
@wberger sorry I missed your update earlier. I'll take a closer later today or tomorrow and then get back to you.

wing328:
@wberger do you mind submitting a PR based on your work so far? That way the community can more easily pick it up and comment on the code if they've any question about the change. Thanks!

(please also rebase on the latest master)

bithavoc:
@wberger any plans to submit a PR? I can take a look too.

wing328:
@wberger I would suggest you still submit a PR for your partial work so far so that the community can help work on it.

wberger:
Currently, I don't have a lot of time available, but I'll try to prepare
the PR.

Good news: in the meantime, I have used the generator with AFNetworking 3
and it seems to work well.



Am 28.04.2016 um 06:38 schrieb wing328 <notifications@github.com>:

@wberger <https://github.com/wberger> I would suggest you still submit a PR
for your partial work so far so that the community can help work on it.

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/swagger-api/swagger-codegen/issues/1981#issuecomment-215308222>

wberger:
[...]

PR #3522, opened 2016-08-04
Feature/objc tasks 2.3.0

wberger:
@wing328 As agreed in PR #3391 I created a new PR for branch 2.3.0.

This replaces PR #3510.


fehguy:
Hi @wberger you may have shared this with @wing328 but what is the goal of this PR?  Please create an issue and reference it in your PR, or put in some comments as to what the goal is with breaking the signatures of the generated clients.

wberger:
@fehguy @wing328 The original motivation was this issue: #1981

The change to `AFHttpSessionManager` has been made since then, but I expanded on it by removing the request IDs for pending requests and returning `NSURLSessionTasks` instead. This allows for real task cancellation and leaves the housekeeping to the OS (see also the explanation below).

Furthermore during integration in our project I ran into the limitation of having only a single global configuration object and not being able to provide a different configurations for different ApiClient instances. Thus I reworked the code to make the configuration a component of the ApiClient which can be passed in during client creation. It results in less global state, higher configurability and less coupling - but only if required, since there are still fallbacks to global defaults.

Thus the only breaking change in the API is the return value of the API calls (`NSURLSessionTask *` instead of `NSNumber *`) and the renaming of the default configuration object from `{{classPrefix}}Configuration` to `{{classPrefix}}DefaultConfiguration`.

The original description of the changes can be found in PR #3391:

> I propose a change to the objc ApiClient. Instead of internally created and maintained request IDs the API implementation should return NSURLSessionTasks. This change has two advantages:
> 
> - Less internal state
> - These task object provide real cancellation as compared to the request ID based implementation (even if the task is currently running)
> 
> Furthermore I changed the configuration object to be protocol based in order to allow for more flexible integration into projects and more complex configuration setups. A default configuration implementation is provided and can be passed in, if no custom implementation is required.

mateuszmackowiak:
@wing328 👍  Looks good to me

wing328:
### Upgrade Note ###
**(1)** {classPrefix}Configuration has been renamed to {classPrefix}DefaultConfiguration, e.g.
*Before*:
```objc
    SWGConfiguration *config = [SWGConfiguration sharedConfig];
```
*After*:
```objc
    SWGDefaultConfiguration *config = [SWGDefaultConfiguration sharedConfig];
```

**(2)** Return type for API functions has been changed from `NSNumber*` to `NSURLSessionTask*`, e.g. 
*Before*
```objc
-(NSNumber*) addPetWithBody: (SWGPet*) body
```
*After*
```objc
-(NSURLSessionTask*) addPetWithBody: (SWGPet*) body
```

Issue #3523, opened 2016-08-04
Use Title as name in any inline model defined (when title is defined)


dmytrobr:
##### Description

Feature has been introduced to support "title" as model name:
https://github.com/swagger-api/swagger-codegen/issues/2868

However it has not been applied to some type of inline models (inline models defined within MapProperty and ArrayProperty. I see it as a bug in the feature introduced with [Issue-2868](https://github.com/swagger-api/swagger-codegen/issues/2868).

##### Swagger-codegen version

2.2.1-SNAPSHOT

##### Swagger declaration file content or url

    ```java
       @Test
    public void resolveInlineArrayModelWithTitle() throws Exception {
        Swagger swagger = new Swagger();

        swagger.addDefinition("User", new ArrayModel()
                .items(new ObjectProperty()
                        .title("InnerUserTitle")
                        ._default("default")
                        .access("access")
                        .readOnly(false)
                        .required(true)
                        .description("description")
                        .name("name")
                        .property("street", new StringProperty())
                        .property("city", new StringProperty())));

        new InlineModelResolver().flatten(swagger);

        Model model = swagger.getDefinitions().get("User");
        assertTrue(model instanceof ArrayModel);

        Model user = swagger.getDefinitions().get("InnerUserTitle");
        assertNotNull(user);
        assertEquals("description", user.getDescription());
    }
  '''


##### Command line used for generation

Used Maven plugin to generate Java client

##### Steps to reproduce

See Unit Test above

##### Related issues
[Issue-2868](https://github.com/swagger-api/swagger-codegen/issues/2868).

##### Suggest a Fix
I'm submitting pull request to fix this

wing328:
PR meged. Thanks for the contribution.

PR #3524, opened 2016-08-04
support Title as model name for any inline object

dmytrobr:
Fixing bug as defined in [Issue-3523](https://github.com/swagger-api/swagger-codegen/issues/3523)

InlineModels defined within MapProerty and ArrayProerty were not using
"title".
Added Unit Tests


wing328:
@dmytrobr thanks for the PR, especially with the test cases. We'll review and let you know if we've any question.

Issue #2858, opened 2016-05-13
[csharp] (Discussion) Refactor and document ApiClient/Configuration


jimschubert:
This is in response to conversation in #2723.

I'd like to discuss the following:
1. Differentiate between "global configuration" and "instance/per-request configuration".
2. Better documentation related to constructing the *Api classes with `Configuration` instances.
3. Clean up the constructors for ApiClient to call through to the other constructors.
4. Maybe deprecate `Configuration.Default` or mark it obsolete and introduce `Configuration.Global`?
5. Maybe remove the circular reference between `Configuration` and `ApiClient`? It makes sense that ApiClient _has a_ Configuration, but it isn't immediately clear why Configuration _has an_ ApiClient.
## References

> 3. Clean up the constructors for ApiClient to call through to the other constructors

```
public ApiClient()
{
    Configuration = Configuration.Default;
    RestClient = new RestClient("http://petstore.swagger.io/v2");
}

public ApiClient(Configuration config = null)
{
    if (config == null)
        Configuration = Configuration.Default;
    else
        Configuration = config;

    RestClient = new RestClient("http://petstore.swagger.io/v2");
}

public ApiClient(String basePath = "http://petstore.swagger.io/v2")
{
    if (String.IsNullOrEmpty(basePath))
        throw new ArgumentException("basePath cannot be empty");

    RestClient = new RestClient(basePath);
    Configuration = Configuration.Default;
}
```

> 4. Maybe deprecate `Configuration.Default` or rename to `Configuration.Global`?

```
public static Configuration Default = new Configuration();
```

> 5. Maybe remove the circular reference between Configuration and ApiClient? 

```
// In Configuration.cs
public ApiClient ApiClient;

public void setApiClientUsingDefault (ApiClient apiClient = null)
{
    if (apiClient == null)
    {
        if (Default != null && Default.ApiClient == null)
            Default.ApiClient = new ApiClient();

        ApiClient = Default != null ? Default.ApiClient : new ApiClient();
    }
    else
    {
        if (Default != null && Default.ApiClient == null)
            Default.ApiClient = apiClient;

        ApiClient = apiClient;
    }
}

// In ApiClient
public Object CallApi(
    String path, RestSharp.Method method, Dictionary<String, String> queryParams, Object postBody,
    Dictionary<String, String> headerParams, Dictionary<String, String> formParams,
    Dictionary<String, FileParameter> fileParams, Dictionary<String, String> pathParams,
    String contentType)
{
[...]

PR #3530, opened 2016-08-05
[csharp] client refactoring

jimschubert:
This is a breaking changes refactoring of the C# client (see #2858)

This cleans up the concept of a "default" configuration versus a per-instance configuration. These are disconnected in such a way that most use cases won't accidental modify global state. To limit the number of breaking changes, there is still a setter on `Configuration#DefaultHeader` as well as `ApiKey` and `ApiKeyPrefix`. In most cases, these will default to a concurrent dictionary. A constructor is provided which will allow users to reference those dictionaries from `Configuration.Default`.

To limit possible instances of misuse, an `ApiClient` can only be built:
- using parameterless constructor, which defaults to `Configuration.Default`
- from a `Configuration` instance, either as `new ApiClient(instance)` or `instance.ApiClient`
- with a provided `basePath` which, again, uses `Configuration.Default`

The instance of configuration assigned to `ConfigurationDefault` is exposed as a new partial class called `GlobalConfiguration`. Properties on `Configuration` are now marked virtual. This allows users to provide clearer defaults. 

Configuration now implements `IReadableConfiguration`, for which the properties do _not_ provide setters. This interface is how the `ApiClient#Configuration` property is defined. This somewhat forces the user to reinstantiate apiClient or reassign the Configuration property, which also helps avoid users accidentally assigning different properties to `Configuration.Default`.
## Questions

1) I'd like to remove the parameter default on the `ApiClient` constructor which takes a `basePath`.  This will remove any future possibly undefined behavior of ApiClient's parameterless constructor and this constructor interfering. Right now, they're doing basically the same work.

Any objections to removing this string default?

```
// reference: ApiClient.mustache
        /// <summary>
        /// Initializes a new instance of the <see cref="ApiClient" /> class
        /// with default configuration.
        /// </summary>
        /// <param name="basePath">The base path.</param>
        public ApiClient(String basePath = "{{{basePath}}}")
        {
           if (String.IsNullOrEmpty(basePath))
                throw new ArgumentException("basePath cannot be empty");

            RestClient = new RestClient(basePath);
            Configuration = Client.Configuration.Default;
        }
```

2) I've left the `ApiClient` property on `Configuration` to provide a new instance of `ApiClient` which back-references to the configuration instance that created it. I think this more cleanly removes any questions about what configuration is in use for a given `ApiClient`. For example:

```
var a = new ApiClient(); // obviously Configuration.Default
Configuration.Default.ApiClient; // obviously Configuration.Default
var a2 = new ApiClient(config); // obviously config instance
config.ApiClient; // obviously config instance
```

Are these too many different ways to achieve the same thing?

3) `RestClient` is publicly exposed from an instance of ApiClient. I've left `Configuration#Timeout` on the configuration instance, but haven't really hooked it up through `ApiClient.RestClient.Timeout` as it was done before. What is the benefit of having the configuration exposed and settable in multiple places?  Should we look at hiding the internal implementation of using RestSharp's `RestClient`?  There is another issue related to refactoring out the RestSharp dependency and using FCL http client functionality, but I think that would be difficult without first deprecating the `RestClient` access.

---

~~**NOTE** This isn't ready to merge until the Timeout question is discussed.~~ Timeout now passes through.


lunat:
For me It's all OK

After these changes documentation file modules\swagger-codegen\src\main\resources\csharp\api_doc.mustache needs two variations:

row 40
`Configuration.Default.ApiKey.Add("{{{keyParamName}}}", "YOUR_API_KEY");`
becomes
`Configuration.Default.AddApiKey("{{{keyParamName}}}", "YOUR_API_KEY");`

row 42
`            // Configuration.Default.ApiKeyPrefix.Add("{{{keyParamName}}}", "Bearer");{{/isApiKey}}{{#isOAuth}}`
becomes
`            // Configuration.Default.AddApiKeyPrefix("{{{keyParamName}}}", "Bearer");{{/isApiKey}}{{#isOAuth}}`

jimschubert:
@lunat thanks for reviewing and for finding the missed docs.

What's your take on the Timeout? Should we leave it as a pass-through property off of Configuration?

staticcat:
Just got redirected to this pr from #4843. Some great work going on here. Thought I would ask how people see the following things being handled in this new update.

1. Name conflicts with classes chosen for core functionality in the client/tests. E.g. The user has a model class called Configuration. 

2. Automatically generating tests (we have a great deal of information about types) and models. I have been experimenting with generating basic ground level tests. E.g. Does the rest call work, can it return the other documented status code. 
I have worked on a static class that provides the TestCase data for each scenario. All the user needs to do is specify the data. 

3. Optional class constructor with all parameters. Haven't found a way to do this. I had to remove the mustache entry for these to achieve what I was looking for. 

Happy to help out on all of this. Just don't have that much experience in the latest Java environments. Most the alterations I have been able to achieve are through updating the mustache files.

staticcat:
Another thing I just came across is that the default generation does not encode enum's as string correctly. 

This stack overflow post explains the issue with using rest sharps default encoder. The article after explains how to create a separate encoder. 

http://stackoverflow.com/questions/38591048/restsharp-not-deserializing-json-correctly

http://bytefish.de/blog/restsharp_custom_json_serializer/

jimschubert:
@staticcat That's some great feedback.

There's not supposed to be any conflicts with these built-in types for the client generators anymore. I just looked, and I don't know if it was a bad merge somewhere or what but `Configuration` isn't using the fully qualified name in every case (see `new Configuration`), so we'd need to fix that:

```
        private IO.Swagger.Client.ExceptionFactory _exceptionFactory = (name, response) => null;

        /// <summary>
        /// Initializes a new instance of the <see cref="FakeApi"/> class.
        /// </summary>
        /// <returns></returns>
        public FakeApi(String basePath)
        {
            this.Configuration = new Configuration(new ApiClient(basePath));

            ExceptionFactory = IO.Swagger.Client.Configuration.DefaultExceptionFactory;

            // ensure API client has configuration ready
            if (Configuration.ApiClient.Configuration == null)
            {
                this.Configuration.ApiClient.Configuration = this.Configuration;
            }
        }
```

This would also require removing the using directive on the `.Client` namespace so any types in this namespace don't conflict with the `.Model` namespace.

I've also thought about generating tests. It is a bit complicated to generate tests given how loose the  Swagger specification is. For instance, if you have an example defined for JSON inputs, it's ridiculously easy to deserialize that input into the target type and do some verification. The problem is that Swagger specs don't require examples, so you'd have to rely on logic to write tests. Since the templates are written in mustache, which is logicless, that poses many problems.

It's further complicated because APIs don't all support only one input format (JSON), and to create tests for each mime type would be difficult. For example, these would be relatively easy:

```
produces:
- application/json
- application/xml
```

while writing tests for both of these would be impossible:

```
produces:
- application/json
- application/vnd.jimschubert+problem.json
```

We could limit to testing JSON endpoints only, and only if there are examples, but I think that would be relatively confusing to an end user of the generator. I'm also not sure there's much value in adding it to the base generator, because for many it would mean more code to replace when they begin implementation. If you have a prototype, I'd suggest opening a PR to get community feedback. We could also put the test generation behind an optional flag.

Lastly, I'm not sure I understand your comment about optional class constructors with all parameters. The model constructors set the properties that are required for a type to be constructed, or which are intended to be readonly after construction. You can double this with object initialization to achieve what you want:

```
var model = new Model(required1, required2) {
    Optional1 = optional1,
    Optional2 = optional2
}
```

Is your question here more related to automatically generating tests? If so, I think you could mimic how the properties are created for these readonly properties vs those with public full accessors.

jimschubert:
@wing328 I don't know if you're getting notifications for those build failures. I attempted to merge 2.3.0 into this branch and git only reported about 10 merge conflicts. After fixing those and pushing, CI showed there were many more that git didn't display. I'll need to attempt later and track down all merge conflicts.

wing328:
@jimschubert yes, I saw those alerts. I was thinking if cherry-picking your commits into a new branch would be a relatively easier approach.

jimschubert:
Yeah, I think that's the route I'll have to go after looking at the conflicts again. I'll try and get this tracking current 2.3.0 before I leave for vacation later this week, so we can try to get it merged. I think this PR will resolve a few issues.

jimschubert:
Closing this PR because it has diverged too far from 2.3.0 and a merge proved too time consuming. Opened #5740 with a new branch which has been rebased against 2.3.0.

PR #3531, opened 2016-08-05
[Java] add enum test cases for Jackson, Gson

wing328:
- add enum test cases for Jackson, Gson
- add symbol mapping (e.g. period)
- fix jackson deserialization issue using @JsonCreator


cbornet:
@wing328 I think the @JsonCreator is not needed if the ObjectMapper is configured with READ_ENUMS_USING_TO_STRING. It would make the POJOS lighter.

PR #3533, opened 2016-08-05
Fix compilation error in Go templates

JackDanger:
When all of the `pathParams` for an API call are empty then the imported
"strings" and "fmt" are unused. This is a compilation error in Go and
the best way to work around it considering the limitations of Mustache
is to add dummy usages of the libraries in the file.


wing328:
@JackDanger thanks for the PR. Please run `./bin/go-petstore.sh` to update the Go Petstore sample so that CI can verify the change.

Would using `{{#hasPathParams}} .. {{/hasPathParams}}` instead also work?

cc @neilotoole @guohuang

JackDanger:
@wing328 I've regenerated the petstore project project and the tests pass. I don't understand the test project or the YAML format well enough to create a failing test case where an endpoint has no path parameters.

I can't do `{{#hasPathParams}} .. {{/hasPathParams}}` because the parameters are nested under a list of operations so I'd have to do `#{{operation}}{{#hasPathParams}} .. {{/hasPathParams}}{{/operation}}` but that would result in multiple entries in the imports list. I'm not sure of a way to ask Mustache if any of the operations have any path params.

guohuang:
@JackDanger I don't think this is the right approach, can you show me more detail about your use case where these imports are not needed? we should programmatically hide those import if they are not being used.

JackDanger:
@guohuang If a resource does not define any path parameters on any endpoints then invalid go is generated. The only places that `fmt` or `strings` are used are in interpolating path parameters so when those lines don't appear in the generated file the imports are unused and the file doesn't compile.

This isn't such a weird pattern in Go, it's often the case that you'll assign a var with an interface type to `_` as a compiler check - these lines don't result in any actual memory allocation.

Here's an example of something quite common:
```
var _ SomeInterface = &ThingIThinkImplementsTheInterface{}
```

The generated files would be cleaner if we added `{{#hasAnyPathParams}} .. {{/hasAnyPathParams}}` as a field in our mustache renderer but I wasn't sure if we wanted to clutter up the renderer like that.

guohuang:
@JackDanger this is how I solve this problem, https://github.com/swagger-api/swagger-codegen/pull/3543, you could try to pull it to test it out. I have verified this will fix your problem.

Thanks

JackDanger:
Oh, I like your solution much better. Thank you!

wing328:
PR merged. Please pull the latest master to give it a try.

PR #3536, opened 2016-08-05
[PHP] Examples on inheritance vs composition

arnested:
**Do not merge**

This is for testing how the PHP client generates code for model definitions with (multilevel) inheritance and composition.

In the [v2.1.6](https://github.com/swagger-api/swagger-codegen/releases/tag/v2.1.6) the PHP client supported multilevel inheritance and composition. Both model definitions with and without discriminators (inheritance vs composition) would be generated as inheritance in the PHP code.

The models defined in this pull requests [example spec](https://github.com/arnested/swagger-codegen/blob/php-inheritance-composition/modules/swagger-codegen/src/test/resources/2_0/inheritanceTest.yaml) would be generated as described in this class diagram:

inheritance-composition-v2 1 6

In current HEAD (and [v2.2.0](https://github.com/swagger-api/swagger-codegen/releases/tag/v2.2.0) as well) something changed and we now generate this class diagram:

inheritance-composition-head

A few things to notice:
- `InheritLevel2` no longer inherits from `InheritLevel1`.
- It would be fine that `compositionLevel2` and `compositionLevel2` no longer is generated with inheritance but they don't include the properties from their ancestors.

I suspect the change in 8740b4d0ea68f9de2bd2849a501e33a5010a11d3 combined with https://github.com/swagger-api/swagger-parser/pull/246 to be the cause of this. No hard feelings, though. I'm not sure multilevel inheritance was ever supposed to be supported.

I know there is at least one project using multilevel inheritance _and_ composition in it's spec :smile:

Currently I haven't found a way to work around this but I'll keep looking into this.

I'm opening this pull request for sharing my findings and collecting feedback.


arnested:
Adding the fix from #3475 definitely improves things:

inheritance-composition-3475

I'll have a look later tonight to see if I can get the last missing piece fixed.

arnested:
So, time went by and I haven't found the time to look further into this issue.

But re-running the test on current master shows it has been fixed at some point:

model

Thank you. Whoever you are!

arnested:
I should probably have posted a version with properties as well:

model

This shows that inheritance is implemented as inheritance and composition as composition. Very nice.

PR #3541, opened 2016-08-05
Fix spelling in ruby ApiClient: Configuraiton -> Configuration

janxious:
Simple fix! I think!


wing328:
@janxious thanks for the fix. Have a nice weekend!

wing328:
FYI. I also did a case sensitive search for `Configuraiton` in the 'samples' directory and confirmed only Ruby API client contained the typo.

janxious:
:100: Yeah I did the same. Should have mentioned that in my PR to save you the trouble!

PR #3546, opened 2016-08-06
[JavaScript] add enum tests cases for JS generator (java)

wing328:
- add enum tests cases for JS generator (java) to cover numeric value and array of enum
- add more symbol mapping


sebciok:
Great, Thanks. Looks great. I will verify soon. Should it be "Quote" insread of "Quota", though?

wing328:
Fixed quote.

Issue #3356, opened 2016-07-12
[C#] add an option to implement INotifyPropertyChanged in generated model


tdjastrzebski:
##### Description

Please add an option to implement INotifyPropertyChanged in generated model.
##### Swagger-codegen version

1.0

wing328:
@tdjastrzebski thanks for the suggestion. May I know if you've cycle to contribute the feature?

jimschubert:
@tdjastrzebski @wing328 This can easily be done using https://github.com/Fody/PropertyChanged. We could then make the `INotifyPropertyChanged` functionality optional as many devs are passionate about not using `INotifyPropertyChanged` when it's not being used.

wing328:
@jimschubert thanks for sharing https://github.com/Fody/PropertyChanged. The solution looks very neat.

I've marked the task as "Need community contribution" to see if anyone's cycle to make the contribution.

I believe we'll start with C# API client to begin with, and then C# server stub (ASP.NET core and NancyFx)

jimschubert:
@wing328 I'll look at doing this using Fody. It should be relatively simple.

jimschubert:
I've implemented this for .NET 4 and higher in PR #3547. It's a lot easier to maintain and test if we limit the feature to the newer framework version.

@wing328 .NET Core support would have to apply the changes manually to the models, unless we wait for .NET Core projects to go back to MSBuild-based. Then we could just use Fody there, too.

wing328:
@tdjastrzebski PR merged into master. Please pull the latest to give it a try.

jimschubert:
@tdjastrzebski I'd be interested to know if you've tried this and if it is working for you.

wing328:
Closing this as there's no update.

PR #3547, opened 2016-08-06
[csharp] PropertyChanged support .NET 4.x+

jimschubert:
This adds support for additional property `generatePropertyChanged=true` which will add an `OnPropertyChanged` method and a `PropertyChanged` event to all models and use Fody code weaving to inject the raised events into property setters.

For more information on code weaving, check out [Fody](https://github.com/Fody/Fody) on github.

see feature request #3356.

@wing328 question... I moved the `$@` in `bin/csharp-petstore.sh` to support running the script with properties passed to CLI. For example:

```
./bin/csharp-petstore.sh --additional-properties=generatePropertyChanged=true
```

Was there a reason it previously went `$@ generate...` instead? Usage of those sample generator scripts doesn't seem to be documented anywhere.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CSharpClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CSharpClientCodegen.java
@@ -1,32 +1,20 @@
 package io.swagger.codegen.languages;
 
-import com.google.common.collect.ImmutableBiMap;
 import com.google.common.collect.ImmutableMap;
-import io.swagger.codegen.CodegenConfig;
 import io.swagger.codegen.CodegenConstants;
 import io.swagger.codegen.CodegenType;
 import io.swagger.codegen.CodegenModel;
-import io.swagger.codegen.DefaultCodegen;
 import io.swagger.codegen.SupportingFile;
 import io.swagger.codegen.CodegenProperty;
-import io.swagger.codegen.CodegenModel;
 import io.swagger.codegen.CodegenOperation;
-import io.swagger.codegen.CodegenParameter;
-import io.swagger.models.properties.*;
 import io.swagger.codegen.CliOption;
 import io.swagger.models.Model;
 
 import java.io.File;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.ArrayList;
 import java.util.Iterator;
 
-import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.text.WordUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -48,6 +36,7 @@ public class CSharpClientCodegen extends AbstractCSharpCodegen {
     protected String targetFrameworkNuget = "net45";
     protected boolean supportsAsync = Boolean.TRUE;
     protected boolean supportsUWP = Boolean.FALSE;
+    protected boolean generatePropertyChanged = Boolean.FALSE;
 
 
     protected final Map<String, String> frameworks;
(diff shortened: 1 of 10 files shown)


wing328:
> Was there a reason it previously went $@ generate... instead? Usage of those sample generator scripts doesn't seem to be documented anywhere.

I've not done anything with $@ in the shell script, which was there before I started working on this project.

If there's a good reason/use case, I think it's ok to remove it.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CSharpClientCodegen.java
```diff
@@ -127,6 +116,10 @@ public CSharpClientCodegen() {
         addSwitch(CodegenConstants.OPTIONAL_EMIT_DEFAULT_VALUES,
                 CodegenConstants.OPTIONAL_EMIT_DEFAULT_VALUES_DESC,
                 this.optionalEmitDefaultValue);
+
+        addSwitch(CodegenConstants.GENERATE_PROPERTY_CHANGED,
+                CodegenConstants.PACKAGE_DESCRIPTION_DESC,
```
wing328: @jimschubert I believe `PACKAGE_DESCRIPTION_DESC` should be replaced by `GENERATE_PROPERTY_CHANGED_DESC` instead.

wing328:
@jimschubert thanks for the PR. I tested with --additional-properties=generatePropertyChanged=true and Xamarin reported the FodyWeavers.xml not found

screen shot 2016-08-12 at 3 10 23 pm

Did you encounter similar issue with Visual Studio or Xamarin in your development environment? Should FodyWeavers.xml be generated under `src/{packageName}/` instead?

wing328:
About `$@ generate`, maybe it was designed for debug flag (e.g. `-DdebugModels=true`) to begin with.

jimschubert:
@wing328. No I didn't  have that problem as I recall. I can look again. The only problem I had with the XML file is that I originally forgot to add the PropertyChanged  reference, but that should be fixed now.

I had written a test and verified in Xamarin and VS. I'll try to see what happened.

Plus for the debug flag, you should be able to pass it like this:

```
JAVA_OPTS=-DdebugModels=true ./bin/csharp-petstore.sh
```

wing328:
@jimschubert then my guess is that the one who created the shell script had forgotten about the JAVA_OPTS option.

wing328:
FYI. I've filed https://github.com/swagger-api/swagger-codegen/pull/3782 to fix the location of the XML file.

jackjwilliams:
So, I'm trying to decipher is this exists in 2.2.1, or if it is to be released soon. It looks like it was merged to master, but it's still slated for a 2.3.0 release, correct? I keep trying to generate with 2.2.1 but no PropertyChanged attributes get generated according to the mustache templates in master. If I need this now, is there a bleeding build available or will I need to build myself?

jimschubert:
@jackjwilliams I downloaded 2.2.1 from maven and tested locally, it looks like it didn't make it to that release.

I have a docker image tracking the master branch. I have it set to trigger a build whenever Swagger Generator from the official repo is built.

I kicked off another build just now since the last build was 21 days ago.

https://hub.docker.com/r/jimschubert/swagger-codegen-cli/

Let me know if you need help workout out the command.  This is how you'd do the sample with property changed:

```
docker run -it -v ~/temp/swagger-generated:/swagger-api/out     \
    jimschubert/swagger-codegen-cli generate     \
    -i /swagger-api/yaml/petstore-with-fake-endpoints-models-for-testing.yaml    \
    -l csharp     \
    -o /swagger-api/out/cs \
     --additional-properties=generatePropertyChanged=true
```

Also, the property changed stuff works from a generated method in the templates and the properties notify via compile time code weaving using Fody. So you won't see these notifications in the mustache templates.

jackjwilliams:
@jimschubert Thank you for the info! I ended up installing maven and building, but opened https://github.com/swagger-api/swagger-codegen/issues/4344 to address an issue with the build.bat that gets generated.

PR #3550, opened 2016-08-07
[aspnet] Honor casing of properties from definition

jimschubert:
See #3262

This would replace #3263, which provides a correct implementation for JSON endpoints only. Although generated code for the `aspnet5` generator outputs to JSON, we still need to support all output types (i.e. from the `produces` definition of the api endpoint).

I've also fixed the formatting in the `ToString` method for models and changed the generic description on enums which was previously worded like an auto-implemented property description.


wing328:
@jimschubert thanks for the PR. The change looks good to me.

PR #3551, opened 2016-08-07
[Java] Fix @JsonCreator annotation issue with Java API clients using gson

wing328:
The issue was due to not using {{#jackson}} tab to handle the JsonCreator (we've done it for import but not `fromValue`) and it's causing issue for Java API clients using gson (okhttp-gson, retrofit, retrofit2)

Ran `./bin/java-petstore-all.sh` to update all Java API client so as to back test API clients using jackson and  to confirm API clients using gson no longer having the issue. 

The issue was caused by https://github.com/swagger-api/swagger-codegen/pull/3531 in an attempt to fix deserializaiton issue with jackson.


sebciok:
I pulled the latest, regenerated my definition, and I could compile the generated code. I can no longer see @JsonCreator attributes. Kudos for the quick fix.

PR #3553, opened 2016-08-08
Amending Path annotation

rynger:
The code generated for CXF was always referring to to @Path("/") in the API.
Changed the templated Path annotation value from "/" to "/{{baseName}}".


wing328:
@rynger thanks for the PR. Please run `./bin/jaxrs-cxf-petstore-server.sh` to update the Petstore sample so that the CI can tests the change.

I still believe it should be {{basePath}} as {{baseName}} refers to the original parameter name (header, form, query)

rynger:
Ah yes, looking at other templates, they refer to the {{basePath}}, in which case all the api.mustache files under the JavaJaxRS resources folder have @Path annotations which should use {{basePath}} and not {{baseName}}. (A seperate issue?)

I will amend the CXF template, run the scripts and try the built snapshot with my code first before attempting another commit.

wing328:
@rynger it's worth sharing with you that there's another variable {{contextPath}} which stores the base path without the host name (e.g. /v2 for Petstore while an endpoint for Petstore looks like this: https://petstore.swagger.io/v2/pet/1)

rynger:
Thanks for that, additionally from inspecting the DefaultGenerator class source (~Line 399) i see there's a {{basePathWithoutHost}} as well and would seem to contain the same value. I will err in favour of your suggested {{contextPath}} as it seems to have a null check in its assignment.

wing328:
@rynger thanks for the contribution. The change looks good to me. When you've time, I wonder if you can add a auto-generated pom.xml similar to this one: https://github.com/swagger-api/swagger-codegen/blob/master/samples/server/petstore/jaxrs/jersey1/pom.xml (with another PR)

PR #3564, opened 2016-08-09
Fix unknown parameter on api_client's update_params_for_auth method.

cliffano:
This should fix warning message when generating Ruby API doc, e.g. using Yardoc.
Also re-gen ruby petstore.


wing328:
@cliffano thanks for the fix. The change looks good to me. I'll see if I can later setup CI to capture the warning/error reported by Yardoc

wing328:
Looks like someone has made a suggestion to yardoc alread but no one has the cycle to contribute the feature yet: https://github.com/lsegal/yard/issues/935

PR #3573, opened 2016-08-10
Go: Replace `.Username` with `.UserName` in generated API methods.

ivucica:
Right now, any generated code where the http basic authentication is
required is broken, as it tries to access nonexistent field in the
generated `Configuration` struct.

The alternative is to fix `configuration.mustache` to replace the
`UserName` field with `Username` field.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/go/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/go/api.mustache
@@ -63,7 +63,7 @@ func (a {{{classname}}}) {{{nickname}}}({{#allParams}}{{paramName}} {{{dataType}
 {{/isApiKey}}
 {{#isBasic}}
 	// http basic authentication required
-	if a.Configuration.Username != "" || a.Configuration.Password != ""{
+	if a.Configuration.UserName != "" || a.Configuration.Password != ""{
 		localVarHeaderParams["Authorization"] =  "Basic " + a.Configuration.GetBasicAuthEncodedString()
 	}
 {{/isBasic}}


ivucica:
It would be great if some Go code would be built as part of CI tests, covering this type of change. What would be a good way of tackling this?

wing328:
@ivucica totally agree leveraging the CI and Go Petstore API client to catch the issue moving forward.

I'll submit a PR to update the test spec.

wing328:
I've added test cases via #3574  to cover the issue and got the following error before your fix:
```
[INFO] --- exec-maven-plugin:1.2.1:exec (go-get-resty) @ Goswagger ---
[INFO] 
[INFO] --- exec-maven-plugin:1.2.1:exec (go-test) @ Goswagger ---
# _/Users/williamcheng/Code/jul2016/swagger-codegen/samples/client/petstore/go/go-petstore
go-petstore/fake_api.go:137: a.Configuration.Username undefined (type Configuration has no field or method Username)
FAIL	_/Users/williamcheng/Code/jul2016/swagger-codegen/samples/client/petstore/go [build failed]
```

wing328:
@ivucica FYI. There's a PR to undo the change: https://github.com/swagger-api/swagger-codegen/pull/4245.

Let me know if you've concern.

ivucica:
This is fine. #4245 seems to also be covering my comment:

> The alternative is to fix configuration.mustache to replace the
> UserName field with Username field.

I do not understand what @paulvollmer's "the UserName variable does not work in CamelCase" means, i.e. why this is necessary, but it seems like a clean fix.

PR #3576, opened 2016-08-11
Issue 3572 Remove duplicate CLI Options

andylowry:
Issue 3572 Remove dupe options in jaxrs-spec

Removed CLI option declarations in JavaJAXRSSpecServerCodegen that also
appear in AbstractJavaJAXRSServerCodegen.


andylowry:
Note that I attempted to tackle the goal of making `DefaultCodegen` check for this situation, but while the implementation was fairly straightforward, I found it difficult to identify a good spot for the check to be kicked off. Two options seemed to be in `Codegen#main` and in `DefaultGenerator#opts`. Former has the disadvantage of not applying in situations where codegen is executed via API rather than via command line. Overall I was just not comfortable making this change, given the confusing (to me) code surrounding the various types of "options" available in the overall project. Someone with deeper knowlege than I might want to tackle this, if it seems important.

In any case, as our project (RepreZen API Studio) upgrades to new swagger-codegen releases in the future, we will spot additional instances of this issue automatically, and can report them as they arise.

wing328:
@andylowry no problem. Your fix is definitely good and we'll review other generators later to see if there are similar issues.

PR #3577, opened 2016-08-11
[Javascript] Fix types

arikon:
There are no Integer and DateTime types in Javascript.


wing328:
@arikon thanks for the PR. I've also updated the JS test cases due to new test for HTTP basic authentication.

PR #3578, opened 2016-08-11
Replaced auto-generated equals/hashCode with apache commons builder

heinrichj:
+ Replaces auto-generated methods for equals and hashCode with builders from apache.commons

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/ClientOpts.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/ClientOpts.java
@@ -1,9 +1,11 @@
 package io.swagger.codegen;
 
+import io.swagger.codegen.auth.AuthMethod;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
 import java.util.HashMap;
 import java.util.Map;
-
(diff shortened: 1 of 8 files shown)


wing328:
@heinrichj thanks for the PR. Would there be performance concern? 

Ref: http://stackoverflow.com/a/11036608/677735

heinrichj:
Hi wing328,

thank you for reviewing my PR.
I took a look at the stackoverflow post and made a very little and simple test. Yes it is slower, as expected. But in my opinion it is not very critical.
There is a trade-off. The commons implementation is less code and the avantage is that it is save against changes. If new fields are introduced these will be taking into consideration. You don't have to regenerate the methods, so you cannot forget to do so as well.

fehguy:
this really doesn't make sense to merge.  The whole idea of auto-generated code is that it will be faster and dependency free.  Why clog it up with a utility to do something that the codegen has done already?

Will leave this to @wing328 but I think this is a bad idea to bring into everyone's projects.

wing328:
@fehguy I also prefer our current approach as it's better in terms of performance.

@heinrichj you can update the mustache template locally and pass the customized templates via the `-t` option.

If there are more demands, we'll make this an option so that developers will have a choice on which one they prefer when generating the code.

wing328:
I've opened #3580 for tracking. Closing this for the time being.

Will reopen if there's a demand from the community.

PR #3582, opened 2016-08-12
Adding support for JMustache compiler customization

regga-ws:
Extension of the CodegenConfig interface in order to use JMustache compiler native features, such as the deactivation of HTML escaping.

My custom generator implements it like this : 
```
@Override
public Compiler processCompiler(Compiler compiler) {
   return compiler.escapeHTML(false);
}
```

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConfig.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConfig.java
@@ -10,6 +10,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import com.samskivert.mustache.Mustache.Compiler;
+
 public interface CodegenConfig {
     CodegenType getTag();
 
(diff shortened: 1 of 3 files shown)


wing328:
Re: 
```
@Override
public Compiler processCompiler(Compiler compiler) {
   return compiler.escapeHTML(false);
}
```
Thanks for the PR. Can you please elaborate on why you want to turn off `escapeHTML`?

regga-ws:
With `escapeHTML` turned on by default, special characters such as simple or double quotes will be transformed into their HTML code (respectively &#39; and &#34;) by the compiler. In my case, I am using a custom codegen generator to transform descriptions from a swagger spec into an human-readable document, and I end up with visible sentences such as : 
...following attributes can&#39;t be updated...
...possible values are &#39;ABP&#39;...

Also, I believe that other compiler configuration features may be interesting for some developers, for example to customize delimiters in some cases.

wing328:
@regga-ws the change sounds reasonable to me. If you want the tag without escaping HTML special characters, you can also use {{{tag_name}}} instead of {{tag_name}}

PR #3585, opened 2016-08-12
Add withCredentials field to ApiClient

### LINKED ISSUE #2442: [JavaScript] Need option for withCredentials

delenius:
This allows the user to change the withCredentials flag on the underlying
superagent request object.

Fixes #2442 

This PR solves this problem in a better and more minimalistic way. Instead of adding a new (JavaScript-only) option to the code generator, we expose a boolean flag on `ApiClient` that allows the user to set this flag (similarly to how e.g. the timeout can be changed).

This is analogous to how one can achieve the same effect in the Java client already.


wing328:
@fehguy here is a PR to update JavaScript to support cookies authentication via a boolean flag in ApiClient.js. Are you happy to merge this into master so that the JS API client can talk to backend server that requires the cookie parameter?

delenius:
I realized that this PR only works in a browser setting, not in Node.js in general.
I have an improved PR that works in both settings, coming up in a minute.

PR #3592, opened 2016-08-16
Allowing generators to be aware of original inputSpec path

regga-ws:
I have two use cases where I would like my custom generator to be aware of the Swagger inputSpec path being used for generation:
- I am generating documentation for my API and I want to indicate the path of the original spec so it can also be downloaded by developers - as the spec also consists in useful documentation
- I want to parse myself the inputSpec in addition to the parsing performed by the SwaggerParser, as the SwaggerParser loses some information - such as the order of the operations in the spec, and I want to display the operations in that same exact order in the generated documentation

Currently the inputSpec path is only kept until the SwaggerParser is used to build the Swagger object, and then it is lost. Generators are not able to access that information.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -509,14 +509,14 @@ public class DefaultCodegen {
     }
 
     public String getInputSpec() {
-		return inputSpec;
-	}
+        return inputSpec;
+    }
 
-	public void setInputSpec(String inputSpec) {
-		this.inputSpec = inputSpec;
-	}
+    public void setInputSpec(String inputSpec) {
+        this.inputSpec = inputSpec;
+    }
 
-	public void setTemplateDir(String templateDir) {
+    public void setTemplateDir(String templateDir) {
         this.templateDir = templateDir;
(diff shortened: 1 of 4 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
```diff
@@ -507,7 +508,15 @@ public String getOutputDir() {
         return outputFolder();
     }
 
-    public void setTemplateDir(String templateDir) {
+    public String getInputSpec() {
+		return inputSpec;
+	}
```
wing328: It seems to me the code is not indented properly (using tab instead of spaces). Would you please fix it?

regga-ws: Done !

Alexandre

2016-08-17 12:20 GMT+02:00 wing328 <notifications@github.com>:

alex-estela: Fixed !

wing328:
@regga-ws I'll review this weekend and let you know if I've further question.

wing328:
@regga-ws PR merged into master. Thanks for the contribution.

PR #3594, opened 2016-08-16
[Spring Boot] use port defined in spec file for server.port value

kevinchabreck:
Title says it all. Sets the `server.port` value in `application.properties` to the port defined in the spec file (or to 8080 by default, which matches the current behavior). The `serverPort` property is an existing value that is already being properly parsed by [SpringCodegen.java](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SpringCodegen.java#L233).


kevinchabreck:
I have [opened a new PR](https://github.com/swagger-api/swagger-codegen/pull/3605) to `master` with identical changes. Closing as this is now a duplicate.

PR #3599, opened 2016-08-17
Don't escape path in generated okhttp-gson client code

ggood:
The mustache templates for the other Java clients (e.g
Java/api.mustache) use 3 curly brackets around the path variable
so the path is not escaped. Make the okhttp-gson template
consistent with them.


wing328:
@ggood thanks for the contribution.

Issue #3081, opened 2016-06-08
Missing header in typescript-angular2


sebastianhaas:
##### Description
`typescript-angular2` doesn't set `Content-Type` and `Accept` in the Header which might cause problems on some servers. In my case, I wasn't able to POST data to a LoopBack (Express) backend.

##### Swagger-codegen version

Latest master, baed578010332a020d3d722dce6cb5f62b862d98

##### Generated code
The generated API class currently sends an empty header:
```typescript
public defaultHeaders : Headers = new Headers();
```

##### Command line used for generation
`java version "1.7.0_79"
Java(TM) SE Runtime Environment (build 1.7.0_79-b15)
Java HotSpot(TM) 64-Bit Server VM (build 24.79-b02, mixed mode)`

##### Steps to reproduce
Just look it up in the generated sample
https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/typescript-angular2/default/api/PetApi.ts

##### Suggest a Fix
I think we want to set `application/json` as default for all requests:
```typescript
    public defaultHeaders : Headers = new Headers({
      'Accept': 'application/json',
      'Content-Type': 'application/json'
    });
```
Fixed the problem in my case. I opened a PR with that two additional lines of code.

wing328:
Have you set a proper `consumes` and `produces` in the spec for the endpoints?

e.g. https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.yaml#L37

sebastianhaas:
My spec does specify proper `consumes` and `produces`:

```json
"host": "localhost:3000",
"basePath": "/api",
"schemes": ["http"],
"consumes": ["application/json", "application/x-www-form-urlencoded", "application/xml", "text/xml"],
"produces": ["application/json", "application/xml", "text/xml", "application/javascript", "text/javascript"],
```

I can't see any efforts made in the respective `.java` and `.mustache.` source to take these into account anyway.
I also do not think that any other content type than json would work for current solution. It is using `JSON.stringify()` to pass the body to Angular2's `Http` class.

https://angular.io/docs/ts/latest/api/http/Http-class.html

wing328:
@sebastianhaas thanks for sharing more. For other API clients (e.g. C#), we've a function to select the property content-type and accepts: 

[...]

PR #3603, opened 2016-08-17
Angular2 rc5

sebastianhaas:
Due to changes in the latest version of Angular ([2.0.0-rc.5](https://github.com/angular/angular/releases/tag/2.0.0-rc.5)), I had to make a few changes to the existing codegen to make it work again for me.
I also believe these changes would make a good starting point for #3081.

### RequestOptions
See https://angular.io/docs/ts/latest/api/http/index/RequestOptions-class.html. A few changes here, the HTTP method can now be specified using an enum instead of a string, which I did to make the client more robust. I also set the response content type to JSON, which is the only type of content we can handle at the moment.

### Emtpy body issue
https://github.com/angular/angular/issues/10612
See line [95](https://github.com/swagger-api/swagger-codegen/compare/master...sebastianhaas:angular2-rc5?expand=1#diff-174bce348e2584ed556f2d1ce0714665R95).
Supplying an empty body (actually an undefined body) with the `content-type` header set, will lead to some null reference error. This issue occurs in the current release only and has already been fixed for the next one.

### Headers shallow copy issue
https://github.com/angular/angular/issues/6845
See line [41](https://github.com/swagger-api/swagger-codegen/compare/master...sebastianhaas:angular2-rc5?expand=1#diff-174bce348e2584ed556f2d1ce0714665R41). 
At the moment, every endpoint uses `this.defaultHeaders` as starting point for its request, which causes all subsequent requests being extended by the headers applied to previous requests.
To get around this, I used `Header`'s copy constructor, which lacks a proper implementation atm since it only creates a shallow copy copying the underlying data structure by reference only. This should also be fixed for the next release, however, for now `new Headers(this.defaultHeaders.values());` is the way to go.

### Removed `use-strict`
I removed the `'use-strict'` directives since TypeScript will automatically add them from 1.8 on.
https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#modules-are-now-emitted-with-a-use-strict-prologue

### TypeScript template string to build URLs
I changed the template to use TypeScript template strings instead of `string.replace()`.
So
```typescript
const path = this.basePath + '/store/{storeId}/order/{orderId}'
   .replace('{' + 'storeId' + '}', String(storeId))
   .replace('{' + 'orderId' + '}', String(orderId));
```
becomes
```typescript
const path = this.basePath + `/store/${storeId}/order/${orderId}`;
```
which I find more readable.

### Consumes and produces
I added collections for consumes and produces for later use.

### Imports
I refactored imports according to Google's heroes tour for angular. https://angular.io/docs/ts/latest/tutorial/

## Open issues
* #3081 
* Proper rxjs import according to latest docs https://angular.io/docs/ts/latest/guide/server-communication.html#!#rxjs
* Use of the new modules system `NgModules`. Angular will stop supporting non modules at some point. https://angular.io/docs/ts/latest/guide/ngmodule.html
* Barrel imports: I had lots of problems with barrel imports lately. It appears that Angular's DI doesn't always like them.
* Naming conventions: Services should be in a file called `something.service.ts` etc.

I will try to submit a separate PR for those.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
@@ -1,7 +1,8 @@
 {{>licenseInfo}}
-import { Http, Headers, Response, URLSearchParams }          from '@angular/http';
-import { RequestMethod, RequestOptions, RequestOptionsArgs } from '@angular/http'; 
 import { Inject, Injectable, Optional }                      from '@angular/core';
+import { Http, Headers, URLSearchParams }                    from '@angular/http';
+import { RequestMethod, RequestOptions, RequestOptionsArgs } from '@angular/http';
+import { Response, ResponseContentType }                     from '@angular/http';
 
 import { Observable }                                        from 'rxjs/Observable';
 
@@ -34,14 +35,8 @@ export class {{classname}} {
      * {{notes}}
      {{#allParams}}* @param {{paramName}} {{description}}
      {{/allParams}}*/
-<<<<<<< 54fe7a731caec554d582b4b0a79defdc52751e05
-    public {{nickname}} ({{#allParams}}{{paramName}}{{^required}}?{{/required}}: {{{dataType}}}, {{/allParams}}extraHttpRequestParams?: any): Observable<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}{}{{/returnType}}> {
-        const path = this.basePath + '{{path}}'{{#pathParams}}
-            .replace('{' + '{{baseName}}' + '}', String({{paramName}})){{/pathParams}};
-=======
     public {{nickname}}({{#allParams}}{{paramName}}{{^required}}?{{/required}}: {{{dataType}}}, {{/allParams}}extraHttpRequestParams?: any): Observable<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}{}{{/returnType}}> {
         const path = this.basePath + `{{path}}`;
->>>>>>> Changes due to changes in Angular's http module.
 
         let queryParameters = new URLSearchParams();
         let headers = new Headers(this.defaultHeaders.values()); // https://github.com/angular/angular/issues/6845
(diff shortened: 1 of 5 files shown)


wing328:
@sebastianhaas thank you for your contribution. I believe this is a non-breaking change as the method signature is not changed.

I'll merge this PR tomorrow if no one has any feedback or question on this PR.

cc @Vrolijkx

sebastianhaas:
Shouldn't introduce breaking changes, no.

wing328:
@sebastianhaas when you've time, would you please rebase on the latest master to resolve the merge conflicts?

modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
```diff
@@ -1,14 +1,16 @@
 {{>licenseInfo}}
-import {Http, Headers, RequestOptionsArgs, Response, URLSearchParams} from '@angular/http';
-import {Injectable, Optional} from '@angular/core';
-import {Observable} from 'rxjs/Observable';
-import * as models from '../model/models';
-import 'rxjs/Rx';
+import { Injectable, Optional }                              from '@angular/core';
+import { Http, Headers, URLSearchParams }                    from '@angular/http';
+import { RequestMethod, RequestOptions, RequestOptionsArgs } from '@angular/http';
+import { Response, ResponseContentType }                     from '@angular/http';
```
Vrolijkx: Breaking change. ResponseContentType is new from Angular RC5 this means that you will be forced to upgrade to Angular RC5 when using this change which makes it breaking. Because Angular RC5 breaks against RC2 in a lot of ways.

Vrolijkx:
Looks great but i think these changes also require that the angular versions in the ```package.json``` file are also changed to ```"^2.0.0-rc.5"``` because it won't work with rc2 anymore.

sebastianhaas:
Thanks for taking a look @Vrolijkx, you are of course right, rc5 is required for these changes to work. But that's the point of the whole PR, and no other changes in the code should be required. I don't know what `swagger-codegen` policy on things like that looks like? I also have another PR in the making, introducing `ngModules` and different filenames, so that will certainly break existing code bases. Afaik, `http` in rc5 should be stable now, so I don't expect anymore breaking changes in the release.

Re `package.json`: Thanks for pointing that out, I haven't taken a look on the packaged version. Will do this later, and also rebase on master @wing328.

sebastianhaas:
Alright, looks good now @wing328

wing328:
@sebastianhaas PR merged into master. Thanks for the contribution.

PR #3604, opened 2016-08-17
[ObjC] Add version define and share default headers of each client

mateuszmackowiak:


wing328:
@mateuszmackowiak thanks for the PR. All the ObjC Petstore tests passed locally.

PR #3605, opened 2016-08-17
[Spring Boot] use port defined in spec file for server.port value

kevinchabreck:
Title says it all. Sets the `server.port` value in `application.properties` to the port defined by the `host` field in the swagger spec file. If no port is provided `server.port` is set to 8080 by default, which matches current template behavior. 

The `serverPort` property is an existing value that is already being parsed by [SpringCodegen.java](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SpringCodegen.java#L233).

### DIFF
--- a/modules/swagger-codegen/src/main/resources/JavaSpring/application.mustache
+++ b/modules/swagger-codegen/src/main/resources/JavaSpring/application.mustache
@@ -1,3 +1,3 @@
 springfox.documentation.swagger.v2.path=/api-docs
 server.contextPath={{^contextPath}}/{{/contextPath}}{{#contextPath}}{{contextPath}}{{/contextPath}}
-#server.port=8090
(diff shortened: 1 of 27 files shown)


wing328:
@kevinchabreck thanks for the PR. Please run `./bin/spring-all-pestore.sh` to update the Petstore sample for Spring MVC and Boot so that CI can test the change.

kevinchabreck:
@wing328 thanks for the response. I ran that script and pushed the changes.

wing328:
@kevinchabreck I'll review tomorrow (Fri) and let you know if I've further question. Thanks again for the contribution.

wing328:
@kevinchabreck PR merged into master. Thanks for the contribution.

PR #3613, opened 2016-08-19
Uses this.{{name}} to scope instance

mikesol:
Otherwise, if {{classVarName}} is the same as a class member name, the function will not work.


wing328:
@mikesol thanks for the PR. Please run the following to update the Android Petstore API clients (Volley and Apache HTTP client) so that the CIs can test the change:
```
bin/android-petstore.sh
bin/android-petstore-httpclient.sh
```
The one you modified is for `httpclient`. Please do the same for `volley` (default HTTP lib) as well.
```
modules/swagger-codegen/src/main/resources/android/libraries/volley/model.mustache
```

To get a list of options for customizing Android library, run `java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar config-help -l`

wing328:
FYI. I've updated the volley's model to use `this` and update Android's Petstore sample.

Issue #3611, opened 2016-08-18
Possible swagger hub codegen issue with Enums: @JsonValue annotation is missing from toString()


ralgor14:
Description

The @JsonValue annotation is missing when codegen creates java code (Server->jaxrs) for Enums.
This annotation should be on toString() method.
The following:
        "state": {
          "type": "string",
          "enum": [
            "active",
            "inactive"
          ]
        }

causes creation of the following Enum code:
  public enum StateEnum {
    ACTIVE("active"),
    
    INACTIVE("inactive");

    private String value;

    StateEnum(String value) {
      this.value = value;
    }

    @Override
     public String toString() {
      return String.valueOf(value);
    }
  }

Which causes "active" and "inactive" in JSON files to be unrecognized.
In order to correct it the annotation @JsonValue should be added before toString():
    @Override
    @JsonValue
    public String toString() {
      return String.valueOf(value);
    }

It was OK in previous versions of swagger hub.
Annotations are used by jackson's fasterXml package.

wing328:
@ralgor14 thanks for reporting the issue and suggesting a fix. I'll look into this tomorrow (Friday) and hopefully file a PR for the fix.

wing328:
@ralgor14 I've filed #3615 to fix the issue.

wing328:
@ralgor14 PR merged into master. Please pull the latest to give it a try.

ralgor14:
I do not see  this version under https://github.com/swagger-api/swagger-codegen
I can only download version 2.2.1 from 2016, Aug 7

wing328:
@ralgor14 please build the JAR locally (`mvn clean install`) before 2.2.2 (next stable release) is released.

ralgor14:
Because of security issues I cannot download maven into my host.
Could you please send me the compiler exe file just rename it from exe to something else?
Or even better create a temporary link for unreleased version that can be downloaded in the same way as a regular released version?

wing328:
Can you use https://generator.swagger.io? If yes, I can setup another online code generator to run the latest version of Swagger Codegen.

ralgor14:
I didn't succeed.
Following an input for swagger: Could you please compile it with your latest version and send me the Server->jaxrs created java files?
{
  "swagger": "2.2.1",
  "info": {
    "description": "demo",
    "version": "1.1.11",
    "title": "demo",
    "contact": {
      "name": "Ruth Algor",
      "email": "ruth_a@rad.com"
    }
[...]

Issue #3509, opened 2016-08-02
Possible swagger hub codegen issue with Spring Data Mongo


mherb63:
##### Description
Mongo not able to map swagger code generated fields with domain model code generated from swaggerhub.

The following JSON is used on swaggerhub:
        "frameTransmissionPeriod": {
          "description": "SLM period for frame transmission",
          "type": "string",
          "enum": [
            "_10ms",
            "_100ms",
            "_1sec",
            "_10sec"
          ],
          "default": "_100ms"
        },

And here is the generated code (Server->Spring), note the uppercase "MS" and "SEC"
  public enum FrameTransmissionPeriodEnum {
    _100MS("_100ms"),
    _1SEC("_1sec"),
    _10SEC("_10sec");

    private String value;

    FrameTransmissionPeriodEnum(String value) {
      this.value = value;
    }

    @Override
    @JsonValue
    public String toString() {
      return value;
    }
  }

When reading from Spring Data mongo we have what's in Mongo:
		"ethOamDmmCfg" : {
			"testId" : "5555",
			"pmCosId" : 1,
			"vlanId" : 100,
			"frameTransmissionPeriod" : "_100ms",
			"frameSize" : "1028"
		},

And the following Mongo error:
2016-08-02 15:46:21.804 ERROR 79727 --- [nio-8680-exec-8] c.c.l.p.exception.ApiExceptionHandler    : No enum constant com.ciena.lion.mef.pm.domain.EthOamDmmCfg.FrameTransmissionPeriodEnum._100ms
java.lang.IllegalArgumentException: No enum constant com.ciena.lion.mef.pm.domain.EthOamDmmCfg.FrameTransmissionPeriodEnum._100ms
	at java.lang.Enum.valueOf(Enum.java:238)
	at org.springframework.data.mongodb.core.convert.MappingMongoConverter.getPotentiallyConvertedSimpleRead(MappingMongoConverter.java:819)

##### Swagger-codegen version
swaggerhub.com

##### Suggest a Fix
at some point in time, the generated Spring code has upper-cased the enums and I believe this has caused the issue.  If I change "MS" and "SEC" to "ms" and "sec" in the generated domain model code, Mongo is happy.

wing328:
@mherb63 are you saying the following enum name and value mapping is incorrect?
```
_100MS("_100ms"),
_1SEC("_1sec"),
_10SEC("_10sec");
```
If yes, what would the correct mapping looks like?

[...]

PR #3615, opened 2016-08-19
[Java][Spring][Inflector][Jax-RS] To fix various enum issues

wing328:
- [spring] to fix deserialization issue by adding `fromValue` method: https://github.com/swagger-api/swagger-codegen/issues/3509
- [jaxrs][inflector] to fix the serialization issue by adding @JsonValue: https://github.com/swagger-api/swagger-codegen/issues/3611

TODO: add test cases to cover the enum issues in spring, inflector and jaxrs moving forward.

### DIFF
--- a/bin/jaxrs-jersey1-petstore-server.sh
+++ b/bin/jaxrs-jersey1-petstore-server.sh
@@ -26,7 +26,7 @@ fi
 
 # if you've executed sbt assembly previously it will use that instead.
 export JAVA_OPTS="${JAVA_OPTS} -XX:MaxPermSize=256M -Xmx1024M -DloggerPath=conf/log4j.properties"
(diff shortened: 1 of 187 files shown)


deveshpujari:
It does not work for language jaxrs .
I tried using 2.2.3 and 2.3.0 lastest snapshot also.

public enum HealthType {

  OVERALL("overall"),

  CONNECTED("connected"),

  ONBOARDED("onboarded");

  private String value;

  HealthType(String value) {
    this.value = value;
  }

  @Override
  public String toString() {
    return String.valueOf(value);
  }

  @JsonCreator
  public static HealthType fromValue(String text) {
    for (HealthType b : HealthType.values()) {
      if (String.valueOf(b.value).equals(text)) {
        return b;
      }
    }
    return null;
  }
}

PR #3622, opened 2016-08-22
For Qt5 server generator, add BaseIntegerProperty to default values

### LINKED ISSUE #3192: [Qt5/C++] Wrong default value for integers
##### Description

datatypes `qint32` and `qint64` are rendered with default value `= NULL;` but should be `0`

##### Swagger-codegen version

latest from editor.swagger.io

##### Suggest a Fix

I think it somehow has to do with https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Qt5CPPGenerator.java#L254 but for some reason it doesnt walk into the `} else if (p instanceof IntegerProperty) `

YetAnotherMinion:
Fixes #3192 

**edit:** I have barely touched Java at all, so please double check that this change makes sense. I just made the default values stop being NULL, without any understanding of how the entire codegen project goes together.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Qt5CPPGenerator.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Qt5CPPGenerator.java
@@ -267,12 +267,14 @@ public class Qt5CPPGenerator extends DefaultCodegen implements CodegenConfig {
             return "0.0";
         } else if (p instanceof FloatProperty) {
             return "0.0f";
-        } else if (p instanceof BaseIntegerProperty) {
-            return "0";
         } else if (p instanceof IntegerProperty) {
             return "0";
         } else if (p instanceof LongProperty) {
             return "0L";
+        } else if (p instanceof BaseIntegerProperty) {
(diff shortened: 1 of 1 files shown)


wing328:
@YetAnotherMinion thanks for the PR and the note. I'll do a test later to verify.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Qt5CPPGenerator.java
```diff
@@ -266,6 +267,8 @@ public String toDefaultValue(Property p) {
             return "0.0";
         } else if (p instanceof FloatProperty) {
             return "0.0f";
+        } else if (p instanceof BaseIntegerProperty) {
```
wing328: @YetAnotherMinion thanks for the PR. I would put this line after the IntegerProperty and LongProperty `else if` check as both IntegerProperty and LongProperty extends BaseIntegerProperty.

wing328:
@YetAnotherMinion thanks for the update. The change now looks good to me.

PR #3624, opened 2016-08-22
[Swift] Include HTTP status code and body data in response errors

jgavris:
This is useful when differentiating different error responses. Example:

if let e = error {
  switch e {
    case ErrorResponse.Error(400, _, _): responseLabel.text = "Unknown error"
    case ErrorResponse.Error(403, _, _): responseLabel.text = "Not logged in"
    default: responseLabel.text = "\(e)"
  }
}

The existing error is straight from Alamofire and only includes the HTTP
status code in the body of the message (not broken out as a separate value).
It also does not include the response body, which may be useful.


samples/client/petstore/swift/default/SwaggerClientTests/SwaggerClientTests/PetAPITests.swift
```diff
@@ -74,15 +74,17 @@ class PetAPITests: XCTestCase {
...
-            guard let error = error where error._code == -6006 else {
+            guard let error = error else {
                 XCTFail("error deleting pet")
                 return
             }
-            
-            expectation.fulfill()
+
+            switch error {
+            case ErrorResponse.Error(200, _, _):
```
jaz-ah: Will there ever be an error w/ a 200 status?  This reads a little odd here.

jgavris: I agree, this is awkard. It's result of the current swift implementation, which doesn't handle `Void` responses currently. Let me see if I can make that go away.

jgavris: @jaz-ah @wing328 Updated. This is now a breaking change with `Respose.body` being optional now to handle `Void` responses. Covers all the petstore examples without ugly hacks now :)

samples/client/petstore/swift/promisekit/SwaggerClientTests/SwaggerClientTests/PetAPITests.swift
```diff
@@ -61,22 +61,8 @@ class PetAPITests: XCTestCase {
...
-                // length." UserInfo={NSLocalizedFailureReason=JSON could not be serialized. Input data was nil or zero
-                // length.}
-                let error = errorType as NSError
-                if error.code == -6006 {
-                    expectation.fulfill()
-                } else {
-                    XCTFail("error logging out")
-                }
+        PetAPI.deletePet(petId: 1000).then {
+            expectation.fulfill()
```
jaz-ah: do we still have to do some type of error checking here?

samples/client/petstore/swift/default/SwaggerClientTests/SwaggerClientTests/PetAPITests.swift
jaz-ah: much better thx!

samples/client/petstore/swift/promisekit/SwaggerClientTests/SwaggerClientTests/PetAPITests.swift
jgavris: There's no more error, and we're using the `then` block.

jaz-ah: ah ok cool thx

jaz-ah:
+1 good to go @wing328

wing328:
Thanks all for the work. All the Petstore tests passed locally.

jgavris:
🎈

PR #3626, opened 2016-08-22
Adds flattener for ComposedModel

diegode:


wing328:
@diegode thanks for the PR. The change looks good to me. It would be nice to add test case(s) in [modules/swagger-codegen/src/test/java/io/swagger/codegen/java/JavaModelTest.java](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/java/io/swagger/codegen/java/JavaModelTest.java) to cover the change in another PR.

Issue #3636, opened 2016-08-24
[All] Add a supportsMixin property to codegens


cbornet:
* currently if supportsInheritance is true, the properties from parent and interfaces will not be included in vars.
* currently if supportsInheritance is true, the properties from self, parent and interfaces will all be included in allVars.
* this works well for Javascript because it uses some "mixin" import mecanism
* this should also work well for languages that support mixins/multi-inheritance (python, ruby, C++, ...)
* but it doesn't work well with Java because it doesn't support mixins/multi-inheritance (java interfaces can't be used because jackson and gson need POJOs/javaBeans).

The consequence is that since we don't have the interface properties in vars, we have to duplicate the code everywhere in the template to include both vars and interfaceModels.allVars
This could be mitigated by putting some parts of the code in imported templates but that would also make the template much harder to read.
So my idea is to add a supportsMixin property for languages that can use it (eg. Javascript)

@wing328 I can do the PR if OK.

cc @demonfiddler

demonfiddler:
Apologies for the delay in responding – have been swamped both at work and at home.

 

For the first bullet I think you meant “currently if supportsInheritance is false, the properties from parent and interfaces will not be included in vars.”

 

I see what you mean in the Java case. It is true that Mustache templates, being logic-less, require all business logic to be implemented by the caller. If you’re confident that your change can address the requirements of JavaScript, Java and other languages whilst preserving or rectifying existing (mis-)behaviour as appropriate then I wish you good luck with the PR.

 

Cheers,

 

--A

wing328:
PR merged into master

Issue #3544, opened 2016-08-05
If allOf used in a base type, inherited types are missing "extends <basetype>" in generated Java code


ggood:
##### Description

When a type is constructed using allOf, and also includes a discriminator property (so it can serve as the base class for inherited types) the inherited types do not extend from it.

If instead, the "allOf" construct is collapsed and all the properties inlined, the inheritance works as expected.

In the example below, the generated code for LevelTwoTypeOne is missing the "extends LevelOne", while the generated code for LevelTwoTypeTwo includes "extends InlinedLevelOne" as expected.

##### Swagger-codegen version

2.2.0

##### Swagger declaration file content or url

 ```yaml
---
swagger: '2.0'
info:
  version: 0.0.0
  title: Simple API
paths:
  /:
    get:
      responses:
        200:
          description: OK
definitions:
  Base:
    type: object
    properties:
      name:
        type: string

  LevelOne:
    allOf:
    - $ref: '#/definitions/Base'
    - type: object
      discriminator: resource_type
      properties:
        color:
          type: string
        resource_type:
          type: string
      required: [resource_type]

  InlinedLevelOne:
    type: object
    discriminator: resource_type
    properties:
      name:
        type: string
      color:
        type: string
      resource_type:
        type: string
    required: [resource_type]

  LevelTwoTypeOne:  
    type: object
    allOf:
    - $ref: '#/definitions/LevelOne'
    - type: object
      properties:
        odor:
          type: string

  LevelTwoTypeTwo:  
    type: object
    allOf:
    - $ref: '#/definitions/InlinedLevelOne'
    - type: object
      properties:
        height:
          type: string
    ```

##### Command line used for generation

export JAVA_OPTS="${JAVA_OPTS} -XX:MaxPermSize=256M -Xmx1024M -DloggerPath=conf/log4j.properties"
ags="$@ generate -i modules/swagger-codegen/src/test/resources/2_0/inheritance.yaml -l java --language-specific-primitives integer,Long -o samples/client/nsx/java/default -DhideGenerationTimestamp=true"

java $JAVA_OPTS -jar $executable $ags

##### Steps to reproduce

After running swagger-codegen, I observe:

LevelTwoTypeOne.java's class declaration is:

public class LevelTwoTypeOne   {

I expected it to inherit from LevelOne.

LevelTwoTypeTwo's class declaration is:

public class LevelTwoTypeTwo extends InlinedLevelOne  {

as expected.

##### Related issues

[...]

PR #3637, opened 2016-08-24
[All] Fix some inheritance/composition issues and add allOf unit tests

### LINKED ISSUE #3629: Java (spring) server code generation of  allOf field in yml seems broken
Using 'allOf' no longer produces classes that extend the allof reference. 
Tested in the current master branch (2.2.2), trying to generate spring-mvc server code.
When using the live swagger editor (2.3.0) things work fine however.

cbornet:
Also adds a supportMixins property for languages that support mixins/multiple inheritance.

Fix #3629 #3544 #3474 #3636

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -100,6 +100,7 @@ public class DefaultCodegen {
     protected List<CliOption> cliOptions = new ArrayList<CliOption>();
     protected boolean skipOverwrite;
     protected boolean supportsInheritance;
+    protected boolean supportsMixins;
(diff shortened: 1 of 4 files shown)


cbornet:
@wing328 we are now selecting the interface that has a discriminator as parent. But for models that don't have a discriminator, should we select the first interface as parent like it was done before ? 
I don't have a strong opinion on this: 
* on one hand, it allows to have some model inheritance without discriminator which some people may find useful. 
* on the other hand, if you have several interfaces, it's a bit disturbing to have different implementation for them. But I guess most of the time, models will have only one interface so that may not be a big issue. 

Eventually, we could also have some "x-extends" or "x-parent" vendor extension where users could explicitely define the interface they want to be used as parent model.

alexandern:
Wondering if addProperties should account for this issue
https://github.com/swagger-api/swagger-codegen/issues/3829

cacti77:
Will this PR fix issue #3904?

Issue #3644, opened 2016-08-25
Go generation is somewhat broken.


AmandaCameron:
##### Description

- [ ] Swagger types that are free-form, such as the generator API's `GeneratorInput.Spec` field are being encoded as an undefined type `Object` -- I'd recommend these get defined as `interface{}`
- [ ] Non-JSON returning functions, such as `ClientsApi.DownloadFile` are always returning an error due to trying to unmarshal a non-json type.

##### Swagger-codegen version

Docker image latest for generator API. swagger.json self-reports as `"2.2.2-SNAPSHOT"`

##### Swagger declaration file content or url

Swagger generator's swagger.json causes these errors.

##### Command line used for generation

API Endpoint

wing328:
Please provide a spec for reproducing the issue.

AmandaCameron:
http://generator.swagger.io/api/swagger.json causes both of the issues I've
mentioned in this issue.

wing328:
Thanks. @guohuang when you've time, would you please take a look?

PR #3645, opened 2016-08-26
[Go] fixing code gen not recognizing 'object' type issue

guohuang:
this is only fixing the first reported issue in #3644 

@wing328 please review


wing328:
@guohuang thanks for the prompt fix. I'll review this weekend.

cc @neilotoole

wing328:
Tested with `type: object` and didn't encounter any compilation error.

PR #3652, opened 2016-08-27
Perl example fixes

scop:


wing328:
@scop thanks for the PR to improve the Perl sample code. I've also updated the Perl Petstore sample to keep the sample up-to-date.

PR #3653, opened 2016-08-27
java: Fix model package import in example

scop:


wing328:
@scop the change looks good to me. 

Merged into master.

PR #3655, opened 2016-08-28
perl: Sync default user agent with other languages

scop:


wing328:
@scop thanks for the enhancement 👍

PR #3657, opened 2016-08-29
Fix to case of propertyClass which pukes on most checkstyle rules in …

brentryan:
PropertyClass should be camelCase.  Simple fix to a small edge case.


wing328:
@brentryan thanks for the fix.

PR #3671, opened 2016-08-29
Python pattern escaping fixes

scop:
### PR checklist

- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (NOTE: no need to update *all* Petstore sample. For instance, only need to update PHP Petstore sample by running `./bin/php-petstore.sh` if updating the PHP code generator or PHP client's mustache templates)

### Description of the PR

Python pattern output has two flaws: 1) replacing of `'` with `\'` doesn't actually happen because of a missing backslash, causing syntax errors if pattern contains a single quote, and 2) the pattern output is HTML entity escaped which is just broken, e.g. `<` in a pattern turns to `&lt;` in the output regex.

With regards to updating the petstore, I ran the script and it caused some changes (`api_key` auth blocks moving around, possibly in error), but that's unrelated to these changes so I'll let someone else debug and fix that if necessary.


scop:
Nevermind the comment about petstore update, seems harmless and similar one was recently done for ruby, so included here too for completeness.

wing328:
@scop thanks for the contribution. PR merged into master.

PR #3672, opened 2016-08-29
Don't HTML escape PHP pattern output

scop:
### PR checklist

- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (NOTE: no need to update *all* Petstore sample. For instance, only need to update PHP Petstore sample by running `./bin/php-petstore.sh` if updating the PHP code generator or PHP client's mustache templates)

### Description of the PR

php and lumen patterns should not be HTML entity escaped in output.

With regards to the updated lumen petstore, I see line 106 in  samples/server/petstore/lumen/lib/app/Http/Controllers/FakeApi.php contains `"/^[A-Z]."` passed to preg_match which will cause a `preg_match(): No ending delimiter '/' found` warning, but that's something not caused by this change.


samples/server/petstore/lumen/lib/app/Http/Controllers/FakeApi.php
```diff
@@ -100,13 +100,13 @@ public function testEndpointParameters()
...
         $double = $input['double'];
 
-        if (!isset($input['string'])) {
-            throw new \InvalidArgumentException('Missing the required parameter $string when calling testEndpointParameters');
+        if (!isset($input['pattern_without_delimiter'])) {
+            throw new \InvalidArgumentException('Missing the required parameter $pattern_without_delimiter when calling testEndpointParameters');
         }
-        if (!preg_match("/[a-z]/i", $input['string'])) {
-            throw new \InvalidArgumentException('invalid value for $string when calling FakeApi.testEndpointParameters, must conform to the pattern /[a-z]/i.');
+        if (!preg_match("/^[A-Z].", $input['pattern_without_delimiter'])) {
```
scop: This regex is missing a trailing slash for some reason.

wing328: Noted. I'll take a look.

wing328: Did some debugging but no luck. Will open a new issue to track this.

PR #3683, opened 2016-08-31
Added functionality to handle optional parameters for Scala

jyotsnakaran:
### PR checklist

### Description of the PR

Added functionality to handle optional parameters for Scala
#3665


wing328:
CI (travis) failed. Please check the result: https://travis-ci.org/swagger-api/swagger-codegen/jobs/156459865

You can also repeat the issue by running `sbt test` locally under the Scala Petstore folder.

modules/swagger-codegen/src/main/resources/scala/api.mustache
```diff
@@ -53,27 +53,57 @@ class {{classname}}(val defBasePath: String = "{{basePath}}",
     {{/isPrimitiveType}}
     {{/required}}
     {{/allParams}}
-    {{#queryParams}}if(String.valueOf({{paramName}}) != "null") queryParams += "{{baseName}}" -> {{paramName}}.toString
+    {{#queryParams}}
+    {{#required}}
+    if(String.valueOf({{paramName}}) != "null") queryParams += "{{baseName}}" -> {{paramName}}.toString
```
wing328: I don't think that's the correct way to do a null check in scala

It would be nice to to do the following:
- add an else clause to throw an exception (e.g. java.lang.IllegalArgumentException)
- add the same null check on required parameter for other type of parameters (e.g. header, form, body)

jyotsnakaran:
Actually, I was just focusing on optional parameters. For required ones, the code is same as before, I have just made changes for the optional params.

And the check for null parameters is already there on line no: [51](https://github.com/jyotsnakaran/swagger-codegen/blob/4aa454abe2fab11d84d2e8086656cc1a74bcc077/modules/swagger-codegen/src/main/resources/scala/api.mustache#L51) which throwing exception for missing required parameters.

Please have a look for the same and let me know.

wing328:
Understood (I didn't mean it's your fault, just want to point out something not right during my review).

I'll take another look and let you know if I've further question.

jyotsnakaran:
No, that's completely fine. :) I was just letting you know about the already existing null parameter handling in the code, so that you can have a look over it.
Let me know, if you need anything from my side.

wing328:
PR merged into master. Thanks for the contribution.

jyotsnakaran:
@wing328 Thanks

PR #3689, opened 2016-08-31
Python fixes

scop:
### PR checklist

### Description of the PR

Implements collection formatting to Python. Previously, collections were not formatted at all, but serialized as e.g. literal `['foo', 'bar']`.

This implements it for form data only for now; I don't think it can be implemented sanely without API changes for query string data. The `query_params` arg for `api_client.call_api` would have to be changed from dict to a list (like `post_params` already is). Is that considered a breaking API change that needs to go to the 2.3.0 branch?


wing328:
@scop looks good to me. Thanks for the contribution.

cc @scottrw93

PR #3697, opened 2016-09-01
Python collection formatting fixes/support

scop:
### PR checklist

### Description of the PR

This reworks support for form data collection formatting, and adds collection formatting support for paths, headers, and query string. The rework is because the previous implementation would fail for non-string collection items, and to provide basis for the other types.

This _does_ contain breaking changes for the methods in `api_client`; however if I understand correctly, the breaking changes meant in the PR checklist would be ones that affect client code that uses the generated api classes. This one doesn't contain any of them, so I suppose it's good for `master`.


wing328:
@scop thanks for the note about the breaking change in which I tend to agree. Later we'll see if any developers have feedback/question on this in particular.

PR #3699, opened 2016-09-01
java: Javadoc fixes

scop:
Simple changes to template, caused lots in generated code (+ the Java petstores seemed to be lagging behind already).


wing328:
I did a grep to confirm only JSON.mustache (okhttp-gson) uses @throw (which you've fixed in this PR) and ObjC uses @throw, which is correct according to https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/Exceptions/Tasks/RaisingExceptions.html

PR #3700, opened 2016-09-01
php: Fix syntax error when pattern contains a single quote

scop:


wing328:
@scop I saw you update the path to `autoload.php`. Did you use composer to install the PHP SDK and other dependencies?

wing328:
cc @arnested

scop:
No, I didn't install anything extra, just ran a PHP script in some dir to try things out.

wing328:
@scop then I think you should use the following instead to load the PHP API client:
```
require_once('/path/to/package-name/autoload.php');
```
Ref: https://github.com/stripe/stripe-php#manual-installation

scop:
Ok, PR rebased and autoload related changes dropped, just the syntax error fix remains.

wing328:
@scop actually the README.md already covers the manual installation procedure: 
https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/php/SwaggerClient-php/README.md#manual-installation

Currently, the PHP example code is not smart enough to tell whether the installation is done manually or via composer.

In the example code, would adding a commented line for manual installation meet your requirement? Something like the following:
```
<?php
//require_once('/path/to/SwaggerClient-php/autoload.php'); // for manual installation
require_once(__DIR__ . '/vendor/autoload.php');

$api_instance = new Swagger\Client\Api\FakeApi();
...
```

scop:
The README.md doc is ok, I'm pretty much indifferent wrt adding the comment. **But** if it is added, then I think a comment should be added to the other require_once vendor line as well, to make clear what that one is for, and that only one of those lines should be uncommented.

wing328:
@scop good point. Let's keep it as it's for now. If there're more report of issue due to confusion on how to require the lib, we'll revisit this.

PR #3702, opened 2016-09-01
ruby: Fix syntax error when pattern contains a single quote

scop:


wing328:
@scop thanks for the fix, which looks good to me.

Issue #3439, opened 2016-07-22
[Java] Serialize OffsetDateTime in UTC


cbornet:
See #2138

PR #3704, opened 2016-09-01
[Java] Serialize OffsetDateTime in UTC timezone

cbornet:
### PR checklist

### Description of the PR

See #3439

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/feign/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/feign/ApiClient.mustache
@@ -1,5 +1,10 @@
 package {{invokerPackage}};
 
+{{#java8}}
+import java.time.OffsetDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+{{/java8}}
 import java.util.LinkedHashMap;
 import java.util.Map;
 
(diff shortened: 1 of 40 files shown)


cbornet:
I'm first doing it for feign to get your comments. Then I'll do it for the other frameworks.

modules/swagger-codegen/src/main/resources/Java/libraries/feign/ApiClient.mustache
```diff
@@ -28,6 +34,10 @@ import {{invokerPackage}}.auth.OAuth.AccessTokenListener;
 public class ApiClient {
   public interface Api {}
 
+  //Formatter with fixed length and timezone
+  public static final DateTimeFormatter ISO_FIXED_FORMAT =
```
cbornet: Issue: this should go in a "java8" block. Fixing it.

cbornet:
I just read from rfc3339
> The offset between local time and UTC is often useful information.
   For example, in electronic mail (RFC2822, [IMAIL-UPDATE]) the local
   offset provides a useful heuristic to determine the probability of a
   prompt response.  Attempts to label local offsets with alphabetic
   strings have resulted in poor interoperability in the past [IMAIL],
   [HOST-REQ].  As a result, RFC2822 [IMAIL-UPDATE] has made numeric
   offsets mandatory.

So I'm not sure about this PR in the end.

cbornet:
On further thought, the best would probably be to expose the date formatter so the user can configure it the way he wants (milli/micro/nano precision, UTC/time-offset, ...). I'll see if that's easily doable.

wing328:
@cbornet totally agree that's a better approach and that's what we did for the C# API client: https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/csharp/SwaggerClient/src/IO.Swagger/Client/Configuration.cs#L288-L290

cbornet:
Closing for now. I'll see how to expose the formatter config.

Issue #3703, opened 2016-09-01
[Ruby, PHP] Errors with slashes in pattern


scop:
##### Description

For this in YAML syntax:
```yaml
pattern: '^a/b$'
```
...the PHP code generated is:
```php
if (!preg_match("/^a/b$/", ...
```
...and for Ruby:
```ruby
Regexp.new(/^a/b$/)
```
The problem with both is that the slash between `a` and `b` should be backslash escaped to avoid errors. I tried this in `DefaultCodegen.addRegularExpressionDelimiter`:
```java
return "/" + pattern.replace("/", "\\/") + "/";
```
...but it makes no difference on the eventual output.

##### Swagger-codegen version

Current master

wing328:
Your fix should work.
```
scala> var a = "^a/b$"
a: String = ^a/b$

scala> a.replace("/", "\\/")
res0: String = ^a\/b$
```
Please submit a PR and I'll do some tests

scop:
I agree it _should_ work, but as said, in my tests it did not end up resulting any changes to those problematic ruby and php code spots -- for some reason, the slash remained unescaped there. Anyway, #3707 is the PR.

wing328:
@scop I did some tests with your PR and have no idea either as it failed to escape forward slash.

@cbornet when you've time, I wonder if you can take a look to see if you've any clue why the replace does not work as intended

ePaul:
I debugged this a bit ... the problem might be that the result of this method gets later passed through DefaultCodegen.escapeText, which does (between others) this: `.replace("\\/", "/"))` – i.e. it undoes the transformation.

input | method | output 
---|------|------
`[A-Z]/.*` | `addRegularExpressionDelimiter` | `/[A-Z]\/.*/`
`/[A-Z]\/.*/` | `escapeJava(...)` | `/[A-Z]\\/.*/`
`/[A-Z]\\/.*/` | `.replace("\\/", "/"))` | ` /^[A-Z]\/.*/`
` /^[A-Z]\/.*/` | `unescapeJava(...)` | `/^[A-Z]/.*/`
`/^[A-Z]/.*/` | `.replaceAll("[\\t\\n\\r]", " ")` | `/^[A-Z]/.*/`
`/^[A-Z]/.*/` | `.replace("\\", "\\\\")` | `/^[A-Z]/.*/`
`/^[A-Z]/.*/` | .replace("\"", "\\\"") | `/^[A-Z]/.*/`
[...]

PR #3707, opened 2016-09-02
Backslash-escape slash when adding leading and trailing slashes to patterns

scop:
See #3703 for discussion


wing328:
I've filed a PR to fix the issue with escaping forward slash in the latest master. Please give it a try.

PR #3710, opened 2016-09-02
[PHP] Corrected PHPDoc type declarations

BeneficialName:
Some PHPDoc declarations was mistyped or just missing


wing328:
@BeneficialName thanks for the fix. I've used `phpdoc` to verify as well.

Issue #3690, opened 2016-08-31
[Go] Add fields URL, Method, Payload, and Operation to APIResponse


neilotoole:
##### Description

As it currently stands, there's no mechanism for a client of the generated code to determine what URL, HTTP method, or swagger operation was invoked when examining the returned `APIResponse` object. This information is readily available, and is useful for the client for debugging/logging. Additionally, because the generated code does not currently handle swagger-specified errors, there is no mechanism for the caller to access the response body payload. Having access to the payload in the `APIResponse` will allow the caller to inspect the response payload and perform some intelligent action.

##### Swagger-codegen version

`master`

##### Suggest a Fix

Create fields `URL`, `Method`, `Payload` and `Operation` in `APIResponse`.

PR #3711, opened 2016-09-02
#3690 adding RequestURL, Method, Operation and Payload fields to APIResponse

neilotoole:
### PR checklist

### Description of the PR

Addresses #3690 

In case it isn't clear in #3690, the motivation for making `RequestURL`, `Method` etc available on the `APIResponse` object, is that sometimes the embedded `*http.Response` can be `nil`, therefore the caller cannot access this information.

Filed against `master` as it does not contain breaking changes.


ePaul:
@neilotoole I don't really have any comments about the PR's content, but could you mention in the initial comment that it is for the language Go? (I.e. replace `{LANG}` with `Go` where applicable.) (Possibly also mention it in the title.) This makes it easier for me to filter out the "uninteresting" Pull requests.

modules/swagger-codegen/src/main/resources/go/api.mustache
```diff
@@ -135,11 +136,20 @@ func (a {{{classname}}}) {{{nickname}}}({{#allParams}}{{paramName}} {{{dataType}
...
+
+	var localVarURL, _ = url.Parse(localVarPath)
+	localVarURL.RawQuery = localVarQueryParams.Encode()
+	var localVarAPIResponse = &APIResponse{Operation: "{{operationId}}", Method: localVarHttpMethod, RequestURL: localVarURL.String()}
+	if localVarHttpResponse != nil {
+		localVarAPIResponse.Response = localVarHttpResponse.RawResponse
+		localVarAPIResponse.Payload = localVarHttpResponse.Body()
+	}
+
+if err != nil {
```
wing328: @neilotoole this line is not indented properly.

wing328:
Please rebase on the latest master to resolve the merge conflicts.

cc @guohuang

wing328:
I've made some minor fixes to indentation and merged the changes into the master.

Issue #3712, opened 2016-09-02
Invalid array enum type validation in Python client


bkryza:
##### Description

Python client generator creates invalid enum validation code in case the enum is a string type inside an array, for instance this data type:

```yaml
  EnumStringRequest:
    required:
      - values
    properties:
      values:
        type: array
        items:
          type: string
          enum:
            - value1
            - value2
            - value3
```

generates the following validation code in the setter:

```python
        allowed_values = ["value1", "value2", "value3"]
        if values not in allowed_values:
            raise ValueError(
                "Invalid value for `values` ({0}), must be one of {1}"
                .format(values, allowed_values)
            )
```

where `values` is an instance of `list`, which raises an error even if `values` contains valid elements from the `allowed_values` list.

##### Swagger-codegen version

branch master, commit: 7a4a430c4302a1601793df96134b69a2cf29a4da

##### Swagger declaration file content or url

```yaml
swagger: 2
info:
  title: Array enum test
  description: API for testing generation of enum types in arrays
  version: "1.0.0"
host: "localhost:8002"
schemes:
  - http
basePath: /v1
produces:
  - application/json
paths:
  /enum-strings:
    post:
      description: "Accepts an array of strings"
      operationId: "enum-strings"
      parameters:
      - name: enumarray
        in: body
        schema:
          $ref: EnumStringRequest
definitions:
  EnumStringRequest:
    required:
      - values
    properties:
      values:
        type: array
        items:
          type: string
          enum:
            - value1
            - value2
            - value3
```

##### Command line used for generation

```bash
cd swagger-coden

vim array-enum.yaml # Paste the above YAML example

mvn package
 
java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -i ./array-enum.yaml -l swagger -o ./output/swagger

java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar   generate -i output/swagger/swagger.json -l python -o ./output/array-enum/python
```

##### Suggest a Fix
The problem is in the `modules/swagger-codegen/src/main/resources/python/model.mustache` in section checking for enum types, which should also check if the type is a container...

P.S. Will try to make a simple PR.

wing328:
PR merged.

PR #3713, opened 2016-09-02
Fix for #3712 - invalid enum array validation in Python client mustache model

bkryza:
### PR checklist

### Description of the PR

A simple fix for issue #3712.

The Python client `model.mustache` template now properly validates arrays of strings which are `enum` types.


wing328:
@bkryza thanks for the PR. Please run ./bin/python-petstore.sh to update the Petstore sample so that the CIs will test the change.

cc @scottrw93

bkryza:
@wing328 Done.

wing328:
@bkryza for `modules/swagger-codegen/src/test/resources/2_0/array-enum.yaml`, we already have test cases to cover array of enum in https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml#L1210

If it's not sufficient, I suggest we update `petstore-with-fake-endpoints-models-for-testing.yaml` instead.

bkryza:
@wing328 I agree that the test you mention should be sufficient - the question is why this wasn't caught by CI - are you sure this method is actually executed during tests: 

https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/python/petstore_api/models/enum_arrays.py#L99

wing328:
@bkryza I think the reason is that there's no test cases to cover it. It would be nice if you can add a test case under https://github.com/swagger-api/swagger-codegen/tree/master/samples/client/petstore/python/tests to cover the issue moving forward.

bkryza:
@wing328 Ok, I'll do that.

bkryza:
@wing328 I've added test cases for `EnumArrays`, I've also noticed that my PR affects map properties whose keys have an enum constraint, for instance in case of `MapTest` model, so I've extended my PR to cover that and also added a test for that.

wing328:
@bkryza thanks for adding the test cases. PR merged into master.

Issue #3695, opened 2016-09-01
TypeScript Angular2 wrong Map usage


ataraxus:
##### Description

Hello, I'm using the editor.swagger.io for code generation. I stumbled upon this: If you define formData parameter the api client generates following code: 
`formParams['key'] = value;`
which leds to an empty request, since later on the request parameters are derived by;
`requestOptions.body = formParams.toString();` 
The 'correct' usage of the TS Map interface would be:
`formParams.set('key', value);`

I would imagine that this could be a fairly simple fix, I even looked at the **TypeScriptAngular2ClientCodegen.java** but wasn't sure if this could be the [problem](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngular2ClientCodegen.java#L123) 

##### Swagger-codegen version

editor.swagger.io

##### Swagger declaration file content or url

```yaml
swagger: '2.0'
info:
  version: 1.0.0
  title: Broken TS Map generation
host: broken.map
basePath: /

paths:
  /brokenMap:
    post:
      parameters:
        - name: limit
          in: formData
          description: number of User to return
          type: integer
         
      responses:
        200:
          description:  List all User
          schema:
            title: Pets
            type: array
            items:
              $ref: '#/definitions/User'

definitions:
  User:
    type: object
    properties:
       roles:
        type: array
        items:
          type: string

```

wing328:
@ataraxus thanks for reporting the issue and suggesting a fix. I believe updating this line should fix it: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache#L90

May I know if you've time to contribute a fix?

ataraxus:
@wing328 Hey, thanks for pointing this out! yes this looks like the correct place to fix it, actually i can see that there are already some other bugs fixed which i have on my list..

I'll try to setup the project on my machine and test the change and also see if the other issues i experienced still exist!

ataraxus:
@wing328 fixed it. good that i had also an edge case in my api, so I had to add explicit casting before calling .set. right now it works like a charm ;) 

By the way, it fixed all other issues i had with the previous release! kudos!

When will this version be released?

wing328:
@ataraxus usually it takes 3 to 4 months for a minor release.

PR #3714, opened 2016-09-02
fix map usage in api generation #3694

ataraxus:
Fix for issue: 

https://github.com/swagger-api/swagger-codegen/issues/3695

### DIFF
--- a/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
@@ -55,10 +55,10 @@ export class {{classname}} {
 {{/allParams}}
 {{#queryParams}}
         if ({{paramName}} !== undefined) {
-            queryParameters.set('{{baseName}}', String({{paramName}}));
+            queryParameters.set('{{baseName}}', <any>{{paramName}});
         }
-
 {{/queryParams}}
+
 {{#headers}}
         headers.set('{{baseName}}', String({{paramName}}));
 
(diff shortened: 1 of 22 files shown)


wing328:
@ataraxus thanks for the PR. Please run `typescript-angular2-petstore-all.sh` to update the Petstore sample as mentioned in the PR checklist.

ataraxus:
with updated petstore

wing328:
@ataraxus thanks. I'll review later.

wing328:
@ataraxus two questions if you don't mind. For query parameters, we've the following:
```
 {{#queryParams}}
         if ({{paramName}} !== undefined) {
             queryParameters.set('{{baseName}}', String({{paramName}}));
         }
 
 {{/queryParams}}
```
Shall we replace String() with <any> instead (similar to what you use for formParams.set)?

For form parameters, do we need to check for `undefined`?

I think we should make the handling of query and form parameters consistent if possible.

ataraxus:
hey @wing328  i agree with the check if the param is set. I also agree with the typeconversion. I didn't noticed it so i went for the "typescript" way of casting. but yours is also valid, see https://stackoverflow.com/questions/32554624/casting-a-number-to-a-string-in-typescript/32607656#32607656 
I changed it to my style, nevertheless ;) 
But feel free to change it

wing328:
cc @Vrolijkx

Vrolijkx:
seem ok for me.

wing328:
@Vrolijkx thanks for reviewing.

@ataraxus PR merged. Thanks for your contribution.

PR #3716, opened 2016-09-02
[Ruby] Fix ambiguous regex

camelmasa:
We should use `/\A(true|t|yes|y|1)\z/i` better  than `/^(true|t|yes|y|1)$/i` 😃 

For example, `"\ntrue\n" =~ /^(true|t|yes|y|1)$/i` doesn't return nil.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/ruby/base_object.mustache
+++ b/modules/swagger-codegen/src/main/resources/ruby/base_object.mustache
@@ -35,7 +35,7 @@
       when :Float
         value.to_f
       when :BOOLEAN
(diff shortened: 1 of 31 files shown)


wing328:
@camelmasa agreed with you that using \A, \z are better.

There's another line using ^:
```
modules/swagger-codegen/src/main/resources/ruby/base_object.mustache:        if type =~ /^Array<(.*)>/i
```
That should not matter much but I'll file a PR later to make the use of regular expression more consistent in the Ruby API client.

camelmasa:
@wing328 Thank you for mrege 👍  And, Good catch about the another line 😃

PR #3717, opened 2016-09-02
[Lumen, PHP, Python, Ruby] Implement exclusiveMinimum/Maximum

scop:


wing328:
FYI. The change covers PHP, Python, Ruby API client, and PHP Lumen server generator.

Issue #3706, opened 2016-09-02
[JAVA] InlineModelResolver drop all vender extensions for any Schema Object


zhengcan:
##### Description

The spec told me "Allows extensions to the Swagger Schema."
ref: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#patterned-objects-9

When I define some schema objects in `definitions` section, and use $ref to reference it in response, I've noticed that all vender extensions works well.
But when I use the vender extensions in 'schema' section in `response` directly, all vender extensions gone.

I've debugged my generator code, and confirmed that the `InlineModelResolver` used by `DefaultGenerator` drops all vender extensions when it try to `flatten` all response schema objects to inline schema.

For example:
```java
    public Model modelFromProperty(ObjectProperty object, String path) {
        String description = object.getDescription();
        String example = null;

        Object obj = object.getExample();
        if(obj != null) {
            example = obj.toString();
        }
        String name = object.getName();
        Xml xml = object.getXml();
        Map<String, Property> properties = object.getProperties();

        ModelImpl model = new ModelImpl();
        model.setDescription(description);
        model.setExample(example);
        model.setName(name);
        model.setXml(xml);

        if (properties != null) {
            flattenProperties(properties, path);
            model.setProperties(properties);
        }

        return model;
    }
```

I'm not sure why it work like this, but it seams like a kind of bug.

As workaround, I have to `cache` all vender extensions for response schema in `preprocessSwagger(Swagger)`, and set them back in `fromOperation(String, String, Operation, Map<String, Model>, Swagger)`.

##### Swagger-codegen version

master

##### Swagger declaration file content or url

```yaml
response:
  200:
    schema:
        type: object
        properties:
            ....
        x-my-ext: the-value
```

In mustache file:
```mustache
{{#operations}}{{#operation}}
  {{#responses}}
    {{schema.venderExtensions}}
  {{/responses}}
{{/operations}}{{/operation}}
```

##### Command line used for generation

Just run generator.

##### Steps to reproduce

Run generator, and check the output file.

##### Related issues

na

##### Suggest a Fix

Modify `InlineModelResolver` to handle vender extensions.

wing328:
@zhengcan thanks for reporting the issue. May I know if you've time to contribute a fix to `InlineModelResolver`?

As a workaround, you can fully define the model instead of doing it inline.

PR #3719, opened 2016-09-03
Make InlineModelResolver resolve vendorExtensions correctly for Schema Object

zhengcan:
### PR checklist

### Description of the PR

Fix this issue: #3706

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/InlineModelResolver.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/InlineModelResolver.java
@@ -95,9 +95,13 @@ public class InlineModelResolver {
                                         Model model = modelFromProperty(op, modelName);
                                         String existing = matchGenerated(model);
                                         if (existing != null) {
-                                            response.setSchema(new RefProperty(existing));
+                                            RefProperty newSchema = new RefProperty(existing);
+                                            this.copyVenderExtensions(property, newSchema);
+											response.setSchema(newSchema);
                                         } else {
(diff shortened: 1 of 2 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/InlineModelResolver.java
```diff
@@ -363,18 +375,45 @@ public Model modelFromProperty(MapProperty object, @SuppressWarnings("unused") S
...
+     * 
+     * @param source
+     * @param target
+     */
+    private void copyVenderExtensions(Property source, AbstractModel target) {
+        Map<String, Object> venderExtensions = source.getVendorExtensions();
+        for (String extName : venderExtensions.keySet()) {
+            target.setVendorExtension(extName, venderExtensions.get(extName));
+        }
+    }
```
wing328: @zhengcan thanks for the PR. Can you tell me more on why we need to copy the vendor extension from a property to the property's model itself?

zhengcan: I think in the following example, the codegen cannot know whether the developer want to set the vendor extendsion for property or property's model.
```java
response:
  200:
    schema:
        type: object
        properties:
            ....
        x-my-ext: the-value
```
It may be ok to let the property and the model share the same vendor extensions.

I read my PR again, and I think it make sense to not set the vendor extensions to model.
And, let me fix some bugs in the PR :)

wing328:
@zhengcan thanks for the contribution. PR merged into master.

PR #3730, opened 2016-09-06
python: Fix model maxLength validation error message

scop:


wing328:
@scop thanks for the fix. I also reviewed the wording for `minLength` in model.mustache, `minLength` & `maxLength` in api.mustache and all looks good.

PR #3734, opened 2016-09-06
[Python, Ruby, PHP] Validate min/maxItems

scop:
This adds validation for min/maxItems in templates. For some reason the checks don't end up in generated models but only to the apis, but I believe that's a problem somewhere else (and models should receive the validation as well once that issue gets fixed, wherever it is).


wing328:
@scop thanks for the PR. I think that's because [CodegenProperty](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenProperty.java) is missing maxItems and minItems 

After adding these new properties, you will need to update `modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java` to populate them.

scop:
I already did that, but to my surprise it didn't have any effect whatsoever. Maybe I missed something; anyway I'd prefer to address that maybe sometime later in a separate PR and have this go in already in its current form if it's fine with you.

wing328:
Maybe the issue is with the Swagger Parser

wing328:
It's actually with swagger-core. [AbstractSerializableParameter](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-models/src/main/java/io/swagger/models/parameters/AbstractSerializableParameter.java#L39-L40) has `maxItems` and `minItems` but [AbstractProperty](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-models/src/main/java/io/swagger/models/properties/AbstractProperty.java) does not have something similar.

scop:
I'm not intimately familiar with swagger-core at all, but AbstractProperty doesn't have string related properties such as min/maxLength or pattern either, yet they work fine in codegen. For strings they're in StringProperty, and similarly ArrayProperty has min/maxItems for arrays, so I suppose it might not be that clear cut that the issue is there.

(Aside, if you're involved with core, there's https://github.com/swagger-api/swagger-core/pull/1907 waiting for action...)

wing328:
OK. No idea off the top of my head. Will take a deeper look later.

Issue #3738, opened 2016-09-07
[Scala] respect the sourceFolder parameter when generating code


ePaul:
##### Description

I wanted to generate just model classes in Scala from an Swagger/OpenAPI definition, in a place which is not a whole maven/sbt/... project, but just some folder generated during a build.

The Scala generators I found (ScalaClientCodegen (`scala`), ScalatraServerCodegen (`scalatra`), AkkaScalaClientCodegen (`akka-scala`) and AsyncScalaClientCodegen (`async-scala`)) all put the output into
```java
outputFolder + "/" + sourceFolder + "/" + modelPackage().replace('.', File.separatorChar);
```
(or `apiPackage` for the API files), where the `sourceFolder` is hardcoded to `"src/main/scala"`, and cannot be changed by parameters (neither of the CLI, web app nor maven plugin).

##### Swagger-codegen version

This happens in 2.2.1, and is the same in current master.

##### Command line used for generation

```
java -DmodelPackage=example -DsourceFolder=bla -Dmodels \
    -jar ../swagger-codegen/modules/swagger-codegen-cli/target/swagger-codegen-cli.jar \
    generate -i code/swagger.yaml  -l scala -o sample
```
Output:
```
[main] INFO io.swagger.parser.Swagger20Parser - reading from code/swagger.yaml
[main] INFO io.swagger.codegen.AbstractGenerator - writing file /home/paulo/projektoj/swagger-codegen-bugreports/sample/src/main/scala/io/swagger/client/model/MDCTrace.scala
```

##### Suggest a Fix

The Java codegens have a setter for this `sourceFolder` property, and set it from additionalProperties in AbstractJavaCodegen.processOpts(). This could be done here too.

This might be an opportunity to see whether there is some common code between all the Scala codegens which could be extracted into a super class (though I guess most duplication is between the template files, which can't be easily shared – correct me if I'm wrong).

ePaul:
An unrelated problem I noted here: The `modelPackage` parameter seems to be only passable via config file (it doesn't work with -D), while the `models` parameter seems to be only working with `-D`, not per config-file (at least `"models":""` is ignored there).

PR #3755, opened 2016-09-08
Feature/#3738 extract abstract scala codegen

ePaul:
### PR checklist

- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run`./bin/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates) – regenerated all scala related samples: `scala`, `async-scala`, `scalatra`, `akka-scala`)
- [x] Filed the PR against the correct branch: master for non-breaking changes ~~and `2.3.0` branch for breaking (non-backward compatible) changes~~.

### Description of the PR

This extracts a common super class for AkkaScalaClientCodegen, AsyncScalaClientCodegen, ScalaClientCodegen, ScalatraServerCodegen, which cleans up some duplicated code. This is a prerequisite to fixing #3738 (which will be done in a second pull request – #3757 – building on this one).

I tried to do it in a minimally invasive way – the results should be the same as before. In the petstore samples the only difference is that some imports for classes in the same package disappeared (only for `scalatra`).

I guess there could be done more, but I'm not so familiar with the Scala generators, and don't know which of the behavior differences is intentional and which not. For example, all classes seem to have different lists of reserved words.

One potential issue:

* I extracted the `languageSpecificPrimitives` initialization into the super class – but it had slightly different lists in each of the subclasses. I chose to use the union of all of them, though I'm not sure what the effect would be here. (`Any` and `Seq` are now primitives for all of them, and `scalatra` had `Integer` instead of `Int`.)

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractScalaCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractScalaCodegen.java
@@ -1,0 +1,150 @@
+package io.swagger.codegen.languages;
+
+import java.io.File;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+import com.google.common.collect.ImmutableList;
+
+import io.swagger.codegen.CodegenConfig;
+import io.swagger.codegen.CodegenConstants;
+import io.swagger.codegen.DefaultCodegen;
+import io.swagger.models.properties.ArrayProperty;
+import io.swagger.models.properties.BooleanProperty;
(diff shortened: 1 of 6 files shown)


wing328:
@ePaul thanks for the PR :) I'm sure your work has laid down a more solid foundation for Scala-related generators to improve.

For languageSpecificPrimitives and reserved words, I'll create separate tasks for tracking.

wing328:
Created #3770, #3771

PR #3758, opened 2016-09-08
[Erlang] Add Erlang server generator

galaxie:
### PR checklist

### Description of the PR
This is the first version of Erlang code generation for an Erlang *server*.
As I see all of the repos here are mostly about clients, but we needed a working server so we created it first.

The purpose of this code generation is to create an Erlang server on top of `cowboy` app and handle all the routing and validation. The user is supposed to add his own handlers to provide any business logic to process requests. Any operation which is not covered by user defined handlers leads to `501 Not implemented` error (instead of `200 OK` approach in e.g `go-server`).

This attempt is FAR from being done. Some significant parts that are still missing or need refactoring that I would like to mention:
- Tests
- Requests validation
- REST flow in `cowboy` handlers

and legions of other stuff.


wing328:
@galaxie thanks for the PR, which looks good as a starting point.

PR #3760, opened 2016-09-09
Add last response code and headers

stevedenman:
### PR checklist

### Description of the PR

Apply same workaround as https://github.com/swagger-api/swagger-codegen/pull/1127 to typescript-angular2 template.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
@@ -24,17 +24,6 @@ export class {{classname}} {
     protected basePath = '{{basePath}}';
     public defaultHeaders: Headers = new Headers();
 
-    private _statusCode: number;
-    private _responseHeaders: Headers;
-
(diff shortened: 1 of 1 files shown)


wing328:
For #1127, we've replaced that with an additional method WithHttpInfo, e.g. https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/java/okhttp-gson/src/main/java/io/swagger/client/api/PetApi.java#L123-L137

Can we do something similar in TypeScript-Angular2 client?

The reason we didn't use variables to store the last status code and response headers was that such solution will have issues with multi-threading environment.

wing328:
cc @Vrolijkx

stevedenman:
Hi @wing328, yeah, that makes sense. I will have a look at changing it.

stevedenman:
@wing328 I've pushed that change - the new functions return the Response, so headers etc can be used.

wing328:
@stevedenman thanks for the contribution. PR merged into master.

PR #3776, opened 2016-09-12
Fixed bug for handling optional header parameters

geetikagupta16:
### PR checklist

### Description of the PR
Refactored code to handle optional header parameters


modules/swagger-codegen/src/main/resources/asyncscala/api.mustache
```diff
@@ -40,9 +40,14 @@ class {{classname}}(client: TransportClient, config: SwaggerConfig) extends ApiC
     {{/required}}
     {{/queryParams}}
 
-    {{#headerParams}}
-    headerParams += "{{baseName}}" -> {{paramName}}.toString
-    {{/headerParams}}
+  {{#headerParams}}
+      {{^required}}
+          if ({{paramName}} != null) {{paramName}}.map { v => headerParams += "{{baseName}}" -> v.toString }
```
wing328: Can we use 2-space indentation instead?

And what if {{paramName}} (variable) is a non-string? Shall we use {{#isString}} ... {{/isString}} to do the null check only if it's a string?

geetikagupta16:
null check is done just to make sure that the code won't break, if the user has provided null values when calling the Api.

wing328:
@clasnake when you've time, please review this change. Thanks!

modules/swagger-codegen/src/main/resources/asyncscala/api.mustache
clasnake: Yes we should use 2-space indentation here.

modules/swagger-codegen/src/main/resources/asyncscala/api.mustache
```diff
@@ -40,9 +40,14 @@ class {{classname}}(client: TransportClient, config: SwaggerConfig) extends ApiC
...
+      {{/required}}
+      {{#required}}
+          if ({{paramName}} != null) headerParams += "{{baseName}}" -> {{paramName}}.toString
+      {{/required}}
```
clasnake: Since the type of {{paramName}} is an Option, why do we need to do the null check?

clasnake:
@wing328 Just reviewed this, I don't think we need the null check because the type of {{paramName}} is an Option.

modules/swagger-codegen/src/main/resources/asyncscala/api.mustache
wing328: My understanding is that {{paramName}} is an Option if it's an optional parameter.

If it's required, then we still need the null check (if the data type is `string`)

clasnake: Ah I see. But why do we use Option for non-required and raw type for required params? Required or not can be achieved by whether providing default value or not. It doesn't matter whether we use Option or not. It's orthogonal.

wing328: I do not know the reason (I was not the one who created the async-scala generator).

`scala` generator (API client) also follows similar convention: Option for non-required and raw type for required.

If Option should be used for both required and non-required parameters, I welcome the (breaking) change as long as it conforms to Scala style guide or best practices.

clasnake: Yes, basically we will never see null in Scala code. And actually that's the reason why we have Option.

geetikagupta16:
@clasnake @wing328 Thanks for the review.
I have made the required changes. Please have a look at it.

wing328:
@geetikagupta16 thanks for the contribution. PR merged into master.

In the future, we may merge the `async-scala` generator into the `scala` generator so that the Scala API client will provide both async and synchronous methods (similar to what we've done for other API clients, e.g. C#, Ruby, Python, etc)

geetikagupta16:
@wing328 Thanks for merging PR

PR #3779, opened 2016-09-12
[Swift] Use ISOFullDate for 'date' format

jgavris:
### PR checklist

- [ x ] Read the [contribution guildelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [ x ] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run`./bin/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [ x ] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

- Adds support for the `date` format (ISO-8601 RFC 3339). A common use case is 'date of birth', where we do not care about units beyond the 'day'.

### DIFF
--- a/modules/swagger-codegen/src/test/resources/2_0/petstore.json
+++ b/modules/swagger-codegen/src/test/resources/2_0/petstore.json
@@ -833,6 +833,10 @@
         "username": {
           "type": "string"
         },
+        "dateOfBirth": {
+          "type": "string",
+          "format": "date"
+        },
         "firstName": {
           "type": "string"
         },
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SwiftCodegen.java
@@ -115,8 +115,8 @@ public class SwiftCodegen extends DefaultCodegen implements CodegenConfig {
         typeMapping.put("array", "Array");
         typeMapping.put("List", "Array");
         typeMapping.put("map", "Dictionary");
-        typeMapping.put("date", "NSDate");
-        typeMapping.put("Date", "NSDate");
+        typeMapping.put("date", "ISOFullDate");
+        typeMapping.put("Date", "ISOFullDate");
         typeMapping.put("DateTime", "NSDate");
         typeMapping.put("boolean", "Bool");
         typeMapping.put("string", "String");
--- a/modules/swagger-codegen/src/main/resources/swift/Extensions.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift/Extensions.mustache
@@ -84,6 +84,99 @@ extension NSUUID: JSONEncodable {
     }
 }
 
+/// Represents an ISO-8601 full-date (RFC-3339).
+/// ex: 12-31-1999
+/// https://xml2rfc.tools.ietf.org/public/rfc/html/rfc3339.html#anchor14
+public final class ISOFullDate: CustomStringConvertible {
+
+    public let year: Int
+    public let month: Int
+    public let day: Int
+
+    public init(year year: Int, month: Int, day: Int) {
+        self.year = year
+        self.month = month
+        self.day = day
+    }
+
+    /**
+     Converts an NSDate to an ISOFullDate. Only interested in the year, month, day components.
+
+     - parameter date: The date to convert.
+
+     - returns: An ISOFullDate constructed from the year, month, day of the date.
+     */
+    public static func from(date date: NSDate) -> ISOFullDate? {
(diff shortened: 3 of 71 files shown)


jgavris:
cc @vineetshah

wing328:
@jgavris thanks for the PR.

cc @Edubits @jaz-ah

jaz-ah:
@jgavris just out of curiosity, why not have the generator still deal w/ basic objects (NSDate) and then you have convertors in your proj that convert to different ISO formats - i.e. what if other people wanted other ISO formats out of the NSDate...

jgavris:
@jaz-ah The swagger spec states that the `date` format represents an ISO-8601 `full-date`, which is unambiguously `YYYY-MM-DD`. That is the format that I've implemented here. I'm not sure what you mean when you say people might want other formats out of the `NSDate`. If you've specified `type=string,format=date` in your API, you should get a `full-date` field.

jaz-ah:
@jgavris agreed that the server will send down an ISO-8601, but the most convenient/extendable class to deal w/ this in the client is a NSDate - i.e. you can deserialize the ISO-8601 and return an NSDate/serialize NSDate and serialize as ISO-8601 format...

jgavris:
@jaz-ah I agree with you in principle, but that's the implementation problem. The current client cannot do that, and that's how I ended up here. I was implementing a 'date' format field and on my first request, the server rightfully rejected it as being in the wrong format. The Swift codegen encodes all types to a json string using the following protocol.

```swift
protocol JSONEncodable {
    func encodeToJSON() -> AnyObject
}
```

`NSDate` in particular is encoded into the full ISO-8601 `date-time` format with the following

```swift
private let dateFormatter: NSDateFormatter = {
    let fmt = NSDateFormatter()
    fmt.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZZZZZ"
    fmt.locale = NSLocale(localeIdentifier: "en_US_POSIX")
    return fmt
}()

extension NSDate: JSONEncodable {
    func encodeToJSON() -> AnyObject {
        return dateFormatter.stringFromDate(self)
    }
}
```

The client only knows it has an `NSDate` field, and does not know what format to serialize to, since this converter function does not take any arguments. Without major (and potentially undesirable) changes in the architecture of the Swift codegen, using a specific type is the way to implement this. 

I've provided an API that is lighter than `NSDateComponents` (because I wrap it) so this is effectively doing the same thing in a more friendly way.

jaz-ah:
@jgavris fair enough - looks good @wing328 let's merge.

modules/swagger-codegen/src/test/resources/2_0/petstore.json
```diff
@@ -833,6 +833,10 @@
         "username": {
           "type": "string"
         },
+        "dateOfBirth": {
+          "type": "string",
```
cbornet: Is it OK to modify the petstore.json ?

wing328: Agreed with @cbornet that we should keep Petstore.json as it's.

@jgavris what I would suggest is to add the test case in [petstore-with-fake-endpoints-models-for-testing.yaml](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml#L989) instead and then update `swift-petstore.sh` to use `petstore-with-fake-endpoints-models-for-testing.yaml`, which covers a lot more edge cases (you may want to do this in another PR as the edge cases in `petstore-with-fake-endpoints-models-for-testing.yaml` may require some fixes in the Swift generators or templates.

jgavris: Thanks! I haven't been able to make the changes yet, but I will.

wing328:
@jgavris I've filed #3945 to remove `dateOfBirth` and therefore I'm closing this PR. Thanks again for the contribution.

(the commit/change is still owned by you)

ahmettaha:
Hello,

Is it possible to make the 

`fmt.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZZZZZ"`

customizable. I would like to send local date to our server. 

Thank you.

jgavris:
@ahmettaha Could you open an issue with your question specifically? This pull request was to add support for 'dates' specifically (where you want to express a single day).

ahmettaha:
i did. thank you. it's issue

#5318

PR #3781, opened 2016-09-13
add authentication to typescript-angular2

timadamson:
### PR checklist

- [x ] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run`./bin/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)

### Description of the PR

Adds authentication to typescript-angular2 clients. Inspired by the csharp equivalent.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
@@ -24,7 +24,7 @@ import { Configuration }                                     from '../configurat
 export class {{classname}} {
     protected basePath = '{{basePath}}';
     public defaultHeaders: Headers = new Headers();
-    public configuration: Configuration = null;
+    public configuration: Configuration = new Configuration();
 
     constructor(protected http: Http, @Optional()@Inject(BASE_PATH) basePath: string, @Optional() configuration: Configuration) {
         if (basePath) {
@@ -106,13 +106,13 @@ export class {{classname}} {
         // authentication ({{name}}) required
 {{#isApiKey}}
 {{#isKeyInHeader}}
-        if (this.configuration && this.configuration.apiKey)
+        if (this.configuration.apiKey)
         {
             headers.set('{{keyParamName}}', this.configuration.apiKey);
         }
 {{/isKeyInHeader}}
 {{#isKeyInQuery}}
-        if (this.configuration && this.configuration.apiKey)
+        if (this.configuration.apiKey)
         {
             formParams.set('{{keyParamName}}', this.configuration.apiKey);
         }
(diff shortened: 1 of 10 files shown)


wing328:
@timadamson thanks for the PR :)

Can you please also run `./bin/typescript-angular2-petstore-all.sh` to update the Petstore sample?

timadamson:
I run a Windows machine, but I'll see if I can sort something out.

wing328:
@timadamson you will find windows batch files under `.\bin\windows\`

wing328:
cc @Vrolijkx

timadamson:
Done!

modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
```diff
@@ -23,11 +24,15 @@ import { BASE_PATH }                                         from '../variables'
...
     protected basePath = '{{basePath}}';
     public defaultHeaders: Headers = new Headers();
+    public configuration: Configuration = new Configuration();
 
-    constructor(protected http: Http, @Optional()@Inject(BASE_PATH) basePath: string) {
+    constructor(protected http: Http, @Optional()@Inject(BASE_PATH) basePath: string, @Optional() configuration: Configuration) {
         if (basePath) {
             this.basePath = basePath;
         }
+        if (configuration) {
```
Vrolijkx: If we start using a configuration it would probably be better to also put the basePath in it instead of passing it sepperatly.

Vrolijkx:
Looks ok to me. But maybe also at the base path in the configuration object instead of passing it separately.

timadamson:
Yeah I wan't sure about that, figured that the Configuration object's properties can be updated dynamically but BASE_PATH is static. The C# template I modelled it on kept them separate in the constructor. Your call, I'm happy to update if you think it's more suitable.

wing328:
Very good feedback and discussion.

If we move the base path to configuration object, then I believe the change will be a breaking one (non-backward compatible), which means the change should be filed against `2.3.0` branch.

One consideration I would suggest is to look at how typescript-node, typescript-angular and typescript-fetch store the base path and authentication-related properties (e.g. username, API key) so as to make the developer experience more consistent across different typescript generators. I'm not a professional typescript developer so I cannot speak on behalf of the typescript developer community but I believe a consistent developer experience will be welcome by any developers.

timadamson:
Unfortunately typescript-angular is also missing authentication. Node is less opinionated than angular2 and typescript-node just using simple but more cumbersome setApiKey methods on the api classes. I feel consistency would be more valued along framework lines as typescript in general (like javascript) offers a wide range of standard practices; as is it fits in quite well from what I've seen of Angular 2. 

Angular 2 is still in beta/RC and conventions might shift. Can we leave it as the non-breaking change and if someone with more experience pipes in later when things are more settled we can update with breaking changes?

wing328:
@timadamson Sure your suggestion sounds good to me. PR merged. Thanks for the contribution.

Issue #3783, opened 2016-09-13
[PHP] Bug in setter validation of non-required properties with null value


xaviapa:
##### Description

When generating an SDK in PHP, there is a problem in the Model generation of an entity/api resource. If a property is **not required** and has the value **null** you get the exception of the minlength condition. In my opinion the null value should be valid for a non-required property.

##### Swagger-codegen version

master

##### Swagger declaration file content or url

```
"Program": {
   "required": [
      "id_program",
      "id_origin",
      "id_platform",
      "id_category"
   ],
   "properties": {
      "id_program": {
          "type": "string",
          "description": "Program UUID or numeric ID",
          "minLength": 1,
          "maxLength": 36
      },
      "id_developer": {
          "type": "string",
          "description": "Developers identifier",
          "minLength": 1,
          "maxLength": 100
      },
      "id_origin": {
          "type": "string",
          "description": "Origins identifier",
          "minLength": 1,
          "maxLength": 40
      },
      "id_platform": {
          "type": "string",
          "description": "Platforms identifier",
          "minLength": 1,
          "maxLength": 20
      },
      "id_category": {
          "type": "string",
          "description": "Categories identifier",
          "minLength": 1,
          "maxLength": 40
      }
   }
},
```

##### Steps to reproduce

Check the SDK https://github.com/softonic/sdk_catalog-api, file https://github.com/softonic/sdk_catalog-api/blob/master/src/Model/Program.php, method setIdDeveloper()
```
public function setIdDeveloper($id_developer)
    {
        if (strlen($id_developer) > 100) {
            throw new \InvalidArgumentException('invalid length for $id_developer when calling Program., must be smaller than or equal to 100.');
        }
        if (strlen($id_developer) < 1) {
            throw new \InvalidArgumentException('invalid length for $id_developer when calling Program., must be bigger than or equal to 1.');
        }
        $this->container['id_developer'] = $id_developer;
        return $this;
    }
```
This is throwing the exception _'invalid length for $id_developer when calling Program., must be bigger than or equal to 1.'_ because _strlen(null) = 0_.

As $id_developer is a non-required property (see swagger content pasted above), the function should be:
```
public function setIdDeveloper($id_developer)
    {
        if (is_null($id_developer)) {
            $this->container['id_developer'] = $id_developer;
            return $this;
        }
        if (strlen($id_developer) > 100) {
[...]

PR #3785, opened 2016-09-13
Made null value for non-required property valid

xaviapa:
### Description of the PR

https://github.com/swagger-api/swagger-codegen/issues/3783

In the setter, if a non-required property has value null, it should be valid.

### DIFF
--- a/samples/client/petstore/php/SwaggerClient-php/lib/Model/MixedPropertiesAndAdditionalPropertiesClass.php
+++ b/samples/client/petstore/php/SwaggerClient-php/lib/Model/MixedPropertiesAndAdditionalPropertiesClass.php
@@ -181,6 +181,11 @@ class MixedPropertiesAndAdditionalPropertiesClass implements ArrayAccess
      */
     public function setUuid($uuid)
     {
+        if (is_null($uuid)) {
+            $this->container['uuid'] = $uuid;
+
+            return $this;
+        }
         $this->container['uuid'] = $uuid;
 
         return $this;
@@ -202,6 +207,11 @@ class MixedPropertiesAndAdditionalPropertiesClass implements ArrayAccess
      */
     public function setDateTime($date_time)
     {
+        if (is_null($date_time)) {
+            $this->container['date_time'] = $date_time;
+
+            return $this;
+        }
         $this->container['date_time'] = $date_time;
 
         return $this;
@@ -223,6 +233,11 @@ class MixedPropertiesAndAdditionalPropertiesClass implements ArrayAccess
      */
     public function setMap($map)
     {
+        if (is_null($map)) {
+            $this->container['map'] = $map;
+
+            return $this;
+        }
         $this->container['map'] = $map;
 
         return $this;
--- a/modules/swagger-codegen/src/main/resources/php/model_generic.mustache
+++ b/modules/swagger-codegen/src/main/resources/php/model_generic.mustache
@@ -258,6 +258,13 @@ class {{classname}} {{#parentSchema}}extends {{{parent}}} {{/parentSchema}}imple
      */
     public function {{setter}}(${{name}})
     {
+        {{^required}}
+        if (is_null(${{name}})) {
+            $this->container['{{name}}'] = ${{name}};
+
+            return $this;
+        }
+        {{/required}}
         {{#isEnum}}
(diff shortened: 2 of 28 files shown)


wing328:
@xaviapa thanks for the PR. Please run `./bin/php-petstore.sh` to update the PHP Petstore sample so that the CIs can verify the change.

cc @arnested

xaviapa:
Commited the PHP Petstore sample. Sorry for not doing it before.

wing328:
@xaviapa your second commit is not linked to your Github account. Let me know if you need help fixing it.

xaviapa:
Added my company e-mail to this account. Looks like it is fixed now. Sorry for that, it's my first contribution to an external project :)

samples/client/petstore/php/SwaggerClient-php/lib/Model/MixedPropertiesAndAdditionalPropertiesClass.php
```diff
@@ -181,6 +181,11 @@ public function getUuid()
      */
     public function setUuid($uuid)
     {
+        if (is_null($uuid)) {
+            $this->container['uuid'] = $uuid;
+
+            return $this;
+        }
         $this->container['uuid'] = $uuid;
```
wing328: @xaviapa I don't think I fully understand the change.

Without the "if" block that you added, `$this->container['uuid']` is still set to `$uuid`

What problem did you encounter when using the setter?

xaviapa: The problem is in the properties that have validations, for example, setInteger(). If you receive a null and execute
```
if ($integer < 10.0) {
             throw new \InvalidArgumentException('invalid value for $integer when calling FormatTest., must be bigger than or equal to 10.0.');
         }
```
null < 10 => true => The exception is thrown.
The same happens in our API:
```
if (strlen($id_developer) < 1) {
            throw new \InvalidArgumentException('invalid length for $id_developer when calling Program., must be bigger than or equal to 1.');
        }
```
strlen(null) => 0, so the Exception is also thrown.

I understand that for setters without validations, the resulting code makes no sense, like in the example you said.
Maybe a better approach could be to add **if (!is_null($var) && (<any_other_condition>)** in all the validations of the setter in the case the property is not required. This would result in the following setInteger() method:
```
public function setInteger($integer)
      {
          if (!is_null($integer) && ($integer > 100.0)) {
              throw new \InvalidArgumentException('invalid value for $integer when calling FormatTest., must be smaller than or equal to 100.0.');
         }
         if (!is_null($integer) && ($integer < 10.0)) {
             throw new \InvalidArgumentException('invalid value for $integer when calling FormatTest., must be bigger than or equal to 10.0.');
         }
 
         $this->container['integer'] = $integer;
 
         return $this;
     }
```

What do you think @wing328 ?

wing328: @xaviapa seems like we've done that already: https://github.com/xaviapa/swagger-codegen/blob/23636ee245817b139e8fe16b2c1c9995aee22dab/modules/swagger-codegen/src/main/resources/php/model_generic.mustache#L132-L175

Are you using the latest master to generate the PHP API client?

xaviapa: @wing328 I saw it and we are using the latest master, but the problem is that the change you say needs to be done also in the setter.

wing328: @xaviapa you're right, it should be done in the setter as well, can you apply the [fix](https://github.com/xaviapa/swagger-codegen/blob/23636ee245817b139e8fe16b2c1c9995aee22dab/modules/swagger-codegen/src/main/resources/php/model_generic.mustache#L132-L175) to the setter instead? (you may consider submitting a new PR for it)

xaviapa: @wing328 sure, I've created https://github.com/swagger-api/swagger-codegen/pull/3797. Can you decline this Pull Request?

PR #3791, opened 2016-09-13
[typescript-angular2] npm publish missing files

savcha:
### PR checklist

### Description of the PR

Removed files property from typescript-angular2 package.json so that all files are uploaded during npm publish.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/typescript-angular2/package.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular2/package.mustache
@@ -7,10 +7,10 @@
     "swagger-client"
   ],
   "license": "Apache-2.0",
-  "main": "./lib/index.js",
-  "typings": "./lib/index.d.ts",
+  "main": "dist/index.js",
+  "typings": "dist/index.d.ts",
   "scripts": {
-    "build": "typings install && tsc",
+    "build": "typings install && tsc --outDir dist/",
     "postinstall": "npm run build"
   },
   "peerDependencies": {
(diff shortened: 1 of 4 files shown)


wing328:
@savcha thanks for the PR. Shall we follow something similar to https://github.com/zyramedia/ng2-facebook-sdk/blob/master/package.json#L5-L12 by specifying `--outDir` during `build` ?

Currently, we still have the following attributes referencing files under `./lib`:
```
    "main": "./lib/index.js",
    "typings": "./lib/index.d.ts",
```

savcha:
Good idea. I'll make the changes.

wing328:
cc @Vrolijkx

Vrolijkx:
I like the idea of uploading all the files and the dist folder. I already ran into problems with debugging because i was missing the ```.ts``` files that where not publish.

wing328:
@savcha @Vrolijkx do we need to make similar changes to other typescript generators (e.g. typescript-fetch, typescript-angular, etc) ?

zhenwenc:
Since the output folder has been changed from `lib` to `dist`, please also update the `exclude` field in `tsconfig.json` to exclude the `dist` folder, and remove the unused `outDir` field to avoid confusion (its optional though). The build failed without excluding `dist` folder, I suspect that this is caused by the transpiler tries to recursively transpile "*.ts" files in `dist` folder. Cheers!

wing328:
@zhenwenc do you mind submitting a PR for the enhancement you mentioned so that we can review the change?

Let me know if you need some pointers on where to start.

zhenwenc:
@wing328 Thanks for the quick reply. It should be an easy fix, I am happy to submit the PR. 

But I am not sure if I should update the sample results manually, or via some script? (e.g.: samples/client/petstore/typescript-angular2/npm/tsconfig.json)

wing328:
Please run `./bin/typescript-angular2-petstore-all.sh` to update the Petstore samples.

PR #3797, opened 2016-09-14
Added !is_null condition in setter conditions for non-required proper…

xaviapa:
### Description of the PR

Issue https://github.com/swagger-api/swagger-codegen/issues/3783

In the setter, if a non-required property has value null, it should be valid.

This Pull Request is done after finding a better solution than https://github.com/swagger-api/swagger-codegen/pull/3785


wing328:
@xaviapa thanks for the PR. The change looks good to me

cc @arnested

PR #3813, opened 2016-09-16
add optional parameters to init when unwrapRequired is true

mwgray:
### PR checklist

### Description of the PR

Previously when running with unwrapRequired=true, the init function would only list required parameters.  I added the optional ones with default values to nil.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/swift/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift/model.mustache
@@ -33,10 +33,10 @@ public class {{classname}}: JSONEncodable {
     public init() {}
 {{/unwrapRequired}}
 {{#unwrapRequired}}
-    public init({{#requiredVars}}{{^-first}}, {{/-first}}{{name}}: {{#isEnum}}{{datatypeWithEnum}}{{/isEnum}}{{^isEnum}}{{datatype}}{{/isEnum}}{{/requiredVars}}) {
-        {{#requiredVars}}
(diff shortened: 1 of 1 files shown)


wing328:
cc @Edubits @jaz-ah

wing328:
@mwgray thanks for the PR. Would it be possible to keep the existing `init` (which only initializes required properties) so that the enhancement is backward compatible?

`{{#hasOptionalParams}}` mustache tag may be helpful to avoid duplicated `init` when there's no optional parameter.

mwgray:
It's possible, but having both generated at the same time creates ambiguity for the compiler.  I'm using Swift 2.3, and it causes a compiler error if both are generated(not sure about 2.2, or 3.0):

```
Ambiguous use of 'init(username:password:)'
```
```
public var username: String
public var password: String?

public init(username: String) {
    self.username = username
}

public init(username: String, password: String?=nil) {
    self.username = username
    self.password = password
}
```

It's not clear to me what this change does to prevent backwards compatibility.  The signature is the same as the previously generated code(hence the ambiguity)

If you don't set optionals in a init, they are set to the default values, which in this case is nil.  In both init functions, all values are being set, it's just the compiler sets password for you in init(username:).

wing328:
OK. Thanks for the explanation.

wing328:
## Upgrade Note ##

When `unwrapRequired` is set to true, the model constructor `init` signature has been updated to include both required and optional parameters (as opposed to just required parameters)

PR #3823, opened 2016-09-19
[Swift] fixes #3788 Support for Swift 3.0

### LINKED ISSUE #3788: Support for Swift 3.0
Is swift 3.0 Supported yet? if not when is the release date?

hexelon:
### PR checklist

- [X] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run`./bin/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)

I've upgrade the swift codegen to swift 3.0


Edubits:
Thanks @hexelon for the contribution! I didn't test it yet, but looks good on first sight. One question for the community: can we from Swagger Codegen 2.3 on only support Swift 3 and up? Or  do we want to keep support for Swift 2.3 a bit longer? This as the changes are quite major.

My personal preference would be to completely switch to 3.

cc: @jaz-ah, @wing328

jaz-ah:
@Edubits I'd say let's move to Swift3 completely.

jaz-ah:
@hexelon nice change thx for the contribution!  @wing328 LGTM +1 I'm ok to pull it and start working w/ it.

zxy198717:
@hexelon Thanks for your works. and I find two issues:
1. `api.mustache`: `open enum` should be `public enum`
2. When entity include enum in Array, the generated code is error, just like this:
```
"SettingsInfo": {
            "type": "object",
            "properties": {
                "accountId": {
                    "type": "string",
                    "description": "User ID"
                },
                "listGender": {
                    "type": "array",
                    "description": "list Gender",
                    "items": {
                        "type": "string",
                        "enum": [
                            "GIRL",
                            "BOY"
                        ]
                    }
                }
            }
        },
```
output code:
```
open class SettingsInfo: JSONEncodable {
    public enum [ListGender]: String { 
        case Girl = "GIRL"
        case Boy = "BOY"
    }

    public var accountId: String?
    public var listGender: [ListGender]?
}
```
Please check~

Thanks

hexelon:
Thanks for the review, will check and get back to you

wing328:
@hexelon thanks for the PR :)

You may want to use [petstore-with-fake-endpoints-models-for-testing.yaml](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml) instead of petstore.yaml/json as the former covers a lot more edge cases.

Shell scripts to update:
```
swift-petstore-rxswift.sh
swift-petstore-promisekit.sh
swift-petstore.sh
```

hexelon:
@zxy198717 i've done some fixes please checkout my latest commit

@wing328 I've added the new endpoints and there were some issues as you've expected. There are a couple of places where where i'm not sure whether my changes are good enough.

https://github.com/swagger-api/swagger-codegen/pull/3823/commits/c932e2d0dd0b8ba70803ef3e5eec696d2736bc34#diff-163ccd7b06b7d19ba3a8ab98a631ae63R389 => I've changed the enum value name handling, the code is a bit ugly and can be refactored later. Also this works with the current swift guideline where enum values should start with a lower case letter.

https://github.com/swagger-api/swagger-codegen/pull/3823/commits/478a57056dfa13a2b093790ad0a9081aab87e8d4#diff-9b85434b30074e84c32aecaafb94e088L51 =>  I've removed enums from being sent to the backend as there's no encoding implementation for enum.

zxy198717:
@hexelon Still have one issue. 
```
"listGender": {
                     "type": "array",
                     "description": "list Gender",
                     "items": {
                         "type": "string",
                         "enum": [
                             "GIRL",
                             "BOY"
                         ]
                     }
                 }
```

`listGender` is defined as array, but output code is one object.
```
public var listGender: ListGender?
```

zxy198717:
@hexelon 
Another issue:
```
let request = manager.request(URLString, method: xMethod!, parameters: parameters, encoding: encoding)
            if let onProgressReady = self.onProgressReady {
                onProgressReady(request.progress)
            }
            processRequest(request: manager.request(URLString, method: xMethod!, parameters: parameters, encoding: encoding), managerId, completion)
```

Two `request` instances, maybe cause the request send twice.

hexelon:
Thanks @zxy198717 I missed the duplicate request. The second issue requires more through refactoring, will try to get to it sometime over the week

tomekc:
Any chance for leaving Swift 2.2 support for some time? Swift 3 is bringing so many breaking changes, that not all projects are able to move on.

jaz-ah:
@tomekc you should be able to use the prior version of Swagger for Swift2.2 - worst case is we create a branch that you can cherrypick into if there are changes that you need...

harryblam:
@hexelon a very small thing, but line 8 in Podspec.mustache reads
`  s.authors = '{{podAuthors}}'{{/podAuthors}}{{#podSocialMediaURL}}
`

rather than 
`  s.authors = '{{podAuthors}}'{{/podAuthors}}{{#podAuthors}}
`

airwareic:
@hexelon I've just used this branch for the first time and it worked pretty smoothly, with just one hang-up. It seems that the defined ErrorResponse is conflicting with Swift.Error.

```
public enum ErrorResponse : Error {
    case Error(Int, Data?, Error)
}
```
had to be corrected to:
```
public enum ErrorResponse : Swift.Error {
    case Error(Int, Data?, Error)
}
```

andriuskl:
I think `APIHelper.mustache` is missing `import Foundation` because, I am getting `Use of unresolved identifier 'NSNumber'`

hexelon:
@andriuskl are you getting this in the petstore project?

andriuskl:
@hexelon, I am getting this on my project (Xcode 8, Swift 3)

hexelon:
@andriuskl I've added the import try now please

hexelon:
@zxy198717 I've done a bunch of changes and refactorings, please test again

ewanmellor:
I have some changes to this branch: some silencing of warnings, and one new feature.

Can we get this branch committed to master soon, so that I can create my own pull requests based on it?  Or @hexelon, do you want to bring these changes into your branch?  The warnings at least should be trivial for you to include.

https://github.com/unitive-inc/swagger-codegen/commits/swift-3-fixes

Thanks for the work everyone!

jaz-ah:
@hexelon I'm hitting some compiler errors still w/ this in Decoders.swift:

When adding decoders like this inside     private static var __once: () = {
....
            let sourceDictionary = source as! [AnyHashable: Any]
            let instance = FooRequest(foo: Decoders.decode(clazz: Int.self, source: sourceDictionary["Foo"]!))

You get a compiler error because Decoders.decode wants an AnyObject not Any:
    static func decode<T>(clazz: [T].Type, source: AnyObject) -> [T] {

do you get the same errors?

hexelon:
@ewanmellor you can submit those changes to my branch, it will be better then handing multiple requests.

hexelon:
@jaz-ah that's a good point i'm not sure this scenario is generated under petstore-with-fake-endpoints-models-for-testing.yaml

I'll test it

jaz-ah:
thx @hexelon

wing328:
> @jaz-ah that's a good point i'm not sure this scenario is generated under petstore-with-fake-endpoints-models-for-testing.yaml

Please feel free to update petstore-with-fake-endpoints-models-for-testing.yaml to cover different edge cases.

jaz-ah:
hmm @hexelon I notice in your PR your generated decoders have the casting as AnyObject and mine don't...:
+        Decoders.addDecoder(clazz: Animal.self) { (source: AnyObject) -> Animal in
 +            let sourceDictionary = source as! [AnyHashable: Any]
 +            let instance = Animal()
 +            instance.className = Decoders.decodeOptional(clazz: String.self, source: sourceDictionary["className"] as AnyObject?)
 +            instance.color = Decoders.decodeOptional(clazz: String.self, source: sourceDictionary["color"] as AnyObject?)
 +            return instance
 +        }

jaz-ah:
ah I see - @hexelon just update your models.mustache to make sure you cast as AnyObject? the non optional decoding.

jaz-ah:
@hexelon one more issue - getting this error w/ AlamoFireImplementations.swift here:

                completion(
                    Response(
                        response: voidResponse.response!,
                        body: nil),
                    nil
                )

Expression type 'Void' (aka '()') is ambiguous without more context

looks like body can't be nil

airwareic:
Thanks for creating this branch! I've been using this along with p2/OAuth2, RxSwift, and AlamoFire and I'm finding the result very compelling.

That said there is just one issue with trying to fit all of these together, and that's that the AlamofireRequestBuilder constructs its own SessionManager internally, making it non-configurable. Perhaps this should be filed separately as an issue, but it would be much more flexible if I could inject my own sessionmanager, or even just my own Retrier/Adapter (as shown here: https://github.com/p2/OAuth2/wiki/Alamofire-4).

ewanmellor:
@airwareic I hit the same thing, I've made the SessionManager configurable by making it possible to override the factory. See https://github.com/unitive-inc/swagger-codegen/commit/30143a084267d6b92005532574216d181bfe1857

My plan is to propose my changes as PRs on hexalon's branch this week, and then hopefully this whole thing will be ready to be merged soon.

zxy198717:
@hexelon Got empty enum value:
```
public enum [ListGender]: String {
        case  = ""
        case  = ""
    }
```

hexelon:
@zxy198717 you can see a working example of enum array here, https://github.com/hexelon/swagger-codegen/blob/swift-3/samples/client/petstore/swift/default/PetstoreClient/Classes/Swaggers/Models/EnumArrays.swift

Let me know what's the difference between your scenario and the one tested in the petstore

zxy198717:
@hexelon I mistake. I used my local template. I will check other features. 
Thanks for your works.

zxy198717:
@hexelon Other issue: The method `encodeToJSON` in `EnumArrays.swift` haven't `ArrayEnum`. And I think the `ArrayEnum` should be encoded to raw value, just like `["fish"]` .

wing328:
One suggestion: What about creating the Swift 3.0 generator as `swift3` instead of replacing the existing one for Swift 2.x? We've done it for Python2.x (`python`) and Python3.x (`python3`) before and thanks to the effort from @geekerzp and the community, `python` generator now supports both 2.x and 3.x without using any switch.

Based on the discussion above, it's not easy to add a `supportSwift3` switch to the existing Swift generator to support both Swift 2.x and 3.0. If we split it into two, the community can decide which one they want to use and to contribute enhancements/bug fixes.

(I'm aware of the drawback/overhead in maintaining both generators and my guess is that the migration from Swift 2.x to Swift 3.x will take some time.)

tomekc:
@wing328 I second that - it is possible that Swift 2.x will live for some time. I was thinking also about other way: would creating just separate template directory suffice? We have choice of "libraries" in Android or Java generators, are they based on different template directories? Does it make sense?

wing328:
@tomekc yes, that's one way to do it but not sure how easy/hard would it take for Swift generator to achieve something similar.

I think this PR has too many changes as I would recommend just renaming it to swift3 and keeping the original `swift` generator and template. Later we can consolidate the `swift3` generator into the `swift` generator to cover both 2.x and 3.x when anyone from the community has cycle to do so.

lightbow:
I love the idea of having separate generators. That will allow the swift3 generator to grow on its own, but the swift2.3 generator to get periodic fixes for those developers whose projects are stuck in the old world (mine, for the next month or so). I only use swagger-codegen for a small part of my project, but once the swift3 generator is ready, I'm looking forward to ditching a few crappy SDKs (for lighting control) and using my own autogenerated swift code.

jaz-ah:
+1 @wing328 so we can get this pushed in.

wing328:
Another benefit of creating a standalone `swift3` generator is that we can give credits to @hexelon for his work in https://github.com/swagger-api/swagger-codegen#template-creator as the template creator of `swift3`

hexelon:
thanks @wing328  😄 

A rebase with the recent changes on the 2.3.0 branch is getting harder, so 👍 for creating a swift3 generator if it will get changes merged faster

samples/client/petstore/swift/default/PetstoreClient/Classes/Swaggers/Models/EnumArrays.swift
```diff
@@ -26,6 +26,7 @@ open class EnumArrays: JSONEncodable {
     func encodeToJSON() -> Any {
         var nillableDictionary = [String:Any?]()
         nillableDictionary["just_symbol"] = self.justSymbol?.rawValue
+        nillableDictionary["array_enum"] = self.arrayEnum?.encodeToJSON()
```
hexelon: Right, something like:

nillableDictionary["array_enum"] = array_enum?.map({$0.rawValue}).encodeToJSON()

Will add it later this week

zxy198717:
@hexelon Two suggests:
1. Could you provider one way to replace `requestBuilderFactory` instance, since `Alamofire 4` isn't support iOS8.
2. Open `Decoders` and model's `encodeToJSON` method, since we need serialization and deserialize our models.

samples/client/petstore/swift/default/PetstoreClient/Classes/Swaggers/Models/EnumArrays.swift
hexelon: Add the change let me if it works for you

hexelon:
@zxy198717 You're welcome to open those as feature requests/issues, this thread is getting too long and it's hard to follow. 

I'm going to submit another pull request as @wing328 suggested with a separate new swift3 generator.

wing328:
Closed via https://github.com/swagger-api/swagger-codegen/pull/4002

Issue #3754, opened 2016-09-08
Java: Latest JMustache breaks bigDecimalAsString


wienczny:
The commit samskivert/jmustache@9eda17f85bfd8849850bdafa92d2f6ffe12bb71f changes the default escape characters used by JMustache. This breaks the code-generation of bigDecimalAsString in the Java language because the annotation 
```@JsonSerialize(using = ToStringSerializer.class)``` 
is written as
```@JsonSerialize(using &#x3D; ToStringSerializer.class)```

You could avoid this by disabling the escaper in JMustache

wing328:
What about using {{{ tag_name }}} (or {{& tag_name }}) for the unescape value instead ?

wienczny:
I didn't know the triple mustache. This should work, too.

PR #3825, opened 2016-09-19
Use triple mustache for vendorExtensions.extraAnnotation

wienczny:
Triple mustache is required because annotations may contain chars like "=" that
would be mistakenly encoded. See #3754 for an example


wing328:
@wienczny thanks for the PR, which has been merged into master.

PR #3853, opened 2016-09-22
Fix custom generator execution example in README template

mattalexx:


wing328:
@mattalexx thanks for the fix.

PR #3860, opened 2016-09-23
add title, description to HTML output

DavidBiesack:
The HTML output does not include the title and description of the schema. This adds them, for better documentation of the API and schema.


wing328:
@DavidBiesack thanks for the contribution.

PR #3865, opened 2016-09-24
Replace ^M with new line (\r) in mustache template

wing328:
### PR checklist

### Description of the PR

Replace ^M with new line (\r) in the mustache templates under `modules/swagger-codegen/src/main/resources`

TODO: update CI (travis) to catch the issue moving forward.


samples/client/petstore/spring-cloud/src/main/java/io/swagger/model/Pet.java
```diff
@@ -13,10 +13,14 @@
...
 
 
+/**
+ * A pet for sale in the pet store
+ **/
 
 /**
- * Pet
+ * A pet for sale in the pet store
  */
```
ePaul: Hmm, now the comment is here twice?

ePaul: This seems to be happening for all the java model files.

ePaul:
@wing328 Just a note about terminology: \r and ^M represent the same thing, the carriage return character (aka CR, 0x0D, U+000D). Newline is also known as Line feed, 0x0A, LF, \n, ^J, U+000A.

I think you meant to say you replaced all CRLF sequences (and possibly some single CRs?) by just LF (\n). (I couldn't really see what changed in the diffs, but that would be my guess as this fixes what Windows editors do wrong sometimes.)

wing328:
@epual you're definitely right. Actually I used \r when using grep and ^M in vim. The title was (incorrectly) "borrowed" from a stackoverflow question and I'll be more careful with the terminology moving forward.

samples/client/petstore/spring-cloud/src/main/java/io/swagger/model/Pet.java
wing328: I don't see duplicated comments. Can you share a URL to the lines of code with duplicated comments in the latest master?

ePaul: [diff](https://github.com/swagger-api/swagger-codegen/pull/3865/files/2083c03b28c009e3cfde9056b15f3a9b8a89451f#diff-dc82ea0f80e6fc11df755bfdccfa57e1), [raw file](https://github.com/wing328/swagger-codegen/blob/2083c03b28c009e3cfde9056b15f3a9b8a89451f/samples/client/petstore/spring-cloud/src/main/java/io/swagger/model/Pet.java). It looks like this:

```java
 /**
  * A pet for sale in the pet store
  **/
  
  /**
   * A pet for sale in the pet store
   */
```

(This new review mode doesn't show enough context, nor does it easily link back to the file.)

wing328: @ePaul thanks for the details. I'll take a deeper look tomorrow.

wing328: @ePaul fixed via https://github.com/swagger-api/swagger-codegen/pull/3887

wing328:
FYI. Added scripts to detect the issue moving forward via https://github.com/swagger-api/swagger-codegen/pull/4526

PR #3873, opened 2016-09-26
[Swift] Use thread safe manager dictionary

jgavris:
### PR checklist

- [ x ] Read the [contribution guildelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [ x ] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run`./bin/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [ x ] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

- Add thread-safe access to Alamofire Manager dictionary. Since we cannot assume the API client will always be invoked on the same thread.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/swift/AlamofireImplementations.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift/AlamofireImplementations.mustache
@@ -12,8 +12,33 @@ class AlamofireRequestBuilderFactory: RequestBuilderFactory {
     }
 }
 
+public struct SynchronizedDictionary<K: Hashable, V> {
+
+    private var dictionary = [K: V]()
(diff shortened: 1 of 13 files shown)


wing328:
@jgavris thanks for the PR.

cc @Edubits @jaz-ah

jaz-ah:
+1 @wing328

wing328:
Tests look good. PR merged. 

As always, thanks for the contribution.

Issue #3773, opened 2016-09-11
[Cpprest] Generated client code does not handle arrays of primitive types, integers with min and max values, or API methods that return a file properly.


sdrichter:
##### Description

When generating a cpprest client using a YAML/JSON API specification that uses arrays made up of primitive types (in my case integers), the cpprest code generated has multiple errors with std::vector that prevent compilation. The generated code is invalid due to some errors in the mustache files for cpprest that result in invalid C++ code being generated for this case.

Another issue is that when using integers with min/max values, the generated client code attempts to set the default value of the integer to nullptr, which is invalid.

Finally, API methods that return a file object are not handled correctly in the generated client code.

##### Swagger-codegen version

2.2.1

##### Swagger declaration file content or url
https://gist.github.com/sdrichter/0e2dfd58eeadf863f303ea1359732c50

##### Command line used for generation
java -jar swagger-codegen-cli-2.2.1.jar generate -l cpprest -i "yaml\issue.yaml" -o ..\src

##### Steps to reproduce
Generate cpprest code using above yaml file and attempt to compile it as a static library in Visual Studio 2015.

##### Related issues
None

##### Suggest a Fix
I am very close to resolving these issues on my fork of swagger-codegen.

I have modified several of the mustache files for cpprest as well as the code generator java. There were logic errors in the mustache files related to the isPrimitiveType check. I will issue a pull request once I have fully tested my changed. I wanted to get the issue logged here so everyone was aware I am working on it.

wing328:
@sdrichter thanks for logging the issue and looking forward to your PR.

PR #3876, opened 2016-09-27
[Cpprest] Fixing issue #3773

sdrichter:
### PR checklist

### Description of the PR
Fixed issues noted in #3773. Verified that it fixes all 3 issues on my generated client and also the petstore sample compiles and I am able to call the methods within successfully.

This should be a non-breaking change. The interface for the calling code has not been affected.

Please use the [YAML file here](https://gist.github.com/sdrichter/0e2dfd58eeadf863f303ea1359732c50) to verify the before and after behavior of the generated client code.


wing328:
@sdrichter thanks for the PR :)

Seems like you're using tab instead of 4-space in the Java files. Do you mind replacing tab with 4-space so as to make it consistent with the rest of the generators?

sdrichter:
Done :+1:

wing328:
@sdrichter thanks. The change looks good.

PR #3893, opened 2016-09-28
[objc] Remove the custom security policy

0xced:
### PR checklist

### Description of the PR

There are several issues with the security policy:

* The security policy is created with `AFSSLPinningModeNone` which means that even if pinned certificates are set (if config.sslCaCert is not nil), they will be ignored. Pinning will not work at all with this security policy.

* The configuration wrapper for the security policy is a bad idea.
  * `verifySSL` controls both invalid certificates and domain validation. A vanilla `AFSecurityPolicy` clearly exposes both `allowInvalidCertificates` and `validatesDomainName`.
  * `sslCaCert` only allows for a single pinned certificate. A vanilla `AFSecurityPolicy` clearly exposes a set of pinned certificates and makes it very convenient to load them with either `+[AFSecurityPolicy policyWithPinningMode:]` or `+[AFSecurityPolicy certificatesInBundle:]`

Since the security policy does not work at all and adds confusion, it is better to just remove it and let the user configure a security policy that fits their needs.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/objc/ApiClient-body.mustache
+++ b/modules/swagger-codegen/src/main/resources/objc/ApiClient-body.mustache
@@ -80,7 +80,6 @@ static NSString * {{classPrefix}}__fileNameForResponse(NSURLResponse *response) 
             @"application/x-www-form-urlencoded": afhttpRequestSerializer,
             @"multipart/form-data": afhttpRequestSerializer
         };
-        self.securityPolicy = [self createSecurityPolicy];
         self.responseSerializer = [AFHTTPResponseSerializer serializer];
     }
     return self;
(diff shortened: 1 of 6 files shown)


wing328:
cc @mateuszmackowiak

0xced:
You set the v2.2.2 milestone but I submitted this PR against the 2.3.0 branch since this is a breaking change.

wing328:
@0xced thanks for the PR. What's the correct way to disable TLS/SSL verification and provide a custom certificate?

wing328:
Given that @mateuszmackowiak has reviewed and approved it, I've merged it into master. Thanks for the contribution.

0xced:
> What's the correct way to disable TLS/SSL verification and provide a custom certificate?

You should simply be doing this:
```objc
manager.securityPolicy = [AFSecurityPolicy policyWithPinningMode:AFSSLPinningModeCertificate];
manager.securityPolicy.allowInvalidCertificates = YES;
```
And then copy your certifcates (.cer files) in your main bundle.

wing328:
@0xced thanks! The code snippet is very helpful and we will show it to those who ask about disabling TLS/SSL certification or providing a custom certificate.

Issue #2208, opened 2016-02-22
[ObjC] Wrong generation of additionalProperties


cbornet:
The following definition
```json
{
  "swagger": "2.0",
  "info": {
    "version": "1.0.0",
    "title": "Swagger Petstore"
  },
  "paths": {
    "/pets": {
      "get": {
        "description": "Returns all pets from the system that the user has access to",
        "produces": [
          "application/json"
        ],
        "responses": {
          "200": {
            "description": "A list of pets.",
            "schema": {
              "type": "array",
              "items": {
                "$ref": "#/definitions/Pet"
              }
            }
          }
        }
      }
    }
  },
  "definitions": {
    "PetMap": {
      "type": "object",
      "additionalProperties": {
        "$ref": "#/definitions/Pet"
      }
    },
    "Pet": {
      "type": "object",
      "properties": {
        "id": {
          "type": "integer",
          "format": "int64"
        }
      }
    }
  }
}
```
Generates SWGPetMap.h
```objective-c
@interface SWGPetMap : NSMutableDictionary
```
and SWGPetMap.m
```objective-c
#import "SWGPetMap.h"

@implementation SWGPetMap

- (instancetype)init {
  self = [super init];

  if (self) {
    // initalise property's default value, if any
    
  }

  return self;
}

/**
 * Maps json key to property name.
 * This method is used by `JSONModel`.
 */
+ (JSONKeyMapper *)keyMapper
{
  return [[JSONKeyMapper alloc] initWithDictionary:@{  }];
}

/**
 * Indicates whether the property with the given name is optional.
 * If `propertyName` is optional, then return `YES`, otherwise return `NO`.
 * This method is used by `JSONModel`.
 */
+ (BOOL)propertyIsOptional:(NSString *)propertyName
{
  NSArray *optionalProperties = @[];

  if ([optionalProperties containsObject:propertyName]) {
    return YES;
  }
  else {
    return NO;
  }
}

/**
 * Gets the string presentation of the object.
 * This method will be called when logging model object using `NSLog`.
 */
- (NSString *)description {
    return [[self toDictionary] description];
}

@end
```
which fails to compile because toDictionary is not declared for NSMutableDictionary (it's a JSONModel method)

cbornet:
The above was generated with 2.1.5.
Also I think #1540 might not be fixed.

wing328:
@cbornet If I understand correctly, PetMap should not be generated as a separate model as it's simply a `map`, right ?

[...]

PR #3894, opened 2016-09-28
2208 fix Objc Mapping Generation

ClayAtWork:
### PR checklist

### Description of the PR

fixes https://github.com/swagger-api/swagger-codegen/issues/2208

Fixes Objective-C generation of maps via additionalProperties, to create a string-to-object map. The JSON Mapping library used by the generated code, JSONModel, see https://github.com/jsonmodel/jsonmodel/issues/341

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ObjcClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ObjcClientCodegen.java
@@ -359,7 +359,7 @@ public class ObjcClientCodegen extends DefaultCodegen implements CodegenConfig {
                         return getSwaggerType(p) + "<NSString*, " + innerTypeDeclaration + "*>*";
                     }
                 }
-                return getSwaggerType(p) + "<NSString*, " + innerTypeDeclaration + ">*";
+                return getSwaggerType(p) + "<" + innerTypeDeclaration + ">*";
             }
         } else {
             String swaggerType = getSwaggerType(p);
--- a/modules/swagger-codegen/src/test/java/io/swagger/codegen/objc/ObjcModelTest.java
+++ b/modules/swagger-codegen/src/test/java/io/swagger/codegen/objc/ObjcModelTest.java
@@ -224,7 +224,7 @@ public class ObjcModelTest {
         final CodegenProperty property1 = cm.vars.get(0);
         Assert.assertEquals(property1.baseName, "children");
         Assert.assertEquals(property1.complexType, "SWGChildren");
-        Assert.assertEquals(property1.datatype, "NSDictionary<NSString*, SWGChildren>*");
+        Assert.assertEquals(property1.datatype, "NSDictionary<SWGChildren>*");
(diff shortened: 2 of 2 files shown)


wing328:
cc @mateuszmackowiak

ClayAtWork:
What's the timeline for this? I would love to get some feedback from the other objc consumers

wing328:
@ClayAtWork After reading https://github.com/jsonmodel/jsonmodel/issues/341, one question I've is whether this change is compatible with iOS 9 only.

ClayAtWork:
@wing328 No, the previous version was the one that used Objective-C generics, which are erased at compile time. This uses a Protocol, which is available to the mapping library at runtime. 

I'm not actually sure why this was in the generator how it was because the code did not actually compile. You cannot mix a generic and a protocol in a single declaration like the previous tests declared:
`-        Assert.assertEquals(property1.datatype, "NSDictionary<NSString*, SWGChildren>*");`
notice SWGChildren is not a pointer type, it is actually a protocol, but NSString* is a pointer type, meaning the <...> is to indicate a generic type.

ClayAtWork:
Hey, can we evaluate this for being merged? Seems like an issue with the project is long term support for individual languages. Is there a single touchpoint for the Objective-C codegen?

fabiendevos:
@wing328 : We are already using swagger in production on Android, and we plan on using swagger in production on both iOS and Web (through ruby) very soon. This could be a great demonstration of the power of using swagger code gen to do faster multi platform development. We published a [blog post promoting the use of swagger in the past](http://eng.wealthfront.com/2016/08/31/enforce-rest-api-standard-with-swagger/) and we plan on publishing more, but we really need this to work for all clients or we will need to consider another technology, or even worse: forking this project. 😞  We [love open source](https://github.com/wealthfront) and would love to avoid the extreme solution of forking.

Please let us know what we can do to get this PR (and hopefully more in the future) get merged! :)

fehguy:
Hi @fabiendevos thanks for the PR.  We'll get this considered based on compatibility--if it's a breaking change and causes clients to force an IOS upgrade (not saying it does!) we can always introduce a new version of the generator.

PS I never saw that blog post!  You should mention `@SwaggerAPI` and tweet it out.

PPS Some of the team is right near your office and we can discuss changes faster in person...

fabiendevos:
Hi @fehguy! Thanks for the quick answer. Sorry for not mentioning @SwaggerAPI when we tweeted. We will in the future.
@ClayAtWork is telling me that the code it was previously generating did not compile. So he is not sure how this could break anything that wasn't already broken?
It would be great to discuss in person! I'll send you an email.
Thanks again!

wing328:
Test results look good. PR merged into master. Thanks again for the contribution.

fabiendevos:
@wing328 thanks a lot for merging this! We might have more contribution to come. :)

wing328:
@fabiendevos thanks! Looking forward to your PRs. Btw, you may find the following useful:

https://github.com/swagger-api/swagger-codegen/wiki/FAQ#how-do-i-run-integration-test-with-petstore-objc-api-client

PR #3892, opened 2016-09-28
Add a new cli command to output version information

MiguelGL:
### PR checklist

- [x] Read the [contribution guildelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [ ] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [x] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

This is as per discussed in #3007. I am using maven resources filtering support to generate a resource file with the ${project.version} stamped on it. 

Please note when running (from example from an IDE) from a non-maven env. this will lead to the 
version be **the literal** `${project.version}` string.

Issue #3007, opened 2016-05-31
CLI should be able to output its own version


ePaul:
##### Description

When getting a codegen CLI binary of unknown origin, it is currently not possible to know which version of codegen it is.

As an example, I yesterday downloaded the docker image `mydock/swagger-codegen-prebuilt` (from https://github.com/mydock/swagger-codegen-prebuilt, I guess) mentioned in #2991, and was unable to determine which version of swagger-codegen it is.
##### Swagger-codegen version

This applies to master from yesterday (2.2.0-SNAPSHOT) as well as all older versions I tested.
##### Suggest a Fix

There could be either an option (common are `-V`, `--version`) or a subcommand (`version`) which outputs the artifact version. Alternatively (or in addition) this could be included in the `help` command's output.

wing328:
@ePaul agree that it would be useful to show the version.

jimschubert:
@ePaul 

I agree with your feature suggestion.

If you ever have a problem like this with an older version and you're trying to debug an issue, you can extract the manifest from he jar and look at the `implementation-version`:

```
$ unzip -p -j swagger-codegen*.jar "META-INF/MANIFEST.MF" | grep version
implementation-version: 2.2.0-SNAPSHOT
```

wing328:
If anyone from the community wants to work on this, please reply to let us know (to avoid duplicated effort)

MiguelGL:
Just for the record, I just PR-ed wrt. this. I did not get to ask and coordinate as I thought it was a simple change.

ePaul:
This is already implemented in #3899, which is merged. Closing this issue now.

PR #3899, opened 2016-09-29
Add a new cli command to output version information (2nd attempt) #3892

MiguelGL:
### PR checklist

### Description of the PR

(Second PR, after fixing issues with my account email references)

This is as per discussed in #3007. I am using maven resources filtering support to generate a resource file with the ${project.version} stamped on it.

Please note when running (from example from an IDE) from a non-maven env. this will lead to the 
version be the literal `unset` string.

### DIFF
--- a/modules/swagger-codegen-cli/pom.xml
+++ b/modules/swagger-codegen-cli/pom.xml
@@ -18,6 +18,7 @@
         <resources>
             <resource>
                 <directory>src/main/resources</directory>
+                <filtering>true</filtering>
                 <excludes>
                     <exclude>logback.xml</exclude>
                 </excludes>
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ out/
 *.iws
 classpath.txt
 version.properties
+!modules/swagger-codegen-cli/src/main/resources/version.properties
 .project
 .classpath
 lib/*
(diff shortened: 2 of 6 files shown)


modules/swagger-codegen-cli/pom.xml
```diff
@@ -18,6 +18,7 @@
         <resources>
             <resource>
                 <directory>src/main/resources</directory>
+                <filtering>true</filtering>
```
ePaul: I fear there might have some side effects as we are filtering all the files under src/main/resources, not just the single one – including all the templates.

To make sure that there are none, could you actually run all the petstore samples and see if there are any changes in the output?

MiguelGL: Unless I'm missing anything build-wise, I think this should not be a problem because the filtering is enabled just for the `swagger-codegen-cli` artifact and should not apply to the resources from `swagger-codegen` artifact (which is where all the templates are included as resources).

Running the samples should be a good practice anyway, so I'll try to find the time to do so, however.

ePaul: Ah, good point, I missed this.

wing328:
@MiguelGL thanks for the PR, which has been merged into master.

Test result looks good:
```
$ java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar version
2.2.2-SNAPSHOT
```

PR #3909, opened 2016-09-30
WIP: Fixup rails5 server

apanzerj:
### PR checklist

### Description of the PR

There were 3 issues (so far) with the autogenerated Rails5 code:

1. The controllers are singular, they need to be plural. In Rails, if you have a model Foo, it looks for FoosController (object) and foos_controller.rb (file, in app/controllers). 

2. The tables in the migration were singular, they needed to be plural as well. The table for the model Foo is foos.

3. The method for validating presence is ```#validates_presence_of``` not ```#validate_presence_of```


apanzerj:
cc @neroleung

modules/swagger-codegen/src/main/resources/rails5/migrate.mustache
```diff
@@ -4,7 +4,7 @@
 
 class InitTables < ActiveRecord::Migration
   def change{{#models}}{{#model}}
-    create_table :{{classFilename}}, id: false do |t|{{#vars}}{{#isContainer}}
+    create_table :{{classFilename}}s, id: false do |t|{{#vars}}{{#isContainer}}
```
neroleung:
```
table_name = '{{classFilename}}'.pluralize.to_sym
create_table table_name, id: false do |t|{{#vars}}{{#isContainer}}
```

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Rails5ServerCodegen.java
```diff
@@ -281,8 +281,8 @@ public String toApiFilename(String name) {
         // replace - with _ e.g. created-at => created_at
         name = name.replaceAll("-", "_"); // FIXME: a parameter should not be assigned. Also declare the methods parameters as 'final'.
 
-        // e.g. PhoneNumberApi.rb => phone_number_api.rb
-        return underscore(name) + "_controllers";
+        // e.g. DefaultController => defaults_controller.rb
+        return underscore(name) + "s_controller";
```
apanzerj: in Java?

apanzerj: Yeah, the issue here is that we are using this for the file name and the code being executed is in Java, not Rails. Rails has the #pluralize method but there is no method in Java from a quick search.

Since this is the filename it can't be generated in Rails land.

wing328: You're right. I got confused during the review.

wing328:
The change looks good to me. Thanks for the contribution.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Rails5ServerCodegen.java
bluengreen: This is incorrect and breaks. The filename and the class name should match. It'd be better to make them whatever the tag is so your explicitly setting it to what you want. Controllers don't have inflections like models.

wing328: @bluengreen please open a new issue with details for tracking.

bluengreen: @wing328 - Already did here - #5053 and a PR here #5057 
Thanks

wing328: @bluengreen thanks for the PR, will review and let you know if we've any question.

PR #3919, opened 2016-10-04
[Java][jersey1] add Java6 support to Java (jersey1) API client

wing328:
### PR checklist

### Description of the PR

Add `supportJava6` flag to remove equals and hashCode methods in the model as these methods are not compatible with Java6.


k1w1m8:
If we wanted to keep the equals and hashcode methods for Java, we could take the approach of replacing the java.util.Objects calls with the calls to other libraries which can do the same as we did for PR #4033 eg commons.lang or Guava.

wing328:
@k1w1m8 your suggestion sounds good to me.

ErnestoLazaroCorredera:
@ApiModelProperty in my model's class is not compatible with java 6: 

java.lang.UnsupportedClassVersionError: io/swagger/annotations/ApiModelProperty : Unsupported major.minor version 51.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClassCond(ClassLoader.java:631)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:615)

Issue #3922, opened 2016-10-04
Ignores 'required' tag in model.mustache in python


vishalsanfran:
##### Description

Swagger ignores 'required' tag in model.mustache in python.
The setter function for these properties does not allow null values, unlike ruby and other languages

##### Swagger-codegen version

2.2.1 (latest)

##### Steps to reproduce

1) Required is false

"license":
{"type":"string","description":"license key", required: false}

Expected output:
@license.setter
    def license(self, license):
        self._license = license

2) Required is true
"license":
{"type":"string","description":"license key", required: true}

Expected output:
@license.setter
    def license(self, license):
        if not license:
              raise ValueError("Invalid value for license, must not be `None`")
        self._license = license

Currently, output is always 2) whether 'required' is true or false

##### Suggest a Fix

Its a simple 2 line fix, very similar to the ruby code
i'll submit it

wing328:
PR merged into master. Thanks for the contribution.

PR #3923, opened 2016-10-04
required tag is used in model, allows null values

vishalsanfran:
### PR checklist

### Description of the PR

Swagger ignores 'required' tag in model.mustache in python.
The setter function for these properties does not allow null values, unlike ruby and other languages

Issue: https://github.com/swagger-api/swagger-codegen/issues/3922

### DIFF
--- a/modules/swagger-codegen/src/main/resources/python/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/model.mustache
@@ -90,7 +90,7 @@ class {{classname}}(object):
 {{^isEnum}}
 {{#hasValidation}}
 {{#required}}
-        if not {{name}}:
+        if {{name}} is None:
             raise ValueError("Invalid value for `{{name}}`, must not be `None`")
 {{/required}}
 {{#maxLength}}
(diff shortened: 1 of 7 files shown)


wing328:
@vishalsanfran thanks for the PR. Would you please run `./bin/python-petstore.sh` to update the Python Petstore API client so that the CI can test the changes?

vishalsanfran:
pr is updated

modules/swagger-codegen/src/main/resources/python/model.mustache
```diff
@@ -89,9 +89,10 @@ class {{classname}}(object):
 {{/isEnum}}
 {{^isEnum}}
 {{#hasValidation}}
-
+{{#required}}
         if not {{name}}:
```
wing328: I think it would be an issue if {{name}} is 0 (integer) or False (boolean)

What about using `if {{name}} is None` to check instead?

vishalsanfran: yes, makes sense, i'll update

jdevera: This particular issue was blocking me, since I got the invalid value due to a 0.

wing328: @vishalsanfran @jdevera FYI. I've filed https://github.com/swagger-api/swagger-codegen/pull/3944 with the fix.

wing328: #3944 merged into master. Please pull the latest master to give it a try.

PR #3927, opened 2016-10-05
[Objc] Added support for lower case discriminator

sylvain-abadie:
### PR checklist

### Description of the PR

Add support for lowercase discriminator. Currently if the discriminator is "cat" (and not "Cat") the Cat class won't be deserialized. This just adds a test to deserialize by adding an uppercase to the discriminator.


modules/swagger-codegen/src/main/resources/objc/model-body.mustache
```diff
@@ -28,6 +28,9 @@
          return [super initWithDictionary:dict error:err];
     }
     Class class = NSClassFromString([@"{{classPrefix}}" stringByAppendingString:discriminatedClassName]);
+    if(!class){
+      class = NSClassFromString([@"{{classPrefix}}" stringByAppendingString:[discriminatedClassName capitalizedString]]);
```
wing328: We prefer 4-space indentation. Please also add a space before `{` in line 31

wing328:
Please also run `./bin/objc-petstore-all.sh` to update ObjC petstore sample.

wing328:
Would this change cater both lowercase discriminator and discriminator in the original form (e.g. PascalCase) ?

sylvain-abadie:
Yes the original behavior is preserved. In case the original behavior doesn't succeed, it tries to add an uppercase to the first letter of the discriminant.

wing328:
@sylvain-abadie thanks for the explanation. PR merged into master.

PR #3935, opened 2016-10-06
Fix duplicate entries (property inheritance)

j4velin:
If 'supportsInheritance' is set to true, a model property is part of both the list 'properties' and 'allProperties'. The DefaultCodegen::addVars method is then called with both lists, so that the property is added twice to the m.requiredVars/m.optionalVars and m.readOnlyVars/m.readWriteVars lists

Example shown here: https://gist.github.com/j4velin/915d5ad70141685e8b588781adf7c4c7

Change-Id: I76e562fb629c28c869c7e4622e369cc79df7620a


wing328:
@j4velin I tried the latest master and couldn't repeat the issue. I wonder if you can give it a try and let me know if it works for you. Thanks!

(sorry for taking this long to review...)

PR #3938, opened 2016-10-06
Add jaxrs-cxf-cdi generator

nickcmaynard:
Generate a JAX-RS server that assumes an Apache CXF framework on a Java EE server supporting CDI.

Similar to the original JAXRS generator, this creates API and Service classes in /src/gen/java and a sample ServiceImpl in /src/main/java.  The API uses CDI to get an instance of ServiceImpl that implements the Service interface.


nickcmaynard:
Bad sample generation, closing.

PR #3949, opened 2016-10-07
Sanitize tuples properly

iCEAGE:
In Python, tuples are fixed size in nature whereas lists are dynamic. So they should be sanitized as same as lists. Failure case here is when you have a query or post parameter as Datetime.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/python/api_client.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/api_client.mustache
@@ -176,12 +176,12 @@ class ApiClient(object):
         :param obj: The data to serialize.
         :return: The serialized form of data.
         """
-        types = (str, float, bool, tuple) + tuple(integer_types) + (text_type,)
+        types = (str, float, bool) + tuple(integer_types) + (text_type,)
         if isinstance(obj, type(None)):
             return None
         elif isinstance(obj, types):
             return obj
-        elif isinstance(obj, list):
+        elif isinstance(obj, (list, tuple)):
             return [self.sanitize_for_serialization(sub_obj)
                     for sub_obj in obj]
         elif isinstance(obj, (datetime, date)):
--- a/modules/swagger-codegen/src/main/resources/python/api_client.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/api_client.mustache
@@ -176,7 +176,7 @@ class ApiClient(object):
         :param obj: The data to serialize.
         :return: The serialized form of data.
         """
-        types = (str, float, bool) + tuple(integer_types) + (text_type,)
+        types = (str, float, bool, bytes) + tuple(integer_types) + (text_type,)
         if isinstance(obj, type(None)):
             return None
         elif isinstance(obj, types):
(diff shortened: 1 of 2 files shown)


wing328:
@iCEAGE thanks for the PR.

cc @scottrw93

wing328:
@iCEAGE please run `./bin/python-petstore.sh` to update the Petstore sample so that the CI can test the change.

wing328:
Seems like CI tests for Python Petstore failed. Here is the log: https://s3.amazonaws.com/archive.travis-ci.org/jobs/166440726/log.txt

Look for "petstore_api.py" and you should see the error messages.

wing328:
@iCEAGE do you need help fixing the issue?

cc @scottrw93

iCEAGE:
@wienczny Yes please. I just tried to skip file sanitization but it seems it doesn't work.

wing328:
@iCEAGE can you share a spec to easily reproduce the issue? (assuming the spec can be shared publicly)

iCEAGE:
@wing328 do you mean the original issue which is datetime doesn't get converted to ISO format?

iCEAGE:
@wing328 Please find the spec [here](https://gist.github.com/iCEAGE/7d26c125f5c6ebd056cacb3c0a6271f0). The issue is story_published_at parameter as date-time doesn't get converted (sanitized) to ISO8601 format.

Hope this helps. Thanks

scottrw93:
@iCEAGE I think there are two issues here

1) In your `elif isinstance(obj, tuple):` section, you return a list instead of a tuple even though a tuple was passed in. You need to return a tuple, not a list here, so just cast it before returning it

2) There also seems to be a bug around handling `bytes` that was not caught by our unit tests until this change. `[This line](https://github.com/swagger-api/swagger-codegen/pull/3949/files#diff-027e164ab608834aec24a0ee3c3fc9c5R179)` should be `types = (str, float, bool, bytes) + tuple(integer_types) + (text_type,)`.

As a side note, you can run the test locally by running:
`mvn clean package`
`./bin/python-petstore.sh`
`cd samples/client/petstore/python`
`virtualenv venv`
`source ./venv/bin/activate`
`./test_python2_and_3.sh` or `nosetests tests/test_pet_api.py:PetApiTests.test_upload_file` to just run the failling test

modules/swagger-codegen/src/main/resources/python/api_client.mustache
```diff
@@ -176,14 +176,20 @@ class ApiClient(object):
...
+        types = (str, float, bool, bytes) + tuple(integer_types) + (text_type,)
         if isinstance(obj, type(None)):
             return None
         elif isinstance(obj, types):
             return obj
         elif isinstance(obj, list):
             return [self.sanitize_for_serialization(sub_obj)
                     for sub_obj in obj]
+        elif isinstance(obj, tuple):
+            if len(obj) == 1 and isinstance(obj[0], list) and len(obj[0]) == 3:
```
scottrw93: I don't think this should be left in, it is not necessary for handling multipart files and could lead to unintended behaviour in the unlikely chance a list had this structure and it wasnt a file

iCEAGE:
@scottrw93 Thanks for your help. I think now all looks good.

wing328:
@scottrw93 thanks for reviewing the PR

@iCEAGE thanks for the contribution, which has been merged into master.

PR #3989, opened 2016-10-13
WIP - Polymorphic Javascript client

triptec:
### PR checklist
### Description of the PR

Something to discuss.
I'm trying to get polymorphic stuff to work, ie when running Animal.constructFromObject with Dog data I want a Dog model back.

There are thing that should probably be within useInheritance and stuff like that. My biggest problem is being able to call Dog.constuctFromObject from within Animal. Dog is importing Animal and Animal would have to import Dog so we got a circular dep =(. Now I use SwaggerPetstore.Dog.constructFromObject(data, obj) but that is dependent on that you SwaggerPetStore = require('./src'), ie. someOtherVariable = require('./src') wouldn't work.

So some pointers please.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenModel.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenModel.java
@@ -1,11 +1,6 @@
 package io.swagger.codegen;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeSet;
-import java.util.Objects;
+import java.util.*;
 
 import io.swagger.models.ExternalDocs;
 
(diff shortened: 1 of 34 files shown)


triptec:
I'm thinking we would like to use something like the factory pattern and add factory classes for parent classes like
{{parentModel}}Factory.constructFromObject()

any thoughts on that?

wing328:
@triptec I'm sorry that I didn't have a chance to review. I"ll take a look soon and provide a detailed reply.

hpatoio:
We really need this feature. Any plan to merge this PR ? 
Is there a workaround for this ?

triptec:
@hpatoio I didn't implement this enough for merging, just something to discuss so I think it would require a considerable amount of work to finish, and that's if this is inline with the project.

hpatoio:
thanks @triptec sound strange to me that the official generator doesn't support all swagger specs.
Starting from the very same swagger.json I've generated the PHP client and it does support _discriminator_

PR #3281, opened 2016-07-02
[CXF] Add CXF test templates & Bean-Validation-API support

jfiala:
* added the CXF client dependency (as this allows to run unittests as well inside the generated server)
* added test case templates
* also added Bean-Validation-API support for CXF (see #2549)

In fact, the current JavaCXFServerCodegen is exactly identical to the CXF client.

Only adding implementation stubs would make it to a the server project.

Issue #2615, opened 2016-04-15
Jaxrs-CXF missing mustache variables for javax.validation framework


FlorentBarreau:
Hello,

I'm using swagger codegen maven plugin to auto-generate endpoints and models for a REST API project. I personalize mustache template to generate javax.validation annotations such as 
@Valid, @Min, @Max, @Size, @Pattern, etc... on Query/PathParam and Models 

So far I'm facing to those 2 Issues : 
- {{minimum}} and {{maximum}} returns Double values because of CodegenProperty minimum and maximum types which are Double (and I understand it could be useful to have Double instead of Integer...)
  Issue is that I can't use @Min({{minimum}}) and @Max({{maximum}}) on my mustache templates : it will break at compilation because it requires Integer values instead of Double
  Now I'm looking for a fancy solution without modifying codegen source code if possible. Is CodegenProperty overridable by config ? Is there any secret param I can use to change that behavior ?
- For Endpoint QueryParams, I have this problem : @NotNull must be added only if required = true, but I can only test if tag 'required' is present or not, through {{#required}}, and so I cannot use same tricks than in java's templates
  So far I can work-around this issue by deleting all 'required = false' in swagger specification yaml and using {{#required}}@NotNull{{/required}} but this will generate functional issues if people modify yaml...
  So I was looking for some kind of isRequired tag or another solution without modifying codegen source if possible, but haven't found one yet.

At worst, if those issues are not solvable by config, is extending AbstractJavaJAXRSServerCodegen a good idea to generate new properties and use them for mustache templates ?

FYI I'm adding my overrides of jaxrs-cxf templates (dev on-going...), maybe it can help to understand/reproduce

[jaxrs-cxf_javax_validation_mustache_templates.zip](https://github.com/swagger-api/swagger-codegen/files/221320/jaxrs-cxf_javax_validation_mustache_templates.zip)

I'll be glad to post those templates and contribute to this project once all issues solved.

Thanks and regards,
Florent Barreau

wing328:
@FlorentBarreau ideally {{minimum}} and {{maximum}}  should be determined by langauges (e..g 3.7F for float value in C#) and I would foresee some changes in the codegen class to accommodate that.

For {{#required}}, are you looking for {{^required}} (the opposite of "required" being set to true) ?

And most importantly thanks for offering help to contribute to this project.

FlorentBarreau:
@wing328 thanks for your reply !

[...]

Issue #2549, opened 2016-04-11
Client-generation: Allow data validation on the client side (using minLength/maxLength)


jfiala:
Currently the generated Java-clients perform required-validation automatically. 
Do you think this makes sense for minLength/maxLength as well, to reject the request on the client side before sending to the server?

For Java the JSR-303 Bean Validation API could be used for this automatically.
For other languages maybe a JSON-schema validation could be used.

wing328:
@jfiala yes, it's definitely something we want to leverage and I'm sure the community would appreciate your contribution on that.

Let me know if you need help.

Ref: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

wing328:
@jfiala may I know if you need help adding the JSR-303 Bean validation to the Java API client?

jfiala:
@wing328 thank you, maybe we can sketch some of the main features:
* I think it should be configurable so users can turn it on/off if they don't want bean validation annotations in the client code
* Bean Validation would make sense for every Java generator, but with the current layout (everything in mustache templates) it means I'd have to add it to all templates, do you have any better ideas?
* More general: Currently there are generators for server/client code kept separately, but shouldn't they be provided together (as for CXF for example where one can switch to generate impl/interface) as the model is always shared between server/client and should be the same to the bigger extent?

jfiala:
Currently we could add support for the following annotations:
http://docs.oracle.com/javaee/6/tutorial/doc/gircz.html

* @NotNull (required=true)
* @Null (required=false)
* @Size (minLength/maxLength)
* @Min (min)
* @Max (min)
* @Pattern (pattern)
* @DecimalMin (min for BigDecimal)
* @DecimalMax (max for BigDecimal)

Pls let me know if other annotations make sense in the scope of Swagger?

wing328:
@jfiala I think your list covers what we've already covered for other generators (e.g. PHP, Ruby).

> Bean Validation would make sense for every Java generator, but with the current layout (everything in mustache templates) it means I'd have to add it to all templates, do you have any better ideas?

Please add it to Java okhttp-gson API client to begin with and we'll do the same for other Java client/server generator one by one.

cbornet:
Duplicate #2412 ?
See also #2663.

Also Java client pojo.mustache only differ by a few annotations between gson and jackson (@Serialized vs @JsonProperty) so it should be very easy to mutualize them under a parent mustache with a "gson" tag.

[...]

Issue #2017, opened 2016-02-02
Generating Java Rest Client based on CXF 


peteryhwong:
As there is now a provision to generate Java CXF-specific Rest **server** stub (https://github.com/swagger-api/swagger-codegen/issues/1854), we would like to see codegen to support generation of Java CXF-specific Rest **client**. This is because we use CXF as our JAX-RS implementation and would like to leverage this implementation for our client libraries (specifically for inter-service interaction). The current client generator uses Jersey which causes conflict when being used in conjunction with CXF.

wing328:
@peteryhwong the Java API client supports many HTTP libraries (e.g. Retrofit, Jersey2.x, etc). Please run `java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar config-help -l java` for a list of HTTP libraries support.

Ref: https://github.com/swagger-api/swagger-codegen#customizing-the-generator

peteryhwong:
@wing328 thanks for this. If I understand correctly, the ```JavaClientCodeGen``` only seems to support the following:

```
library template (sub-template) to use:
	    <default> - HTTP client: Jersey client 1.18. JSON processing: Jackson 2.4.2
	    jersey2 - HTTP client: Jersey client 2.6
	    feign - HTTP client: Netflix Feign 8.1.1.  JSON processing: Jackson 2.6.3
	    okhttp-gson - HTTP client: OkHttp 2.4.0. JSON processing: Gson 2.3.1
	    retrofit - HTTP client: OkHttp 2.4.0. JSON processing: Gson 2.3.1 (Retrofit 1.9.0)
            retrofit2 - HTTP client: OkHttp 2.5.0. JSON processing: Gson 2.4 (Retrofit 2.0.0-beta2)
```
which does not include CXF client. Unless you are suggesting an alternative to CXF?

wing328:
Yes, that's the list of HTTP libraries supported. CXF client is not supported yet and we would welcome contribution from the community to support that.

You mentioned Jersey (1.x) has conflicts when being used in conjunction with CXF and that's why I suggested other HTTP libraries, which might work with CXF. Do you also encounter conflicts when using other HTTP libraries with CXF?

wing328:
And here is the contributing guideline: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

wing328:
@peteryhwong do you need help from us to add CXF client support to the Java client?

[...]

PR #4003, opened 2016-10-14
Java BeanValidation + JAXRS CXF server generator

jfiala:
### Overview
* added interfaces to model language feature support (io.swagger.codegen.languages.features)
* add beanvalidation annotation support
* add new language for CXF client (jaxrs-cxf, server moved to jaxrs-cxf-server)

### Language Java

* add beanvalidation annotation support to language java #2549 

### Language jaxrs-cxf-server (renamed from jaxrs-cxf)
* add beanvalidation annotation support  to language jaxrs-cxf #2615 (using centralized beanValidation.mustache file)

* add spring server generator to jaxrs-cxf (see also #2017 )
* spring server: support CXF features using CLI-switches:
   * SwaggerFeature
   * WadlGenerator
   * Multipart
   * Gzip
   * Logging
   * automatic BeanValidation (JEE 7+ - BeanValidation 1.1)
* support CXF features for Unittests using CLI switches:
   * Gzip
   * Logging
   * automatic BeanValidation (BeanValidation 1.1)

* support Spring Boot embedded server
* support Spring Boot integration tests

### Language jaxrs-cxf (new)
* implement new language and reuse cxf templates (see #2017)
* generate complete client including pom.xml
* support CXF features for Unittests using CLI switches (see above)

for more details also see former PR #3281

### CXF minor improvements
* improve CXF QueryParam BeanValidation formatting (currently reused from beanValidation.mustache, but will add a newline after each beanvalidation annotation).
* add separate flags for controlling junit test gzip/logging/... support (is currently generated automatically for server + junit-tests if switched on)* add option to generate Jboss files (jboss-web.xml etc.)
* add a JbossFeature and flag to only generate jboss-specific files if requested
* add flag for switching JAXB annotations on/off
* check for void in apiServiceImpl to suppress the return statement (instead of returnType, which is void for no returnType...)

### Other languages
* check for void in test-templates to generate the correct statement (currently: void response = ....)
* add BeanValidation annotation support to all Java languages
* add JbossFeature to other languages generating jboss-web.xml
* check JAXRS/CXF languages for consolidation of mustache templates
* it would be great to have automatic contract tests for the different CLI options, pls let me know if I should supply ones (I'd prefer to have automatic unit tests for this, so the mvn build will fail if anything is changing).

### Further CXF improvements
* maybe add further CXF configuration scenarios (blueprint)


wing328:
@jfiala thanks! This one looks a lot cleaner. Will merge over the weekend.

nickcmaynard:
Watching this as the CXF stuff interests me (and I'd like to investigate merging cxf/cxf-cdi some time).  Looking at the diff there are quite a lot of removed files in here... @jfiala are you sure this is good?

jfiala:
@rynger I had to remove the @Path annotation at the interface level, as the @Path annotation including the full path is already part of the operations. Having the @Path annotation at both positions leads to wrong results.

With the annotation at the interface level, the generated code looks like:
``` java
@Path("/contextname/myservices")
public interface {{classname}}  {
    @Path("/myservices/mymethod")
    public Response mymethod (String param1);
}
```
Then uses have to access the REST api using /contextname/myservices/myservices/mymethod (duplicate myservices part).
The simplest solution is to simply remove the @Path annotation at the interface level.

jfiala:
@nickcmaynard  thx for the hint, cleaned everything up now.

jfiala:
@wing328: regarding DefaultCodegen.toRegularExpression: this has been extended to call addRegularExpressionDelimiter(). However, the Java BeanValidation patterns don't need the "/" to be added automatically, so I removed the call in AbstractJavaCodegen.java. Please let me know if it is needed somewhere else in the Java language with the slashes added.

jfiala:
@wing328: Finished the CXF part so far, some ideas for further issues:
* add BeanValidation annotation support to all Java languages
* it would be great to have automatic contract tests for the different CLI options, pls let me know if I should supply ones (I'd prefer to have automatic unit tests for this, so the mvn build will fail if anything is changing).
* check JAXRS/CXF languages for consolidation of mustache templates
* maybe add further CXF configuration scenarios (blueprint)

jfiala:
@sberyozkin Can you pls review the CXF features I've added and let me know if you're missing something?

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaCXFServerCodegen.java
```diff
@@ -36,14 +64,98 @@ public JavaCXFServerCodegen()
...
+		    additionalProperties.put(propertyKey, booleanValue);
+		}
+		
+		return booleanValue;
+	} 
+    
+    @Override
+    public String getName()
+    {
+        return "jaxrs-cxf-server";
```
wing328: @jfiala  Shall we keep `jaxrs-cxf` and for the client, rename it to jaxrs-cxf-client` instead? (to make this PR backward compatible)

wing328:
> regarding DefaultCodegen.toRegularExpression: this has been extended to call addRegularExpressionDelimiter(). However, the Java BeanValidation patterns don't need the "/" to be added automatically, so I removed the call in AbstractJavaCodegen.java. Please let me know if it is needed somewhere else in the Java language with the slashes added.

Yes, please override meet the language requirement

jfiala:
I'll do later today: 
* replace the tabs (sorry I'll adapt IDE settings)
* change the language to jaxrs-cxf-client for backwards compatibility 
* remove TODO comment on toRegularExpression
* add a JbossFeature and flag to only generate jboss-specific files if requested

I'll let you know then.

wing328:
> add a JbossFeature and flag to only generate jboss-specific files if requested

I would suggest we do that after we merge this PR into master as this PR already contains too many changes.

jfiala:
yes, thx I added it to the tasks postponed to later. 
I have updated everything according to your comments (tabs are still present at other files I didn't touch, I'd suggest to setup a maven task to do checkstyle for tabs (e.g. the apache cxf project uses this))!

wing328:
About using maven plug-in to enforce coding style, it's tracked [here](https://github.com/swagger-api/swagger-codegen/issues/3972)

jfiala:
moved convertPropertyToBooleanAndWriteBack to DefaultCodegen.java.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaCXFServerCodegen.java
jfiala: yes, I renamed it to stay backwards compatible.

jfiala:
@wing328 pls let me know if you need anything else here, it would be great if you can pull this in soon ...

wing328:
@jfiala yup, will continue the review this weekend and merge if no further question from me.

jfiala:
@wing328 I just did some integration tests using swagger-codegen-maven-plugin, version 2.2.1. 
If I run using the maven-plugin, it is unable to find the subtemplate referenced at JavaJaxRS/cxf/queryParams.mustache:
{{>../../java/beanValidation}}
java.lang.RuntimeException: can't load template JavaJaxRS\cxf\..\..\java\beanValidation.mustache

However, at JavaJaxRS/cxf/pojo.mustache the same reference works without troubles.
And both references are working if I use the CLI instead of the maven plugin.
Also setting the templateDirectory at the configuration of the maven plugin doesn't change the behaviour.

Subtemplating across directories in mustache-files hasn't been used in the project so far, so this hasn't popped up yet.

Pls let me know if I should use a copy of beanValidation.mustache for cxf for now until this is analyzed/fixed ...? Should I create a separate issue for this?

wing328:
> Pls let me know if I should use a copy of beanValidation.mustache for cxf for now until this is analyzed/fixed ...? Should I create a separate issue for this?

Please avoid using relative path and use a copy of the mustache file instead.

jfiala:
@wing328 changed to local copy + successfully verified with maven.

wing328:
@jfiala I generated the `jaxrs-cxf-client` using your branch and got the following error when running `mvn test`:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:testCompile (default-testCompile) on project swagger-jaxrs-client: Compilation failure: Compilation failure:
[ERROR] /private/var/tmp/jaxrs-cxf-client/tmp5/src/test/java/io/swagger/api/StoreApiTest.java:[110,24] no suitable constructor found for Long()
[ERROR] constructor java.lang.Long.Long(java.lang.String) is not applicable
[ERROR] (actual and formal argument lists differ in length)
[ERROR] constructor java.lang.Long.Long(long) is not applicable
[ERROR] (actual and formal argument lists differ in length)
[ERROR] /private/var/tmp/jaxrs-cxf-client/tmp5/src/test/java/io/swagger/api/PetApiTest.java:[97,22] no suitable constructor found for Long()
[ERROR] constructor java.lang.Long.Long(java.lang.String) is not applicable
[ERROR] (actual and formal argument lists differ in length)
[ERROR] constructor java.lang.Long.Long(long) is not applicable
[ERROR] (actual and formal argument lists differ in length)
[ERROR] /private/var/tmp/jaxrs-cxf-client/tmp5/src/test/java/io/swagger/api/PetApiTest.java:[114,31] java.util.List is abstract; cannot be instantiated
[ERROR] /private/var/tmp/jaxrs-cxf-client/tmp5/src/test/java/io/swagger/api/PetApiTest.java:[130,29] java.util.List is abstract; cannot be instantiated
[ERROR] /private/var/tmp/jaxrs-cxf-client/tmp5/src/test/java/io/swagger/api/PetApiTest.java:[146,22] no suitable constructor found for Long()
[ERROR] constructor java.lang.Long.Long(java.lang.String) is not applicable
[ERROR] (actual and formal argument lists differ in length)
[ERROR] constructor java.lang.Long.Long(long) is not applicable
[ERROR] (actual and formal argument lists differ in length)
```

Here is the command I used:
```
-i modules/swagger-codegen/src/test/resources/2_0/petstore.yaml -l jaxrs-cxf-client \
-o /var/tmp/jaxrs-cxf-client/tmp5/ 
```
Did you get similar error when using the `jaxrs-cxf-client` client?

jfiala:
@wing328 I did a clean install across the whole swagger-codegen project and it ran successfully without errors, probably because the generated projects themselvers are not verified with mvn clean install automatically...

I verified the errors you had now directly using:
io.swagger.codegen.Codegen -i C:\projects\swagger-codegen_jfiala\modules\swagger-codegen\src\test\resources\2_0\petstore.yaml -l jaxrs-cxf-client -o c:\temp\swagger-codegen-x_cxf_client -c c:\temp\swagger_config.json

swagger_config.json has all CLI flags turned on:
```
{
"useBeanValidation": "true",
"useGzipFeature": "true",
"useLoggingFeature": "true"
"generateApiTests": "true"
}
```
And then run mvn clean install on the generated project:
At the test client, currently placeholders are generated for all the parameters with new <type> and the method invocation is not commented out. This is causing the errors during the compilation.

I can replace them with "null" and comment out the invocation itself so the generated test cases can be compiled (like this is the case for the Java test cases)?

wing328:
> I can replace them with "null" and comment out the invocation itself so the generated test cases can be compiled (like this is the case for the Java test cases)?

Yes, please.

wing328:
FYI. For `jaxrs-cxf`, I got the following errors related to implementation class:
```
[ERROR] /private/var/tmp/jaxrs-cxf/tmp5/src/main/java/io/swagger/api/impl/PetApiServiceImpl.java:[25,16] cannot return a value from method whose result type is void
[ERROR] /private/var/tmp/jaxrs-cxf/tmp5/src/main/java/io/swagger/api/impl/PetApiServiceImpl.java:[31,16] cannot return a value from method whose result type is void
[ERROR] /private/var/tmp/jaxrs-cxf/tmp5/src/main/java/io/swagger/api/impl/PetApiServiceImpl.java:[55,16] cannot return a value from method whose result type is void
[ERROR] /private/var/tmp/jaxrs-cxf/tmp5/src/main/java/io/swagger/api/impl/PetApiServiceImpl.java:[61,16] cannot return a value from method whose result type is void
[ERROR] /private/var/tmp/jaxrs-cxf/tmp5/src/main/java/io/swagger/api/impl/UserApiServiceImpl.java:[24,16] cannot return a value from method whose result type is void
[ERROR] /private/var/tmp/jaxrs-cxf/tmp5/src/main/java/io/swagger/api/impl/UserApiServiceImpl.java:[30,16] cannot return a value from method whose result type is void
[ERROR] /private/var/tmp/jaxrs-cxf/tmp5/src/main/java/io/swagger/api/impl/UserApiServiceImpl.java:[36,16] cannot return a value from method whose result type is void
[ERROR] /private/var/tmp/jaxrs-cxf/tmp5/src/main/java/io/swagger/api/impl/UserApiServiceImpl.java:[42,16] cannot return a value from method whose result type is void
[ERROR] /private/var/tmp/jaxrs-cxf/tmp5/src/main/java/io/swagger/api/impl/UserApiServiceImpl.java:[60,16] cannot return a value from method whose result type is void
[ERROR] /private/var/tmp/jaxrs-cxf/tmp5/src/main/java/io/swagger/api/impl/UserApiServiceImpl.java:[66,16] cannot return a value from method whose result type is void
[ERROR] /private/var/tmp/jaxrs-cxf/tmp5/src/main/java/io/swagger/api/impl/StoreApiServiceImpl.java:[24,16] cannot return a value from method whose result type is void
[INFO] 11 errors 
```
I believe these errors already present in the latest master so I'll create another issue to track it.

jfiala:
@wing328: fixed the tests, now you can generate with the petstore without results after mvn install.

Regarding Impl.: The "return null;" will be generated only if the returnType is set, I didn't get any errors here! The problem is that "void" is set as returnType, so we need a flag to indicate that the returnType is void, then we can check for this in the mustache templates.

The other server stub generators evade this by using a generic "Response", but IMHO it is much better to deal with the "real" objects instead of the generic "Response", the Response is then wrapped automatically by CXF.

jfiala:
@wing328 the current master jaxrs-server doesn't include the apiServiceImpl.mustache, this was not supported there (https://github.com/swagger-api/swagger-codegen/tree/master/modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf)

wing328:
> To assemble the implementation class correctly (without compile-errors) I need a flag to check if the returnType is void or not to generate correctly.

I'll look into that after this PR is merged into master and add the flag if needed.

jfiala:
great, thank you!

wing328:
```
   /**
   * Get integer
   * minimum: 10.0
   * maximum: 100.0
   * @return integer
  **/
  //@Min(10.0)
  //@Max(100.0)
  @ApiModelProperty(example = "null", value = "")
  public Integer getInteger() {
    return integer;
  }
```
I noticed that `@Min` and `@Max` are commented out, probably due to decimal value used instead of integer.

If we can fix the min and max value with respect to the type, we should be able to generate uncommented constraints, right?

jfiala:
Yes exactly, this has been discussed here #2615, if Codegen will output integer values here, we can remove the comments in the template (or use a different mustache variable).

wing328:
@jfiala I've created the branch `jfiala-jaxrs_cxf_2017b` with resolved conflicts and will squash and merge the commits (owned by you) into master after all CI tests pass.

wing328:
@jfiala merged via #4068

PR #4033, opened 2016-10-19
Add Java6 compatibility for Jersey2

k1w1m8:
### PR checklist

### Description of the PR

Java6 compatibility is being added via supportJava6 option. This PR makes a small change to Jersey2 to replace usage of Java 7 class java.nio.file.Files with reference to Apache commons.io.FileUtils. 

I'm not sure exactly how to manage the additional dependency for commons io. Originally I saw it as part of swagger-codegen but of course that is only for the generation process. I doubt we want to add it as a dependency on the swagger runtime code, are you happy for it be up to users of supportJava6 to add manually?

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
@@ -22,7 +22,13 @@ import org.glassfish.jersey.media.multipart.MultiPartFeature;
 
 import java.io.IOException;
 import java.io.InputStream;
+
+{{^supportJava6}}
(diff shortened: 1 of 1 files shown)


wing328:
@k1w1m8 thanks! The change looks good to me.

k1w1m8:
Great, thanks. 

Are you happy for it be up to users of supportJava6 to add commons.io dependency manually? Is there somewhere we should document that?

wing328:
Ideally we should do it automatically in the pom.xml, build.gradle, sbt.build so as to save the developer's time.

k1w1m8:
Ah yes, of course. Doh. Forgot we generated those too.

Issue #4025, opened 2016-10-18
[TYPESCRIPT] typescript-fetch cannot pass fetch specific request options


alechill:
##### Description

Through the current fetch api template the fetch options is created internally by declaring simply the request method, and appended to in special cases such as headers/body/file etc. These cases not however exhaustive, and a bunch of request options in the spec (https://fetch.spec.whatwg.org/#concept-request) that may need be supplied are not accounted for. 

For example, cookies are not automatically sent via fetch api, and must instead have the `{ credentials: 'include' }` option set to inform it to do so.
##### Swagger-codegen version

2.2.1 (not a regression)
##### Suggest a Fix

Allow passing in custom request options which are `assign`ed to the fetch options at their point of creation, so that any of the already accounted for special cases are not overridden. This is flexible enough to encompass all use cases in a single implementation, and can be done so without breaking backwards compatibility

taxpon:
@alechill @wing328 I think this issue can be closed

wing328:
@taxpon thanks for reviewing (and sorry I missed your message)

PR #4038, opened 2016-10-19
[TYPESCRIPT] Support passing custom RequestInit options in typescript-fetch client…

alechill:
…, to enable fetch API specific options (eg including credentials, enabling CORS etc)
### PR checklist
### Description of the PR

Fixes ticket https://github.com/swagger-api/swagger-codegen/issues/4025

Tests updated to include calling with and without options, and run both locally and on travis (https://travis-ci.org/alechill/swagger-codegen)

Would have liked to have made the options a `RequestInit` type but if generated api package is included as an npm module in another project (as the test suite does) it would require an extra ambient typing for the parent project (for isomorphic-fetch), and so would make it a breaking change (mildly - not in code but in typings WTFs). However, should still get the type safety (when it is internally cast to `RequestInit`) if the generated api is used directly within the same project. Best compromise I hope


wing328:
cc @Vrolijkx

samples/client/petstore-security-test/typescript-fetch/api.ts
```diff
@@ -115,11 +115,26 @@ export const FakeApiFp = {
...
 export class FakeApi extends BaseAPI {
     /** 
-     * To test code injection *_/ &#39; \&quot; &#x3D;end \\r\\n \\n \\r
-     * @param test code inject * &#39; &quot; &#x3D;end rn n r To test code injection *_/ &#39; \&quot; &#x3D;end \\r\\n \\n \\r
+     * To test code injection *_/ &#39; \&quot; &#x3D;end -- \\r\\n \\n \\r
+     * @param test code inject * &#39; &quot; &#x3D;end  rn n r To test code injection *_/ &#39; \&quot; &#x3D;end -- \\r\\n \\n \\r
      */
-    testCodeInjectEndRnNR(params: {  test code inject * &#39; &quot; &#x3D;end rn n r?: string; }) {
-        return FakeApiFp.testCodeInjectEndRnNR(params)(this.fetch, this.basePath);
+    testCodeInjectEndRnNR(params: {  test code inject * &#39; &quot; &#x3D;end  rn n r?: string; }, options?: any) {
```
Vrolijkx: This line confuse me. Can Typescript compile this? The added options part seems ok but does somebody know what is expected here as params?

alechill: @Vrolijkx nope it doesn't compile - it appears to be a more general problem whereby the security tests YAML is specified without string delimiters and has spaces, which isn't being handled by the template. 

Looks like these params need be string delimited in the signature - which seems horrible style-wise when used in this typescript interface context, but actually is perfectly valid as its defining the interface for some JSON that can absolutely have spaces in a string key

To compile it should become... 
```testCodeInjectEndRnNR(params: {  "test code inject * &#39; &quot; &#x3D;end  rn n r"?: string; }, options?: any) {
            return FakeApiFp.testCodeInjectEndRnNR(params, options)(fetch, basePath);
        }```

It is a strange context for the problem to have been surfaced - I can't see why I as an API developer would introduce malicious injection code as an expected parameter key?! As it stands it wouldn't make the API vulnerable, as it wouldn't even compile in the first place! However it would be perfectly reasonable to declare to expect a JSON param with spaces as key ie. `{"Full name": "Nic Cage" }` so should probably be addressed anyway

I reckon this should be raised and looked at in it's own ticket, as probably merits a little more discussion - and its also broken in the other typescript modules

Comment (alechill) [inline samples/client/petstore-security-test/typescript-fetch/api.ts]: @Vrolijkx (cc @wing328 ) nope it doesn't compile - it appears to be a more general problem whereby the security tests YAML is specified without string delimiters and has spaces, which isn't being handled by the template.

Looks like these params need be string delimited in the signature - which seems horrible style-wise when used in this typescript interface context, but actually is perfectly valid as its defining the interface for some JSON that can absolutely have spaces in a string key

To compile it should become... 
testCodeInjectEndRnNR(params: { "test code inject * &#39; &quot; &#x3D;end rn n r"?: string; }, options?: any) { return FakeApiFp.testCodeInjectEndRnNR(params, options)(fetch, basePath); }

It is a strange context for the problem to have been surfaced - I can't see why I as an API developer would introduce malicious injection code as an expected parameter key?! As it stands it wouldn't make the API vulnerable, as it wouldn't even compile in the first place! However it would be perfectly reasonable to declare to expect a JSON param with spaces as key ie. {"Full name": "Nic Cage" } so should probably be addressed anyway

I reckon this should be raised and looked at in it's own ticket, as probably merits a little more discussion - and its also broken in the other typescript modules

**Comment by JLLeitschuh** [comment]
Perhaps adding a base set of parameters that get merged with ones specific to the method call?

That way you can reuse the same `DefaultApi` object that has parameters set like the `credentials: 'include'` flag and you don't have to supply that option on every single `DefaultApi` call?

**Comment by alechill** [comment]
@JLLeitschuh I see what you are getting at, but I think that would actually introduce another level of options that aren't necessary to gain functionality, those that you want applied to all requests, and those that you want to introduce only to specific calls. The options added in this PR will therefore still need to be available in order to do request specific options, or to override any defaults that you provide.

I think at this point it starts getting a bit option inception and confusing! The point being that without the above PR the options cannot physically be changed at all (as under the surface the OO API uses the FP API which still needs them passed in this same way), whereas this suggestion would be an extra nicety on top so save some keystrokes, welcome but not necessary

You could always provide defaults in your application by wrapping the API in a service, something like...

```
class MyApiService {

   private defaultOpts: any
   private whateverApi: WhateverApi

   constructor(fetch?: FetchAPI, basePath?: string, defaultOpts: any = {}) {
        this.whateverApi = new WhateverApi(fetch, basePath)
        this.defaultOpts = defaultOpts
   }
    
   private provideOpts = (opts: any = {}): any => Object.assign({}, this.defaultOpts, opts)

   getWhatever(requestOpts?: any): Promise<Whatever> {
       return this.whateverApi.getWhatever(this.provideOpts(requestOpts))
   }

}
```

JLLeitschuh:
Is there any way to to this generically in typescript that wraps all of the method calls or do you have to wrap each one independently?

The best way to solve the "default opts" problem now as I see it and with your change is to hard code your default value in the generator (I'm using a local copy of the `mustache` files with some other fixes already).

Perhaps another service class could be added to the default generator `mustache` file that adds the functionality you propose.

wing328:

PR #4049, opened 2016-10-21
[dart] updated intl library version

ircecho:
The version of the "intl" library is now more than a year old and is starting to conflict with other libraries' dependencies, in our case with angular 2 for dart. Only date parsing, i.e. a very limited subset of "intl" features, is used in the generated libraries. We find it unlikely, that this feature would be changed in a major way in future, and therefore and to avoid this problem in future, we find it justified to set the version constraint to "any".

### DIFF
--- a/modules/swagger-codegen/src/main/resources/dart/pubspec.mustache
+++ b/modules/swagger-codegen/src/main/resources/dart/pubspec.mustache
@@ -4,7 +4,7 @@ description: {{pubDescription}}
 dependencies:
   http: '>=0.11.1 <0.12.0'
   dartson: "^0.2.4"
-  intl: "^0.12.4+2"
+  intl: any
 
 dev_dependencies:
   guinness: '^0.1.17'


modules/swagger-codegen/src/main/resources/dart/pubspec.mustache
```diff
@@ -4,7 +4,7 @@ description: {{pubDescription}}
 dependencies:
   http: '>=0.11.1 <0.12.0'
   dartson: "^0.2.4"
-  intl: "^0.12.4+2"
+  intl: any
```
wing328: What about using `">=0.12.4"` instead of `any` to ensure a minimal version is installed?

wing328: @ircecho are you ok with my suggestion or you would insist on using `any`?

ircecho: That would work just as well, so do as you think!

However, the choice of that version seems quite arbitrary. I am quite sure the features that are used have been present in much older versions of the library as well.

wing328: @ircecho if the users want to use an older version of the lib, they can always use the customized template via the `-t` option.

wing328:
Closed via #4111

PR #4050, opened 2016-10-21
Fix sanitizeTag to retain numbers and underscore in tag names

### LINKED ISSUE #2657: Generator - numbers getting stripped from tag names
I have a swagger path tagged with "v1-apiname", and would like to generate an api (with api.mustache) that is named v1_apiname.rb (according to my settings).

My first thought was to override `toApiFilename`, but it seems that a function higher in the call chain, `apiFilename` (DefaultCodegen.java line 2404), already gets "Vapiname" as the tag parameter using the above example, meaning that I am unable to keep the numbers and do my own conversion.

Since the functions deeper in the stack seem problematic to override, I wonder if it's possible to get the name by some other method or delegate the stripping of numerals and dashes to an overridable function.

Thanks :)

mbohlool:
### PR checklist

**I've run bin/run-all-petstore but there are lots of unrelated changes, and in a glance I could not see any change related to this PR. I am just guessing petstore samples are out of date. Let me know if you want me to add those (huge) changes.**

### Description of the PR

I've changed the pattern sanitizeTag uses to remove invalid characters and also added underscore to tags starting with a number. Also added some test cases.

fixes #2657

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -3171,7 +3171,12 @@ public class DefaultCodegen {
                 buf.append(StringUtils.capitalize(part));
             }
         }
-        return buf.toString().replaceAll("[^a-zA-Z ]", "");
+        String returnTag = buf.toString().replaceAll("[^a-zA-Z0-9_]", "");
+        if (returnTag.matches("\\d.*")) {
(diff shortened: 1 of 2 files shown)


wing328:
@mbohlool thanks for the contribution and adding various tests to cover the change.

mbohlool:
Happy to contribute. How can I get this change into my maven workflow? (sorry not a maven expert), I have plugin version "<version>2.2.2-SNAPSHOT</version>" but it doesn't work.

PR #4057, opened 2016-10-24
Added support for string responses

plankswert:
### PR checklist
### Description of the PR

When a method/URL/response is defined to return string:
- If no content types are define, default to 'text/plain' instead of
  'application/json'
- Add response handler, that returns the reponse string as-is if response
  content-type is 'text/plain'


plankswert:
Associated with Issue-4058

wing328:
Linked to #4058

modules/swagger-codegen/src/main/resources/cpprest/api-source.mustache
```diff
@@ -53,7 +53,22 @@ pplx::task<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}void{{/r
     utility::string_t responseHttpContentType;
     
     // use JSON if possible
-    if ( responseHttpContentTypes.size() == 0 || responseHttpContentTypes.find(U("application/json")) != responseHttpContentTypes.end() )
+    if ( responseHttpContentTypes.size() == 0 )
+    {
+        {{#vendorExtensions.x-codegen-response.isPrimitiveType}}
```
wing328: Instead of using {{#vendorExtensions.x-codegen-response.isPrimitiveType}, what about using `{{#primitiveType}}` in the response object instead?

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenResponse.java#L15

modules/swagger-codegen/src/main/resources/cpprest/api-source.mustache
```diff
@@ -53,7 +53,22 @@ pplx::task<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}void{{/r
...
+        responseHttpContentType = U("application/json");
+        {{/vendorExtensions.x-codegen-response.isPrimitiveType}}
+        {{^vendorExtensions.x-codegen-response.isPrimitiveType}}
+        {{#vendorExtensions.x-codegen-response.isString}}
```
wing328: What about adding `isString` to CodegenResponse, similar to [isBinary](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenResponse.java#L18)?

modules/swagger-codegen/src/main/resources/cpprest/api-source.mustache
plankswert: Looks better. I will try it.

wing328: Btw, the mustache template variables can be obtained via debug flags: https://github.com/swagger-api/swagger-codegen/wiki/Mustache-Template-Variables

plankswert: Thanks for the link. I am reviewing this today.

modules/swagger-codegen/src/main/resources/cpprest/api-source.mustache
plankswert: Adding isString looks considerably more complex, but doable. The question that comes to mind is if this has not been needed until now, is it the right feature to add the the CodegenResponse for a single language generator?

plankswert: It looks like the isString would not be similar to isBinary. It is closer to isResponseBinary under 
https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java#L17

modules/swagger-codegen/src/main/resources/cpprest/api-source.mustache
plankswert: Since the original code was not iterating over the responses array of nodes for picking the preferred "Accept" content type, primativeType is not available without rethinking the approach to the code generation.

NOTE: I am still ramping on mustache.

PR #4062, opened 2016-10-24
Fixed toUpper and toLower usage in haskell-servant gen.

AndrewRademacher:
### PR checklist
### Description of the PR

This PR addresses an issue with the `haskell-servant` generator. The generator didn't account for the possibility of single character identifiers when applying `toUpper` and `toLower` casing rules. Calls to `word.substring(1)` would crash as a result.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/HaskellServantCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/HaskellServantCodegen.java
@@ -158,6 +158,26 @@ public class HaskellServantCodegen extends DefaultCodegen implements CodegenConf
         return name + "_";
     }
 
+    public String firstLetterToUpper(String word) {
+        if (word.length() == 0) {
+            return word;
+        } else if (word.length() == 1) {
+            return word.substring(0, 1).toUpperCase();
+        } else {
+            return word.substring(0, 1).toUpperCase() + word.substring(1);
+        }
+    }
+
+    public String firstLetterToLower(String word) {
+        if (word.length() == 0) {
+            return word;
+        } else if (word.length() == 1) {
+            return word.substring(0, 1).toLowerCase();
+        } else {
+            return word.substring(0, 1).toLowerCase() + word.substring(1);
+        }
+    }
+
     @Override
     public void preprocessSwagger(Swagger swagger) {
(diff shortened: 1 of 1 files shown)


wing328:
@AndrewRademacher thanks for the PR. What about using the `camelize` function instead (https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L2908) ?

AndrewRademacher:
I've just tested it with camelize, and the camelize function seems to suffer the same behavior. For example lines 2950-2952.

```java
if (lowercaseFirstLetter) {
    word = word.substring(0, 1).toLowerCase() + word.substring(1);
}
```

This function also assumes that the string is at least 2 characters.

wing328:
@AndrewRademacher do you mind giving 1 or 2 examples showing the difference so as to illustrate the fix?

AndrewRademacher:
Sure. For clarity I want to compare the following two snippets.

```java
if (word.length() == 0) {
    return word;
} else if (word.length() == 1) {
    return word.substring(0, 1).toUpperCase();
} else {
    return word.substring(0, 1).toUpperCase() + word.substring(1);
}
```

and

```java
newword = word.substring(0, 1).toLowerCase() + word.substring(1);
```

In the second case if the word supplied is "x", you would expect to receive "X" in return. However, word.substring(1), on a string of length 1 causes the following exception.

```
Exception in thread "main" java.lang.StringIndexOutOfBoundsException: String index out of range: 1
        at java.lang.String.substring(String.java:1907)
        at io.swagger.codegen.languages.HaskellServantCodegen.firstLetterToUpper(HaskellServantCodegen.java:167)
        at io.swagger.codegen.languages.HaskellServantCodegen.preprocessSwagger(HaskellServantCodegen.java:208)
        at io.swagger.codegen.DefaultGenerator.generate(DefaultGenerator.java:139)
        at io.swagger.codegen.cmd.Generate.run(Generate.java:223)
        at io.swagger.codegen.SwaggerCodegen.main(SwaggerCodegen.java:41)
```

This is because a string of length 1 has no index 1, it only has an index 0. The first code sample show a fix where you check for the length of a string to decide if using word.substring(1) is possible.

wing328:
@AndrewRademacher thanks for the explanation. PR merged into master.

PR #4063, opened 2016-10-24
ISSUE-1388 - Do not NPE when array properties items are not specified

nddipiazza:
### Description of the PR

Do not NPE on an array type with no "items". Just return null.


wing328:

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java
```diff
@@ -422,6 +422,9 @@ public String getTypeDeclaration(Property p) {
         if (p instanceof ArrayProperty) {
             ArrayProperty ap = (ArrayProperty) p;
             Property inner = ap.getItems();
+            if (inner == null) {
+              return null;
```
wing328: Instead of returning `null`, what about defaulting to `string` and log (LOGGER.WARN) a warning message so that the user is aware of the missing inner type)?

wing328: @nddipiazza any feedback on my comment above?

Issue #4066, opened 2016-10-25
Uppercase property names lead to java code that produces wrong json.


schnabel:
##### Description
If the swagger declaration contains properties with names containing only uppercase letters and underscores, the generated java code is not java bean conform. As a result jackson maps the properties twice. This actually is part of [#4051](https://github.com/swagger-api/swagger-codegen/issues/4051) but the code currently handles this case earlier.

##### Swagger-codegen version
Tested with 2.2.1 and 2.2.2-SNAPSHOT

##### Swagger declaration file content or url
```yaml
swagger: '2.0'
schemes:
  - http
basePath: '/v0.1/foo'
consumes:
  - 'application/json'
produces:
  - 'application/json'
paths:            
  '/foo':
    get:
      responses:
        200:
          description: "successful operation"
          schema:
            type: "array"
            items:
              $ref: "#/definitions/Pet"

definitions:

  Pet:
    description: >
      Pet
    type: object
    properties:
      ATT_NAME:
        description: >
          Name of the pet
        type: string
```

##### Command line used for generation
swagger-codegen-maven-plugin 

##### Steps to reproduce
Build a jaxrs or spring server. Implement controller to return Pet array.
Or use jackson mapper and serialize with the generated Pet model code.

##### Related issues
https://github.com/swagger-api/swagger-codegen/issues/4051

##### Suggest a Fix
```java
    @Override
    public String toVarName(String name) {
        // sanitize name
        name = sanitizeName(name); // FIXME: a parameter should not be assigned. Also declare the methods parameters as 'final'.

        if ("class".equals(name.toLowerCase())) {
            return "propertyClass";
        }

        if("_".equals(name)) {
          name = "_u";
        }

        // if it's all uppper case, do nothing
//        if (name.matches("^[A-Z_]*$")) {
//            return name;
//        }

        // camelize (lower first character) the variable name
        // pet_id => petId
        name = camelize(name, true);

        // for reserved word or word starting with number, append _
        if (isReservedWord(name) || name.matches("^\\d.*")) {
            name = escapeReservedWord(name);
        }

        return name;
    }
```

schnabel:
The fix depends on the fix suggested in [#4051](https://github.com/swagger-api/swagger-codegen/issues/4051) so it is actually:
```java
    public String toVarName(String name) {
        // sanitize name
        name = sanitizeName(name); // FIXME: a parameter should not be assigned. Also declare the methods parameters as 'final'.

[...]

PR #4073, opened 2016-10-25
bugfix: issue-4051

schnabel:
### PR checklist
### Description of the PR

The fix converts the first two letters of a property name to lower case if the property name started with two upper case letters.
Another way to handle the problem jackson has with uppercase properties would be to put the JsonProperty annotation to the getter methods.


wing328:
@schnabel thanks for the PR, which has been merged into master.

schnabel:
@wing328 Hi, I will provide a fix for #4066 now.

PR #4074, opened 2016-10-26
Jdk6 jersey2 support

scannerscan:
### PR checklist
### Description of the PR

supportJava6 flag was added to the server side model classes generators,
licenseInfo ttemplate was added to JavaJaxRS section,
javaJaxRS templates jdk6 suport was added,
shouldOverride method was commented out to allow server classes to be overriden - what if new methods or signatures changes?


scannerscan:
@wing328 could you please review?

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaJAXRSServerCodegen.java
```diff
@@ -201,9 +201,10 @@ private String implFileFolder(String output) {
         return outputFolder + "/" + output + "/" + apiPackage().replace('.', '/');
     }
 
-    @Override
-    public boolean shouldOverwrite(String filename) {
-        return super.shouldOverwrite(filename) && !filename.endsWith("ServiceImpl.java") && !filename.endsWith("ServiceFactory.java");
-    }
+// TODO: @scannerscan: I think we need to allow overriding because the number of methods and signature can be changed.
```
wing328: We want to move away from `shouldOverwrite` to .swagger-codege-ignore. Would `.swagger-codege-ignore` better meet your requirement?

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaJAXRSSpecServerCodegen.java
```diff
@@ -24,7 +24,7 @@
 	public JavaJAXRSSpecServerCodegen()
 	{
         super();
-        sourceFolder = "src/main/java";
+      //  sourceFolder = "src/main/java";
```
wing328: I think we can simply remove this line (instead of just commenting it out)

scannerscan: Ok, sounds good.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaJAXRSServerCodegen.java
wing328: https://github.com/swagger-api/swagger-codegen/wiki/FAQ#how-to-skip-certain-files-during-code-generation

scannerscan: I like that approach

scannerscan:
@wing328 please review

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaJAXRSServerCodegen.java
wing328: Shall we keep `shouldOverwrite` for the time being before we move to .swagger-codegen-ignore?

scannerscan:
hi @wing328, did you get my response regarding shouldOverride ? the thing is I have commented out the part that belongs to the server side generation only because it has not been working properly. Basically if any Impl related file exists it just skipped regeneration and did not pay any attention to the flag itself. the method in the base class still there.

scannerscan:
if we have this method in the other class. that is why i decided to
commented out it here because the method is overriding the original method
and does not allow us to override server impl classes

wing328:
@scannerscan thanks for the PR, which has been merged into master.

scannerscan:
Thanks @wing328

willgorman:
Removing io.swagger.codegen.languages.AbstractJavaJAXRSServerCodegen#shouldOverwrite doesn't seem right to me.  The *ServiceImpl files are always going to be modified with user-written code after they are generated.  We use the maven plugin to regenerate code on every build and this immediately clobbers our implementation.  Even if it's possible to avoid this with `.swagger-codegen-ignore` it seems reasonable to me that those files should never be overwritten by the generator.

scannerscan:
Your point is right if you are not actively changing your api. I mean
number of methods in the services and params for them.

Here is what works for us.  2 things. 1. Create a class delegate for every
server site generated service implementation (that is actual implementation
and should not be touched by the code gen). 2. create your own template for
the server side impl where the delegate will be specified.

willgorman:
The *ServiceImpl classes are already delegates of the *Api classes ([example](https://github.com/swagger-api/swagger-codegen/blob/master/samples/server/petstore/jaxrs/jersey2/src/gen/java/io/swagger/api/PetApi.java#L33)) so adding another layer of delegation seems unnecessary.  I'm also not sure how it solves the problem when adding methods or changing parameters.  Even if the *ServiceImpl classes get regenerated you still have to manually make the changes to your own delegate so why not just make those same changes in the *ServiceImpl class instead?  Those classes seem to me as clearly intended to have user-written code added to them as indicated by the `// do some magic!` comments.

willgorman:
Logged #4230 to discuss the removal of AbstractJavaJAXRSServerCodegen#shouldOverwrite

Issue #3829, opened 2016-09-19
discriminator generated in both base and child


AlexNolasco:
##### Description

Should the discriminator property be also a property of the child class even if it is already in the base class?
##### Swagger-codegen version

2.2
##### Swagger declaration file content or url

[ticket.yaml.txt](https://github.com/swagger-api/swagger-codegen/files/481027/ticket.yaml.txt)

##### Command line used for generation

Objective-C   and C# client generators generate the petType in the base and child class. Thus generating compilation warnings. OTOH, the Java generator does not duplicate the property.
##### Steps to reproduce

Simply generate C# or Objective-C client code for the YAML aforementioned
##### Suggest a Fix

Child classes should not duplicate properties found in the base class. The work around is to only define the discriminator in the base class and then copy/paste the property on each child class.

wing328:
@alexandern am I understanding correctly that that the Java client does not have the issue and only ObjC, Swift and C# client has the issue?

alexandern:
@wing328 

> am I understanding correctly that that the Java client does not have the issue and only ObjC, Swift and C# client has the issue?

Correct,

I've attached the Java client generated code for the yaml in question which does what you will expect. Base properties are _not_ copied to the child class.
 
[Dog.java.txt](https://github.com/swagger-api/swagger-codegen/files/485164/Dog.java.txt)
[Pet.java.txt](https://github.com/swagger-api/swagger-codegen/files/485165/Pet.java.txt)

On the other hand, the Objective-C does duplicate the properties
[SWGPet.h.txt](https://github.com/swagger-api/swagger-codegen/files/485182/SWGPet.h.txt)
[SWGCat.h.txt](https://github.com/swagger-api/swagger-codegen/files/485183/SWGCat.h.txt)

Possible workaround would be to mark the child properties as dynamic to avoid the compilation warnings produced by the current output. Not having the parent properties would probably be best.

The C# client generation also copies the properties found in the parent class
[Pet.cs.txt](https://github.com/swagger-api/swagger-codegen/files/485200/Pet.cs.txt)
[Dog.cs.txt](https://github.com/swagger-api/swagger-codegen/files/485201/Dog.cs.txt)

alexandern:
@wing328 

I've tested Swift Client and it does not use inherit the Pet class,  it  uses composition instead. So, it should not be part of this issue.

cbornet:
[...]

PR #4080, opened 2016-10-27
[csharp] Don't duplicate parent properties in derived classes

jimschubert:
### PR checklist
### Description of the PR

referring to #3829, this sets parent properties only on base types in the C# client generator.  It looks like there's a similar solution in the NancyFX generator, but it didn't port correctly to `AbstractCSharpCodegen`.

I'd like to move logic from this PR into `AbstractCSharpCodegen` to be shared across generators, but I'd need to look more into how NancyFX is doing its inheritance (because it doesn't have `supportsInheritance = true` set).

One of my main concerns is that the code block I have commented "Cleanup possible duplicates." may or may not indicate a bug elsewhere (wherever `readWriteVars` is used). Without this cleanup block, when `supportsInheritance = true` and I debug `./bin/csharp-petstore.sh`, I get `readWriteVars` containing duplicates for all parent properties .  My assumption is that the `name` property of these codegen properties isn't being compared in a case-insensitive way, or post-processing of property names isn't being done before the sorting. I spent some time trying to track this down and couldn't find it. 

I was wondering if someone could:
- verify this code generates the correct C# code?
- remove the "Cleanup possible duplicates" block and `supportsInheritance = true;` to see if readWriteVars does indeed include duplicates?

I plan to take another look, but it may be a week or so before I get around to it.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -2848,6 +2848,7 @@ public class DefaultCodegen {
                 if (Boolean.TRUE.equals(cp.isReadOnly)) {
                     m.readOnlyVars.add(cp);
                 } else { // else add to readWriteVars (list of properties)
+                    // FIXME: readWriteVars can contain duplicated properties. Debug/breakpoint here while running C# generator (Dog and Cat models)
                     m.readWriteVars.add(cp);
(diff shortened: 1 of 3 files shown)


manuc66:
@jimschubert I leverage your work to add subtype inheritance support in https://github.com/manuc66/swagger-codegen/tree/feature/csharp-subtypes could you have a look ?

wing328:
Closed via #5922

Issue #3908, opened 2016-09-30
[JAVA] CodegenParameter instance copy is not safe


RaphC:
Hi,
##### Description

I'm using swagger codegen maven plugin to generate from swagger file our Java class with my own generator. I extended JavaJAXRSSpecServerCodegen to add some feature like ignoring some of parameters in the methods which are described in the swagger file. To do that, I set a vendor extensions and then when the plugin invokes fromOperation It ignores the parameters with these extension. In my code I have to make a copy of the CodegenParameter.

List<CodegenParameter> printableParams = Lists.newLinkedList();
        for (CodegenParameter p : op.allParams) {
            ....
                printableParams.add(p.copy());
    }

Unfortunately when copying CodegenParameter instance, the attribute vendorExtensions is copied by reference.

output.vendorExtensions = this.vendorExtensions;

So If the current instance p modify its  vendorExtensions Map (add a new element for instance), my new cloned instance is affected by whereas it shouldn't be.
##### Swagger-codegen version

2.2.1
##### Suggest a Fix

Just make a copy of attribute vendorExtensions in the method copy() of CodegenParameter instead of  copying reference

wing328:
@RaphC thanks for the suggestion. May I know if you've time to contribute the enhancement?

RaphC:
Hi,

No problem, I can provide an enhancement.

Regards,

RaphC:
Hi @wing328 
I've just created the following PR : https://github.com/swagger-api/swagger-codegen/pull/4093

PR #4093, opened 2016-10-27
#3908 make a copy of vendorExtensions map instead of copying the reference

RaphC:
### PR checklist
### Description of the PR

In the method copy of the class CodegenParameter, we make a copy of vendorExtensions map instead of copying the reference to prevent unwanted instance modification of a copied instance.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenParameter.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenParameter.java
@@ -123,7 +123,7 @@ public class CodegenParameter {
             output.items = this.items;
         }
         if(this.vendorExtensions != null){
-        	output.vendorExtensions = new HashMap<String, Object>(this.vendorExtensions);
+            output.vendorExtensions = new HashMap<String, Object>(this.vendorExtensions);
         }
         output.hasValidation = this.hasValidation;
         output.isBinary = this.isBinary;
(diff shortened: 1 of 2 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenParameter.java
```diff
@@ -122,7 +122,9 @@ public CodegenParameter copy() {
         if (this.items != null) {
             output.items = this.items;
         }
-        output.vendorExtensions = this.vendorExtensions;
+        if(this.vendorExtensions != null){
+        	output.vendorExtensions = new HashMap<String, Object>(this.vendorExtensions);
```
wing328: Please use 4-space instead of tab.

Also what about `vendorExtensions` in other objects (e.g. CodegenProperty) ?

wing328: @RaphC have you committed/pushed the change? I only see 1 commit in this PR.

RaphC: @wing328  Not yet. I'm waiting for your feedback about CodegenProperty

wing328: > We should rewrite the clone method to be prevent the copy by reference.

Yes, please help contribute the enhancement.

RaphC: Right. I'll make a PR ASAP.

PR #4099, opened 2016-10-29
Newline escaping disabled for HTML and HTML2 generators

bazhenov:
### PR checklist
### Description of the PR

Newline escaping disabled for HTML and HTML2 generators. Escaping failing Markdown processors to work correctly on descriptions.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/StaticHtml2Generator.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/StaticHtml2Generator.java
@@ -66,6 +66,12 @@ public class StaticHtml2Generator extends DefaultCodegen implements CodegenConfi
     }
 
     @Override
+    public String escapeText(String input) {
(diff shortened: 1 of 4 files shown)


wing328:
Your suggestion sounds reasonable. I'll review your PR shortly. Thanks for your contribution.

GlenDC:
Does this issue fixes the bug where newline characters (`\n`) in a string value (e.g. `info.description`) get ignored, instead of forcing an HTML linefeed?

e.g. I would expect `foo\nbar` to be translated to:
```html
foo<br>bar
```
or however that should be done in the generated (static) HTML. But for now it just seems to render `foo\nbar` 

Or should I open a new issue for this and, later, open a PR with a fix for it?

bazhenov:
Unfortunately, it doesn't, as far as I can tell.

GlenDC:
So I should open a new issue for this?

bazhenov:
Definitely

PR #4101, opened 2016-10-31
[python] Add packageUrl option, replacing infoUrl in setup.py

cliffano:
### PR checklist

- [v] Read the [contribution guildelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [v] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [v] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

Replace infoUrl with packageUrl in Python's setup.py .
infoUrl is the contact URL, which is more suitable for an author URL, however the url field in setup.py is for the package.
This PR introduces packageUrl option and used in setup.py mustache template.


wing328:
@cliffano the change looks good to me. Thanks for the contribution.

Issue #2848, opened 2016-05-12
java/jersey2 


jerrylawson:
When running swagger-codegen with -l java and library jersey2, the generated ApiClient.java does not allow for "PATCH" in invokeAPI(). PATCH support should be added to the ApiClient.mustache: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache

wing328:
@jerrylawson thanks for reporting the issue. Would you have cycle to contribute the enhancement?

Ref: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

For the time being, please generate Java API client with `okthtp-gson` library instead, which should support PATCH operation.

jerrylawson:
Yes, I can contribute this change

wing328:
@jerrylawson do you need any help with the change? Let us know.

sutdisi:
I have a fix for this. I am assuming no one else is working on it.

sutdisi:
#4110 should fix this.

PR #4110, opened 2016-11-02
Jersey2 Patch Support

sutdisi:
### Description of the PR
Adding PATCH support for Jersey2 java client. This ports over PR-2128 to Jersey2. Original fix only applied to Jersey client.

Fixes issue #2848


wing328:
@sutdisi thanks for the PR, which has been merged into master.

Moving forward, I would suggest you to create a new branch for the change.

sutdisi:
@wing328 thanks will do so in the future.

Is there a suggested repo with jars of nightly builds (are there nightly builds)? Or should I wait for a release to consume this change?

wing328:
We didn't do nightly builds We'd a discussion on that before but decided not to do it.

Please wait for the upcoming stable release (2.2.2) or build it locally by `mvn clean install`, which still install 2.2.2 snapshot into your local maven repo.

Issue #4077, opened 2016-10-26
Code generation for spring-mvc failed: Index out of bounds


hensb:
##### Description

Code generation failed for spring:

code: 1
type:  "error"
message:  "Unable to build target: String index out of range: 1"
##### Swagger-codegen version

2.2.1

##### Swagger declaration file content or url

```
---
swagger: "2.0"
info:
  version: "0.1.0"
  title: ""
basePath: "/foo-auth/rest/unicorns/riders"
tags:
- name: "unicorns/riders"
schemes:
- "http"
paths:
  #version
  /version:
    get:
      tags:
      - "unicorns/riders"
      description: Get version
      responses:
        200:
          description: Successful response
          # A schema describing your response object.
          # Use JSON Schema format
          schema:
            title: Version
            type: object
            items:
              $ref: '#/definitions/Version'
        default:
          description: Unexpected error
          schema:
            $ref: '#/definitions/Error'

  /contents:
    get:
      tags:
      - "unicorns/riders"
      summary: "unicorn Request returning the unicorn Decision Content of a specific User."
      description: "blergh"
      operationId: "getunicornsridersContents"
      produces:
      - "application/json"
      parameters:
      - name: "UserID"
        in: "query"
        description: "blergh"
        required: true
        type: "string"
      - name: "CupCakeID"
        in: "query"
        description: "blergh"
        required: true
        type: "integer"
        format: "int32"
      - name: "fooID"
        in: "query"
        description: "blergh"
        required: true
        type: "integer"
        format: "int32"
      - name: "barID"
        in: "query"
        description: "blergh"
        required: false
        type: "integer"
        format: "int32"
      - name: "RequestContextInfo"
        in: "query"
        description: "blergh"
        required: false
        type: "array"
        collectionFormat: multi
        items: 
          type: "string"
      responses:
        200:
          description: "blergh"
          # A schema describing your response object.
          # Use JSON Schema format
          schema:
            title: AuthorizedContent
            type: "object"
            items:
              $ref: '#/definitions/Content'
        400:
          description: "blergh"
        404:
          description: "blergh"
        default:
          description: Unexpected error
          schema:
            $ref: '#/definitions/Error'

  /Thingys:
    get:
      tags:
      - "unicorns/riders"
      summary: "unicorn Request returning the unicorn Decision Thingys of a specific User."
      description: "blergh"
      operationId: "getunicornsridersThingys"
      produces:
      - "application/json"
      parameters:
[...]

PR #4115, opened 2016-11-02
Added String.length() > 0 check to avoid IndexOutOfBoundException

wwn:
see issue #4077: An IndexOutOfBoundExeception is thrown during comiling yaml to spring for the element title: ""


wing328:
@wwn thanks for the PR, which has been merged into master.

PR #4125, opened 2016-11-04
DART: multi-word class names work properly now, fixes #4117

### LINKED ISSUE #4117: DART: wrong filename used when importing api file
The dart generator creates api files with underscored names, but uses non-underscored names to import them which leads to problems if the file names multi-word.

JSON example: https://gist.github.com/ircecho/32f4836c38f9a316dd39630d1b9b2833

Here the file name generated is ([from here](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/DartClientCodegen.java#L229-L231)): 

> .../lib/api/two_words_api.dart

The filename used to import the file is ([from here](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/dart/apilib.mustache#L19-L19)):
```dart
part 'api/twoWords_api.dart';
```

two_words_api.dart is the correct filename and should be used in both instances.

ircecho:
Fixes #4117. I'm not sure if this is a proper fix. I tested it for dart only.


wing328:
@ircecho thanks. I'll review and let you know (I seem to recall other code generators, eg. ruby, python, also need access to the class filename)

wing328:
@ircecho looks good to me. Thanks for your contribution.

Issue #4129, opened 2016-11-04
[ObjC] Generated code is not actual with JSONModel library


Dreddik:
##### Description

Generated objective-c code uses old JSONModel library method for init, which is deprecated. 
```
SwaggerClient/Model/SWGSearchModel.m:20:33: 'initWithDictionary:' is deprecated: use initWithModelToJSONDictionary:
```

##### Swagger-codegen version

2.2.1

##### Swagger declaration file content or url
```yaml
swagger: '2.0'

info:
  version: "1"
  title: "Hello warning"

basePath : "/api/v1"
produces:
  - application/json
schemes:
  - https
        
paths:
  /categories:
    get:
      description: get categories
      responses:
        200:
          description: Successful response
          schema:
            type: array
            items:
              $ref: '#/definitions/UserSkill'
           
definitions:
  Category:
    description: Some category for user
    type: object
    properties:
      id:
        type: integer
        format: int64
      alias:
        type: string
      localized_name:
        type: string
  UserSkill:
    description: User skill description based on category
    allOf:
      - $ref: '#/definitions/Category'
      - properties: 
          year_started:
            description: Year since user has that skill
            type: integer
```

##### Command line used for generation

`swagger-codegen generate -i ./swagger.yaml -l objc -o ./SwaggerClient`

##### Steps to reproduce

1) Generate code
2) Create a test project, create Podfile with contents:
```
use_frameworks!
platform :ios, '9.0'

target 'TestSwagger' do
    pod 'SwaggerClient', :path => './SwaggerClient'
end
```
3) Run `pod install`
4) Build project

##### Suggest a Fix

Use a suggested method initWithModelToJSONDictionary

wing328:
@Dreddik thanks for reporting. May I know if you've time to contribute a fix? A good starting point is https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/objc/model-body.mustache

cc @mateuszmackowiak

wing328:
@Dreddik PR merged. Thanks for the contribution.

PR #4133, opened 2016-11-04
[ObjC] Fix deprecated JSONModel API usage

Dreddik:
### PR checklist

### Description of the PR
Replaced deprecated call `-[JSONKeyMapper initWithDictionary:]` with `-[JSONKeyMapper initWithModelToJSONDictionary:]`, inverting keys and values of dictionary as required.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/objc/model-body.mustache
+++ b/modules/swagger-codegen/src/main/resources/objc/model-body.mustache
@@ -43,7 +43,7 @@
  * This method is used by `JSONModel`.
  */
 + (JSONKeyMapper *)keyMapper {
-  return [[JSONKeyMapper alloc] initWithDictionary:@{ {{#vars}}@"{{baseName}}": @"{{name}}"{{#hasMore}}, {{/hasMore}}{{/vars}} }];
+  return [[JSONKeyMapper alloc] initWithModelToJSONDictionary:@{ {{#vars}}@"{{name}}": @"{{baseName}}"{{#hasMore}}, {{/hasMore}}{{/vars}} }];
 }
 
 /**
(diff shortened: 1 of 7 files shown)


Dreddik:
Fix for #4129

wing328:
@Dreddik thanks for the PR. Do we also need to update `jsonmodel` to the latest stable version (1.7) in the podspec: https://github.com/Dreddik/swagger-codegen/blob/9510ba5462317734b8237754a73e871dfba506af/modules/swagger-codegen/src/main/resources/objc/podspec.mustache#L34 ?

Btw, please run `objc-petstore-all.sh` to update ObjC Petstore samples (CoreData) as well.

cc @mateuszmackowiak

Dreddik:
@wing328 
> Do we also need to update jsonmodel to the latest stable version (1.7) 

Actually, API changed in v1.4, so we can use it, I suppose? In any way, clients will use version 1.7, because minor version is not specified in `podspec`.
> Btw, please run objc-petstore-all.sh to update ObjC Petstore samples (CoreData) as well.

Already done.

wing328:
> Actually, API changed in v1.4, so we can use it, I suppose? In any way, clients will use version 1.7, because minor version is not specified in podspec.

Yes, please update it to v1.4 as that is the minimal version required for the auto-generated ObjC SDK after this change.

Also to confirm my understanding, I believe developers can still do the following after the change:
```
    pet = [[SWGPet alloc] initWithDictionary:petDict error:nil];
```
and they will receive a warning saying that `initWithDictionary` has been deprecated. In other words, this PR (change) is backward compatible. Please kindly correct me if I misunderstood.

Dreddik:
>Also to confirm my understanding, I believe developers can still do the following after the change...

This PR is backward compatible, because it is only affects initialization of JSONKeyMapper inside of model classes. 
`[[SWGPet alloc] initWithDictionary:petDict error:nil]` won't be affected (and actually not deprecated).

wing328:
@Dreddik PR merged into master. Thanks for your contribution.

Issue #2885, opened 2016-05-16
C# Why does it make properties nullable?


Serjster:
I noticed that the autogen automatically changes many properties to nullable type! I assume that this is to allow the constructor to default to null to nonsupplied params...unfortunately this breaks my deal. Is there anyway to prevent the constructor generation as well as prevent the making of those non-nullable properties on source to be nullable?

kbrooks:
Swagger has a "required" property for parameters. If this is not set to true, the codegen will make the corresponding parameters nullable.

Serjster:
So you're saying I have to decorate my properties on my server with
required, even for things like public long ID? That doesn't make sense! Why
is it changing the type at all? Leave it as is in the source! Is there no
other way?

wing328:
@Serjster can you elaborate a bit more on the issue cased by nullable type?

kbrooks:
What does your swagger.json look like?

Serjster:
Sure. I've written a script that adds an IType (or whatever Type happens to
be)to each of the genned classes. The problem is that because those
properties are nullable, the compiler throws an error because "interface is
not implemented". So, instead of being able to use the genned class as the
concrete implementation for my IType interface, I have to go back and
remove all of these nullables first. I am dealing with some complex objects
here, so in my case I have over 260 errors to fix! This will not work if I
have to fix them each time I re-gen the classes.

Serjster:
Here's the JSON:

{"swagger":"2.0","info":{"version":"v1","title":"iUS Claim Service
API","description":"iUS Claim Service API for retrieving all things... wait
for it... Claim
[...]

PR #4145, opened 2016-11-06
[csharp] Use default rather than null in ctor

jimschubert:
### PR checklist

### Description of the PR

See #3608

This improves the defaults on constructor parameters by using `defaults(type)` rather than `null` for all inputs.

~~There seems to have been some weird usage of nullable types, possibly on parameters that aren't defined in the target swagger definition as nullable although I'm not 100% sure and didn't spend a lot of time investigating this.~~
Edit: I looked more closely and the nullable types seem to only occur when a primitive or enum is not marked as required. I had actually [previously explained this behavior for properties](https://github.com/swagger-api/swagger-codegen/issues/2885#issuecomment-220203786), but wasn't thinking broadly enough when I made comment about constructor parameters above.

This PR changes constructor parameters in models to:

1. Default to a default if present
2. Set to `default(TheType)` if no default value is defined

### Old Behavior: Dog

```
public Dog(string ClassName = null, string Color = null, string Breed = null)
{
    // to ensure "ClassName" is required (not null)
    if (ClassName == null)
    {
        throw new InvalidDataException("ClassName is a required property for Dog and cannot be null");
    }
    else
    {
        this.ClassName = ClassName;
    }
    // use default value if no "Color" provided
    if (Color == null)
    {
        this.Color = "red";
    }
    else
    {
        this.Color = Color;
    }
    this.Breed = Breed;
}
```

### New Behavior: Dog

```
 public Dog(string ClassName = default(string), string Color = "red", string Breed = default(string))
{
    // to ensure "ClassName" is required (not null)
    if (ClassName == null)
    {
        throw new InvalidDataException("ClassName is a required property for Dog and cannot be null");
    }
    else
    {
        this.ClassName = ClassName;
    }
    // use default value if no "Color" provided
    if (Color == null)
    {
        this.Color = "red";
    }
    else
    {
        this.Color = Color;
    }
    this.Breed = Breed;
}
```

### New Behavior: Other constructor examples

```
public Cat(string ClassName = default(string), 
            string Color = "red", 
            bool? Declawed = default(bool?)
) { }

public EnumArrays(JustSymbolEnum? JustSymbol = default(JustSymbolEnum?), 
            List<ArrayEnumEnum> ArrayEnum = default(List<ArrayEnumEnum>)
) { }

public Pet(long? Id = default(long?), 
           Category Category = default(Category), 
           string Name = default(string), 
           List<string> PhotoUrls = default(List<string>), 
           List<Tag> Tags = default(List<Tag>), 
           StatusEnum? Status = default(StatusEnum?)
) { }
```

I didn't include regenerated samples in this PR.

## Question

If this looks good, should I go ahead and update aspnetcore and NancyFx generators to do the same for models?

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -2192,7 +2192,7 @@ public class DefaultCodegen {
         p.jsonSchema = Json.pretty(param);
 
         if (System.getProperty("debugParser") != null) {
-            LOGGER.info("working on Parameter " + param);
+            LOGGER.info("working on Parameter " + param.getName());
         }
 
         // move the defaultValue for headers, forms and params
@@ -2212,7 +2212,7 @@ public class DefaultCodegen {
             String collectionFormat = null;
             String type = qp.getType();
             if (null == type) {
-                LOGGER.warn("Type is NULL for Serializable Parameter: " + param);
+                LOGGER.warn("Type is NULL for Serializable Parameter: " + param.getName());
             }
             if ("array".equals(type)) { // for array parameter
                 Property inner = qp.getItems();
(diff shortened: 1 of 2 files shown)


jimschubert:
@wing328 I'll try to get the same fix in place for aspnetcore and NancyFX within a day or so.

PR #4152, opened 2016-11-07
Select application/json content-type in python generated client, if */* is in the list of content-types

mbohlool:
### PR checklist

### Description of the PR

select_header_content_type in generated python client's api_client prefers application/json if it is in the list of content types. But the list may content */* that means server accepts any content-type. In that case we should still prefer application/json content type.


wing328:
@mbohlool thanks for the PR, which has been merged into master.

PR #4157, opened 2016-11-08
add file response support

### LINKED ISSUE #4034: [ALL] lack of recieving file response support
##### Description

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#responseObject

`Response Object`'s schema can be a single file type.
But swagger-codegen does not support it.

https://github.com/swagger-api/swagger-codegen/blob/v2.2.1/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
https://github.com/swagger-api/swagger-codegen/blob/v2.2.1/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenResponse.java
https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenProperty.java

`CodegenOperation` has `isResponseBinary` but does not `isResponseFile`.
`CodegenResponse` has `isBinary` but does not `isFile`.
`CodegenProperty` has `isBinary` but does not `isFile`.

https://github.com/swagger-api/swagger-codegen/blob/v2.2.1/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L1892-L1894

Therefore swagger-codegen supports binary response but does not file response.
##### Swagger-codegen version

swagger-codegen 2.2.1 or above

``` yaml
swagger: '2.0'
info:
  title: File Response Api
  version: "1.0.0"
paths:
  /file:
    get:
      summary: File
      description: Produces File
      parameters:
        - name: id
          in: query
          description: file id
          required: true
          type: number
          format: int32
      produces:
        - application/octet-stream
      responses:
        200:
          description: An array of products
          schema:
            type: file
```
##### Suggest a Fix

We should add relevant properties and some code flagments.

Thanks.

www2k:
### PR checklist

### Description of the PR

 - Closes #4034
 - Add file response support for typescript-node client generator

Sorry for late action. I did try to treat both `binary` and `file` as `Buffer in Node.js`
However I could not find a good solution for it so I just make a PR for `file`.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -36,6 +36,7 @@ import io.swagger.models.properties.DateProperty;
 import io.swagger.models.properties.DateTimeProperty;
 import io.swagger.models.properties.DecimalProperty;
 import io.swagger.models.properties.DoubleProperty;
+import io.swagger.models.properties.FileProperty;
 import io.swagger.models.properties.FloatProperty;
 import io.swagger.models.properties.IntegerProperty;
 import io.swagger.models.properties.LongProperty;
@@ -765,6 +766,7 @@ public class DefaultCodegen {
         typeMapping.put("integer", "Integer");
         typeMapping.put("ByteArray", "byte[]");
         typeMapping.put("binary", "byte[]");
+        typeMapping.put("file", "File");
 
 
         instantiationTypes = new HashMap<String, String>();
@@ -1069,6 +1071,8 @@ public class DefaultCodegen {
             datatype = "ByteArray";
         } else if (p instanceof BinaryProperty) {
             datatype = "binary";
+        } else if (p instanceof FileProperty) {
+            datatype = "file";
         } else if (p instanceof BooleanProperty) {
             datatype = "boolean";
         } else if (p instanceof DateProperty) {
@@ -1536,6 +1540,9 @@ public class DefaultCodegen {
         if (p instanceof BinaryProperty) {
             property.isBinary = true;
         }
+        if (p instanceof FileProperty) {
+            property.isFile = true;
+        }
         if (p instanceof UUIDProperty) {
             property.isString = true;
         }
@@ -1964,6 +1971,9 @@ public class DefaultCodegen {
                 if (r.isBinary && r.isDefault){
                     op.isResponseBinary = Boolean.TRUE;
                 }
+                if (r.isFile && r.isDefault){
+                    op.isResponseFile = Boolean.TRUE;
+                }
             }
             op.responses.get(op.responses.size() - 1).hasMore = false;
 
@@ -2163,6 +2173,7 @@ public class DefaultCodegen {
             }
             r.dataType = cm.datatype;
             r.isBinary = isDataTypeBinary(cm.datatype);
+            r.isFile = isDataTypeFile(cm.datatype);
             if (cm.isContainer != null) {
                 r.simpleType = false;
                 r.containerType = cm.containerType;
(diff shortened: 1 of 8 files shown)


baronfel:
First off, this is fantastic and I want to help out pushing this over the line however I can.

However, I don't think it's quite there.  I tried running the generation against the test file response spec that you added at `modules/swagger-codegen/src/test/resources/2_0/fileResponseTest.json` using the typescript-node generator and the output I get is still not taking file into account.  I've got a gist [here](https://gist.github.com/baronfel/41c49bbaa3917ea1433cb5cd5eafe006#file-api-ts-L145) that shows the result.   I'm running `java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -l typescript-node -i modules/swagger-codegen/src/test/resources/2_0/fileResponseTest.json` as my testing command.

I've been trying to debug this to no great success, but so far I haven't been able to verify that the `isResponseFile` boolean even gets set on line 1975 of `DefaultCodgen.java`.

www2k:
@baronfel Thank you for reporting the problem. It was my fault. I removed necessary code snippet (`getTypeDeclaration`) accidentally. It was fixed now.

baronfel:
That's great to hear! I'll try this with my own spec and see if I can perform POSTS using files.

wing328:
@baronfel please keep us posted with the test result.

baronfel:
I'll try to spend more time on this over the weekend.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptNodeClientCodegen.java
```diff
@@ -27,6 +29,8 @@
     public TypeScriptNodeClientCodegen() {
         super();
 
+        typeMapping.put("file", "Buffer");
```
wing328: @www2k If a model is also named "Buffer", it would break the TS API client. I think we need to add `Buffer` as a reserved keyword in the abstract typescript client codegen class.

www2k: Your idea is partially right but `Buffer` comes from 
[Node.js](https://nodejs.org/api/buffer.html#buffer_class_buffer). So I think I should make the diff below:

```diff
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptNodeClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptNodeClientCodegen.java
@@ -29,6 +29,8 @@ public class TypeScriptNodeClientCodegen extends AbstractTypeScriptClientCodegen
     public TypeScriptNodeClientCodegen() {
         super();

+        languageSpecificPrimitives.add("Buffer");
+
         typeMapping.put("file", "Buffer");

         // clear import mapping (from default generator) as TS does not use it
```

wing328: @www2k I will provide an example to explain the issue a bit more.

wing328:
@www2k PR merged into master. Thanks for your contribution.

modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
```diff
@@ -3257,6 +3275,9 @@ public void setParameterBooleanFlagWithCodegenProperty(CodegenParameter paramete
         } else if (Boolean.TRUE.equals(property.isBinary)) {
             parameter.isByteArray = true;
             parameter.isPrimitiveType = true;
+        } else if (Boolean.TRUE.equals(property.isFile)) {
+            parameter.isFile = true;
+            parameter.isPrimitiveType = true;
```
wing328: Fixed via https://github.com/swagger-api/swagger-codegen/pull/4421

PR #4159, opened 2016-11-08
fix[dart]: send ISO8601 dates

ircecho:
Dates for parameters were sent by using DateTime#toString. This would give us a date of the format "2016-11-08 14:33:37.270". Now I changed it to use DateTime#toIso8601String to give us a value like "2016-11-08T14:34:29.520". There were functions to format the date correctly, but those were not used anymore, which is why I assume that ISO8601 should have been used in the first place.

In other news, I decided not to use the provided code which relied on the intl package to format a ISO 8601, but a function from the DateTime class in core, thereby eliminating the necessity for using the external "intl" library.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/dart/api_client.mustache
+++ b/modules/swagger-codegen/src/main/resources/dart/api_client.mustache
@@ -17,7 +17,6 @@ class ApiClient {
 
   final dson = new Dartson.JSON()
                    ..addTransformer(new DateTimeParser(), DateTime);
-  final DateFormat _dateFormatter = new DateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
 
   final _RegList = new RegExp(r'^List<(.*)>$');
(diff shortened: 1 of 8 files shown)


wing328:
@ircecho if I understand correctly, this is a breaking change so please submit it against 2.3.0 branch instead.

ircecho:
@wing328 I would see this as a bugfix. The code was adapted to work as originally intended. The intention can be seen by the inclusion of the intl library by the original author and the provision of the format function to format dates for iso8061, which was no longer in use, due to some change in the past. Additionally, our backend accepts iso8061 dates only, and to me it has been working with the dart client library at least last year, and although i could not find any particular change that changed the date format (I did not search very long), I assume that some commit within the last year changed the library to erroneously use non standard date formats, which is why it would be okey to change it back.

If you still insist, that the changes be merged into the 2.3.0 branch instead, I will of course change the pull request.

wing328:
@ircecho It's perfectly fine to merge this PR into current master as you've provided very good reasons.

PR #4166, opened 2016-11-10
Support _preload_content flag in python client

mbohlool:
### PR checklist

### Description of the PR
In order to support streaming responses, we need the API Client not to read response content and return a Response object. urllib3 has a flag called preload_content that if set to False, it will return a urllib3.HTTPResponse. This PR add support for this flag in rest client and api client and generated api classes. The flag is optional and defaulted to True to keep current behavior.


wing328:
@mbohlool thanks for the contribution. If it's not too much to ask, do you mind adding a test case for `_preload_content` in https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/python/tests/test_pet_api.py so that the test case won't pass in case some future changes break it?

cc @scottrw93

mbohlool:
@wing328 done #4172

PR #4178, opened 2016-11-13
[Flask] Add models support to Flask Connexion codegen

cbornet:
* Generates the models with type hints (using typing which becomes a dependency in py2)
* Adds docstring and type hints in controllers
* Generate calls to deserialize functions in controllers when appropriate
* Adds the 'items' CodegenProperty to CodegenParameter when the CodegenParameter is an array or a map (needed by this PR to get the isString, isInteger, isXXX... attributes of the boxed type)


cbornet:
Note for later : the Python and Flask codegen common methods should be mutualized in an AbstractPythonCodegen class.

PR #4179, opened 2016-11-14
add support for relative path in specs files for SpringMVC.   

YaoYao-dd:
### PR checklist
when we define basePath in spec like this   petstore/v2,   it normally means in the web container,   petstore is the context path,  v2  is global  relative path for all resource path.  so it is good to generate API with this relative path for @RequestMapping(value = "/v2/dogs"),   also it is better to define the controller level  mapping since controller is resource based.

### Description of the PR


wing328:
@beingcoder thanks for the PR. Please use 4-space instead for code indentation in the Java files.

beingcoder:
@wing328   ,  adjusted the format,  thx.

wing328:
@beingcoder there are still tabs in the Java files.

Please provide an example and retarget this for 2.3.0 branch instead as it might be a breaking change.

wing328:
cc @cbornet

cbornet:
I'm not sure I understand what this does and I'm not sure I agree with it...

PR #4183, opened 2016-11-14
[csharp] Escape special characters in the API doc

iCEAGE:
Special characters like <, > should be HTML escaped. When the data type contains these characters (e.g.  `List<long?>`), should be converted to the escaped version(e.g. `List&lt;long?&gt;`).


wing328:
@iCEAGE How did you display the document (markdown)? (< and > need to be escaped in HTML but not markdown)

iCEAGE:
@wing328 As Markdown supports inline HTML, I think open angle bracket following a close angle bracket should be escaped. Please refer to [Automatic Escaping for Special Characters](https://daringfireball.net/projects/markdown/syntax#autoescape) in Markdown documentation.

> How did you display the document (markdown)?

Github

wing328:

PR #4187, opened 2016-11-15
Allow multiple requests in parallel in Python client

hermansc:
### PR checklist

### Description of the PR

If you tried to do two parallel calls to the same API object in the Python client you would get an error from urllib3 connection pool:

> Connection pool is full, discarding connection: ***

Because the [default maxsize=1](https://github.com/shazow/urllib3/blob/f9409436f83aeb79fbaf090181cd81b784f1b8ce/urllib3/connectionpool.py#L162)

By defaulting to a higher maxsize we mitigate for the common use case where a user is running a couple of requests in parallel.

Ideally, in the future, this should be a configuration parameter together with the pool size.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/python/rest.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/rest.mustache
@@ -49,10 +49,11 @@ class RESTResponse(io.IOBase):
 
 class RESTClientObject(object):
 
-    def __init__(self, pools_size=4):
+    def __init__(self, pools_size=4, maxsize=4):
         # urllib3.PoolManager will pass all kw parameters to connectionpool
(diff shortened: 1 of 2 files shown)


wing328:
@hermansc thanks for the PR. Please run `./bin/python-petstore.sh` to update the Python Petstore sample so that the CIs can verify the change.

hermansc:
@wing328 Thanks for picking up this one. I've added the sample code now - just forgot to add it to my commit earlier.

wing328:
@hermansc thanks for the PR. 

May I know if you've time to add `maxsize` to [configuration.py](https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/python/petstore_api/configuration.py)?

hermansc:
@wing328 I'll have a look at it this weekend :)

PR #4201, opened 2016-11-16
fix(swift3): rename reserved enum values

gierschv:
### PR checklist

### Description of the PR

fix(swift3): rename reserved enum values

Example:
```diff
- case public = "public"
+ case _public = "public"
```


jaz-ah:
looks like for other reserved words we do:

    public String escapeReservedWord(String name) {
        return "_" + name;  // add an underscore to the name                                                                                                                                           
    }

gierschv:
@jaz-ah ok, updated.

jaz-ah:
+1 @wing328

PR #4202, opened 2016-11-16
fix(swift3): multi-level inheritance support

gierschv:
### PR checklist

### Description of the PR

Supports multi-level inheritance to avoid having the variables from the parents of level > 2

This commit also fixes the comparison of the variables when skipping the variables in the children classes by only comparing their names (in some cases some parent variables were present in children classes)

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Swift3Codegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Swift3Codegen.java
@@ -427,10 +427,18 @@ public class Swift3Codegen extends DefaultCodegen implements CodegenConfig {
         if(codegenModel.description != null) {
             codegenModel.imports.add("ApiModel");
         }
-        if (allDefinitions != null && codegenModel.parentSchema != null) {
-            final Model parentModel = allDefinitions.get(codegenModel.parentSchema);
-            final CodegenModel parentCodegenModel = super.fromModel(codegenModel.parent, parentModel);
+        if (allDefinitions != null) {
+          String parentSchema = codegenModel.parentSchema;
+
+          // multilevel inheritance: reconcile properties of all the parents
+          while (parentSchema != null) {
+            final Model parentModel = allDefinitions.get(parentSchema);
+            final CodegenModel parentCodegenModel = super.fromModel(codegenModel.parent, parentModel, allDefinitions);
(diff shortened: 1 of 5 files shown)


gierschv:
cc @wing328 @jaz-ah

jaz-ah:
+1 @wing328

wing328:
@gierschv thanks for the PR. Do you mind sharing the spec so that I can more easily reproduce the issue?

I'm think we may need similar fix in other language generators.

gierschv:
@wing328 Sure, here is a short example.

3 different views of an object "User":
- `User`: Minimal profile (we can return as a nested object for example)
- `UserPublicProfile`: Full public profile (inherited from `User`)
- `UserDetails`: Private profile with settings & preferences (inherited from `UserPublicProfile`)

```yaml
swagger: "2.0"

info:
  title: Multi-level Inheritance
  version: "v1.0"

schemes: ['http']
consumes: ['application/json']
produces: ['application/json']

paths:
  /me:
    get:
      summary: User account
      responses:
        200:
          description: The user settings
          schema:
            $ref: '#/definitions/UserDetails'

definitions:
  User:
    type: object
    discriminator: userType
    properties:
      id:
        type: string
      username:
        type: string

  UserPublicProfile:
    allOf:
      - $ref: '#/definitions/User'
      - type: object
        properties:
          organization:
            type: string
          htmlUrl:
            type: string

  UserDetails:
    allOf:
      - $ref: '#/definitions/UserPublicProfile'
      - type: object
        properties:
          email:
            type: string
            format: email
```

Without this PR, the model `UserDetails` generated will have the properties of `User`: `id` and `username`, which will not compile since those are already defined in the parent model `User`.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Swift3Codegen.java
```diff
@@ -582,7 +590,7 @@ private static CodegenModel reconcileProperties(CodegenModel codegenModel, Codeg
           Iterator<CodegenProperty> iterator = codegenProperties.iterator();
           while (iterator.hasNext()) {
               CodegenProperty codegenProperty = iterator.next();
-              if (codegenProperty.equals(parentModelCodegenProperty)) {
+              if (codegenProperty.baseName == parentModelCodegenProperty.baseName) {
```
wing328: So moving forward, the `baseName` of the property becomes the key of the property, which is reasonable as I can't think of a language which allows both parent and child property/member having the same name.

PR #4204, opened 2016-11-16
[html2] Add missing enums to parameters

bshamblen:
`enum` values were not being displayed in the **Parameters** section. Enums are now displayed next to the **Enums:** label, as a comma separated list.

I also cleaned up the HTML formatting if the index.mustache page, to make it easier to read and debug.

Lastly, I hid the **Schema** block below the response code, if no schema was defined. For example, when a 401 error is defined, if there's no `schema` specified, the empty schema tab and panel body below that header are no longer rendered. It was taking up a lot of vertical space having a bunch of empty schema blocks for each endpoint.


wing328:
@bshamblen thanks for the PR, which has been merged into the master.

PR #4209, opened 2016-11-17
[Flask] Add generated tests

cbornet:


cbornet:
The generated tests for the petstore pass so they probably could be used to test Flask in CI

cbornet:
Conflicts rebased and:
* modify to use the generated models in the tests
* fix default model package to ` models`

PR #4215, opened 2016-11-18
[Flask] fix parameter naming

cbornet:
Flask-connexion maps directly the param name in the swagger spec to the param name in the controller method so it must not be changed in the method during generation.
If santitization is needed then a warning log will be emitted.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/FlaskConnexionCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/FlaskConnexionCodegen.java
@@ -10,8 +10,7 @@ import io.swagger.models.HttpMethod;
 import io.swagger.models.Operation;
 import io.swagger.models.Path;
 import io.swagger.models.Swagger;
-import io.swagger.models.parameters.BodyParameter;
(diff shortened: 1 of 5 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/FlaskConnexionCodegen.java
```diff
@@ -317,6 +316,15 @@ public void preprocessSwagger(Swagger swagger) {
                                     controllerPackage + "." + toApiFilename(tag)
                             );
                         }
+                        for (Parameter param: operation.getParameters()) {
+                            // sanitize the param name but don't underscore it since it's used for request mapping
```
cbornet: Flask is a little special in that the swagger file is part of the generated code so this code is to sanitize the param name in the yaml file itself.

wing328: That means the developers need to use the **updated** spec to generate the API clients, right? otherwise the auto-generated API client cannot talk to the Flask backend as the parameter name may have been renamed.

PR #4219, opened 2016-11-19
fix(javascript): add check to see if the parsed response is really empty.

jordanyaker:
### PR checklist

### Description of the PR

For non-JSON responses (i.e., strings, etc.) SuperAgent was still returning an empty prototype for `response.body`.  This change makes it so both a null result AND an empty prototype (i.e., one without keys) will cause `response.body` to be replaced by `response.text`.


wing328:
@jordanyaker thanks for the PR, which has been merged into master.

Issue #4228, opened 2016-11-21
[CodeGen] boolean values from JSON are treated as strings


yohanakh:
##### Description

Boolean values from JSON are treated as string. I believe this happens in every language as it is in the generator code.

This applies to dynamicProperties and additionalProperties.

##### Swagger-codegen version

I am using the latest code in master branch with commit 1968bead432aac0adbf3cc576758947bf0803e11 (21/11/2016).

##### Swagger declaration file content or url

```json
{
        ...
	"java8": false
}
```

##### Command line used for generation

swagger-codegen generate -l spring -i api-v1.yaml -o rest -c config.json

##### Steps to reproduce

1. Run the generate command with the JSON above.
2. Open the generated rest/pom.xml file and note that the java.version property is 1.8 instead of 1.7

##### Related issues

#3318

##### Suggest a Fix

1. Change the value types in additionalProperties and dynamicProperties to Object instead of String.
2. Change methods that insert values to these maps to receive Object as the type of the value instead of String. 

I already have a fix and updated the existing tests to cover it.
https://github.com/yohanakh/swagger-codegen/tree/issue-4228

And created a pull request: 
https://github.com/swagger-api/swagger-codegen/pull/4229

wing328:
@yohanakh thanks for the PR. Are you referring to those boolean value in the config.json not being parsed correctly?

yohanakh:
@wing328 yes, the boolean values in config.json

wing328:
PR merged into master. Please pull the latest.

Thanks for your contribution.

yohanakh:
Thank you!
When are you planning to release v2.2.2 ?

wing328:
We're targeting Jan/Feb 2017

yohanakh:
That's great! Thanks.
Should I close this issue ?

wing328:
Yes please ( I thought I did )

frol:
I cannot follow the logic behind the fix because I am not a Java dev, but it seems that it broke the native boolean types in JSON config:

Having this config for Javascript client:

```json
{
  "usePromises": true
}
```

and running Swagger Codegen like this:

```
$ java -jar /opt/swagger-codegen/modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate --input-spec ./swagger.json --lang javascript --output ./dist --config ./config.json
```

I receive a bunch of odd errors. It starts with this one:

```
[...]

PR #4229, opened 2016-11-21
boolean values from JSON are treated as strings

yohanakh:
### PR checklist

### Description of the PR

For #4228 

Running the *petstore.sh scripts has unwanted result, it changed many files where it should not.  Any suggestion? 

These are a non-breaking changes.

##Changes:
Change the value types in additionalProperties and dynamicProperties to Object instead of String.
Change methods that insert values to these maps to use Object as the type of the value instead of String.


wing328:
@yohanakh I did some tests and the result looks good. Thanks for your contribution.

Issue #4236, opened 2016-11-22
[Java] Bug generating Java model with '_class' property


menchauser:
##### Description

This issue is similar to #2841. swagger-codegen may produce invalid code for model like
```
  test:
    type: object
    properties:
      '_class':
        type: string
```

Result code contains inelligible overload for method Object.getClass():
```
    public String getClass() {
        return _class;
    }
```

##### Swagger-codegen version

Latest (2.2.2-SNAPSHOT)

menchauser:
I've implemented simple fix: #4237. That could be not the best solution and, I'm afraid, I haven't checked all language/framework combinations (except Java clients).

PR #4237, opened 2016-11-22
Fix support for 'class' properties in Java codegen

menchauser:
Currently Java codegen works successfully for property named 'class' but
fails on '_class', '__class', etc, because of 'Object.getClass()'
overloading.

This fix is intended to avoid all Object method overloading cases.

Fix is similar to #3177
Resolves: #4236

(Sample code regeneration caused big changeset)


menchauser:
Seems like I've broken existing tests by regenerating samples.

wing328:
@menchauser thanks for the PR. Your suggested fix looks good to me.

The enum issue caused by updated Java petstore sample is not related to the change you made. I'll look into it.

wing328:
I fixed the test cases via #4329

This PR has been merged into master. Thanks for the contribution.

Issue #4082, opened 2016-10-27
Truncated PHP example code in the html2 generated client


David-Vander-Elst:
##### Description

When I generate the HTML2 client, we have a lot of code example for several languages. All langages examples looks great except for PHP. A part of the PHP code is truncated above
##### Swagger-codegen version

Current one.
##### Steps to reproduce

Simply generate an html2 client and check the PHP code example.

Thanks.

wing328:
@David-Vander-Elst thanks for reporting the issue. May I know if you've time to contribute the enhancements?

David-Vander-Elst:
it looks like this is simply an opened tag at line 3 witch is commented and not the closed one.
I'm not able to test it right now as I'm a PHP developer only, but I can try later to set an environment, but not my priority right now.

jhitchcock:
This should resolve it; 
https://github.com/swagger-api/swagger-codegen/pull/4212

The opening PHP tag seems to be screwing up the HTML, and tbh its not really needed anyway in this context so I removed it. I may look into it in more detail in the future as it could cause issues with new languages

wing328:
PR merged into master. Please pull the latest master to give it a try.

PR #4243, opened 2016-11-22
[html2] Fix import statements for most languages

bshamblen:
Most of the code examples in the html2 template had various issues with their import statements. 

I also fixed the issue with the php example by escaping the question mark in the opening `<?php` tag. Related to #4082.


wing328:
@bshamblen thanks for the contribution. 

cc @jhitchcock

modules/swagger-codegen/src/main/resources/htmlDocs2/sample_php.mustache
```diff
@@ -1,18 +1,18 @@
-
+<&#63;php
 require_once(__DIR__ . '/vendor/autoload.php');
 {{#hasAuthMethods}}{{#authMethods}}{{#isBasic}}
 // Configure HTTP basic authorization: {{{name}}}
-{{{invokerPackage}}}\Configuration::getDefaultConfiguration()->setUsername('YOUR_USERNAME');
-{{{invokerPackage}}}\Configuration::getDefaultConfiguration()->setPassword('YOUR_PASSWORD');{{/isBasic}}{{#isApiKey}}
+Swagger\Client\Configuration::getDefaultConfiguration()->setUsername('YOUR_USERNAME');
```
wing328: Would it be better to use `{{{invokerPackage}}}` as PHP SDK owners will not always use the default (Swagger\Client) ?

bshamblen: @wing328 Sorry I didn't get back to you sooner. The problem is that the `{{{invokerPackage}}}` value is currently only correct for the Java example, but would not be accurate for any of the other languages. If you look at an existing example for PHP, you'll notice it's currently wrong (showing the default Java `invokerPackage` value):

screen shot 2016-12-05 at 2 28 44 pm

But, if I generate the `php` client, the README that's generated shows the following example:

screen shot 2016-12-05 at 2 35 31 pm


Even if someone were to specify a custom `invokerPackage` command line build value for the htmlDocs2 template, it would only be valid for one of the six examples. 

Short of creating 6 different invoker package command line arguments, I don't see a good way to make them 100% accurate for each language.

wing328: @bshamblen thanks for the explanation. What about using `--additional-properties=phpInvokerPackage="Swagger\Client"` (in which {{phpInvokerPacakge}} can be accessed in the mustache template) instead so as to allow package naming with respect to different languages?

bshamblen:
@wing328 I've set up default values for the phpInvokerPackage and packageName properties and tested it locally with custom values for each. I'll work on #4358 once this is merged. I didn't want to add too many changes to this PR that weren't related.

wing328:
@bshamblen the change looks good to me 👍

wing328:
FYI. I've updated `html2` Petstore sample via `./bin/html2-petstore.sh` (2bf3d051a9ae1af89e3b83ba43aa66c5aa997ab7)

PR #4259, opened 2016-11-25
Still respect super.equals even for models without own vars

j4velin:
### Description of the PR

The Java equals method does not call super.equals but directly returns true if the class has no own vars


wing328:
The AppVeyor issue (due to caching) can be ignore

wing328:
@j4velin PR merged into master. Thanks for your contribution.

wing328:
Likely we'll need something similar for other languages. I'll create another task for tracking.

PR #4287, opened 2016-11-29
[Typescript/fetch][Issue4284] Handle query parameters containing colons

### LINKED ISSUE #4284: [Typescript/fetch] Can't use colon in query parameter name
##### Description

If a query parameter contains a colon, the generated code is broken since they are interpreted as type information by the typescript compiler. 

##### Swagger-codegen version

2.2.1 - This is not a regression

```json
{
  "swaggerVersion": "1.2",
  "basePath": "http://localhost:8000/greetings",
  "apis": [
    {
      "path": "/hello/{subject:main}",
      "operations": [
        {
          "method": "GET",
          "type": "string",
          "parameters": [
            {
              "name": "subject:main",
              "type": "string",
              "paramType": "path"
            }
          ]
        }
      ]
    }
  ],
  "models": {}
}
```

swagger-codegen generate -i hello.json -l typescript-fetch -o generated

##### Steps to reproduce

Grep for `subject:main` in the generated code. 

##### Suggest a Fix
- escape all usage of the dot notation to refer to parameters in `api.mustache`. So replacing  `params.{{paramName}}` by `params["{{paramName}}"]`
- Also replace the param names when they are referred to as interface keys: `{paramName: xxx}` becomes `{"paramName": xxx}`

ChrisJamesC:
Put query parameters between quotes and access object keys using square
brackets to make sure special characters are handled in parameter
names.
Closes #4284


wing328:
@ChrisJamesC thanks for the PR. The change looks good to me. Please rebase on the latest master when you've time.

We might need to apply the same fix to other typescript generators (node, angular, angular2)

cc @Vrolijkx

ChrisJamesC:
@wing328 I updated the branch from master. 

Please let me know if there is anything I can do.

Vrolijkx:
@wing328 This fix is little bigger in the angular2 version
We pass our arguments as function parameters and not as parameter object. Changing this would be breaking and should be discussed a little bit more first.

wing328:
@Vrolijkx sure, I've opened https://github.com/swagger-api/swagger-codegen/issues/4290 for tracking.

Issue #3925, opened 2016-10-04
Codegen does not read all attributes of schema property


abhijith-prabhakar:
Swagger specification for schema object specifies a list of attributes which are valid for a given schema object.  But codegen(CodegenProperty.java) does not read in all of this attributes.  If declared on the schema, these attributes are ignored.  

Codegen should read these properties and use it for model generation with bean validation enabled.  Is linked to https://github.com/swagger-api/swagger-codegen/issues/2549

JasonNorth:
This fixes maxItems and minItems for me although it needs a release of swagger-core first:

https://github.com/swagger-api/swagger-codegen/pull/4293/commits/4d55d25f9f5180f6cbb685d4d5c4b85405f7a719

PR #4309, opened 2016-12-02
Allow templates to use maxItems and minItems for Arrays

JasonNorth:
This is related to issue 3925 but only addresses maxItems and minItems.  A release of swagger-core is required.


wing328:
@JasonNorth thanks for the PR, which has been merged into master.

(linked to #3925)

PR #4325, opened 2016-12-06
Add hasDiscriminator property to CodegenProperty and CodegenParameter.

sreeshas:
### PR checklist

### Description of the PR

This change adds hasDiscriminator property to CodegenModel. 
hasDiscriminator can be used in model.mustache to identify a model which declares discriminator.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenParameter.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenParameter.java
@@ -23,6 +23,11 @@ public class CodegenParameter {
     public Boolean hasValidation;
 
     /**
+     * Set to true, if type associated with <code>CodegenParameter</code> contains a discriminator.
+     */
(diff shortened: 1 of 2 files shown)


jeff9finger:
This sounds like it could be useful, but isn't there already a way to do this?
In the Java templates, there is this line...
```{{>generatedAnnotation}}{{#discriminator}}{{>typeInfoAnnotation}}{{/discriminator}}```

sreeshas:
what @jeff9finger suggested works. This PR might not be required after all. I'll close it once i make sure it works for my use case.

sreeshas:
CodegenModel did not require hasDiscriminator field. CodegenParameter and CodegenProperty needs hasDiscriminator to identify if type associated with them contains a discriminator.

sreeshas:
I was able to solve this problem by using vendor extensions. While this could be beneficial in some use cases, i don't need this PR since i have workaround. Closing it.

PR #4326, opened 2016-12-06
Refactor Boolean properties to boolean in CodegenProperty, CodegenOperation and CodegenModel classes

sreeshas:
### PR checklist

### Description of the PR

There is no good reason to use Boolean instead of boolean.
Using Boolean will lead to handling "null" state which is confusing.
This change replaces Boolean properties to boolean.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenModel.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenModel.java
@@ -39,8 +39,8 @@ public class CodegenModel {
     public Set<String> allMandatory;
 
     public Set<String> imports = new TreeSet<String>();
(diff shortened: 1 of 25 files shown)


jeff9finger:
IIRC, using Boolean is needed, at least, in cases where a PUT operation is updating an object. The server needs to be able to detect if the property is not set, or has a value. Similar to using Optional for non-primitive types.

sreeshas:
@jeff9finger This change is only for Boolean properties in CodegenModel, CodegenProperty and CodegenOperation. It shouldn't affect server behavior or API definitions in any way.

wing328:
Personally I don't have a strong opinion on which one to use. I did some research and seems like primitive is more preferable than class: http://softwareengineering.stackexchange.com/questions/203970/when-to-use-primitive-vs-class-in-java

If no one has further feedback or question on this change, I'll merge on coming Friday.

PR #4352, opened 2016-12-08
[Python] Make the code look Pythonic

frol:
### PR checklist

### Description of the PR

The `api_client.py` looks non-Pythonic to me in some places, so I went ahead and fixed at least the oddest ones:

* `eval` is dangerous and it was trivial to implement it in a safe way.
* list-comprehensions are faster and easier to read than `list(map(lambda: ))`.
* `if "file" == response_type:` is a C-style.


samples/client/petstore-security-test/python/petstore_api/api_client.py
```diff
@@ -1,44 +1,32 @@
...
-
 """
-Copyright 2016 SmartBear Software
-
-   Licensed under the Apache License, Version 2.0 (the "License");
-   you may not use this file except in compliance with the License.
-   You may obtain a copy of the License at
-
-       http://www.apache.org/licenses/LICENSE-2.0
+    Swagger Petstore */ ' \" =end -- \\r\\n \\n \\r
```
frol: I have no idea what caused the change, but it was certainly not my commit.

samples/client/petstore-security-test/python/petstore_api/api_client.py
```diff
@@ -1,44 +1,32 @@
 # coding: utf-8
-
 """
-Copyright 2016 SmartBear Software
-
-   Licensed under the Apache License, Version 2.0 (the "License");
-   you may not use this file except in compliance with the License.
-   You may obtain a copy of the License at
+    Swagger Petstore */ ' \" =end -- \\r\\n \\n \\r
```
wing328: @frol it's due to license removal and additional security test case. No worry :)

wing328:
@frol thanks for the PR 😃 

cc @scottrw93

modules/swagger-codegen/src/main/resources/python/api_client.mustache
```diff
@@ -165,10 +163,9 @@ class ApiClient(object):
         :param obj: The data to serialize.
         :return: The serialized form of data.
         """
-        types = (str, float, bool, bytes) + tuple(integer_types) + (text_type,)
-        if isinstance(obj, type(None)):
+        if obj is None:
             return None
-        elif isinstance(obj, types):
+        elif isinstance(obj, (float, bool, bytes, text_type) + integer_types):
```
wing328: @frol What about storing `(float, bool, bytes, text_type) + integer_types)` in a class static variable so as to avoid the tiny overhead in concatenation?

frol: I thought about it, but chose to minimize my patch, but since you have asked, I went ahead and moved those into the class static variable.

wing328:
@frol PR merged into master. Thanks for your contribution.

PR #4357, opened 2016-12-09
fix(swift3): lowercase enum value before checking reserved words

gierschv:
### PR checklist

### Description of the PR

When using alpha-2 ISO country codes in uppercase (e.g. IN, AS, ...), they were not escaped.

\+ Added test
\+ Fix ident


modules/swagger-codegen/src/test/java/io/swagger/codegen/swift3/Swift3CodegenTest.java
```diff
@@ -14,13 +14,13 @@
     Swift3Codegen swiftCodegen = new Swift3Codegen();
 
     @Test
-    public void testReservedWord() throws Exception {
-	Assert.assertEquals(swiftCodegen.toEnumVarName("Public", null), "_public");
+    public void testCapitalizedReservedWord() throws Exception {
+      Assert.assertEquals(swiftCodegen.toEnumVarName("AS", null), "_as");
```
wing328: @gierschv I would suggest using 4-space instead of 2-space for indentation to make it consistent with the rest of the file.

gierschv: @wing328 My bad, updated.

jaz-ah:
@gierschv unfortunately won't this break the change I was originally trying to fix w/ the word "Public" that you removed the test case for?

gierschv:
@jaz-ah Right, I failed my rebase early today. The test is back and passing.

wing328:
FYI. The Appveyor CI issue due to caching can be ignored.

PR #4366, opened 2016-12-12
fix the value for @Generated java annotation

trigan-d:
com.my.Generator.class.toString() returns "class com.my.Generator", and this value is then used in the javax.annotation.Generated annotation like that: @Generated(value = "class com.my.Generator").
Should use Generator.class.getName() instead, so we end up with @Generated(value = "com.my.Generator")


wing328:
@trigan-d The change looks good to me. Thanks for the PR to fix the issue.

wing328:

PR #4368, opened 2016-12-12
[html] Add type anchor to body param block

wy-z:
- [html] Add type anchor to body param block

Signed-off-by: weiyang weiyang.ones@gmail.com


wing328:
@wy-z thanks for the PR. Are you also aware of another documentation generator `html2`? Would that better meet your requirement?

wy-z:
Uh...It will cause my browser to crash...
I will take some time to find out the reasons.

wing328:
Does it crash all browsers (Chrome, IE, FF, Safari) in your environment?

wy-z:
It should not be a compatibility issue~

PR #4308, opened 2016-12-02
Add support for optional queryParams.

antihax:
### PR checklist

- [X] Read the [contribution guildelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [X] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [X] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

Few Golang fixes.

Add support for optional queryParams.

Remove parameterToString from exports (end user should not need this function).

Add comma back to localVarHttpContentTypes and localVarHttpHeaderAccepts as it refuses to compile without it.

Rewrite ParameterToString to handle other slice types other than just string.

PR #4306, opened 2016-12-02
Add support for optional queryParams.

antihax:
### PR checklist

- [X] Read the [contribution guildelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [X] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [X] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

Add support for optional queryParams.
Remove parameterToString from exports (end user should not need this function).
Make sure the object is not nil before validating.

[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO]
[INFO] swagger-codegen-project ............................ SUCCESS [  1.777 s]
[INFO] swagger-codegen (core library) ..................... SUCCESS [ 12.324 s]
[INFO] swagger-codegen (executable) ....................... SUCCESS [  4.993 s]
[INFO] swagger-codegen (maven-plugin) ..................... SUCCESS [  3.545 s]
[INFO] swagger-generator .................................. SUCCESS [  4.652 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS

D:\swagger\swagger-codegen\samples\client\petstore\go>go test
PASS
ok      _/D_/swagger/swagger-codegen/samples/client/petstore/go 5.861s

PR #4369, opened 2016-12-12
[go] Optimize model and api tmpl

wy-z:
- [api] Make the parameters optional
- [model] Use pointers whenever possible

Signed-off-by: weiyang weiyang.ones@gmail.com


wy-z:
And there is a bug in `/bin/go-petstore.sh`
image

NfNitLoop:
If I'm reading this right, (I'm not familiar w/ the codegen code base), it updates it so that function calls instead of being, ex: 

    GetFoo(int, string)

They're now:

    GetFoo(map[string]interface{})

So the type *and* number of the parameters is completely hidden from the call interface, and only checked at runtime.  (While #4308 only hides the type of parameters, but keeps their number and name as part of the generated code.)

NfNitLoop:
Also, in https://github.com/swagger-api/swagger-codegen/pull/4306 it was pointed out that API-breaking changes need to go in a next `_.X._` semantic version release, (2.3.0), so I'd assume that applies here.

wy-z:
They're now:
```
GetFoo(map[string]interface{}{
        "id": 6,
        "name: "foo",
})
```
~

wing328:
Yes, this is a breaking change and it should file against `2.3.0` branch instead.

wy-z:
ok~

Issue #4359, opened 2016-12-09
[NODE-SERVER] Date stub value set with ISO8601 date-time format


crazyfrozenpenguin:
##### Description

Swagger stub generator for node stubs date fields with date-time format.

date           ISO8601 full-date	2016-04-01
date-time  ISO8601 date-time	2016-04-16T16:06:05Z

##### Swagger-codegen version

2.2.1

##### Command line used for generation

java -jar swagger-codegen-cli.jar generate -i swagger.json -l nodejs-server -o stub

##### Steps to reproduce

Create a swagger endpoint with a format":"date" field. Generate stub using command above.

##### Related issues

None found

##### Suggest a Fix

The generated field should be a date (2000-01-23) not a date-time (2000-01-23T04:56:07.000+00:00).

wing328:
@crazyfrozenpenguin thanks for reporting the issue. May I know if you've time to contribute the fix?

Can you also show what the actual (problematic) code looks like?

A good starting point maybe is to define `typeMapping` in https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NodeJSServerCodegen.java similar to [this](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java#L123-L143)

bmsantos:
@wing328, I've submitted a fix for this. It was simpler than initially thought.

wing328:
Thanks for the fix from @bmsantos 

PR merged into master. Please pull the latest master to give it a try.

PR #4385, opened 2016-12-13
Issue #4359: Date type should not include time

bmsantos:
Removed time section from swagger date type in example generators.

Issue #4359


wing328:
@bmsantos thanks for the fix. Looks good to me.

PR #4406, opened 2016-12-15
Introduce NodeJS codegen for Google Cloud Functions.

jmuk:
### PR checklist

### Description of the PR

This PR introduces a new language "nodejs-google-cloud-functions." This is quite similar to nodejs server but the index.js is formed to work within [Google Cloud Functions]( https://cloud.google.com/functions/docs/).


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NodeJSGoogleCloudFunctionsCodegen.java
```diff
@@ -0,0 +1,92 @@
...
+        //   "controller.js")
+        // );
+        supportingFiles.add(new SupportingFile("swagger.mustache",
+                        "api",
+                        "swagger.yaml")
+        );
+        writeOptional(outputFolder, new SupportingFile("index.mustache", "", "index.js"));
+        writeOptional(outputFolder, new SupportingFile("package.mustache", "", "package.json"));
+        writeOptional(outputFolder, new SupportingFile("README.mustache", "", "README.md"));
+        if (System.getProperty("noservice") == null) {
```
wing328: @jmuk thanks for the PR. Instead of using `noservice` system property, what about using `java -Dmodels -DsupportingFiles` instead? https://github.com/swagger-api/swagger-codegen#selective-generation

You may also consider using `.swagger-codegen-ignore` to skip certain files when regenerating the code: https://github.com/swagger-api/swagger-codegen#ignore-file-format

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractNodeJSCodegen.java
```diff
@@ -0,0 +1,283 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+
+public abstract class AbstractNodeJSCodegen extends DefaultCodegen implements CodegenConfig {
```
wing328: @jmuk thanks for refactoring NodeJS generator into a common abstract class 👍

wing328:
@jmuk thanks for the contribution.

> This PR introduces a new language "nodejs-google-cloud-functions." This is quite similar to nodejs server but the index.js is formed to work within Google Cloud Functions.

If only index.js is different, what about adding a CLI option to use different index.js mustache template instead of introducing another generator.

jmuk:
To be more precise, not only `index.js` but also `package.json` (because it doesn't have to depend on some) and `README.md` are slightly different from nodejs-server -- but yeah, the contents are mostly common. Let me check how I can converge them into a single generator.

wing328:
@jmuk One way is to pass `--additional-properties=enableGoogleCloudFunction=true` via command line argument and then you can use the following in the mustache template
```
{{#enableGoogleCloudFunction}}
... code for enabling Google cloud function here ...
{{/enableGoogleCloudFunction}}
```

jmuk:
Updated. Now the single `NodeJSServerCodegen.java` does the codegen for GCF upon `googleCloudFunctions=true` additional property.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NodeJSGoogleCloudFunctionsCodegen.java
jmuk: re: `.swagger-codegen-ignore`: removed

re: `noservice`: well, this is what currently nodejs-server is doing. I think it's better to do things like that in another PR.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NodeJSServerCodegen.java
```diff
@@ -241,6 +249,46 @@ public String apiFileFolder() {
...
+                        "swagger.yaml")
+        );
+        if (getGoogleCloudFunctions()) {
+            writeOptional(outputFolder, new SupportingFile("index-gcf.mustache", "", "index.js"));
+        } else {
+            writeOptional(outputFolder, new SupportingFile("index.mustache", "", "index.js"));
+        }
+        writeOptional(outputFolder, new SupportingFile("package.mustache", "", "package.json"));
+        writeOptional(outputFolder, new SupportingFile("README.mustache", "", "README.md"));
+        if (System.getProperty("noservice") == null) {
```
wing328: Instead of using `noservice`, what about using selective generation as mentioned in https://github.com/swagger-api/swagger-codegen#selective-generation? Would that meet your need?

I prefer not to introduce another system property if the same output can already be achieved.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NodeJSServerCodegen.java
```diff
@@ -76,27 +82,13 @@ public NodeJSServerCodegen() {
...
-        writeOptional(outputFolder, new SupportingFile("package.mustache", "", "package.json"));
-        writeOptional(outputFolder, new SupportingFile("README.mustache", "", "README.md"));
-        if (System.getProperty("noservice") == null) {
-            apiTemplateFiles.put(
-                    "service.mustache",   // the template to use
-                    "Service.js");       // the extension for each file to write
-        }
+        cliOptions.add(CliOption.newBoolean(GOOGLE_CLOUD_FUNCTIONS,
+            "When specified, it will generate the code which runs within Google Cloud Functions "
+                + "instead of standalone Node.JS server."));
```
wing328: SUGGESTION: A URL to tutorial on how to deploy the auto-generated code to Google Cloud Functions would be helpful for developers who are new to Google Cloud Functions.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NodeJSServerCodegen.java
jmuk: Actually this comes from the original NodeJSCodegen.java, I just moved the code location. I'm not adding new.

Also, the `noservice` is for a different purpose from the selective generation.

The nodejs-server codegen generates two files per API. For the example of petstore, it generates files like `controllers/Pet.js` and `controllers/PetService.js` (see https://github.com/swagger-api/swagger-codegen/tree/master/samples/server/petstore/nodejs/controllers).

When this `noservice` is specified, it omits the generation of `PetService.js` file or any other -Service.js files. Only `Pet.js` or those files.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NodeJSServerCodegen.java
jmuk: Great idea. Done.  Also added a section in README.mustache template, so people can refer to the URL after the code is generated.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NodeJSServerCodegen.java
wing328: @jmuk thanks for the clarification. I'll review and let you know if I've further question.

wing328:
@jmuk PR merged into master. Thanks for your contribution.

wing328:
FYI. I merged https://github.com/swagger-api/swagger-codegen/pull/4411 into master to improve the NodeJS code quality a bit.

modules/swagger-codegen/src/main/resources/nodejs-google-cloud-functions/index.mustache
```diff
@@ -0,0 +1,44 @@
+'use strict';
```
wing328: @jmuk we'll move this file to nodejs/google-cloud-functions/index.mustache (and update the code accordingly). so that all nodejs (server) related files are under the folder.

Are you OK with the proposed change?

jmuk: Sure, I don't care the exact file location as long as the code works. But thank you for telling me!

wing328: I found that this file is the same as `index-gcf.mustache` so I simply removed it via https://github.com/swagger-api/swagger-codegen/pull/4808. I did some tests and the generator is still working properly with or without the option for Google cloud function.

PR #4417, opened 2016-12-17
Additional parameters to test additional cases.

antihax:
### PR checklist

- [!!!] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)

### Description of the PR
Adds some observed cases that cause compile errors on some clients.

1. required queryParam of array integers 
2. optional queryParam of array integers 
3. optional bodyParam 

This will legitimately break some of the affected clients.

- Go failed: PR #4415 fixes.
- Java feign passed.

### DIFF
--- a/modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml
+++ b/modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml
@@ -559,7 +559,29 @@ paths:
           description: Invalid username supplied
         '404':
           description: User not found
-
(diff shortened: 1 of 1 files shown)


wing328:
@antihax thanks for adding more test cases. I think what I will do is to cherry-pick the 1st commit as not all generators are ready for the new test cases.

antihax:
I was trying to determine the best way to handle this since I do not use many of the languages it would break. Cherry picking sounds good.

Issue #4420, opened 2016-12-18
isFile, isPrimitiveType should not be both set to true


wing328:
##### Description
`isFile`, `isPrimitiveType` should not be both set to true

```
{
        "isFormParam" : true,
        "secondaryParam" : true,
        "isPrimitiveType" : true,
        "baseName" : "file",
        "paramName" : "file",
        "dataType" : "file",
        "description" : "file to upload",
        "unescapedDescription" : "file to upload",
        "example" : "'/path/to/file.txt'",
        "jsonSchema" : "{\n  \"name\" : \"file\",\n  \"in\" : \"formData\",\n  \"description\" : \"file to upload\",\n  \"required\" : false,\n  \"type\" : \"file\"\n}",
        "isFile" : true,
        "isEnum" : false,
        "vendorExtensions" : { }
      }
```
##### Swagger-codegen version

Latest master

##### Command line used for generation
```
java -DdebugOperations=true -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate \
  -i modules/swagger-codegen/src/test/resources/2_0/petstore.json \
  -l python \
  -o /var/tmp/python/is_file/ | less
```
##### Suggest a Fix

The behaviour has changed recently (likely due to a PR) and we'll need to restore the old behaviour to set `isPrimitiveType` to false if `isFile` is true.

PR #4421, opened 2016-12-18
Fix `isPrimitiveType` for file property

wing328:
### PR checklist

### Description of the PR

To fix https://github.com/swagger-api/swagger-codegen/issues/4420

cc @www2k


modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
```diff
@@ -3309,7 +3309,8 @@ public void setParameterBooleanFlagWithCodegenProperty(CodegenParameter paramete
             parameter.isPrimitiveType = true;
         } else if (Boolean.TRUE.equals(property.isFile)) {
             parameter.isFile = true;
-            parameter.isPrimitiveType = true;
+            // file is *not* a primitive type
```
www2k: I am confused with the term `primitive`. (I did not notice `additional`.)

> An additional primitive data type "file" is used by the Parameter Object and the Response Object to set the parameter type or the response as being a file.

from http://swagger.io/specification/#data-types-12

To clarify, I want to know whether `additional primitive data type` is `isPrimitiveType` or not.

wing328: > To clarify, I want to know whether additional primitive data type is isPrimitiveType or not.

@www2k we've been treating `isFile` as a non-primitive type (which means `isPrimitiveType` set to false). It's definitely arguable as you've pointed out the following:
```
An additional primitive data type "file" is used by the Parameter Object and the Response Object to set the parameter type or the response as being a file.
```
What I noticed is that some mustache values rely on `isFile` set to true and `isPrimitiveType` set to false. My take is to keep the old logic and to revise it with a valid use case in `2.3.0` branch.

Does your enhancement rely on both `isFile` and `isPrimitiveType` set to true?

PR #4434, opened 2016-12-19
Update DefaultGenerator.java to call close()

chasetec:
Need to call out.close() after IOUtils.copy(in, out); when writing supporting files since IOUtils does not close the stream.

I'm using Java code generation in-process and I need to delete the generated files within the same VM they were created in. Since the streams are not closed I am unable to delete gradle\wrapper\gradle-wrapper.jar. Issue seen on a Windows 7 x64 system with Java 1.8.0_102.


wing328:
@chasetec thanks for the PR, which has been merged into master.

PR #4435, opened 2016-12-19
2.3.0 tizen patch

leaxdc:
patch for cpprest backend
-several bug fixes
-new accurate code to check MIME types
-don't replace * to _* in MIME types

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.iml
+generated/
 out/
 *.ipr
 *.iws
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 *.iml
 generated/
-generated_old/
 out/
 *.ipr
 *.iws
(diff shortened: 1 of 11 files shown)


.gitignore
```diff
@@ -1,4 +1,5 @@
 *.iml
+generated/
```
wing328: API client samples should be put under `./samples/client/petstore` as an example.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CppRestClientCodegen.java
```diff
@@ -391,7 +389,8 @@ public String escapeQuotationMark(String input) {
 
     @Override
     public String escapeUnsafeCharacters(String input) {
-        return input.replace("*/", "*_/").replace("/*", "/_*");
+        // escape nothing for CppRest
```
wing328: We need this to prevent code injection and the latest master should not escape `*/*`

wing328:
@leaxdc may I know if you've a chance to review my feedbacks?

wing328:

Issue #4360, opened 2016-12-09
[NODE-SERVER] Integer stub value not generated if format not defined


crazyfrozenpenguin:
##### Description

Integer swagger field definitions not generating value if format not defined. Should set value to an integer value (1 or 0).

##### Swagger-codegen version

2.2.1

##### Swagger declaration file content or url

The following definition does not generate an integer value:

```json
  "id":{
    "readOnly":true,
    "description":"Internal id.  This field is read only.",
    "type":"integer"
  },
```

Adding "format":"int64" to swagger definition makes codegen set the value.

##### Command line used for generation

java -jar swagger-codegen-cli.jar generate -i swagger.json -l nodejs-server -o stub

##### Steps to reproduce

Create a swagger endpoint with  "type":"integer" field. Generate stub using command above.

##### Related issues

None found.

##### Suggest a Fix

"type":"integer" should be sufficient for codegen to set the field with a stub value of either 1 or 0.

wing328:
@crazyfrozenpenguin do you mind pulling the latest and build the JAR locally to give it another try as the latest master uses a newer version of Swagger Parser?

crazyfrozenpenguin:
@wing328, no, this is still an issue if the definition does not include a format.

wing328:
@crazyfrozenpenguin thanks for performing more tests to confirm.

Please add the `format` as a workaround for the time being while we (the community) look into fixing the issue.

fultonm:
I am fixing this

fultonm:
I am attempting to reproduce this error with the PetStore YAML example. I have added an integer parameter to a path without specifying the format. PetsService.js never sets such an integer to anything.
`  /**
   * parameters expected in the args:
  * petId (String)
  * coolness (Integer)
  **/`

crazyfrozenpenguin:
@fultonm, that is the problem, that it never sets the value to anything. Its an integer and as such it should be enough to set it to at least "1".

fultonm:
In the PetsService.js the value coming in is a parameter with its value presumably being set in the query or path on the client side. Especially if the field is required, it is sure the be the correct type checked by swagger-tools.

You mentioned if you specify the format of the integer to be int32 or int64 that it will initialize the variable? In which file do you see the initialization happening?

crazyfrozenpenguin:
I run it on my own swagger file. Let me have a look if it is reproducible in the PetService.js. Give me a few moments.

[...]

PR #4436, opened 2016-12-19
Fix Integer stub value defaulting to String in nodejs-server if format not specified

fultonm:
fixes issue [[NODE-SERVER] Integer stub value not generated if format not defined](https://github.com/swagger-api/swagger-codegen/issues/4360)

### PR checklist

### Description of the PR

Fixed an issue where properties specifying Integer as the type but not specifying a format (int32 or int64) would default to String when generated as nodejs-server.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/examples/ExampleGenerator.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/examples/ExampleGenerator.java
@@ -97,7 +97,7 @@ public class ExampleGenerator {
             return "2000-01-23";
         } else if (property instanceof DateTimeProperty) {
             return "2000-01-23T04:56:07.000+00:00";
-        } else if (property instanceof DecimalProperty) {
+        }  else if (property instanceof DecimalProperty) {
             return new BigDecimal(1.3579);
         } else if (property instanceof DoubleProperty) {
             return 3.149;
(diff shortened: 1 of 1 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/examples/ExampleGenerator.java
```diff
@@ -106,6 +107,8 @@ protected Object resolvePropertyToExample(String mediaType, Property property, S
             return 1.23f;
         } else if (property instanceof IntegerProperty) {
             return 123;
+        } else if (property.getClass().equals(BaseIntegerProperty.class)) {
```
fultonm: If we use `instanceof` it will return true for `LongProperty` and `IntegerProperty` which inherit from `BaseIntegerProperty`. I don't know if it matters specifically for nodejs-server, but swagger-parser makes the distinction and this change affects all languages. Ints without a format are passed from swagger-parser as BaseInetgerProperties and nothing more.

wing328: Other generators have similar fix in place already: https://github.com/swagger-api/swagger-codegen/blob/f70569987a8545b89cfdbb85b2edf1ca8c529f4c/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Qt5CPPGenerator.java#L275-L282

fultonm: I see. Languages may have different needs when it comes to default values. `NodeJSServerCodegen` does not override the `toDefaultValue` method.

Do you think it is enough to have this extra case in `ExampleGenerator`? I think `toDefaultValue` in `DefaultCodegen` should become abstract so each language must implement their own default values.

wing328: @fultonm eventually we want to make `toDefaultValue` abstract so that all language generators must implement it as the default might not work for the language. Before that, we'll need to add `toDefaultValue` in each language generator, which will probably take quite amount of work as the project supports many generators.

wing328:
@fultonm thanks for the contribution. PR merged into master.

PR #4439, opened 2016-12-20
Allows for generation of spring controller code using the delegate pattern

JLLeitschuh:
### PR checklist

### Description of the PR

Adds the ability to generate Spring Server code using the decorator pattern.

Because annotations on interface parameters aren't inherited in java and aren't accessible using reflection spring requires you to have annotations like `@RequestBody` on the concrete class.

This means that as a developer you can not just override the methods in your controller, you must also copy the annotations from the `Api` interface's methods.

This fixes the problem by allowing you to generate interfaces that can be implemented and don't require you to copy the annotations from the parent.

Note: It seems like someone didn't generate the samples for one of their changes so there is a bit more noise in those files than this PR should create for consumers.


JLLeitschuh:
Some checkstyle things that clearly need to be cleaned up

modules/swagger-codegen/src/main/resources/JavaSpring/apiDecorator.mustache
```diff
@@ -0,0 +1,38 @@
+package {{package}};
+
+{{#imports}}import {{import}};
+{{/imports}}
+
+import io.swagger.annotations.*;
+{{#jdk8}}
+    import org.springframework.http.HttpStatus;
```
wing328: Minor suggestion: what about removing the leading spaces?

wing328:
(the Appveyor issue can be ignored)

modules/swagger-codegen/src/main/resources/JavaSpring/apiDecorator.mustache
JLLeitschuh: Can do!

JLLeitschuh:
Fixed. Other than that is this good?

wing328:
@JLLeitschuh I did some tests using the following commands:
```
  -i modules/swagger-codegen/src/test/resources/2_0/petstore.yaml \
  -l spring  -DdecoratorPattern=false \
  -o /var/tmp/spring/decoratorPattern_false/
```

```
  -i modules/swagger-codegen/src/test/resources/2_0/petstore.yaml \
  -l spring  -DdecoratorPattern=true \
  -o /var/tmp/spring/decoratorPattern_true/
```

Both failed to compile with different error messages (while the generated code without `decoratorPattern` works fine without issue)

Can you give it a try to see if you can repeat the issue?

JLLeitschuh:
You're right, I had some logic that was wrong.
Good catch.
Those should work now.

cbornet:
I think this is not a decorator pattern. It looks more like a delegation.
Also, shouldn't there be an implementation of the delegation interface generated by default ? For instance, doesn't it crash if you don't provide a PetApiDelegator bean ?

JLLeitschuh:
I suppose that you are right, this may be more like delegation.
I was expecting the users to implement the interfaces. (That's what I plan to do). Also, one reason for using interfaces is because now users are free to use one class to implement many controllers or can use one class per controller.
I don't really want to also generate concrete services.

cbornet:
Well, I think the generated server should at least start without crashing and provide basic "200 OK" endpoints. After that you can add the implementations to swagger-ignore so that they are not overridden by further regen.

JLLeitschuh:
So you are saying add an additional flag flag with something like `-DdecoratorPattern=true -DdecoratorPatternImplementations=false` to get what functionality I want and `-DdecoratorPattern=true -DdecoratorPatternImplementations=true` to get what you want?

I mean, I can understand the desire but I don't really want to implement this as I have no use for it.
You are more than welcome to open a PR against this branch to add that feature if you want and I'll merge it into this one.

JLLeitschuh:
@wing328 Are you looking for anything else on this PR?

wing328:
@JLLeitschuh I'll do more tests later today. Will keep you posted.

cbornet:
@JLLeitschuh can you rename the class/option xxxDecorator to xxxService ?
As for the option, I wouldn't have too many possibilities and always generate the concrete class. Don't you need at least one concrete class yourself ?

wing328:
@JLLeitschuh Performed more tests and didn't find any issues.

Please kindly reply to @cbornet 's suggestions: https://github.com/swagger-api/swagger-codegen/pull/4439#issuecomment-273032264

JLLeitschuh:
Yea, but I want to be free to implement it as I want, perhaps I want to create a class that implements two of the controller interfaces? Maybe 4? I don't want another concrete implementation on the classpath that I have to fight spring to get it to not autowire.
I'm using the generated code as a source set for a larger project.

JLLeitschuh:
I can do the renaming of `xxxDecorator` to `xxxService`

JLLeitschuh:
Wouldn't `xxxDelegate` make more sense @cbornet?

JLLeitschuh:
I just changed everything to use the word `delegate` instead.
I also added a sample with the delegate pattern combined with java 8

cbornet:
I feel service would be better in the Spring world since you would probably use a component annotated by @Service

wing328:
@JLLeitschuh thanks for creating the Petstore sample test for Java8 but I got compilation errors when running `mvn test` under ``. Here is part of the error message:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project swagger-spring: Compilation failure: Compilation failure:
[ERROR] /private/var/tmp/pr/JLLeitschuh-feat/swagger-codegen/samples/server/petstore/springboot-delegate-j8/src/main/java/io/swagger/api/PetApiController.java:[17,137] cannot find symbol
[ERROR] symbol:   class Pet
[ERROR] location: class io.swagger.api.PetApiController
[ERROR] /private/var/tmp/pr/JLLeitschuh-feat/swagger-codegen/samples/server/petstore/springboot-delegate-j8/src/main/java/io/swagger/api/PetApiController.java:[17,12] cannot find symbol
[ERROR] symbol:   class ResponseEntity
[ERROR] location: class io.swagger.api.PetApiController
[ERROR] /private/var/tmp/pr/JLLeitschuh-feat/swagger-codegen/samples/server/petstore/springboot-delegate-j8/src/main/java/io/swagger/api/PetApiController.java:[22,12] cannot find symbol
[ERROR] symbol:   class ResponseEntity
[ERROR] location: class io.swagger.api.PetApiController
[ERROR] /private/var/tmp/pr/JLLeitschuh-feat/swagger-codegen/samples/server/petstore/springboot-delegate-j8/src/main/java/io/swagger/api/PetApiController.java:[28,240] cannot find symbol
[ERROR] symbol:   class List
```

Did you get similar error message when running `mvn test` locally in your machine?

samples/server/petstore/spring-mvc-j8-async/src/main/java/io/swagger/api/FakeApiController.java
```diff
@@ -1,10 +1,80 @@
...
+import java.util.concurrent.Callable;
 
 
 
 @Controller
 public class FakeApiController implements FakeApi {
 
+
+
+    public Callable<ResponseEntity<Client>> testClientModel(@ApiParam(value = "client model" ,required=true ) @RequestBody Client body) {
```
JLLeitschuh: This should not be being generated. I need to fix this.

JLLeitschuh:
@wing328
You were right. My logic was still wrong.
I didn't understand that even in mustache when a value is `false` `#[value]` still evaluates to true because the value exists.
I fixed that issue.
That made the diff against the other repos smaller.

wing328:
@JLLeitschuh I did more tests and no longer found issues. Thanks for your contribution.

JLLeitschuh:
Thank you @wing328 Awesome work on this project.
Any ETA on the next release?
I'll continue to use my internally published version for now but I'd love to not have to rely on it for too long.

wing328:
ETA for 2.2.2 release is Jan/Feb 2017

cbornet:
No 2.3 ?

wing328:
@cbornet yup, I also want to release 2.3.0 right after 2.2.2.

JLLeitschuh:
I think I may have accidentally forgotten to change the `rm` lines of the scrips when I copied them.
I'll open another PR later tomorrow if I have time

bin/spring-delegate-j8.sh
```diff
@@ -0,0 +1,34 @@
...
+then
+  mvn clean package
+fi
+
+# if you've executed sbt assembly previously it will use that instead.
+export JAVA_OPTS="${JAVA_OPTS} -XX:MaxPermSize=256M -Xmx1024M -DloggerPath=conf/log4j.properties"
+ags="$@ generate -t modules/swagger-codegen/src/main/resources/JavaSpring -i modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml -l spring -o samples/server/petstore/springboot-delegate-j8 -DdelegatePattern=true,hideGenerationTimestamp=true,java8=true"
+
+echo "Removing files and folders under samples/server/petstore/springboot-delegate-j8/src/main"
+rm -rf samples/server/petstore/springboot/src/main
```
JLLeitschuh: This is wrong. Should be the same as the line above.

bin/spring-delegate.sh
```diff
@@ -0,0 +1,34 @@
...
+then
+  mvn clean package
+fi
+
+# if you've executed sbt assembly previously it will use that instead.
+export JAVA_OPTS="${JAVA_OPTS} -XX:MaxPermSize=256M -Xmx1024M -DloggerPath=conf/log4j.properties"
+ags="$@ generate -t modules/swagger-codegen/src/main/resources/JavaSpring -i modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml -l spring -o samples/server/petstore/springboot-delegate -DdelegatePattern=true,hideGenerationTimestamp=true"
+
+echo "Removing files and folders under samples/server/petstore/springboot-delegate/src/main"
+rm -rf samples/server/petstore/springboot/src/main
```
JLLeitschuh: This is wrong. Should be the same as the line above.

PR #4445, opened 2016-12-21
Added missing package name for Configuration in api.mustache.

gdurand:
This is needed when the package contains "Configuration", e.g. MyApp.Configuration.Network.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/csharp/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/csharp/api.mustache
@@ -83,7 +83,7 @@ namespace {{packageName}}.{{apiPackage}}
         /// <returns></returns>
         public {{classname}}(String basePath)
         {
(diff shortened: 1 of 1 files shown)


wing328:
@gdurand thanks for the PR. The following line imports `{{pacakgeName}}.Client`:

wing328:
@gdurand closing this as there's no update from you. Let us know if you still want this PR to be merged into master.

gdurand:
Sorry for the delay, I was on holidays.
The line you refer to was not removed.
If packageName contains "Configuration", the compiler cannot determine if the class Configuration is a class or a package, despite the import. It is what I observe with VS2015, and this kind of thing is a common PITA with C# when a class has the same name as a package.
So I think this PR is valid.
BTW, I think all class names should be fully qualified to avoid such naming problem.

PR #4450, opened 2016-12-22
Support the empty string as an enumeration value

kevinoid:
When a string enumeration has the empty string as one of its available values, the generated code for many languages is invalid because the empty string can not be used as an identifier.  As with numbers and symbols, provide a mapping to an English name which can be used as a replacement.  In this case, "empty" for the empty string/empty value.

**Reviewer Note:** I have updated all of the languages that I could find and updated the samples by running `./bin/run-all-petstore` and everything looks good.  However, I have been unable to get Travis CI to run on my repo due to [`nexus.codehaus.org: Unknown host nexus.codehaus.org](https://travis-ci.org/kevinoid/swagger-codegen/jobs/186034333) and I was unable to run the integration tests on my machine.  So some additional sanity checking would be greatly appreciated.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -303,6 +303,10 @@ public class DefaultCodegen {
      * @return the sanitized variable name for enum
      */
     public String toEnumVarName(String value, String datatype) {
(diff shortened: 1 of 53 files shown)


wing328:
@kevinoid thanks for the contribution to cover this additional edge case. The change looks good to me.

kevinoid:
@wing328 Great.  Thanks for reviewing and merging it!

PR #4458, opened 2016-12-23
[JAX-RS/jersey2] Fix for incorrect JSON field name capitalization

y-higuchi:
### PR checklist

### Description of the PR

JSON representation of model object, which had field with all capitals, resulted in same field content getting outputted twice in 2 different capitalization form.  

This patch add `@JsonProperty` annotation to the template for getter to avoid them.  
Also added model object to `petstore-with-fake-endpoints-models-for-testing.yaml`
and unit test `CapitalizationTest.java` to high light the issue.  

I should've searched thoroughly before working on it, 
but I've found #4066 which seems to be reporting same problem, 
which proposes a fix in a different way.  

Since this effective part of this patch is template only, 
I'm using this version of template with swagger-codegen-maven-plugin 2.2.1, 
specifying templateDirectory and additional-properties, jackson=true to configOptions.  
So I thought it might be worth submitting PR anyways.


wing328:
@y-higuchi thanks for the fix and the additional test cases to cover it moving forward.

Issue #4234, opened 2016-11-22
Java DateTimeTypeAdapter class is not using the right formatter.


junewoo-suk:
##### Description

DateTimeTypeAdapter class has a variable `private final DateTimeFormatter formatter = ISODateTimeFormat.dateTime();` but it should be `private final DateTimeFormatter formatter = ISODateTimeFormat.dateTimeParser();`. For example, dates like `yyyy-MM-ddTHH:mm:ssZ` will not be parsable in the current version, but it will be possible with dateTimeParser function. The previous date format is a default date format from InnoDB passed to Django models

##### Swagger-codegen version
2.2.1

##### Suggest a Fix
DateTimeTypeAdapter.java
`private final DateTimeFormatter formatter = ISODateTimeFormat.dateTime();`
to
`private final DateTimeFormatter formatter = ISODateTimeFormat.dateTimeParser();`

junewoo-suk:
right now work around is using `setDateFormat`

cbornet:
Yes, but:
* It would rather be dateOptionalTimeParser()
* dateTime() should probably still be used for serializing (as the time is not optional in rfc3339)

cbornet:
Fixed by #4473

wing328:
@junewoo-suk please pull the latest master to give it a try.

Thanks @kevinoid for the fix.

(we've just added Swagger Codegen to ProductHunt: https://www.producthunt.com/posts/swagger-codegen, pleae upvote when you've time)

PR #4473, opened 2016-12-30
Fix Gson parsing of Joda DateTime without millis

kevinoid:
### PR checklist

### Description of the PR

The `DateTimeFormatter` returned by `ISODateTimeFormat.dateTime()` only parses dates with millisecond values, and throws `IllegalArgumentException` when milliseconds are not present.  The
date-time construct from RFC 3339 Section 5.6 referenced by the Swagger/OpenAPI spec allows fractional second values to be omitted.  This results in valid date-time values being rejected by the generated code.

This PR fixes the problem by using `ISODateTimeFormat.dateTimeParser()` for parsing, which correctly handles date-time values without fractional seconds.  Note that `.dateTime()` must
still be used for printing, which is not supported by `.dateTimeParser()`.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/JSON.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/JSON.mustache
@@ -172,14 +172,15 @@ class DateAdapter implements JsonSerializer<Date>, JsonDeserializer<Date> {
  */
 class DateTimeTypeAdapter extends TypeAdapter<DateTime> {
 
-    private final DateTimeFormatter formatter = ISODateTimeFormat.dateTime();
+    private final DateTimeFormatter parseFormatter = ISODateTimeFormat.dateOptionalTimeParser();
+    private final DateTimeFormatter printFormatter = ISODateTimeFormat.dateTime();
 
     @Override
     public void write(JsonWriter out, DateTime date) throws IOException {
         if (date == null) {
             out.nullValue();
         } else {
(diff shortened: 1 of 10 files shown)


cbornet:
Yes but dateOptionalTimeParser should be used instead.
Related to #4234

kevinoid:
@cbornet Why `dateOptionalTimeParser` instead of `dateTimeParser`?  The `date-time` format references the `date-time` construct from RFC 3339, which requires a time part.  The `date` format, which does not have a time part, appears to be parsed by `ISODateTimeFormat.date()` in `LocalDateTypeAdapter` which I haven't changed.

kevinoid:
@cbornet I took a closer look at the grammars accepted by `dateOptionalTimeParser` and `dateTimeParser`.  I had misread the documentation.  It looks like the only difference is `dateOptionalTimeParser` does not accept a time without a date.  I'll update the PR to use `dateOptionalTimeParser` shortly.

kevinoid:
@cbornet I noticed that `dateTimeParser` is currently used by the akka-scala generator.  Should I change that to `dateOptionalTimeParser` as well?

kevinoid:
PR updated and CI passed.  I included the analogous akka-scala change as a separate commit that can easily be dropped if it should be handled separately.

cbornet:
👍

wing328:
@kevinoid thanks for the contribution, especially to the `akka-scala` generator.

kevinoid:
@wing328 Thanks for reviewing and merging it!

PR #4480, opened 2017-01-03
Issue 4416 (Assist escapeReservedWord with custom mappings)

alexandern:
### PR checklist

### Description of the PR

The escapeReservedWord  method does its job fairly well, but in some cases  "reserved words" make some properties look a bit odd in some languages. 

The method escapeReservedWord could be assisted through an additional command line parameter (reserved-words-mappings).

In the case of Objective-C, the definition below would create a property named _id which could mistakenly be interpreted as an iVar.

```
definitions:
  Pet:
    required:
      - id
    properties:
      id:
        type: integer
        format: int64
     name: 
        type: string
```

which in turn yields

```
@interface SWGPet : SWGObject
@property(nonatomic) NSNumber* _id;
@property(nonatomic) NSString* name;
@end
```
The escapeReservedWord method could look up what _id should be substituted with instead.

For example,
`java -jar swagger-codegen-cli.jar generate -i pet.yaml -l objc --reserved-words-mappings id=identifier -o output`

The new output

```
@interface SWGPet : SWGObject
@property(nonatomic) NSNumber* identifier;
@property(nonatomic) NSString* name;
@property(nonatomic) NSString* tag;
@end
```

GoLang can also benefit from this, for example
```
  Car:
    required:
      - id
      - name
      - range
    properties:
      range:
        type: array
        items:  
          type: integer
      id:
        type: integer
        format: int64
      name:
        type: string
```

Instead of this
```
type Car struct {
	Range_ []int32 `json:"range,omit empty"
	Id int64 `json:"id,omit empty"
	Name string `json:"name,omit empty"
}
``````
It could generate

```
type Car struct {
	Ranges []int32 `json:"range,omit empty"
	Id int64 `json:"id,omit empty"
	Name string `json:"name,omit empty"
}
``````
by using
`java -jar swagger-codegen-cli.jar generate -i car.yaml -l go --reserved-words-mappings Range=Ranges -o output`


wing328:
@alexandern thanks for the enhancement. I'm sure other Swagger Codegen users will find this new feature useful in customizing the output.

Issue #2330, opened 2016-03-07
[golang] Default to pointers for non-primitive types


treeder:
It generates code like this currently:

```go
type JobWrapper struct {
    Job  Job  `json:"job,omitempty"`
}
```

Same with slices and what not. 

It would be nice (and probably what most people would want) to have it use pointers, eg:

```go
type JobWrapper struct {
    Job *Job  `json:"job,omitempty"`
}
```

wing328:
@treeder thanks for the feedback. I think we need to provide an option here or use vendor extension to determine using pointer for model property or not.

When I created the Go API client generator, I looked to Stripe, Twilio GO API client (and a couple more) and seems like it really depends on use case. For example, in https://github.com/stripe/stripe-go/blob/master/order.go, you will find some properties refer to the actual model while the others use model pointers.

treeder:
Interesting. Seems pretty random there, although I'm sure it's not. They may have wanted the ability to check for nil's on certain fields rather than the zero value for them seeing as how they even have pointers on primitives. I might argue that pointers for everything (including primitives) is best for an API so you can check for nil's on every field. Without pointers, you don't know whether the user passed in an empty string or nothing for string a field, or if they passed in a 0 or nothing for a number field, etc.

neilotoole:
I recently ran into this issue with a [PR](https://github.com/google/go-github/pull/325) I created for go-github client. I initially implemented the datatypes without pointers, but there's a valid reason the Google guys are using pointers: sometimes you want to send the zero value (e.g. an empty string) back to the API, and sometimes you want to omit a field entirely. Pointers allow you to do this.

However, and this is a big design issue, I don't see any mechanism (I'm still a Go n00b though!) wherein swagger-codegen can support polymorphic return types if the generated code is returning a plain struct (pointers or not). I think we'll have to return an interface type, which can then be cast (by the caller) to the desired sub-type based upon the discriminator field (or using the Go type interrogation mechanisms). If we decided to go with interfaces, we could also provide getter/setter/isSet functions which would allow the user to determine if a field is really present vs just being at the zero value, and also allow the generated client code to determine whether or not a field should be sent or omitted when communicating with the endpoint.

This would likely break any existing code that relies upon the beta Go code generator. But I guess that's why it's a beta...

wing328:
@neilotoole thanks for the feedback on this.

I'd a look at https://github.com/google/go-github/ and found the following:

[...]

PR #4488, opened 2017-01-04

wy-z:
### PR checklist

- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/go-petstore.sh` and `./bin/security/go-petstore.sh` if updating the go (e.g. php, ruby, python, etc) code generator or go client's mustache templates)

### Description of the PR

- Use pointers to avoid circular references
- Abandon `omitempty` tag when the field is required


wing328:
@wy-z thanks for the PR but I think it should be filed against 2.3.0 as it's a breaking change.

Here is the related discussion: https://github.com/swagger-api/swagger-codegen/issues/2330

cc @neilotoole @guohuang @antihax

PR #4489, opened 2017-01-04
[html2] Open the description field that contains the description

frodrigo:
Open the description field that contains the description and required switch, in place of only show the field type by default.


wing328:
@frodrigo thanks for your PR, which has been merged into master.

Issue #4481, opened 2017-01-03
[html2] produces/consumes headers missing from curl example


bshamblen:
The necessary headers are not included in the curl example if the either the `consumes` or `produces` option is set, either globally or for the specific path.

For example, a POST endpoint that expects a JSON body and returns a JSON response should look like this: 

```cli
curl -X POST -H "api-key: [[apiKey]]" -H "Accept: application/json" -H "Content-Type: application/json" -H "Authorization: Basic [[basicHash]]" "https://www.example.com/v1/some-action"
```

... but it currently looks like this:

```cli
curl -X POST -H "api-key: [[apiKey]]" -H "Authorization: Basic [[basicHash]]" "https://www.example.com/v1/some-action"
```

I already have a fix ready to submit, but want to make sure you agree with this before I submit the PR.

wing328:
@bshamblen I agree with you that the curl command needs to include the `Content-Type` and/or `Accept` as defined in `consumes`, `produces` via `-H` command line switch in `curl`.

PR #4493, opened 2017-01-04
[html2] Add curl headers for consumes/produces options

bshamblen:
Fixes issue #4481


wing328:
@bshamblen looks good to me. Thanks for the enhancement.

PR #4504, opened 2017-01-06

wy-z:
### PR checklist

### Description of the PR

- Use pointers to avoid circular references
- Abandon `omitempty` tag when the field is required


wy-z:
@wing328 It seems that #4440 has not modified `model.mustache`, so this PR may be useful~

wy-z:
@wing328 After reading #2330 carefully, I found that the problem was more complex than I thought~
I closed the PR.

Issue #4511, opened 2017-01-06
[PHP] API classes override default config unnecessarily


toomuchpete:
##### Description

PHP Base API classes call `setPath()` on the ApiServer by default. Here's a sample generated constructor:

```php
    public function __construct(\SwaggerUserService\ApiClient $apiClient = null)
    {
        if ($apiClient == null) {
            $apiClient = new ApiClient();
            $apiClient->getConfig()->setHost('http://localhost/v1');
        }

        $this->apiClient = $apiClient;
    }
```

Calling `setHost()` is unnecessary because ApiClient() constructor configures itself based on the default. A related issue is that the code assumes (and hardcodes) `localhost` as the host when one isn't provided. 

This combination makes wrapping the Swagger-generated client in our host-discovery code unnecessarily complex.

##### Swagger-codegen version

Unsure what version is being used, not sure where to find that. I installed codegen through Homebrew. Looking at the code in master, though, this appears to be an issue there as well.

##### Swagger declaration file content or url

https://gist.github.com/toomuchpete/38ad55e570e103954d404c5df73c0e00

##### Command line used for generation

```bash
swagger-codegen generate -l php -i swagger.yaml
```

##### Related issues

None found.

##### Suggest a Fix

Remove this line: https://github.com/swagger-api/swagger-codegen/blob/1ea9865a4498e26bb1f4ecf490aaad19d19357a8/modules/swagger-codegen/src/main/resources/php/api.mustache#L52

wing328:
@toomuchpete thanks for reporting the issue and your suggested fix looks good to me.

Please kindly file a PR so that we can review the change.

toomuchpete:
@wing328 -- I don't want to mess with the workflow for this issue, if one exists, but the fix for this issue has been merged in. Should this ticket be closed now?

wing328:
@toomuchpete I just closed it. Thanks for the reminder.

Issue #4518, opened 2017-01-06
[PHP] lib/Api constructors ignore configuration and set host to localhost 


philip:
##### Description

The constructors within `lib/Api/*.php` ignore the host definition in `lib/Configuration.php`, whether it's set within `lib/Configuration.php` itself or by using `Swagger\Client\Configuration::getDefaultConfiguration()->setHost(https://example.org/);` in the code. 

Either way, the likes of `$foo = new Swagger\Client\Api\Bar()` always resets host to `https://localhost`. 

##### Swagger-codegen version

2.2.1

##### Command line used for generation
```
swagger-codegen generate \
 --input-spec ./swagger.json \
 --lang php \
 --output /tmp/foo
```
##### Steps to reproduce

1. Use swagger_codegen to generate client, see above
2. Edit `lib/Configuration.php` by setting the `$host` there
3. Copy and edit example code from generated README.md
4. Optionally set host in code using `Swagger\Client\Configuration::getDefaultConfiguration()->setHost('https://example.org/);` -- generated code already does this for other stuff like username / password, which is why it's expected that this would work fine
5. Execute
6. Fails, as it attempts to connect to localhost instead of my desired host

##### Suggest a Fix

Code in the generated constructors looks like this:

```php
        if ($apiClient == null) {
            $apiClient = new ApiClient();
            $apiClient->getConfig()->setHost('https://localhost');
        }
```

One option is to change it like so:

```php
        if ($apiClient == null) {
            $apiClient = new ApiClient();
            if (empty($apiClient->getConfig()->getHost())) {
                $apiClient->getConfig()->setHost('https://localhost');
            }
        }
```

With that change, host can be set in `lib/Configuration.php` or within the code itself via the aforementioned `setHost()`, and not reset to localhost.

That said, I'm not familiar enough with the codebase to know if that's ideal, but it's working for me. But, defining the default host like this in multiple files feels odd so another solution is to define it somewhere else (like `lib/Configuration.php`) instead of within each constructor.

philip:
Forgot to mention, an approach people must do now is pass in an API client, so something like:

```
$api_client = new Swagger\Client\ApiClient();
$foo = new Swagger\Client\Api\Bar($api_client);
```

That is a required step if you're not using `https://localhost`. I don't think it should be as the above code isn't much different if you're wanting to rely on `lib/Configuration.php` instead of passing in configuration options directly to ApiClient(). The above will use `lib/Configuration.php` values.

[...]

PR #4525, opened 2017-01-09
Remove unnecessary call to setHost() in the constructor

toomuchpete:
The default host will be automatically set on the client by the ApiClient constructor.

### PR checklist

### Description of the PR

As reported in #4511, Calling `setHost()` is unnecessary because ApiClient() constructor configures itself based on the default. This PR simply removes that line. 

Resulting clients should be equivalent at runtime.


wing328:
@toomuchpete thanks for the PR. Please also run `./bin/php-petstore.sh` to update the Petstore sample so that the change can be verified by the CIs.

wing328:
@toomuchpete you may also want to add test case(s) to cover the issue moving forward: https://github.com/swagger-api/swagger-codegen/tree/master/samples/client/petstore/php/SwaggerClient-php/tests

wing328:
cc @arnested

toomuchpete:
Hi @wing328, thanks for the feedback! I didn't include the petstore updates because they appeared to include more changes than just mine so I thought I might be doing the wrong thing.

I've pushed those changes up in two separate commits. The first contains the changes I expected to see, and the second has changes that, I think, are left over from other code changes. 

I'll look into adding a test for this situation this week. Thank you for your patience with my first contribution. 😄

wing328:
@toomuchpete your first PR definitely looks good 👍 

Please take your time and let us know if you need any help from us.

toomuchpete:
Hello again, @wing328 -- happy to report that I added a test for this situation. I wasn't sure exactly where to put it, but the `PetApiTest` file seemed to have the other tests that were generic to all API classes, so I put it in there with them. 

Let me know if there's anything else I can do to get this PR ready to be merged!

wing328:
@toomuchpete the PR with test cases look good to me. Thanks for your contribution 👍

philip:
@wing328 @toomuchpete Looks good. I filed the same bug with a different PR, but your change is simpler and much preferred. FWIW, it was Issue #4540 and PR #4518. I tested your change and it works great. You even added a test case, what a good model citizen! :) 

I also ran into the creating unrelated `php-petstore.sh` output and wondered, but it seems okay. Kind of your to commit those unrelated changes separately, though.

To summarize: now we can set host at runtime by either editing `lib/Configuration.php` or using `getDefaultConfiguration()->setHost()`, or we no longer have to instantiate `ApiClient` separately at runtime to do that. I see this as a low risk BC change, as I doubt many rely on the current behavior that sometimes ignores configuration settings. Most commonly, users that don't set host in the swagger file definition were affected by this bug.

Issue #4532, opened 2017-01-10
[Scala] Overriden objectmapper wasn't used in deserialisation


lachatak:
##### Description

Overriden objectmapper is not used in deserialisation because in the api.mustache there is a typo so there is a new ApiInvoker creation every time you deserialise a response via the companion object. The side effect is that you cannot override the object mapper in the invoker.

api.mustache
{{#returnType}} Some(ApiInvoker.deserialize(s, "{{returnContainer}}", classOf[{{returnBaseType}}]).asInstanceOf[{{returnType}}])

##### Steps to reproduce

Generate any Scala client from any swagger document. See samples..

##### Suggest a Fix

Change api.mustache to be the following
{{#returnType}} Some(apiInvoker.deserialize(s, "{{returnContainer}}", classOf[{{returnBaseType}}]).asInstanceOf[{{returnType}}])

I have a fix ready to PR.

wing328:
@lachatak thanks for reporting the issue. Please kindly submit a PR so that we can review the change.

lachatak:
PR was created:
https://github.com/swagger-api/swagger-codegen/pull/4533
I hope it is fine.

wing328:
Closed via #4534

PR #4534, opened 2017-01-10
Overriden objectmapper wasn't used in deserialisation (Scala) issue #4532

lachatak:
Use previously provided apiInvoker instead of creating a new instance at deserialisation stage
See issue #4532


wing328:
@lachatak the change looks good. Thanks for your contribution.

PR #4540, opened 2017-01-11
Always use configured host at runtime 

philip:
* If host is defined/configured at runtime, then set host to the value set 
   at runtime, otherwise use the default
* Currently the runtime value is ignored unless you do something like
  the following (instantiate the api client, empty or otherwise):
    $api_client = new Swagger\Client\ApiClient();
    $foo = new Swagger\Client\Api\Bar($api_client);
  But now all you need is:
    $foo = new Swagger\Client\Api\Bar();
* This use case is common if host is not defined in the swagger file
* This now allows the host to be set at runtime under any condition,
  for example using either:
  * lib/Configuration.php $host
  * Swagger\Client\Configuration::getDefaultConfiguration()->setHost($host)
* This could change behavior for some, as it means the configured host
is used instead of the default, unless api client is instantiated first.
Not sure why someone would do that, likely by accident.
* This resolves swagger-api/swagger-codegen#4518

### PR checklist

### Description of the PR


philip:
This was fixed in PR #4525 so closing this PR now.

PR #4545, opened 2017-01-12
When reading mustache template files assume UTF-8 encoding, fixes #4544.

### LINKED ISSUE #4544: [ALL] Unicode characters in mustache template files aren't treated correctly
##### Description
There seems to be an inconsistency to the way files are treated when they're being read (like the mustache templates) and when their being written (the processed files after applying the mustache template). When files are written back to disk **UTF-8** is enforced, but when the mustache files are first read **UTF-8** is not enforced and the default OS value is used instead, which seems to mess things up on Windows systems.
I think it would make sense to assume that the input files are **UTF-8** since we're enforcing that charset when writing files.

##### Swagger-codegen version
master @ 0fb154e9a220e3ec3358893c04dace30c312f62d

Let's take the _JavaSpring_ template, _spring-boot_ library and modify the Readme.md file to something like:
`Spring＿Boot＿Server`

`-t <new, modified template>`

##### Steps to reproduce

1. Generate code for the `JavaSpring` language, `spring-boot` library using the new, modified template.
2. The generated Readme.md file will not have the correct `Spring＿Boot＿Server` text (at least on Windows systems) as the template was read using a different charset than **UTF-8**.

eblis:
### PR checklist

### Description of the PR
Enforce **UTF-8** encoding when reading mustache template files, similar to how **UTF-8** is enforced when the resulting files are written to disk. Fixes #4544.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/AbstractGenerator.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/AbstractGenerator.java
@@ -56,7 +56,7 @@ public abstract class AbstractGenerator {
             if (is == null) {
                 is = new FileInputStream(new File(name)); // May throw but never return a null value
             }
-            return new InputStreamReader(is);
+            return new InputStreamReader(is, "UTF-8");
         } catch (Exception e) {
             LOGGER.error(e.getMessage());
         }


wing328:
@eblis thanks for the PR, which looks good to me. I'll merge it later today if no one else has further feedback on it.

wing328:
@eblis thanks for your contribution. PR merged into master.

Issue #4486, opened 2017-01-04
Nancyfx prefix interfaces with I in generated code


Razzeee:
##### Description

Not exactly a problem, but its a convention to prefix interfaces with `I` in c#

##### Swagger-codegen version

Latest master

##### Suggest a Fix

https://github.com/Razzeee/swagger-codegen/tree/nancyfx-interface
Might be considered a breaking changes, as users might be derriving from the interfaces and now need to change the name. So I can understand if you don't want to do this.

wing328:
@Razzeee it's definitely reasonable and making the change in `2.3.0` is definitely a good idea as it's a breaking change. If later more people want the interfaces named without the `I`, we can add an option as a fallback.

cc @mstefaniuk (the author of the C# Nancyfx generator)

mstefaniuk:
I'm against that convention and against that change in particular. It exposes in naming differences that are should not be exposed. As a consequence of that convention interfaces are prefixed with `I` but concrete class (first one) has no implementation indicator. It should go opposite - interfaces should have clean names but implementation should have implementation specific suffix. In Java there was also such bad convention (as adding `A` to abstracts) but it was dropped many years ago.
This idea has been discussed deeply in _Clean code_ of Robert Martin and also pops out on SO sites http://softwareengineering.stackexchange.com/questions/117348/should-interface-names-begin-with-an-i-prefix and http://stackoverflow.com/questions/5816951/prefixing-interfaces-with-i.

Razzeee:
As much as I want to go with you on this, the one comment on SO pretty much summed it up for me:

> All of this being said, I tend to prefix my C# interfaces with I anyway because that is the .NET convention used and advocated by Microsoft. And when I'm explaining coding conventions to a new developer, it's far less hassle to just use Microsoft's rules than to explain why we have our own "special" rules.

wing328:
What about adding an option (default to false) to prefix the interface with `I` so as to accommodate difference developer style/preference?

cc @jimschubert

jimschubert:
I'm all for this change, even if it means using an option that defaults to the I prefix.

Microsoft's [Framework Design Guidelines](https://msdn.microsoft.com/en-us/library/ms229040(v=vs.110).aspx) suggest: 
--- 
✓ DO prefix interface names with the letter I, to indicate that the type is an interface.

[...]

PR #4557, opened 2017-01-15
[nancyfx/csharp] Customize interface prefix

jimschubert:
### PR checklist

### Description of the PR

Per #4486, this allows user to specify the use of a standard or custom
prefix for interfaces. For C# based languages, this follows Microsoft's
Framework Design Guidelines and uses an I- prefix. However, to avoid
breaking changes with existing nancyfx generated code, the default is
unset.

The option supports true, false, or a custom prefix.

To test:

**NancyFX**:

```
./bin/nancyfx-petstore-server.sh --additional-properties interfacePrefix=I
```

NOTE: true/false don't apply here because there is no default set. This assumes NancyFX community preference is no prefix.

**C# Client**

```
# Default behavior, prefix with I-
./bin/csharp-petstore.sh --additional-properties interfacePrefix=true

# Customize without the default prefix
./bin/csharp-petstore.sh --additional-properties interfacePrefix=false

# Customize using your own prefix
./bin/csharp-petstore.sh --additional-properties interfacePrefix=ZZ
```

Questions...

I'd rather have NancyFX default to the I prefix and allow for the generator to also support the true/false/custom in the same way as the C# generator here.  If this is cool, we'll just need to remove the `setInterfacePrefix("")` line (and its comment) in `NancyFXServerCodegen`.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java
@@ -27,6 +27,8 @@ public abstract class AbstractCSharpCodegen extends DefaultCodegen implements Co
     protected String packageCompany = "Swagger";
     protected String packageCopyright = "No Copyright";
 
+    protected String interfacePrefix = "I";
+
     protected String sourceFolder = "src";
 
     // TODO: Add option for test folder output location. Nice to allow e.g. ./test instead of ./src.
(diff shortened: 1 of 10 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java
```diff
@@ -254,6 +256,18 @@ public void processOpts() {
         if (additionalProperties.containsKey(CodegenConstants.OPTIONAL_EMIT_DEFAULT_VALUES)) {
             setOptionalEmitDefaultValue(Boolean.valueOf(additionalProperties.get(CodegenConstants.OPTIONAL_EMIT_DEFAULT_VALUES).toString()));
         }
+
+        if (additionalProperties.containsKey(CodegenConstants.INTERFACE_PREFIX)) {
+            String useInterfacePrefix = additionalProperties.get(CodegenConstants.INTERFACE_PREFIX).toString();
```
wing328: Minor suggestion: what about using `Boolean.valueOf` to store `CodegenConstants.INTERFACE_PREFIX` as boolean instead of string so as to avoid case sensitivity issue?

jimschubert: @wing328 `Boolean.valueOf` would return true if true, and false otherwise. This would remove support for custom prefixes, as in the example in the PR description which prefixes with `ZZ`.

wing328: @jimschubert you're right. I missed that.

Issue #4401, opened 2016-12-14
[CSharp] Provide option for configurable access modifiers


kbedel:
##### Description

I would like be to able to configure the access modifiers on all generated classes. In my case I am using the c# client, but this is may be applicable to many of the generators.

The reason is that I want to include the generated code into an existing project which exposes a public API. The generated code is completely hidden and should not be part of the exposed API. I believe this should just be an option of the codegen process like the packageName option, and not tied to the specification itself. It may be applicable even in use cases where the person consuming the API is not the author of the specification, but may still want to generate the client as internal

##### Swagger-codegen version

Tested with version 2.2.1. I've also confirmed in master that nothing has changed around this.

##### Related issues

This one is somewhat similar, but is tied to the specification and is a bit more broad: #1467

wing328:
@kbedel Thanks for the feedback. Have you considered modifying the [C# templates](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/csharp/api.mustache) and apply the customized tempalates via the `-t` CLI option?

kbedel:
Yes, that is what I plan to do in the meantime, but I thought I would submit because it would be nice not to have to maintain my own set of templates, and thought it would likely be of use to others as well. Would there be interest in a PR adding a C# config option for this?

jimschubert:
@kbedel The only other access modifier that could work across all classes on the generated client is `internal`. `private` and `protected` wouldn't make sense as options because then even most classes within the generated output wouldn't be able to  access each other.

Would a `internalClassesOnly` option work for you?

When you say you're exposing a "public API" and the generated classes shouldn't be part of that, are you using Swashbuckle or something to generate your Swagger Definition? If so, you can exclude these types via namespace in your ApiExplorer object prior to the generation of the swagger docs. I think this would be a much better approach in your situation, but I can also see the value in generating classes as `internal`.

kbedel:
@jimschubert Yeah, internal is the only other one that makes sense here. That option sounds good.

By public API, I mean as part of an SDK 3rd party developers can use to customize the application. Some of the implementation of this public API calls out to the swagger web services, but other parts are just concerned with the application itself. All the generated types should be encapsulated and hidden from the 3rd party developers. So we need the objects to be generated and consumed in the implementation (and therefore in the Swagger Definition), but the types should not be public in our built assembly.

[...]

PR #4560, opened 2017-01-15
[csharp] Support internal access of generated code

jimschubert:
### PR checklist

### Description of the PR

Allows users to specify 'nonPublicApi' additional property (and C#
client CLI switch) to reduce visibility of classes created by the
generator. This includes API and Models as well as supporting code like
ApiClient and other infrastructure.

The requirement is to support codegen generated code to be embedded
within other applications where the generated code is not intended to be
publicly consumable or publicy exposed. An example would be an SDK which
internally consumes an API via the generated code; we wouldn't want the
internal API implementation exposed as part of that SDK.

Reducing visibility of the classes effectively makes the entire
implementation internal, regardless of the public modifier on methods or
static members. To fully make all members internal it would require
explicit interface implementation, which is not ideal.

see #4401

To evaluate, run the following then inspect the diff:

```
./bin/csharp-petstore.sh --additional-properties nonPublicApi=true
```


wing328:
@jimschubert thanks for the enhancement, which looks good to me.

PR #4573, opened 2017-01-16
[aspnetcore] use default in model constructors, supports enums

jimschubert:
### PR checklist

### Description of the PR

This addresses a couple of issues.  As originally reported in #3608 and added to the C# client generator in #4145, this updates model constructors to use `default()` rather than attempting to initialize parameters to nulls. I didn't modify the NancyFX generator here, because it relies on object initialization and doesn't perform the same types of constructor validations or DataAnnotations/JSON validation.

The `petstore.yaml` has inner enums, so to get the sample code working against this yaml I've also added support for enums and inner enums, and regenerated the sample.

Please note if you're having issues with newer versions of `dotnet` CLI, there's a known issue (https://github.com/dotnet/corefx/issues/9171) with OS X/macOS support which requires linking openssl manually:

```
$ ln -s /usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib /usr/local/lib/
$ ln -s /usr/local/opt/openssl/lib/libssl.1.0.0.dylib /usr/local/lib/
```

To test:

```
cd samples/server/petstore/aspnetcore
sh build.sh
```

NOTE:

If you receive the following error on build, you'll need to perform the openssl symlinks mentioned above:

```
Unhandled Exception: System.TypeInitializationException: The type initializer for 'Crypto' threw an exception. ---> System.TypeInitializationException: The type initializer for 'CryptoInitializer' threw an exception. ---> System.DllNotFoundException: Unable to load DLL 'System.Security.Cryptography.Native': The specified module could not be found.
 (Exception from HRESULT: 0x8007007E)
   at Interop.CryptoInitializer.EnsureOpenSslInitialized()
   at Interop.CryptoInitializer..cctor()
   --- End of inner exception stack trace ---
   at Interop.Crypto..cctor()
   --- End of inner exception stack trace ---
   at Interop.Crypto.GetRandomBytes(Byte* buf, Int32 num)
   at System.IO.Path.GetCryptoRandomBytes(Byte* bytes, Int32 byteCount)
   at System.IO.Path.GetRandomFileName()
   at Microsoft.DotNet.InternalAbstractions.TemporaryDirectory..ctor()
   at Microsoft.Extensions.EnvironmentAbstractions.DirectoryWrapper.CreateTemporaryDirectory()
   at Microsoft.DotNet.Configurer.NuGetPackagesArchiver..ctor()
   at Microsoft.DotNet.Cli.Program.ConfigureDotNetForFirstTimeUse(INuGetCacheSentinel nugetCacheSentinel)
   at Microsoft.DotNet.Cli.Program.ProcessArgs(String[] args, ITelemetry telemetryClient)
   at Microsoft.DotNet.Cli.Program.Main(String[] args)
build.sh: line 8: 45627 Abort trap: 6           dotnet restore src/IO.Swagger/
```

### DIFF
--- a/modules/swagger-codegen/src/main/resources/aspnetcore/enumClass.mustache
+++ b/modules/swagger-codegen/src/main/resources/aspnetcore/enumClass.mustache
@@ -1,0 +1,14 @@
+        /// <summary>
+        /// {{^description}}Gets or Sets {{{name}}}{{/description}}{{#description}}{{{description}}}{{/description}}
+        /// </summary>{{#description}}
+        /// <value>{{{description}}}</value>{{/description}}
+        public enum {{#datatypeWithEnum}}{{.}}{{/datatypeWithEnum}}{{^datatypeWithEnum}}{{classname}}{{/datatypeWithEnum}}
+        {
+            {{#allowableValues}}{{#enumVars}}
+            /// <summary>
+            /// Enum {{name}} for {{{value}}}
+            /// </summary>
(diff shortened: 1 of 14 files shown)


wing328:
@jimschubert thanks for the PR and the solution to address the SSL issue.

Issue #3350, opened 2016-07-12
[C#] automatically generate nuspec file


wing328:
##### Description

We want to automatically generate nuspec file ([example](https://github.com/twilio/twilio-csharp/blob/master/Twilio.nuspec)) for C# API client so that the C# API client can be easily packaged into NuGet library using the following command:

```
nuget pack <project_file> -OutputDirectory <temp_path>
```

Ref: https://docs.nuget.org/create/nuspec-reference
##### Swagger-codegen version

Latest master
##### Suggest a Fix/Solution

If anyone wants to work on this, please reply to let us know. Thanks!

jimschubert:
I can pick this up.

wing328:
Closed via #4576. Thanks for the contribution from @jimschubert

PR #4576, opened 2017-01-16
[csharp] Client nuspec

jimschubert:
### PR checklist

### Description of the PR

Per #3350, this generates a nuspec for C# client generated code.

To test, run:

```
./bin/csharp-petstore.sh
```

On Windows:

```
cd samples\client\petstore\csharp\SwaggerClient\src\IO.Swagger
nuget pack -Build -OutputDirectory out IO.Swagger.csproj
```

This will create the `.nupkg` file under a new directory called `out`.

This implementation bundles the README.md and docs.


wing328:
@jimschubert thanks for the PR but the CI reported the following issue:
```
$ /bin/bash ./bin/utils/detect_carriage_return.sh
modules/swagger-codegen/src/main/resources/csharp/nuspec.mustache
Templates contain carriage return '/r'. Please remove it and try again.
```
I believe the NuGET packing will still work after removing the carriage return from nuspec.mustache.

jimschubert:
@wing328 I can save with line feed only. The file looks terrible in Windows without carriage returns. Is that alright? I don't understand the need to warn for carriage returns, does it cause problems somehow?

wing328:
@jimschubert another way is to change your git setting so that git will clean it up for you during commit: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md#tips

We did receive complains about auto-generated files containing '/r' before.

Personally I think it would be a good idea to standardize the line break in the project unless there's a specific use case that we must use '/r'. I'm definitely open to discussion to allow '/r' in the code or template files.

jimschubert:
@wing328 I think files predominately used by Windows users like the C# stuff should probably allow carriage returns. It might make sense to disallow it in the templates, but output with carriage returns as an option?

wing328:
@jimschubert one thing I also consider is the experience programming in Mac/Linux (using Mono) with the C# client/server generated by Swagger Codegen. I think it depends on what IDE/editor is being used and some editors may not have the (default) setting to hide/convert "\r\n" properly.

Issue #4569, opened 2017-01-16
[JAXRS-CXF] basePath ignored since #4068


markus-wa:
##### Description

In the CXF generator the basePath gets ignored since #4068 and all paths changed from 'host/basePath/operationPath' to 'host/operationPath'.

##### Swagger-codegen version

2.2.2-SNAPSHOT

##### Swagger declaration file content or url

```yaml
swagger: '2.0'
info:
  version: 1.0.0
  title: missing basePath example
host: example.com
basePath: /exa
schemes:
  - http
consumes:
  - application/json
produces:
  - application/json
paths:
  /mple:
    get:
      responses:
        200:
          description:  Example
```

generates

```java
...
@Path("/")
@Api(value = "/", description = "")
public interface DefaultApi  {

    @GET
    @Path("/mple")
    @Consumes({ "application/json" })
    @Produces({ "application/json" })
    @ApiOperation(value = "", tags={  })
    public void mpleGet();
}
```

instead of
```java
...
@Path("/exa")
...
```

##### Command line used for generation

java -jar swagger-codegen-cli-2.2.2-SNAPSHOT.jar generate -i schema.json -l jaxrs-cxf

##### Steps to reproduce

Use a basePath in your swagger spec and generate the CXF server code.

##### Related issues

Pull #4068

##### Suggest a Fix

Change 
```java
@Path("/")
```
in https://github.com/swagger-api/swagger-codegen/blob/14bca6cd8922e6134c5848378e2c86151c439486/modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf/api.mustache#L21
back to
```java
@Path("{{contextPath}}")
```

wing328:
@markus-wa thanks for reporting the issue. Please submit a PR so that we can review the fix.

sbarinov:
Am I the only one who still experiences this issue with 2.3.1?

staraise:
still have this in 2.4.13, as I see the spring version got fixed, but all of the jax-rs ones are still bugged.

PR #4580, opened 2017-01-17
[JAXRS-CXF] Issue 4569 - Re-added usage of contextPath in api.mustache (basePath)

markus-wa:
### PR checklist

### Description of the PR

Fix for #4569 - basePath was ignored for jaxrs-cxf


wing328:
@markus-wa Thanks for the PR but your commit (as shown in the Commits tab) is not linked to your Github account, which means this PR won't count as your contribution in https://github.com/swagger-api/swagger-codegen/graphs/contributors.

wing328:
@jfiala please take a look at this PR when you've time.

markus-wa:
I have added the email from the commit to my github account.

jfiala:
@wing328 I removed this when I reworked the cxf template, because using the basePath in the @Path is not the usual configuration for CXF.
CXF uses spring and the basePath is configured in web.xml/spring configuration files:

@markus-wa
Please take a look at the generated web.xml file (with config generateSpringApplication=true)
``` 
<servlet-mapping>
        <servlet-name>CXFServiceServlet</servlet-name>
        <url-pattern>/rest/*</url-pattern>
    </servlet-mapping>
``` 
There the servlet mapping is done.

The next step is the JAXRS-server which is configured in Spring (see ApplicationContext.xml):
``` 
<jaxrs:server id="restServer" address="/services">
``` 
Here the services are now located at http://<servername>/<appcontext-name>/rest/services/<myservices>

The @Path(..) always is connected to the root ("/"), as there are already two other places where you can configure the servlet-mapping and the jaxrs:server-endpoint.

This is how I worked with the cxf language so far, other usages usually caused troubles with other features, so that was the most stable and also understandable solution to start with.

If you really need the basePath in @Path, I'd suggest to customize the template or if you think it makes really sense for the cxf-language, please also submit how the spring configuration should change in your opinion and verify all the features work (including swagger-ui, Swagger2Feature etc.).

markus-wa:
Until now we have been using a single CxfNonSpringServlet for multiple services with different basePaths. The @Path seems to be the only option in this case as far as I'm aware.

I will get back to you with more details - maybe a command line option might be the way to go?

jfiala:
@markus-wa I'm green with the CLI option, in fact the solution you proposed seems the cleaner way in JaxRS terms (contract -> jaxrs interface). You can also provide an alternative to the spring configuration I did by supporting a flag for the CxfNonSpringServlet, this would make it complete. Pls let me know if you need assistance in powering up the cxf language support...

jfiala:
@markus-wa great, this looks good!

jfiala:
for the generated samples, you generated them with the CLI flag turned on. In fact we would need a separate sample for all off (=default), one with all spring flags turned on and one with the generateNonSpringServlet turned on. @wing328  pls can you advise here how we should handle this in your opinion?

markus-wa:
Woops, what happened here? https://travis-ci.org/swagger-api/swagger-codegen/jobs/195578396

wing328:
@markus-wa I just pushed a fix to the latest master. Pls pull the latest to give it a try.

wing328:
> @wing328 pls can you advise here how we should handle this in your opinion?

A good starting is by looking at the following 2 shell scripts:
[spring-delegate-j8.sh](https://github.com/swagger-api/swagger-codegen/blob/master/bin/spring-delegate-j8.sh#L29)
[spring-delegate.sh](https://github.com/swagger-api/swagger-codegen/blob/master/bin/spring-delegate.sh#L29)

The differences are the output folder (samples/server/petstore/springboot-delegate-j8) and the additional option (java8=true).

For your case, you can do something similar with the additional option `generateNonSpringServlet`.

If you want, I can do it later after merging this PR into master.

markus-wa:
@jfiala @wing328 I've added the sample generation, let me know if i missed something.

wing328:
@markus-wa I've resolved the merge conflicts and merge the changes into master. Thanks for your contribution.

PR #4552, opened 2017-01-12
Swift3 Clener template of client methods

tomekc:
### PR checklist

- [x] Read the [contribution guildelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [x] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

Fixed an error which in some cases could lead to generate improper API function, for example not uploading files.

Done a little makeover of `api.mustache`. Limited capabilities of mustache templates led to extremely convoluted and hardly comprehensible templates. 

The template has been reformatted a little, added a couple of newlines, as I believe that readibility of template is much more important than resulting code having a couple extra newlines.

I tested on following use-cases, created privately against httpbin.org:

* upload a file with one more form param and query params
* POST an object with query params
* GET with couple of parameters

PR #4490, opened 2017-01-04
[swift3] allow POST with both body and query parameters

tomekc:
### PR checklist

- [x] Read the [contribution guildelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [x] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

Made POST requests allow use query parameters, besides of body. Currently generated code ignores all query parameters when body exists. 

Reported in #2557 and also in #2483 

**Note**
This is a workaround of flawed Alamofire API which is actual limitation. Next time I will post a Alamofire-less template :)

PR #4587, opened 2017-01-17
[Swift] two fixes to latest param mapping code

jaz-ah:
### PR checklist

### Description of the PR

1) extra ?'s at end of some url's 
-if you had nil parameters passed into a function, then you would get an api call like:
http://foo.com?

2) enums not being called out w/ rawValue to get the proper string name
-if you defined an enum:
enum {
    foo = "Foo"
}
the new mapping wouldn't call rawValue so you would get the lowercase foo instead of Foo


jaz-ah:
cc'ing @tomekc who made the original change in https://github.com/swagger-api/swagger-codegen/pull/4552 if you could look @ this one 
cc'ing @jgavris @Edubits @hexelon

modules/swagger-codegen/src/main/resources/swift3/api.mustache
```diff
@@ -130,7 +130,7 @@ open class {{classname}}: APIBase {
         {{#hasQueryParams}}
         url?.queryItems = APIHelper.mapValuesToQueryItems(values:[
         {{#queryParams}}
-                "{{paramName}}": {{paramName}}{{#hasMore}}, {{/hasMore}}
+                "{{paramName}}": {{paramName}}{{#isEnum}}{{#isContainer}}{{{dataType}}}{{/isContainer}}{{^isContainer}}{{^required}}?{{/required}}.rawValue{{/isContainer}}{{/isEnum}}{{#hasMore}}, {{/hasMore}}
```
tomekc: I believe it is now correct, but I'd address the readability of this fragment in the future. 
What  `isContainer` means by the way?

jaz-ah: y - I actually have to make one more fix here - this isn't all the way correct yet.  let me see if I cleanup the readability as well.

jaz-ah:
@wing328 @tomekc still feels like there is an error here where my new code is messing w/ arrays/container types when it shouldn't be (i.e. adding a rawValue when it shouldn't)

samples/client/petstore/swift3/default/PetstoreClient/Classes/Swaggers/APIs/FakeAPI.swift
```diff
@@ -208,9 +208,9 @@ open class FakeAPI: APIBase {
 
         let url = NSURLComponents(string: URLString)
         url?.queryItems = APIHelper.mapValuesToQueryItems(values:[
-                "enumQueryStringArray": enumQueryStringArray, 
-                "enumQueryString": enumQueryString, 
-                "enumQueryInteger": enumQueryInteger
+            "enum_query_string_array": enumQueryStringArray, 
+            "enum_query_string": enumQueryString?.rawValue, 
+            "enum_query_integer": enumQueryInteger?.encodeToJSON()
```
jaz-ah: @tomekc @wing328 is this correct for the integer value - was this broken as well?  I notice the double above in the formParams wasn't encoded...

jaz-ah: This works for me locally - looks like the correct change.

jgavris:
I can check this out tomorrow as a smoke test in our project. Would love to improve the test coverage since https://github.com/swagger-api/swagger-codegen/pull/4490 was passing. @wing328 could you give us an example of using the fake petstore?

jaz-ah:
@jgavris did the new code work out for you?

jgavris:
@jaz-ah We're actually running the 2.3.0 branch and there are conflicts with master at the moment. I might have to defer to someone else.

wing328:
>  could you give us an example of using the fake petstore?

C#, PHP, Python, Ruby and more are using [fake petstore](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml) for testing.

https://github.com/swagger-api/swagger-codegen/blob/master/bin/csharp-petstore.sh
https://github.com/swagger-api/swagger-codegen/blob/master/bin/python-petstore.sh
https://github.com/swagger-api/swagger-codegen/blob/master/bin/php-petstore.sh
https://github.com/swagger-api/swagger-codegen/blob/master/bin/ruby-petstore.sh

Is that what you're looking for?

Totally agree with you about adding an additional test cases, if not too difficult, to cover this issue.

jaz-ah:
@wing328 might be a little more difficult - I'd need to run testEnumParametersWithRequestBuilder and I need a different cased enum, and the the server to fail if the case of the enum isn't correct... unless you have an idea of a different way to test this one - I agree we should have something to make sure this doesn't break in the future.   (also not sure why the build broke on this)

wing328:
@jaz-ah right, definitely not easy. I'll create a task to track various test cases we want to include.

Petstore (fake or original) definitely is not the best spec for testing various test cases. Moving forward, we want to leverage one of the server stub generators to generate server code covering more scenarios.

jaz-ah:
@wing328 not sure why this build isn't building though...

wing328:
@jaz-ah I've restarted the jobs and the results now look good.

PR merged into master.

PR #4593, opened 2017-01-18
Feat/travis fail if generator changes

JLLeitschuh:
### PR checklist

### Description of the PR

Travis now runs all of the code generation to ensure that the person opening the PR has updated the samples with any changes that they have made.

If the individual has not run the generation scripts then travis will fail the build.


JLLeitschuh:
The perl project needs to have the date removed from the readme for this to work correctly.

JLLeitschuh:
I don't know why travis is failing the way it is unless there exists a problem in the generated code as it is now.

wing328:
@JLLeitschuh thanks for the PR. It's definitely a good attempt.

If I only update README.md (in which I did quite often), will your mechanism to detect changes report errors?

JLLeitschuh:
It detects any diff that running the code generator creates.
So if you forget to run the code generator and there is a change that would have happened in the samples this will fail.

JLLeitschuh:
My company has decided to go a different route than use the swagger codegenerator.
I don't think I'm going to have the time to clean this up ever. This PR can be closed unless you think that someone else would be willing to take on cleaning this up.

PR #4602, opened 2017-01-19
change html2 moustache to fix parameter name

johnluan:
### PR checklist

### Description of the PR

Considering html2 as generating document instead of code, so there should be baseName in the page.
Otherwise, the parameter will be changed if the parameter doesn't follow the element rules.
e.g. parameter name as aa_bb_cc, then the document will change it as  aaBbCc, which is confused for the document users.  So highly recommend to change it to baseName.


wing328:

PR #4507, opened 2017-01-06
[Jaxrs-cxf-cdi] Add beanvalidation annotations

jfiala:
### PR checklist

- [x] Read the [contribution guildelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [x] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

add beanvalidation annotation support to jaxrs-cxf-cdi

for details see #4091

Issue #4091, opened 2016-10-27
[Java*] BeanValidation for all Java languages


jfiala:
##### Description

BeanValidation support has been added in #4003 for Java and jaxrs-cxf/jaxrs-cxf-client.
This needs to be added to all other Java languages

- [x] jaxrs-spec
- [x] jaxrs (Jersey2 + library jersey1)
- [x] jaxrs-resteasy
- [x] jaxrs-cxf-cdi
- [x] spring (default library spring-boot + library spring-mvc, spring-cloud)

##### Related issues

PR Beanvalidation annotation for Java and CXF  #4003
Issues with language jaxrs-spec #4509 

##### Issues with language jaxrs-spec (see #4509)

* the api is generated as class, but IMHO this should be an interface!?
* Jaxrs Response is used instead of the real response data type
* @ApiResponse uses wrong response type (the return type, not the response type of the contract)
* src/gen/java is not added to the build path
* "Attachment" is used for attachments, but no import is added for this (I think org.apache.cxf.jaxrs.ext.multipart.Attachment is referenced, but this is CXF-specific)

Update: moved all the issues with jaxrs-spec to #4509

##### Support for @DecimalMin/@DecimalMax
Also add support for @DecimalMin/@DecimalMax 

- [x] jaxrs-spec
- [x] jaxrs (Jersey2 + library jersey1)
- [x] jaxrs-resteasy
- [x] jaxrs-cxf-cdi

##### Final Cleanup (PR #4635)
- [x] move beanvalidation code to AbstractJavaJAXRSServerCodegen.java
(so it will stay redundant in AbstractJavaJAXRSServerCodegen.java and SpringCodegen.java) or to AbstractJavaCodegen?
- [x] Backport support for @DecimalMin/@DecimalMax
backport to languages which already have beanvalidation annotation support:
   - [x] java
   - [x] jaxrs-cxf / jaxrs-cxf-client

Final Cleanup implemented in PR #4635

jfiala:
@wing328 If you like I can cycle through the language templates and add (optional) beanvalidation support there. Or do you prefer the language maintainers to do the job?

wing328:
@jfiala if you've time, please add the bean validation support. I'm sure the community and the template owner would appreciate your effort.

cbornet:
Small remark: if bean validation is an option, it should be activated by default (since you expect the generated code to follow the contract established by the swagger spec)

deveshpujari:
@jfiala 

[...]

PR #4615, opened 2017-01-21
[Jaxrs-cxf-cdi] Add beanvalidation annotations updated

jfiala:
### PR checklist

### Description of the PR

add beanvalidation annotation support to jaxrs-cxf-cdi
for details see #4091 

This is a re-creation of the PR #4507 to have a clean comparison view.


wing328:
cc @nickcmaynard

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/api.mustache
```diff
@@ -48,7 +52,7 @@ public class {{classname}}  {
     @ApiResponses(value = { {{#responses}}
         @ApiResponse(code = {{{code}}}, message = "{{{message}}}", response = {{{returnType}}}.class{{#returnContainer}}, responseContainer = "{{{returnContainer}}}"{{/returnContainer}}){{#hasMore}},{{/hasMore}}{{/responses}} })
     public Response {{nickname}}({{#allParams}}{{>queryParams}}{{>pathParams}}{{>headerParams}}{{>bodyParams}}{{>formParams}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) {
-    	return delegate.{{nickname}}({{#allParams}}{{#isFile}}inputStream, fileDetail{{/isFile}}{{^isFile}}{{paramName}}{{/isFile}}, {{/allParams}}securityContext);
+    	return delegate.{{nickname}}({{#allParams}}{{#isFile}}fileInputStream, fileDetail{{/isFile}}{{^isFile}}{{paramName}}{{/isFile}}, {{/allParams}}securityContext);
```
nickcmaynard: Good point, though don't we mean `{{paramName}}InputStream`?

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/beanValidation.mustache
```diff
@@ -0,0 +1,53 @@
+{{#required}}
```
nickcmaynard: This, `beanValidationPathParams.mustache`, and `beanValidationQueryParams.mustache` are *very* similar.  I understand the arguments for avoiding reuse across language generators, but I simply don't agree that needs to be the case within a language generator.  I think these three files need refactoring to remove the duplication.

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/enumClass.mustache
```diff
@@ -1,16 +1,33 @@
-import javax.xml.bind.annotation.XmlEnum;
-import javax.xml.bind.annotation.XmlType;
+@XmlType(name="{{datatypeWithEnum}}")
```
nickcmaynard: Why do we have enum changes in here?

nickcmaynard:
Bunch of code review comments above.  Also, still seeing jaxrs-spec sample regeneration...

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/api.mustache
jfiala: without this change there currently is a compile error in PetApi.java, as the parameter is declared with name fileInputStream instead of inputStream
``` java 
 @Multipart(value = "file", required = false) InputStream fileInputStream
```
So, the fix is to modify the returned parameter from inputStream to fileInputStream.
Of course also the parameter name in the method signature could be changed, pls let me know if you prefer it the other way.

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/beanValidation.mustache
jfiala: OK, the only difference is that PathParams are always required, so they don't need a check for NotNull, I commented this in beanValidationPathParams.mustache. So we could extract all other annotation checks into a beanValidationCommonParams.mustache.

Regarding cross-language templates using lambdas this is a nice feature of Mustache, but we need green light from @fehguy /@wing328 for cross-language template reuse.

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/enumClass.mustache
jfiala: @nickcmaynard  With the old version enumClass.mustache there are compile errors, as the generated sample files were out of date. Pls simply try with the old version and then you see the compilation errors.

jfiala:
Responded to all review comments, pls let me know if I should adjust the templates in my branch or if you prefer to do a PR on my repo etc.

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/api.mustache
nickcmaynard: Yes, because

```
{{#isFile}} @Multipart(value = "{{paramName}}"{{^required}}, required = false{{/required}}) InputStream {{paramName}}InputStream, @Multipart(value = "{{paramName}}" {{^required}}, required = false{{/required}}) Attachment {{paramName}}Detail{{/isFile}}
```

The parameter name is *generated* by `{{paramName}}InputStream`.  `fileInputStream` is the result of that.  So shouldn't we generate here, too?

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/beanValidation.mustache
nickcmaynard: Sounds good, within the language.

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/enumClass.mustache
nickcmaynard: My point is that this is a simple case of one PR, two issues.  We should get into the habit of one PR, one issue.

jfiala: I'm absolutely with you (if the samples are up to date and compiling).

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/api.mustache
jfiala: Yes, thx I'll fix that.

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/beanValidation.mustache
jfiala: OK, I'll change that.

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/api.mustache
jfiala: thx for the hint, the templates are now reused in beanValidationParams.mustache.

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/beanValidation.mustache
jfiala: thx for the hint, the templates are now reused in beanValidationParams.mustache.

jfiala:
@nickcmaynard fixed all review comments, pls review :).

jfiala:
The consolidation of beanValidationParams.mustache should be backported to all other languages, created #4636 as reminder.

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/api.mustache
```diff
@@ -48,7 +52,7 @@ public class {{classname}}  {
     @ApiResponses(value = { {{#responses}}
         @ApiResponse(code = {{{code}}}, message = "{{{message}}}", response = {{{returnType}}}.class{{#returnContainer}}, responseContainer = "{{{returnContainer}}}"{{/returnContainer}}){{#hasMore}},{{/hasMore}}{{/responses}} })
     public Response {{nickname}}({{#allParams}}{{>queryParams}}{{>pathParams}}{{>headerParams}}{{>bodyParams}}{{>formParams}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) {
-    	return delegate.{{nickname}}({{#allParams}}{{#isFile}}inputStream, fileDetail{{/isFile}}{{^isFile}}{{paramName}}{{/isFile}}, {{/allParams}}securityContext);
+    	return delegate.{{nickname}}({{#allParams}}{{#isFile}}{{paramName}}InputStream, fileDetail{{/isFile}}{{^isFile}}{{paramName}}{{/isFile}}, {{/allParams}}securityContext);
```
nickcmaynard: Sorry - just spotted that by the same logic, we probably also want `{{paramName}}Detail`!

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/beanValidation.mustache
nickcmaynard: Can beanValidation.mustache also bring in `{{>beanValidationParams}}`?  There looks like there's more common ground here...

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/pom.mustache
```diff
@@ -76,6 +76,16 @@
             <artifactId>swagger-annotations</artifactId>
             <version>[1.5.3,2)</version>
         </dependency>
+
```
nickcmaynard: This section incorrectly indented </picky>

nickcmaynard:
@jfiala added a couple more review comments - apologies.  Additionally, I'm still seeing sample updates for the jaxrs-spec generator, which is unexpected.

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/pom.mustache
jfiala: you probably mean the beanvalidation dependency, this is fixed now.

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/api.mustache
jfiala: fixed - added {{paramName}}Detail

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/beanValidation.mustache
jfiala: the difference here is that in beanValidation.mustache every annotation gets a new line whereas in the params template all annotation share the same line. So I fear we have to keep them separate.

jfiala:
@nickcmaynard updated according to your comments and reset the jaxrs-spec samples to master.

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf-cdi/beanValidation.mustache
nickcmaynard: OK - because this template is used for method annotations, you're adding newlines.  However, to my knowledge, there's no Java syntax requirement for newlines on method/property annotations - indeed, much like parameter annotations, newlines are about style, rather than syntax.

So - syntactically I think this would be OK without the newlines.  What do you think?

I've also noticed you've got some clever integer/decimal handling in here that's missing in `beanValidationParams.mustache`- should that be ported?

jfiala: I'm not sure if its better to drop the newlines in favor of having one beanValidationCore.mustache. 
Currently they are all duplicated with newlines (beanValidation.mustache) and all in one line (beanValidationParams.mustache).
Let's have a second opinion before I change something here - @wing328  can you give us a thought?

jfiala:
I refined the handling of DecimalMin/DecimalMax to correctly use Min/Max for Integer+Long and DecimalMin/Max for all others (it would be great to have a flag for non-decimal types in CodegenProperty) and added it to both beanValidation.mustache + beanValidationParams.mustache.

@wing328 Pls advise if we should merge the two beanValidation template files at the cost of losing the newlines for the ApiModels (the beanvalidation annotations would then always be a one-liner for params and models, but maybe that's even better for readability...).

jfiala:
@wing328 can you pls take a look here and merge this into master, then I we can conclude the beanvalidation (PR #4635).

wing328:
> Pls advise if we should merge the two beanValidation template files at the cost of losing the newlines for the ApiModels (the beanvalidation annotations would then always be a one-liner for params and models, but maybe that's even better for readability...).

I would suggest we go with merging the 2 template files for better readability and see if the community has any feedback.

wing328:
@jfiala thanks for the PR, which has been merged into master.

@nickcmaynard thanks for reviewing the change.

jfiala:
Beanvalidation annotations fully implemented for jaxrs-cxf-cdi

Issue #4627, opened 2017-01-23
Embedded mustache files are used even if TemplateFolder option is specified and contains a different version of the mustache file (in a library folder)


eblis:
##### Description

I'm trying to generate Spring-Boot server sources from a swagger.json file and I'm specifying my own TemplateFolder which was copied from the original `JavaSpring` template folder and subsequently modified. I've modified the `pom.mustache` file for all 3 Spring libraries (`spring-boot`, `spring-cloud` and `spring-mvc`) and none are taken into consideration while generating for `spring-boot` library.

##### Swagger-codegen version

2.2.2-SNAPSHOT

##### Command line used for generation

./java -jar swagger-codegen-cli.jar generate -i swagger.json -c spring-config.json -l spring -o samples/server/test -t JavaSpring-Modified

eblis:
This seems to affect all languages that have libraries, not limited to Spring.

anselmos:
@eblis any suggestion how to fix this issue - I'm currently trying to debug and hotfix this but no success so far... 
Related issues : #9521 #9363

eblis:
I haven't been following this for a while, but have you tried my changes that I made ? It worked for me at the time.

HugoMario:
this is fixed for codegen 2.x version, so closing now, but please, let me know if issue it's still present and i work on it asap.

PR #4628, opened 2017-01-23
Not using files from template folders with libraries inside them

eblis:
### PR checklist

### Description of the PR

#4627 - When looking for template files also look in the libraries subfolder of the supplied templateFolder, before checking the embedded locations.

If the template files reside in the library subfolder in the template dir they are never checked and used, instead the embedded files (from the library subfolder) are used instead.


eblis:
While working on this pull request I discovered another issue, when library folders are present.
If specifying an external template folder which uses libraries, if both the main template folder and the library folder contain the same file, the main template file will override the one in the library, but it feels like the library version of the file should win (like it does for embedded template folders).

I have modified the pull request to take this into consideration, while also adding a unit test to check that this does indeed happen.

1. 1st the code will check if there's a _\<template folder\>_/libraries/_\<library\>_ folder containing the file
2. 2nd it will check for the file in the specified _\<template folder\>_ folder
3. 3rd it will check if there's an _\<embedded template\>_/libraries/_\<library\>_ folder containing the file
4. 4th and last it will assume the file is in _\<embedded template\>_ folder.

wing328:
@eblis thanks for the PR with additional tests and test cases. I did some tests and the result looks good.

Issue #4647, opened 2017-01-25
invalid nodejs titles can lead to illegal project names


fehguy:
##### Description

Per https://docs.npmjs.com/files/package.json#name, the title of a project is constrained to certain characters.  Currently the codegen will let essentially anything go into the name, with the exception of replacing some underscores to dashes.  Needs to be more aggressive and adhere to the npm naming policy

##### Swagger-codegen version

2.2.2-SNAPSHOT

wing328:
Should be fixed by https://github.com/swagger-api/swagger-codegen/pull/4648

PR #4648, opened 2017-01-25
tests, fix for #4647.  Proper escaping project name per npmjs

fehguy:
### PR checklist

### Description of the PR

Properly escape project name

### DIFF
--- a/samples/server/petstore/nodejs/api/swagger.yaml
+++ b/samples/server/petstore/nodejs/api/swagger.yaml
@@ -356,8 +356,8 @@ paths:
         description: "ID of pet that needs to be fetched"
         required: true
         type: "integer"
-        maximum: 5
-        minimum: 1
+        maximum: 5.0
+        minimum: 1.0
         format: "int64"
       responses:
         200:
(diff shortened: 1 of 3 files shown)


samples/server/petstore/nodejs/api/swagger.yaml
```diff
@@ -356,8 +356,8 @@ paths:
         description: "ID of pet that needs to be fetched"
         required: true
         type: "integer"
-        maximum: 5
-        minimum: 1
+        maximum: 5.0
+        minimum: 1.0
```
wing328: @fehguy The output is now a decimal instead of integer: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.yaml#L343-L344

I don't think it's related to the change in this PR, probably due to changes in the Swagger Parser.

wing328: Created https://github.com/swagger-api/swagger-codegen/issues/4649 for tracking.

Issue #4651, opened 2017-01-25
[JAVA] [okhttp-gson] Model class not imported in generated client code


brucegodden:
##### Description

The uploaded [example.zip](https://github.com/swagger-api/swagger-codegen/files/730033/example.zip) contains a simple swagger definition and configuration file that generates a client for the Java okhttp-gson library that fails to compile because an import of the model class used is missing.

The part of the definition that triggers this is the api response being of type List&lt;List&lt;Person&gt;&gt;.

The DefaultApi.java file in the uploaded [example.zip](https://github.com/swagger-api/swagger-codegen/files/730033/example.zip) is the file I generated and you can see that it has no model class imports in it at all. 

##### Swagger-codegen version

v2.2.1

##### Swagger declaration file content or url

Definition: swagger.json in the uploaded [example.zip](https://github.com/swagger-api/swagger-codegen/files/730033/example.zip).

Configuration: java.json in the uploaded [example.zip](https://github.com/swagger-api/swagger-codegen/files/730033/example.zip).

##### Command line used for generation

java -cp /tmp/lib/swagger-codegen-cli-2.2.1.jar io.swagger.codegen.SwaggerCodegen generate -l java -c /tmp/example/java.json -i /tmp/example/swagger.json -o /tmp/build

##### Steps to reproduce

1. Put the json files from the uploaded [example.zip](https://github.com/swagger-api/swagger-codegen/files/730033/example.zip) into `/tmp/example`.
2. Run the generator as above. (You may need a different path to  the cli jar.)
3. Cd into the `/tmp/example/build` directory.
4. Run `./gradlew compileJava`
5. Notice the compile errors due to the missing import for the `Person` class.

##### Related issues

I can't see any similar reports.

##### Suggest a Fix

**Simple**

The vanilla java `api.mustache` template file includes the line

```
import {{modelPackage}}.*;
```

This should be put into all the library specific copies of the `api.mustache` files. I have created [pull request 4652](https://github.com/swagger-api/swagger-codegen/pull/4652) to do just this.

**Complex**

Fix the code that generates the `imports` property to find classes inside more than one level of wrapping by list. (And also, I suspect, set and map.)

wing328:
@brucegodden thanks for reporting the issue and filing a PR for the fix.

I'll review and let you know if I've any question.

wing328:
Closed via https://github.com/swagger-api/swagger-codegen/pull/5891

PR #4652, opened 2017-01-25
Add import of all model classes to java library api.mustache.

brucegodden:
This is a quick fix for https://github.com/swagger-api/swagger-codegen/issues/4651
which reports that under some circumstances imports of model classes my be omitted.

### PR checklist

Done and the results of running the various test scripts added.

### Description of the PR

This change imports all model classes in the Java library specific api.mustache template files. This is needed because the code generator fails to generate an import for model classes in response types like List&lt;List&lt;Model&gt;&gt;.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/feign/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/feign/api.mustache
@@ -5,6 +5,7 @@ import {{invokerPackage}}.ApiClient;
 import {{invokerPackage}}.ParamExpander;
 {{/legacyDates}}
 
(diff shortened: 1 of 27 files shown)


wing328:
@brucegodden thanks for the fix. I could repeat the issue, which is due to 2D array of object.

Instead of hard coding the model import, I think I can come up with a fix to check for 2D array/map of object. Will keep you posted.

wing328:
@brucegodden I've fixed the issue via https://github.com/swagger-api/swagger-codegen/pull/5891

Please pull the latest master to give it a try.

(thanks again for filing a fix for this)

PR #4659, opened 2017-01-26
[PHP] Validate parent's model first, if any

### LINKED ISSUE #4530: [PHP] Model's valid() method should validate parents too
##### Description

When some Swagger definitions using `allOf` for inheritance, the models are generated correctly with `extends` keyword and `parent::__construct($data)` in the constructor. But the `valid()` method doesn't check the validity of the parent's attribute, only the model's own one.

IMO, since the model inherits all the parent's attribute, it'll more make sense to validate those too.

##### Swagger-codegen version

2.2.1

http://pastebin.com/k2EtN43n

```
swagger-codegen generate -i swagger.json -l php
```

##### Steps to reproduce

Using definitions above, the generated models would be like this:

BaseDevice.php
```php
    public function valid()
    {
        if ($this->container['uuid'] === null) {
            return false;
        }
        if ($this->container['platform'] === null) {
            return false;
        }
        $allowed_values = array("android", "ios");
        if (!in_array($this->container['platform'], $allowed_values)) {
            return false;
        }
        return true;
    }
```

DeviceDetail.php
```php
    public function valid()
    {
        if ($this->container['id'] === null) {
            return false;
        }
        return true;
    }
```

That way, if I construct `DeviceDetail` with only `id`, the `valid()` method will return `true`

```
php > include 'autoload.php';
php > $dev = new \DeviceDetail(['id' => 123]);
php > var_dump($dev->valid());
bool(true)
```

##### Suggest a Fix

IMO the "correct" `valid()` method for `DeviceDetail` should be like this:
```php
    public function valid()
    {
        if (!parent::valid()) {
            return false;
        }

        if ($this->container['id'] === null) {
            return false;
        }
        return true;
    }
```

nazieb:
### PR checklist

### Description of the PR

On PHP generated models, if they have parent, the `valid()` & `listInvalidProperties()` method will lookup the parent's first.

Close #4530


wing328:
@nazieb the PR looks good to me. I'll merge it later today.

cc @arnested

wing328:
@nazieb PR merged. Thanks for your contribution.

PR #4666, opened 2017-01-27
[JavaScript] Fixed handling of empty arrays/empty strings/booleans/single numbers as valid API responses

### LINKED ISSUE #4653: [JavaScript] Handling empty array
Hi! Thank for your great work once again! I wonder if you can help me.

##### Description

I'm using generated javascript client in my project and I've met a bug, I suppose.

If my API returns an empty array client is returning it to me as a string and fails with message:

> TypeError: data.map is not a function

It occurs in `src/ApiClient.js:exports.convertToType()` 

```javascript
} else if (Array.isArray(type)) {
          // for array type like: ['String']
          var itemType = type[0];

          return data.map(function(item) {
            return exports.convertToType(item, itemType);
          });
        } else if (typeof type === 'object') {
```

I suppose this data is handled somewhere in `exports.prototype.deserialize`

Because of 
```javascript
if (data == null || !Object.keys(data).length) {
      // SuperAgent does not always produce a body; use the unparsed response as a fallback
      data = response.text;
    }
```

Can you help me?

frol:
### PR checklist

### Description of the PR

Fixes #4653.

/cc @khorolets @wing328 @jordanyaker

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Javascript/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Javascript/ApiClient.mustache
@@ -317,7 +317,7 @@
     // Rely on SuperAgent for parsing response body.
     // See http://visionmedia.github.io/superagent/#parsing-response-bodies
     var data = response.body;
-    if (data == null || !Object.keys(data).length) {
+    if (data == null || (typeof data === 'object' && typeof data.length === 'undefined' && !Object.keys(data).length)) {
       // SuperAgent does not always produce a body; use the unparsed response as a fallback
       data = response.text;
     }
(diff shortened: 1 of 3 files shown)


wing328:
@frol thanks for the quick fix, which looks good to me.

I'll merge it over the weekend if no one has further feedback/question.

frol:
@khorolets Please, test this fix (you can just apply the patch manually).

khorolets:
@frol @wing328 I've patched manually my code and I can confirm the fix works great.

wing328:
PR merged into master. Thanks again for the quick fix from @frol

Issue #4664, opened 2017-01-27
[go] incorrect code generated, if there are more than 1 array parameters 


vvalchev:
##### Description

I'm having a problem, when a method contains two array parameters.
The generated code looks like:
```go
	var collectionFormat = "csv"
	if collectionFormat == "multi" {
		for _, value := range param1 {
			queryParams.Add("param1", value)
		}
	} else {
		queryParams.Add("param1", a.Configuration.APIClient.ParameterToString(param1, collectionFormat))
	}
		var collectionFormat = "csv"
	if collectionFormat == "multi" {
		for _, value := range param2 {
			queryParams.Add("param2", value)
		}
	} else {
		queryParams.Add("param2", a.Configuration.APIClient.ParameterToString(param2, collectionFormat))
	}
```

The problem is that the following line appears twice:
```
var collectionFormat = "csv"
```

##### Swagger-codegen version

2.2.1
2.2.2-SNAPSHOT

##### Swagger declaration file content or url
```yaml
swagger: '2.0'
info:
    title: dummy
    version: "1.0.0"
host: dummy

paths:
    /dummy:
        get:
            parameters:
                - name: param1
                  in: query
                  type: array
                  items:
                    type: string
                - name: param2
                  in: query
                  type: array
                  items:
                    type: string
```
##### Command line used for generation
```
java -jar codegen.jar generate -i collection-format-test.yaml --lang go --output test-generated
```

##### Steps to reproduce

Use the provided yaml and the command line to generate to go sources and then compile.

##### Suggest a Fix

The easiest fix would be to create a code block:
```
  { // param 1
    var collectionFormat = "csv"
    ..
  }
  { // param2
   var collectionFormat = "csv"
  }
```

This could be easily achieved, if the following patch is applied to go/api.mustache:
```diff
--- go/orig/api.mustache	2016-08-08 15:53:46.000000000 +0000
+++ go/api.mustache	2017-01-27 11:30:38.000000000 +0000
@@ -82,13 +82,15 @@
 	{{#hasQueryParams}}
 	{{#queryParams}}
 	{{#isListContainer}}
-	var collectionFormat = "{{#collectionFormat}}{{collectionFormat}}{{/collectionFormat}}"
-	if collectionFormat == "multi" {
-		for _, value := range {{paramName}} {
-			queryParams.Add("{{baseName}}", value)
+	{
[...]

PR #4683, opened 2017-01-31
Fixed issue #4664

vvalchev:
Fixed issue #4664: [go] incorrect code generated, if there are more than 1 array parameters

Took the suggestion by wing328 and reworked the fix. This patch also applied to the latest master, while the original issue was about 2.2.1 code.


wing328:
@vvalchev thanks for the PR. I tested the update Petstore sample for Go and the result looks good.

PR #4686, opened 2017-01-31
[JAXRS-CXF] Add constructors to POJOs

MatanRubin:
Before this change, POJOs where generated without regular constructors.
This would mean you have to use the default constructor and setters or
builder-style methods. Some people consider regular constructors safer
to use due to compiliation failing when changing members of an object.
This change adds a regular constructor to all POJOs, and specifically to
the generated models.


wing328:
@MatanRubin thanks for the PR. I'll take a look by this week and let you know if I've any feedback.

FYI. For C# constructor, we've something more complicated: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/csharp/modelGeneric.mustache#L50

cbornet:
The problem is that you now need to pass every attributes when constructing an object, even the ones you don't care about. This should only be done for required params.

MatanRubin:
@cbornet, thanks for your review, you made me realize something important:
Before I made this change the POJOs didn't have a constructor at all, so the compiler generated the default one. Now that the POJO has a constructor, the default one is gone, and that would be a breaking change.
I will amend the commit to generate both a constructor with parameters and an empty constructor, for backward compatibility.

As to your point regarding optional parameters:
My impression is that most people do not use required parameters in models that much (also apparent from swagger's examples). Am I wrong?
If people usually use optional parameters, than generating a cosntructor for only the required parameters would miss the point...
In any case, I'm open for suggestions. If we want we can simply generate three constructors:
1. An empty constructor.
2. A constructor for required params only
3. A constructor for all params.

And let the user decide which one he wants to use.
What do you think?

cbornet:
Well, I do use required and other validations in my models and probably everybody should...
Otherwise there are two questions here:
* do we generate a constructor with all params (3) ? I won't use it because I prefer fluent setters but that's only my taste... In this case, either (1) or (2) is needed.
* do we generate a constructor based on required params (2) ? I think that would be very useful. In this case the empty constructor (1) is useless if you have required parameters. And (2) is (1) if you don't.

Also note: I think Jackson will need @JsonCreator if (1) is not present.

MatanRubin:
Hi @cbornet,
If we generate (2) and/or (3), we must generate (1) for backward compatibility (and for Jackson). In general, I don't see a reason not to always generate (1).
I believe (2) and (3), as you said, are more a matter of taste, so I'd like to give users the freedom to decide.

So I think the way to go is to generate all 3 constructors. Do you have an objection to that?

P.S. you're right that in some cases there (3) can be the same as (2) (if there are no optionals) and (2) can be the same as (1) (if there are no fields at all). I will handle these cases and only generate the required constructors.

MatanRubin:
I have updated the pull request to generate the 3 constructors discussed.

wing328:
@MatanRubin I just review again and the change looks good to me. I'll resolve the merge conflicts and merge it tomorrow if no one has further feedback on this.

wing328:
I've reviewed some Java SDKs (e.g. Facebook, Twitter) and they do not always provide constructors with all fields (as the object may contain many fields

The best way I think can think of to add this is via a CLI option.

wing328:
@MatanRubin closing this for the time being. We'll revise this when the CLI option is added. Thanks for the work.

shpandrak:
Hi, we are interested in the feature suggested here for pojo constructors, can this PR be reopened?

wing328:
@shpandrak reopened. You can clone the change:
```
git checkout -b MatanRubin-constructors master
git pull https://github.com/MatanRubin/swagger-codegen.git constructors
```
Then you can add the CLI option locally in your branch and submit a separate PR.

JFCote:
@wing328 
I think that by default, the empty constructor and the required constructor should always be there. As for the "all" params, it should be an option in the CLI. If the option is set to true, it should replace the required constructor.

Also, POJO is almost the same in all java generators (duplicated .mustache files). It would be nice to find a way to re-use code between java generators. For example, my generator (JavaPlayFramework) would benefits this feature of constructor.

shpandrak:
After discussing with @MatanRubin, I took over this with a new PR:
https://github.com/swagger-api/swagger-codegen/pull/6687

MatanRubin:
As @shpandrak wrote, he will continue working on this in a separate PR. Closing this one.
Thanks @shpandrak!

PR #4692, opened 2017-02-02
Default directory for generated sources is "src/gen/java" and not "src/main/java"

deblockt:
Default directory for generated sources is "src/gen/java" and not "src/main/java"

### Description of the PR

If i use this configuration : 
``` xml
<executions>
	<execution>
		<goals>
			<goal>generate</goal>
		</goals>
		<configuration>
		    <inputSpec>src/main/resources/contextPortalPanoramix-swagger.json</inputSpec>
		     <language>jaxrs-cxf-client</language>
                     <addCompileSourceRoot>true</addCompileSourceRoot>
		</configuration>
	</execution>
</executions>
```

The generated sources are in "generated-sources/swagger/src/gen/java" directory, but maven pulgin add "generated-sources/swagger/src/main/java" on the maven build directory.

### DIFF
--- a/modules/swagger-codegen-maven-plugin/src/main/java/io/swagger/codegen/plugin/CodeGenMojo.java
+++ b/modules/swagger-codegen-maven-plugin/src/main/java/io/swagger/codegen/plugin/CodeGenMojo.java
@@ -346,7 +346,7 @@ public class CodeGenMojo extends AbstractMojo {
 
         if (addCompileSourceRoot) {
             final Object sourceFolderObject = configOptions == null ? null : configOptions.get(CodegenConstants.SOURCE_FOLDER);
-            final String sourceFolder =  sourceFolderObject == null ? "src/main/java" : sourceFolderObject.toString();
+            final String sourceFolder =  sourceFolderObject == null ? "src/gen/java" : sourceFolderObject.toString();
 
             String sourceJavaFolder = output.toString() + "/" + sourceFolder;
             project.addCompileSourceRoot(sourceJavaFolder);


wing328:

deblockt:
Thanks for this informations, i'have  correct thé issue

wing328:
@ePaul when you've time, would you please review this change? Thanks.

ePaul:
Sorry, I was a bit occupied with other stuff the last time (and I'm a bit out-of-the loop with Swagger-Codegen).

Okay, so for understanding what the touched code does ... if you have the parameter ` <addCompileSourceRoot>true</addCompileSourceRoot>` (which actually is the default), it tries to guess the location where the generated sources were generated, and then adds this path to the compile class path for further compilation steps.

The change in this PR is now (relative to the output directory) from `src/gen/java` to `src/main/java` (when there is no source folder indicated in the options), which seems to work better for all the java-based "languages", and is just as bad as before for the non-java based ones.

For this, I can give a :+1:.

In general, I don't think it is really a good idea to generate a full maven project under target/generated-sources/swagger, and then just include some folders of that generated project into the main project's compile path (basically ignoring the generated pom.xml and stuff around it).

I see two plausible ways to use the maven plugin for code generation:

* Use this in a regular build (each time), generate your sources directly under target/generated-sources/swagger/‹package› (i.e. basically source folder is the empty string, don't generate any project files) and use addCompileSourceRoot=true to add the result into the source path of the main project. (Of course this only works if you have a template which doesn't require editing any files afterwards.)
* Do it once: generate a whole maven project from a swagger definition file, and then afterwards use just the generated files for building. This potentially also allows editing the generated files afterwards.
  In this case there is no point of adding the generated sources to the original project's compile source root.

wing328:
Targeting 2.3.0 instead as it's a breaking change

Issue #4690, opened 2017-02-01
[aspnetcore] packageName or sourceFolder specified in the config do not change the project folder name


abkordata:
After generating aspnetcore server code the packageName provided in the configuration isn't reflected on the source project folder name, instead the folder name uses the default name: "IO.Swagger".
Same thing happens when defining the sourceFolder in the configuration json file.
The structure should be:
Api/src/Title.Case/ 
Instead it defaults to
Api/src/IO.Swagger/

with config.json being:
```json
{
"packageName": "Title.Case"
}
```

Note: All the templates seem to update correctly to the specified packageName.

##### Swagger-codegen version
using swagger-codegen 2.2.1 as well as the master branch and 2.3 branch 

##### Command line used for generation
`java -jar swagger-codegen-cli.jar generate -i swagger.json -l aspnetcore -c config.json -o Api`

wing328:
@abkordata thanks for reporting the issue. 

Have you tried passing the `sourceFolder` option to setup a proper source folder based on package name?
```
	sourceFolder
	    source folder for generated code (Default: src/IO.Swagger)
```
Ref: `java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar config-help -l aspnetcore`

abkordata:
Yes I have tried with the sourceFolder option like so:
```json
{
"sourceFolder": "src/Tilte.Case"
}
```
And I also tried having both packageName and sourceFolder like so:

```json
{
"packageName": "Title.Case"
"sourceFolder": "src/Tilte.Case"
}
```
None of these changed the folder name properly. The folder name still defaults to IO.Swagger

wing328:
@abkordata the following lines show that the `sourceFolder` is used for API and model folders:

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AspNetCoreServerCodegen.java#L118-L126

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AspNetCoreServerCodegen.java#L48-L50 obtains the setting and put it into `sourceFolder`

I can't really tell what's causing the issue at the moment.

cc @jimschubert

jimschubert:
Yeah that's weird. I've generated with different source folder to evaluate an older PR. Maybe something has changed in DefaultCodegen.

[...]

PR #4699, opened 2017-02-03
[aspnetcore] resolve source/packageName issues

jimschubert:
Outputs warnings about ASP.NET Core's src/ requirement, and adjusts
sourceFolder if the input doesn't start with src/.

As a consequence of this change, apiFileFolder and modelFileFolder may
possibly return different values if users are setting both sourceFolder
and packageName. This could be considered a breaking change in those
sceanrios.

### PR checklist

### Description of the PR

see #4690 

This fix resolves the issue with `packageName` not being respected. I've also added some log messages in a new `setSourceFolder` override. However, ASP.NET Core requires a rigid source directory structure, and it appears as though changing `sourceFolder` can cause problems. We may want to override `setSourceFolder` to prevent overwriting it from "src".


wing328:
@jimschubert thanks for the prompt fix.

Issue #4698, opened 2017-02-03
[JAVA] RxJava2 support 


ber4444:
##### Description

When using the configuration settings, 
```
  "library": "retrofit2",
  "useRxJava": true,
```
we are getting these dependencies:
```
ext {
    retrofit_version = "2.0.2"
    rx_java_version = "1.1.3"
}
dependencies {
    compile "com.squareup.retrofit2:retrofit:$retrofit_version"
    compile "com.squareup.retrofit2:converter-gson:$retrofit_version"
    compile "com.squareup.retrofit2:adapter-rxjava:$retrofit_version"
    compile "io.reactivex:rxjava:$rx_java_version"
```
However, we would like to add a useRxJava2 setting which would result in:
```
ext {
    retrofit_version = "2.0.2"
    rx_java_version = "2.0.1"
}
dependencies {
    compile "com.squareup.retrofit2:retrofit:$retrofitVersion"
    compile "com.squareup.retrofit2:converter-gson:$retrofitVersion"
    compile "com.jakewharton.retrofit:retrofit2-rxjava2-adapter:1.0.0"
    compile "io.reactivex.rxjava2:rxjava:$rx_java_version"
```

##### Related issues

Similar to this PR:
https://github.com/swagger-api/swagger-codegen/pull/1928/files

cbornet:
Would you do a PR for that ?

wing328:
Another option is to customize the template: https://github.com/swagger-api/swagger-codegen/blob/1dd9ee39af6c4e2f9c016a94d16ca47a88a9cb0d/modules/swagger-codegen/src/main/resources/Java/libraries/retrofit2/build.gradle.mustache#L102 and then use the `-t` option with the customized templates.

ber4444:
I'll do the PR. Not as simple as @wing328 suggests because the class and package names changed as well.

wing328:
@ber4444 thanks! I think we'll need to provide a CLI option, e.g. useRxJava2, so that users still have the option to use RxJava 1.x. Let me know if you need help on that.

wing328:
@ber4444 I'll review your PR and let you know if I've any question.

ber4444:
I updated the PR based on your feedback.

wing328:
@ber4444 I've found 2 minor issues: https://github.com/swagger-api/swagger-codegen/issues/4708. Let me know if you need help fixing it.

wing328:
Closed via #4743 

[...]

PR #4708, opened 2017-02-03
rx2 support

ber4444:
### PR checklist

### Description of the PR

https://github.com/swagger-api/swagger-codegen/issues/4698

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
@@ -48,7 +48,6 @@ public class JavaClientCodegen extends AbstractJavaCodegen
 
         cliOptions.add(CliOption.newBoolean(USE_RX_JAVA, "Whether to use the RxJava adapter with the retrofit2 library."));
         cliOptions.add(CliOption.newBoolean(USE_RX_JAVA2, "Whether to use the RxJava2 adapter with the retrofit2 library."));
-        cliOptions.add(CliOption.newBoolean(DO_NOT_USE_RX, "If you do not use RxJava[2], Retrofit's default Call type will be generated instead.")); // setting this to false will have undefined results
         cliOptions.add(CliOption.newBoolean(PARCELABLE_MODEL, "Whether to generate models for Android that implement Parcelable with the okhttp-gson library."));
         cliOptions.add(CliOption.newBoolean(USE_PLAY24_WS, "Use Play! 2.4 Async HTTP client (Play WS API)"));
         cliOptions.add(CliOption.newBoolean(SUPPORT_JAVA6, "Whether to support Java6 with the Jersey1 library."));
@@ -90,7 +89,9 @@ public class JavaClientCodegen extends AbstractJavaCodegen
     public void processOpts() {
         super.processOpts();
 
-        if (additionalProperties.containsKey(USE_RX_JAVA)) {
+        if (additionalProperties.containsKey(USE_RX_JAVA) && additionalProperties.containsKey(USE_RX_JAVA2)) {
+            LOGGER.warn("You specified both RxJava versions 1 and 2 but they are mutually exclusive. Defaulting to v2.");
+        } else if (additionalProperties.containsKey(USE_RX_JAVA)) {
             this.setUseRxJava(Boolean.valueOf(additionalProperties.get(USE_RX_JAVA).toString()));
         }
         if (additionalProperties.containsKey(USE_RX_JAVA2)) {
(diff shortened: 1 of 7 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
```diff
@@ -43,6 +47,8 @@ public JavaClientCodegen() {
         modelPackage = "io.swagger.client.model";
 
         cliOptions.add(CliOption.newBoolean(USE_RX_JAVA, "Whether to use the RxJava adapter with the retrofit2 library."));
+        cliOptions.add(CliOption.newBoolean(USE_RX_JAVA2, "Whether to use the RxJava2 adapter with the retrofit2 library."));
+        cliOptions.add(CliOption.newBoolean(DO_NOT_USE_RX, "If you do not use RxJava[2], Retrofit's default Call type will be generated instead.")); // setting this to false will have undefined results
```
wing328: @ber4444 I suggest we don't expose the option `DO_NOT_USE_RX` as it should be not set by the users if I understand correctly.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
```diff
@@ -87,6 +93,12 @@ public void processOpts() {
         if (additionalProperties.containsKey(USE_RX_JAVA)) {
             this.setUseRxJava(Boolean.valueOf(additionalProperties.get(USE_RX_JAVA).toString()));
         }
+        if (additionalProperties.containsKey(USE_RX_JAVA2)) {
+            this.setUseRxJava2(Boolean.valueOf(additionalProperties.get(USE_RX_JAVA2).toString()));
+        }
```
wing328: If both USE_RX_JAVA and USE_RX_JAVA2 are incorrectly set to true, maybe it would be a good idea to log a warning/error and only set one of them to true only.

wing328:
@ber4444 I've done some tests and the results look good.

Thanks for your contribution.

wing328:
Actually I discovered some issues. Let me roll back first before working on a fix.

modules/swagger-codegen/src/main/resources/Java/libraries/retrofit2/pom.mustache
```diff
@@ -213,6 +213,18 @@
...
     {{/useRxJava}}
+    {{#useRxJava2}}
+    <dependency>
+      <groupId>io.reactivex.rxjava2</groupId>
+      <artifactId>rxjava</artifactId>
+      <version>${rxjava-version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.jakewharton.retrofit</groupId>
+      <artifactId>com.jakewharton.retrofit</artifactId>
```
wing328: I think this should be `<artifactId>retrofit2-rxjava2-adapter</artifactId>` instead.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
```diff
@@ -84,9 +89,17 @@ public String getHelp() {
...
+            LOGGER.warn("You specified both RxJava versions 1 and 2 but they are mutually exclusive. Defaulting to v2.");
+        } else if (additionalProperties.containsKey(USE_RX_JAVA)) {
             this.setUseRxJava(Boolean.valueOf(additionalProperties.get(USE_RX_JAVA).toString()));
         }
+        if (additionalProperties.containsKey(USE_RX_JAVA2)) {
+            this.setUseRxJava2(Boolean.valueOf(additionalProperties.get(USE_RX_JAVA2).toString()));
+        }
+        if (additionalProperties.containsKey(DO_NOT_USE_RX)) {
+            this.setDoNotUseRx(Boolean.valueOf(additionalProperties.get(DO_NOT_USE_RX).toString()));
+        }
```
wing328: I think these 3 lines can be removed as we no longer allow setting DO_NOT_USE_RX via option.

And after processing all the RX-related option, we will need to update "doNotUseRx" in additionalProperties so that it has a proper value (true or false)

I tested without any RX-related option and got the following compilation errors:
```
[ERROR] /private/var/tmp/java/useRxJava20/src/main/java/io/swagger/client/api/PetApi.java:[127,3] cannot find symbol
[ERROR] symbol:   class Observable
[ERROR] location: interface io.swagger.client.api.PetApi
```

wing328:
I've discovered 2 minor issues. Please work on those and submit a new PR for the enhancement.

Issue #4509, opened 2017-01-06
[Jaxrs-spec] Improvements for Jaxrs-spec


jfiala:
##### Description

* the api is generated as class, but IMHO this should be an interface!?
* javax.ws.rs.core.Response is used instead of the real response data type
* @ApiResponse uses wrong response type (the return type, not the response type of the contract)
* src/gen/java is not added to the build path
* "Attachment" is used for attachments, but no import is added for this (I think org.apache.cxf.jaxrs.ext.multipart.Attachment is referenced, but this is CXF-specific)
(see also #4327)

##### Swagger-codegen version

2.2.2-SNAPSHOT

##### Related issues
#4327

##### Suggest a Fix
* generate api as interface
* use real response data types instead of javax.ws.rs.core.Response
* fix * @ApiResponse data types
* add src/gen/java to the buildpath
* look for alternatives for JAXRS-compliant alternatives to org.apache.cxf.jaxrs.ext.multipart.Attachment or comment it out

wing328:
> generate api as interface (which then can be easily implemented)

I believe this will be a breaking change so better target 2.3.0 for this enhancement.

jfiala:
If you don't want a breaking change, we could also add the interface with a suffix (e.g. "I" - "MyServiceI") and the class implementing the interface would only generate if it doesn't exist, thus making it non-breaking or add the interface approach as a library or separate language?

jfiala:
Implemented solution at PR #4717, please review and advise if you prefer a non-breaking copy of the jaxrs-spec language or a breaking change for 2.3.0.

ePaul:
Please use ``{{baseName}}` instead of ``{{paramName}}` in all annotations which refer to the actual on-the-wire parameter name. (See #4898.)

mamghari:
Hi @jfiala,

About the point:
-  "javax.ws.rs.core.Response is used instead of the real response data type"

I think it is generally a bad practice for RESTful web services because whitout the Response object how do you manage the HTTP code returned ? How do you add HTTP Headers in your response?

It could be another option for the plugin but it is a bad practice... this is my point of view.

Tillerino:
Hi @mamghari 

[...]

PR #4717, opened 2017-02-05
[Jaxrs-spec] Update jaxrs-spec to interface/impl pattern #4509

jfiala:
### PR checklist

> I filed against master, please review & let me know if I should do against 2.3.0. 
I'd prefer setting up a temporary language so we can have the change in 2.2.2.

### Description of the PR

see #4509


jfiala:
@wing328 I filed against master for review, I'd like to have this in 2.2.2 - should I create a temporary language for this or do you want to stick with 2.3.0?

wing328:
@jfiala from what I can tell it's a breaking change so I suggest we file it against 2.3.0 instead (we'll release 2.3.0 after the 2.2.2 release)

jfiala:
@wing328 you're right, I can create a copy of the whole language so we can make it into 2.2.2 (as "jaxrs-spec-interface" or something like that) or we file it against 2.3.0, what do you prefer? I have a slightly preference for getting this into 2.2.2 if you agree (by copying).

wing328:
@jfiala actually I want to freeze the current master and release 2.2.2. The community is very active (which is definitely a good thing) and we're seeing at least several new PRs (enhancements, bug fixes) every day so my take is to release 2.2.2 earlier and push those enhancements, bug fixes to the next release instead.

jfiala:
@wing328 OK, then lets schedule this for 2.3.

jfiala:
closed, will do again against 2.3.0.

moshelav:
Does the suggestion from [https://github.com/swagger-api/swagger-codegen/issues/4509] was entered to specific release, if so which one?

As we are experiencing jersey (2) validation errors as swagger file type generated ReST with @Multipart Inputstream AND Attachment arguments.  By manually removing the generated Attachment argument seems it work... so does the following suggestion entered.

_look for alternatives for JAXRS-compliant alternatives to org.apache.cxf.jaxrs.ext.multipart.Attachment or comment it out
as there is currently no standardized way for supporting multipart/form-data (see https://java.net/jira/browse/JAX_RS_SPEC-413), I changed formParams.mustache to

**@FormParam(value = "{{paramName}}") InputStream {{paramName}}InputStream**
This should be a sufficient standard hint, although it will have to be adapted for CXF/Jersey/Resteasy then._

Issue #4422, opened 2016-12-18
[JAXRS-CXF] When setting dateLibrary to "legacy" java.util.Date is overridden by javax.xml.datatype.XMLGregorianCalendar


MatanRubin:
Running config-help for the jaxrs-cxf generator shows the following options for dateLibrary:

``` bash
$ java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar config-help -l jaxrs-cxf
...
    dateLibrary
            Option. Date library to use
                joda - Joda
                legacy - Legacy java.util.Date
                java8-localdatetime - Java 8 using LocalDateTime (for legacy app only)
                java8 - Java 8 native
...
```

When using `joda`, `java8` and `java8-localdatetime` everything looks great.
However, when using `legacy`, instead of getting `java.util.Date` type for dates, I get `javax.xml.datatype.XMLGregorianCalendar`.

Looking at the code, it looks like `DateTime` is purposefuly overridden (see `modules\swagger-codegen\src\main\java\io\swagger\codegen\languages\JavaCXFServerCodegen.java +79`.
I don't really understand why `DateTime` is overridden - not sure if this is a bug or just something intentional that I don't understand. In any case, the help message and actual behavior of the generator are not consistent, so this should be treated as a bug.

Using swagger-codegen-2.2.1

jfiala:
The reason for the usage of javax.xml.datatype.XMLGregorianCalendar is that this is the default for the CXF XML based processing. 

For CXF I think there should be more flavours:
- legacy =  javax.xml.datatype.XMLGregorianCalendar as this is the default date type used by CXF itself using wadl2java
- java8
- joda
- java.util.Date?

If there is support for java.util.Date needed, I think this should be done as a fourth library.

The current implementation at master though is actually a mixture of joda and XMLGregorianCalendar as @MatanRubin wrote, so a cleanup is a good idea.

PR #4724, opened 2017-02-06
[JAXRS-CXF] [bug #4422] use java.util.Date when selecting dateLibrary=legacy

MatanRubin:
according to the help message setting dateLibrary to 'legacy' should
make the code generator use java.util.Date.
before this change, when setting dateLibrary to 'legacy' made the
generator use javax.xml.datatype.XMLGregorianCalendar.
now, the generator uses java.util.Date such that the documentation and
behavior are consistent.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaCXFClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaCXFClientCodegen.java
@@ -21,30 +21,30 @@ import io.swagger.codegen.languages.features.LoggingTestFeatures;
 import io.swagger.models.Operation;
 
 public class JavaCXFClientCodegen extends AbstractJavaCodegen
-        implements BeanValidationFeatures, JaxbFeatures, GzipTestFeatures, LoggingTestFeatures 
-{   
+        implements BeanValidationFeatures, JaxbFeatures, GzipTestFeatures, LoggingTestFeatures
+{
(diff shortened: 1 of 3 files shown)


MatanRubin:
@wing328 and @hiveship, I thought this was the only change needed to make codegen work as documented, but now I see other things that I do not understand, maybe you guys could help shed some light...

For example, in `JavaCXFServerCodegen` +80 I see the following:

```java
        typeMapping.put("date", "LocalDate");
        importMapping.put("LocalDate", "org.joda.time.LocalDate");
```

This forces the user to use joda time without regard to what was set in the `dateLibrary` configuration.  

I see the same thing in `JavaCXFClientCodegen.java` and `JavaJAXRSSpecServerCodegen.java`.
Am I missing something? My inclination is to remove these type and import mappings altogether...

Thanks in advance for your assistance.

wing328:
I think those datetime option has been added to Java API client but not for **all** Java server generators.

cc @cbornet to see if he's any input on this.

jfiala:
The reason for the usage of javax.xml.datatype.XMLGregorianCalendar is that this is the default for the CXF XML based processing. 

For CXF I think there should be more flavours:
- legacy =  javax.xml.datatype.XMLGregorianCalendar as this is the default date type used by CXF itself using wadl2java
- java8/joda

If there is support for java.util.Date needed, I think this should be done as a fourth library.

The current implementation at master though is actually a mixture of joda and XMLGregorianCalendar as @MatanRubin wrote, so a cleanup is a good idea.

wing328:
@jfiala thanks for reviewing the change

@MatanRubin thanks for the PR, which has been merged into master.

PR #4747, opened 2017-02-08
[Java] Change Boolean fields of CodegenParameter and CodegenResponse …

sreeshas:
…to boolean.

CodegenOperation and CodegenProperty classes have boolean fields instead of Boolean.
This fix changes Boolean fields of CodegenParameter and CodegenResponse to boolean as well.

Boolean fields are not necessary. Other classes interacting with them have to check for null
before interacting with them which is unnecessary and leads to ugly code.

### PR checklist

### Description of the PR

CodegenOperation and CodegenProperty classes have boolean fields instead of Boolean.
This fix changes Boolean fields of CodegenParameter and CodegenResponse to boolean as well.

Boolean fields are not necessary. Other classes interacting with them have to check for null
before interacting with them which is unnecessary and leads to ugly code.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenParameter.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenParameter.java
@@ -6,28 +6,29 @@ import java.util.Map;
 import java.util.List;
 
 public class CodegenParameter {
-    public Boolean isFormParam, isQueryParam, isPathParam, isHeaderParam,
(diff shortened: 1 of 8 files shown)


fehguy:
But I think it's not the same in the templates

{{booleanValue}} with null => "".  With a boolean, false => "false" and true => "true"

sreeshas:
@fehguy I checked them with templates. It is the same behavior. 
I also ran all tests and they worked fine.

wing328:
@sreeshas I ran some tests and don't see any change to the auto-generated code due to this PR.

PR #4749, opened 2017-02-08
[C#] remove initialization of default temp folder in Configuration object

### LINKED ISSUE #4748: CSharp template generates SDK requiring access to temp folder that raise exception when environment has restricted access.
##### Description
The C# SDK generated from the template requires system permission to get the temp folder. However some hosting environment doesn't grant permission to access system folders. That makes the ApiClient initialization fails.

##### Swagger-codegen version
latest.

##### Related issues

https://github.com/square/connect-csharp-sdk/issues/5

##### Suggest a Fix
Two options here:
1) Let caller handle the `Steam` instead of saving it in temp folder. So the SDK doesn't need to obtain the temp folder path. **(Preferred)**
2) Still save the `Stream`, but only obtain the temp folder path inside the `Deserialize` when it's needed. This is actually just a small change: [set `_templateFolderPath` in `Configuration` to be `null` instead of `Path.GetTempPath()` on line 237](https://github.com/swagger-api/swagger-codegen/blob/90512e6326d5e7aff4dd21a649b8616012da9974/modules/swagger-codegen/src/main/resources/csharp/Configuration.mustache#L237)

wing328:
### PR checklist

### Description of the PR

To fix #4748


bunnyc1986:
lgtm. ship it!

jimschubert:
Looks good. iirc, I did the same thing in my configuration refactor branch.

wing328:
Thanks for reviewing the PR, which has been merged into master.

Issue #4727, opened 2017-02-06
Package names in the finch generator should not contain petstore


clasnake:
I found that in FinchServerCodegen, the packageName and invokerPackage are set to "io.swagger.petstore...".
petstore is just the sample and I guess the package names here should not contain "petstore".

clasnake:
@wing328 @jimschubert Just had a quick look at the finch generator and found the above issue.

clasnake:
@wing328 I guess even the default should not contain petstore. Can be "io.swagger"

wing328:
@clasnake you're definitely right. What default would you suggest?

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/FinchServerCodegen.java#L18

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/FinchServerCodegen.java#L23

clasnake:
@wing328 How about "io.swagger"?

wing328:
@clasnake sounds better to me as it doesn't contain "petstore" 😄 

Please feel free to submit the PR with updated petstore sample for Finch.

wing328:
Closed via #4755

PR #4755, opened 2017-02-08
Fix finch package name.

clasnake:
As described in https://github.com/swagger-api/swagger-codegen/issues/4727#issuecomment-278319767


clasnake:
@wing328 pr ready. Please take a look.

wing328:
@clasnake I ran some tests and the result is good. Thanks for the PR.

PR #4547, opened 2017-01-12
[Swift3] Don't crash on invalid responses

philippelatulippe:
### PR checklist

- [x] Read the [contribution guildelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. Code generation succeeds and the tests pass, but in the PromiseKit version there seems to be errors that seem to indicate an outdated version of PromiseKit. Also the rxswift version has a test that expects a library to return an error when receiving a null body from a DELETE, but this the null body is handled just fine.
- [x] Filed the PR against the correct branch: `2.3.0` since there are breaking changes

### Description of the PR

Currently, generated swift3 code will call fatalError() when it encounters data that does not match the spec. See issue #4382. This PR implements my suggested fix. Decode methods return an enum with two cases: success, which contain a result, and failure, which contain an error. 

All the relevant changes are in the files *modules/swagger-codegen/src/main/resources/swift3/Models.mustache* and *swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache*.

### API Changes

The enum `ErrorResponse` now has a new case: `DecodeError`, which itself contains an enum describing the error. To accommodate this change, `ErrorResponse.Error` has been renamed `ErrorResponse.HttpError`. I have also incorporated the labels from PR #4502, but I reverted the change from an enum to a struct, since we now have multiple kinds of errors. Has the swift3 generator been released? If yes, then this PR is a minor **breaking change**.

The definitions relevant for end users are:

```swift
public enum ErrorResponse : Error {
    case HttpError(statusCode: Int, data: Data?, error: Error)
    case DecodeError(response: Data?, decodeError: DecodeError)
}

public enum DecodeError {
    case typeMismatch(expected: String, actual: String)
    case missingKey(key: String)
    case parseError(message: String)
}

public enum Decoded<ValueType> {
    case success(ValueType)
    case failure(DecodeError)
}
```

PR #4762, opened 2017-02-09
[Swift3] null enumerations should not cause failure

wwwjfy:
### PR checklist

### Description of the PR

As all other the fields can be null, I think an enumeration field can also be null. The users can decide whether that's a problem. This is to be consistent with previous behaviors.
This was introduced in #4547 (@philippelatulippe)

The real change is only [one line](https://github.com/swagger-api/swagger-codegen/pull/4762/files#diff-940535e66d162a893df844fd9b6b45ffR122), but I cannot help but fix format issues.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/swift3/APIs.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift3/APIs.mustache
@@ -36,7 +36,7 @@ open class RequestBuilder<T> {
     let isBody: Bool
     let method: String
     let URLString: String
-    
+
     /// Optional block to obtain a reference to the request's progress instance when available.
     public var onProgressReady: ((Progress) -> ())?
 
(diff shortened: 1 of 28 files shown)


philippelatulippe:
Good catch, if the field is null it will return a _typeMismatch_, which is incorrect for _decodeOptional()_. But will this also return `nil` in the case that the field in the data is of the wrong type? I think the best behaviour is to have type mismatches return an error and only null or missing fields return `nil`. Perhaps this needs to be debated? The pros are that errors can be caught early, but the parser is less permissive.

wwwjfy:
Thanks for reviewing the code. You're right, it shouldn't return success if the type mismatched.
My intention is to return success when it's null, missing the case of mismatched type.
I'll commit a fix later.

wing328:
cc @jaz-ah @jgavris

wwwjfy:
I pushed a fix.

Another similar place (which is left intact) is https://github.com/swagger-api/swagger-codegen/pull/4762/files#diff-940535e66d162a893df844fd9b6b45ffL202
The function signature is
`static func decodeOptional<T: RawRepresentable, U: AnyObject where T.RawValue == U>(clazz: T, source: AnyObject) -> Decoded<T?>`

I don't know when this function will be called. I'm a beginner in Swift. Isn't this case already covered in the function without `U: AnyObject`?

jaz-ah:
+1 thx for the contribution @wwwjfy

Issue #4796, opened 2017-02-14
Generate better example code


bartkummel:
##### Description

The code generated does not leverage all example and default values from the Swagger spec.

##### Swagger-codegen version

2.2.1

##### Suggest a Fix

Use examples, default values and enum values from Swagger spec to generate better example values.

wing328:
@bartkummel I think some generators do have better support for default and example values.

May I know which generators you find inadequate support of example/default value?

bartkummel:
I have tried the nodejs generator. I have also created a pull request on the DefaultGenerator, which is subclassed by the specific generators, as I understand it. I have tested my PR with the nodejs generator and it works as I expect. See https://github.com/swagger-api/swagger-codegen/pull/4797

wing328:
Closed via #4797

PR #4797, opened 2017-02-14
Improved ExampleGenerator

bartkummel:
### PR checklist

See issue #4796 

### Description of the PR

Improved ExampleGenerator:
- Now takes into account enum and uri/url formats for strings.
- Uses example for referenced objects if available.
- Proper examples get generated for specific numeric formats, because more specific formats now get checked before generic format.
- Honors min and max values for numerical properties, if set.


modules/swagger-codegen/src/main/java/io/swagger/codegen/examples/ExampleGenerator.java
```diff
@@ -20,71 +22,107 @@
...
-import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
 public class ExampleGenerator {
+    private static final Logger log = LoggerFactory.getLogger(ExampleGenerator.class);
```
wing328: @bartkummel thanks for the PR. Just a minor suggestion. We use `logger` instead of `log` in other Java class files. Do you mind using the same (ie. `logger`) to make it consistent with other files?

bartkummel: Sure! I've renamed `log` to `logger` in the latest commit.

wing328:
@bartkummel thanks for the PR. I like the new feature you added: randomized example values (numeric), but this may introduce extra noise during change review. I'll merge this and see if the community also has feedback on this.

ePaul:
The random integers have the effect that now after each generation we get different numbers. Currently those examples seem to be not used in many languages (just NodeJS, as far as I can see), but it is already annoying.

If we need randomness, could we make it deterministic? I.e. start the RNG always with the same seed (per ExampleGenerator instance, e.g.), so same input gets same output? This would turn down the noise.

Issue #4513, opened 2017-01-06
[JaxRsCxfServer] Improvements for generated Unittests and Attachments


jfiala:
##### Description
From #4088: 

1.) currently in the tests the api invocations are commented out and thus cannot be compiled.
Wouldn't it be good to activate them and run the CI maven build with -DskipTests?

2.) Handling CXF Attachments is not totally intuitive, do you think it makes sense to add an assistance snippet here like this at api_test.mustache (for #isFile):

``` java
InputStream mybytesInput = new FileInputStream("....");
org.apache.cxf.jaxrs.ext.multipart.ContentDisposition cd = new org.apache.cxf.jaxrs.ext.multipart.ContentDisposition("attachment;filename=file.txt");
Attachment attachment = new Attachment("file", mybytesInput, cd);
```

or does that go too far and we should leave with the default Attachment attachment = null;?

##### Swagger-codegen version

2.2.2-SNAPSHOT

##### Related issues
#4088 Minor improvements JaxRsCXFServer

wing328:
1) If I remember correctly, there are certain scenarios (edge cases) in which the test cases won't compile right out of the box so that's why the tests are commented out to start with.

2) I like the idea of the assistance snippet so as to save the developer's time 👍

jfiala:
For JaxRS-CXF with the contracts we are using for tests everything compiled smoothly, what about activating the tests for Jaxrs-CXF so we proof they compile?

wing328:
@jfiala ok, let's proceed with that.

jfiala:
added PR to compile the api invocations in unittests for cxf clients.

jfiala:
Added the snippet to deal with CXF Attachments.

Sidenote: the currently used PetApi.uploadFile with the FormParam additionalMetadata doesn't work with CXF as it only supports one Attachment parameter. If the parameter additionalMetadata  is removed, the unittest works.

PR #4823, opened 2017-02-18
[Jaxrs-cxf] Unittests: uncomment api invocations, add Multipart snippet #4513

jfiala:
### PR checklist

### Description of the PR
see #4513

* Uncomment api invocations in unittests 
* Also add processing for void to AbstractJavaCodegen.java so it is available to all languages.
* Add snippet to deal with CXF Attachments

Sidenote: the currently used PetApi.uploadFile with the FormParam additionalMetadata doesn't work with CXF as it only supports one Attachment parameter. If the parameter additionalMetadata  is removed, the unittest works.


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java
```diff
@@ -816,6 +816,23 @@ public void postProcessParameter(CodegenParameter parameter) { }
...
+        @SuppressWarnings("unchecked")
+        Map<String, Object> operations = (Map<String, Object>) objs.get("operations");
+        if (operations != null) {
+
+            @SuppressWarnings("unchecked")
+            List<CodegenOperation> ops = (List<CodegenOperation>) operations.get("operation");
+            for (CodegenOperation operation : ops) {
+
+                if (operation.returnType == null) {
+                    operation.returnType = "void";
```
wing328: This will break the Java (okhttp) API client doc
```
 ### Return type
 
-null (empty response body)
+[**void**](.md)
```
We'll need to use `x-java-is-response-void` to check instead. I'll create a separate issue for track this.

wing328: The code for `sync` call will also break:
```
l, null);
-        return apiClient.execute(call);
+        Type localVarReturnType = new TypeToken<void>(){}.getType();
+        return apiClient.execute(call, localVarReturnType);
     }
```

@jfiala what about doing it in the Java-related generators one by one instead of doing it in "abstract" Java generator.

For Java API client, any concern to keep the current approach? Can you also remind me the limitation with the current approach?

jfiala: No problem lets move it to jaxrs first and then do it one by one for each language.
So I can scope it for jaxrs-cxf here and move it to the CXF language?

wing328: sounds good to me.

jfiala:
@wing328 I moved the check from AbstractJavaCodegen.java to JavaCXFClientCodegen.java.
It is already present in AbstractJavaJAXRSServerCodegen.java, so it only had to be moved to the client codegen.

jfiala:
closing as compare view is not refreshed correctly.

PR #4843, opened 2017-02-23
Added csharp Configuration = {{packageName}}.Client.Configuration using clause

staticcat:
Added csharp Configuration = {{packageName}}.Client.Configuration using clause. This allows the models to include a class called Configuration. This change could be done by fully specifying the location of the Configuration class each time its referenced, however this would mean a number of updates through out the mustache files, and for developers to remember each subsequent update. 

Tested against the petstore example. Generation works and resulting client compiles.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/csharp/README.mustache
+++ b/modules/swagger-codegen/src/main/resources/csharp/README.mustache
@@ -49,6 +49,7 @@ Then include the DLL (under the `bin` folder) in the C# project, and use the nam
 ```csharp
 using {{packageName}}.{{apiPackage}};
 using {{packageName}}.Client;
+using Configuration = {{packageName}}.Configuration;
 using {{packageName}}.{{modelPackage}};
 ```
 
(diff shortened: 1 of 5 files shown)


wing328:
@staticcat thanks for the PR. what about adding `Configuration` to the list of reserved words in [this line](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java#L68) instead?

staticcat:
Can be done for sure :)

What about the comment about it being a work around until using a fully
qualified name.

// set "client" as a reserved word to avoid conflicts with IO.Swagger.Client
// this is a workaround and can be removed if c# api client is updated to
use
// fully qualified name

What is preferred?

wing328:
We'll do that (fully-qualified names) in the 2.3.0 release as 2.3.0 will have the C# API client refactor: https://github.com/swagger-api/swagger-codegen/pull/3530

Please give https://github.com/swagger-api/swagger-codegen/pull/3530 a try and share with us if you've any feedback.

staticcat:
Will do. Thanks for the redirect to the refactor.

staticcat:
The branch looks good. Some great work going on there. Its lacking things however that we require for our client and test generation in csharp. I will move to #3530 pr and discuss things there. I think we can close this pr down and group everything up into one pr.

PR #4153, opened 2016-11-07
Update TypeScript Fetch TypeScript version to 2.0.7

leonyu:
### PR checklist

- [ ] Read the [contribution guildelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [ ] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [ ] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

(details of the change, additional tests that have been done, reference to the issue for tracking, etc)
* Update TypeScript Fetch TypeScript version to 2.0.7
* Fix support for underscore casing in POST body

Issue #3921, opened 2016-10-04
TypeScript Fetch client doesn't support multipart/form-data body


janslow:
##### Description

If a POST operation accepts `multipart/form-data`, the `typescript-fetch` client posts an `application/x-www-form-urlencoded` body instead of form data.

##### Swagger-codegen version

Version currently deployed on [editor.swagger.io] 

##### Swagger declaration file content or url
```yaml
swagger: '2.0'
info: { version: 0.1.0, title: Multipart Upload }
paths:
  /:
    post:
      operationId: upload
      consumes:
        - multipart/form-data
      parameters:
        - name: file
          in: formData
          required: true
          type: file
      responses: { '200': { description: Successful upload } }
```

##### Command line used for generation

"Generate Client" -> "TypeScript Fetch" on [editor.swagger.io]

##### Steps to reproduce

1. Generate and compile the sample config
2. Call it with a file
3. Log the request, it has a `Content-Type` of `application/x-www-form-urlencoded` instead of `multipart/form-data`.

##### Suggest a Fix

Instead of setting the Content-Type and using `querystring` to serialize the contents (for `multipart/form-data` requests), it should just set the body to an instance of `FormData` instead of 

e.g., in `DefaultApiFetchParamCreactor#upload`, instead of :

```ts
let contentTypeHeader: Dictionary<string>;
contentTypeHeader = { "Content-Type": "application/x-www-form-urlencoded" };
fetchOptions.body = querystring.stringify({ 
  "file": params.file,
});
if (contentTypeHeader) {
  fetchOptions.headers = contentTypeHeader;
}
```

it should generate (as `Content-Type` is set to `multipart/form-data` automatically if using `FormData`):

```ts
const formData = new FormData();
formData.append("file", params.file);
fetchOptions.body = formData;
```

wing328:
@janslow thanks for reporting the issue. May I know if you've time to contribute the fix?

russtacular:
@wing328 I actually ran into the same issue myself and was in the process of putting together a PR for this and #4153 

I can try to pull something together in the next few days, if that's useful for you and @janslow

bvwells:
Just wondering whether there is any update with this issue?

Alexuy:
Hey, guys! What about this issue? When you planning to fix that bug?

Splash34:
+1

ds1371dani:
Any news on a fix?

currently we can't use any endpoint which uploads a file because of this :cry: .

[...]

PR #4852, opened 2017-02-24
Typescript-fetch: file upload and Typescript dependency updates

russtacular:
### PR checklist

### Description of the PR

This updates the typescript-fetch client with the latest dependencies based on work originally done in #4153 but in need of updates due to additional changes to `@types/isomorphic-fetch` that were causing build errors.

Additionally this adds the ability to upload files using the typescript-fetch client as requested in #3921. Given the limitations of the mustache template, I assumed that form parameters did not include files but, if so, it set a flag that caused the form parameters to be appended to a `FormData` object which is uploaded with a `Content-Type: 'multipart/form-data'` header opposed to the standard `'application/x-www-form-urlencoded'`


wing328:
@russtacular thanks for the PR.

@Vrolijkx @leonyu  would you please take a look when you've time? Thanks.

modules/swagger-codegen/src/main/resources/TypeScript-Fetch/package.json.mustache
```diff
@@ -6,16 +6,18 @@
   "browser": "./dist/api.js",
   "typings": "./dist/api.d.ts",
   "dependencies": {
-    {{^supportsES6}}"core-js": "^2.4.0",
-    {{/supportsES6}}"isomorphic-fetch": "^2.2.1"
+    "@types/node": "6.0.46",
```
Vrolijkx: Would be nice to also have this in the angular2 && angular clients.

modules/swagger-codegen/src/main/resources/TypeScript-Fetch/api.mustache
```diff
@@ -93,12 +93,25 @@ export const {{classname}}FetchParamCreator = {
...
-        {{/formParams}}
-        });
+        let formData: Dictionary<any> = {};
+        let isMultipartData = false;
+{{#formParams}}
+        {{#isFile}}isMultipartData = true;
+        {{/isFile}}formData["{{baseName}}"] = params["{{paramName}}"];
+{{/formParams}}
+
+        if (isMultipartData) {
```
Vrolijkx: Why not instead of this if statement in all the generated code. Use a Mustache if. This way you don't need the next strange code:
```
let isMultipartData = false;
isMultipartData = true;
```
maybe do something like this:
```
{{#isFile}}
  let body = new FormData();
  for (let key in formData) {
     body.append(key, formData[key]);
  }
 fetchOptions.body = body;
 contentTypeHeader = { "Content-Type": "multipart/form-data" };
{{/isFile}}
{{^isFile}}
   fetchOptions.body = querystring.stringify(formData);
   contentTypeHeader = { "Content-Type": "application/x-www-form-urlencoded" };
{{#isFile}}
```

russtacular: I originally used a very similar implementation to the one you suggested and also mentioned in #3921 but noticed that, since `{{#isFile}}` is scoped to the individual parameter, it didn't work for requests that included a file and regular form parameters. I checked again and, unless I'm looping through the individual `formParams`, `{{#isFile}}` always maps to false.

Is there a Mustache template variable available that tracks if _any_ of the included `FormData` attributes is of type file? If so, that would allow me to simplify the mustache template to essentially match your suggestion and get rid of the seemingly extraneous code.

modules/swagger-codegen/src/main/resources/TypeScript-Fetch/package.json.mustache
russtacular: I started looking at updating angular2 but I'm having issues building the copy in master using either TS 1.8.10 or my installed typescript@next developers build. I'm sure my unfamiliarity with building and using the angular generated libraries isn't helping either!

I can look into this but I think it may require more extensive changes and that it's likely better suited for a follow up PR. Is that okay with you?

modules/swagger-codegen/src/main/resources/TypeScript-Fetch/api.mustache
wing328: > Is there a Mustache template variable available that tracks if any of the included FormData attributes is of type file?

Not that I'm aware of. If it helps, we can introduce the variable.

modules/swagger-codegen/src/main/resources/TypeScript-Fetch/package.json.mustache
Vrolijkx: sure.

russtacular:
Hold off on this PR for a bit… I think it needs a small update and I'm also going to see about adding that form-level variable.

wing328:
@russtacular the CI (travis) failed with the following errors:
```

  StoreApiFactory
    without custom request options
      âœ“ should get inventory
    with custom request options
      âœ“ should get inventory


  26 passing (180ms)
  2 failing

  1) PetApi without custom request options should not set a content-type header when uploading a file:
     ReferenceError: File is not defined
      at Context.<anonymous> (dist/test/PetApi.js:49:120)

  2) PetApi with custom request options should not set a content-type header when uploading a file:
     ReferenceError: File is not defined
      at Context.<anonymous> (dist/test/PetApi.js:49:120)
```
Would you please take a look when you've time?

zetlan:
Is there a timeline on this one? It would solve the same problem I'm having with file uploads via generated code.

russtacular:
@wing328 @zetlan — Thanks for prodding me on this, I definitely still want to get this fix merged. The failing test, however, points out a bit of a problem with using `isomorphic-fetch` and trying to maintain compatibility with both browsers and Node.

Node doesn't support `FormData` which causes the node test to fail but the browser-based tests to pass. The options for fixing this are:
- Use a shim to try and support `FormData` in a Node environment. This doesn't seem like a great option except for the fact that should maintain compatibility with any server-side usage but makes the TS Fetch client reliant on more niche and also not as actively maintained shim libraries.
- Remove the expectation that TS Fetch work in Node environments unless they provide a `FormData` shim. That would keep client size down and we could conceivably still rely on packages like JSDom or similar to provide a `FormData` shim to enable Node-based tests to run.
- Replace `isomorphic-fetch` with `whatwg-fetch` and break its ability to run in Node. Given that there's already a TS Node client, maybe this is fine, although it does kind of kill the dream of isomorphic type/javascript.

Those are the options I see, although I could definitely be missing something obvious. I happy to submit a fix that uses any of the options I identified above but I don't feel like I can make that decision unilaterally since I don't know how others are using this client (at Bitly, we're only using the generated TS Fetch client within browsers).

AmazingTurtle:
Can someone have a look at this please? @russtacular

samples/client/petstore/typescript-fetch/tests/default/test/PetApi.ts
```diff
@@ -36,6 +43,18 @@ describe('PetApi', () => {
...
       });
 
+      it('should update Pet using url-encoded form', () => {
+        return new PetApi(parameterInspector).updatePetWithForm({ petId: fixture.id.toString(), name: "Snoopy" }).then(({arguments: [url, init]}) => {
+          return expect(init.headers['Content-Type']).to.equal('application/x-www-form-urlencoded');
+        });
+      });
+
+      it('should not set a content-type header when uploading a file', () => {
+        return new PetApi(parameterInspector).uploadFile({ petId: fixture.id, file: new File(['snoopy'], 'Snoopy.png')}).then(({arguments: [url, init]}) => {
```
AmazingTurtle: File is not defined.

wing328: @AmazingTurtle good catch!

wing328:
cc @TiFu  @taxpon @sebastianhaas @kenisteward  @Vrolijkx to review and see how we can proceed with this PR.

Alexuy:
Hey, guys! What about this issue? When you planning to fix that bug?

Place1:
Have a look at a fork I’ve made over at
https://github.com/Place1/swagger-codegen-typescript-browser

You can download a jar from the releases page

It works with multipart form data requests and fixes a bunch of other bugs.

TiFu:
@Alexuy  I personally currently don't have time to work on this. If you want to work on this, we'd be happy to accept an updated PR for this issue. We can also provide you with guidance on how to implement a suitable solution.

PR #4855, opened 2017-02-25
Feature/scala client async support

clasnake:
First version to provide async apis and the sync api implementation if actually a block call of the async api.
Next steps can be:
1. Add gradle and maven support in addition to sbt support.
2. Clean up useless imports.
3. Let users to choose which data types to use as the java clients.
4. More to go.


wing328:
@clasnake thanks for the PR. I'll cherry-pick the commits and let you know how it goes.

wing328:
Closed via https://github.com/swagger-api/swagger-codegen/pull/5115

PR #4858, opened 2017-02-26
Implicit headers for SpringCodegen 

huningd:
### PR checklist

### Description of the PR

This PR adds a new feature which makes it possible to choose between implicit headers or explicit headers. If you set the option implicitHeaders the generator will add @ApiImplicitParams and for each header a @ApiImplicitParam annotation instead of method parameters. 
Example:
_implicitHeaders = true_
`    
@ApiImplicitParams({
    @ApiImplicitParam(name = "enumHeaderStringArray", value = "Header parameter enum test (string array)",  dataType = "List<String>", paramType = "header"),@ApiImplicitParam(name = "enumHeaderString", value = "Header parameter enum test (string)",  dataType = "String", paramType = "header")
})
`
_implicitHeaders = false_
`
 @ApiParam(value = "Header parameter enum test (string array)"  , allowableValues="GREATER_THAN, DOLLAR") @RequestHeader(value="enum_header_string_array", required=false) List<String> enumHeaderStringArray
`
This feature is helpful because in spring mvc you often use filters and interceptors to handle header parameters. When you use filters and interceptors explicit headers in methods are needless.


wing328:
@huningd thanks for the contribution.

@cbornet @jfiala when you've time, please take a look. Thanks.

jfiala:
Looks good, although I didn't have such a use case so far.
In case this is useful for another language this could be ported then.

wing328:
@jfiala thanks for reviewing

@huningd thanks for the contribution, which has been merged into master.

PR #4874, opened 2017-02-28
Update spring samples with current code.

ePaul:
This is done so I can see my own changes better.

### PR checklist

- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)

### Description of the PR

This doesn't change anything, just updates the samples.
Feel free to merge or not, I'll base my next pull-requests affecting spring on this commit.

Issue #4870, opened 2017-02-28
Spring Boot Path Parameter in the Spec gets renamed into CamelCase in Code and the spec thats generated!


techlogix:
##### Description

Path parameter defined in the yaml spec for example "student-id" on the path and named as such in the parameter name - gets incorrectly renamed to "studentId" in the Spring Controller.

##### Swagger-codegen version

This is the latest version thats hooked to the editor at editor.swagger.io.
##### Steps to reproduce

Create any POST URI path with a path paramenter and have a name that's hypenated.

e.g. POST /students/student-id

Then generate Spring Boot - the API and Controller do not have that name for the parameter - instead it expects a Camel Case.

digitaldivvy:
This issue is a BLOCKER - a fix should be to respect the exact name of the path parameter in the Spring Code that's generated.

techlogix:
in the YAML snippet below - the path parameter is called 'transaction_id' - but when the Spring code is generated - the code 

***Java Code snippet ***
@PathVariable("transactionId") String transactionId)

this is incorrect

it SHOULD be

@PathVariable("transaction_Id") String transactionId)


*** the YAML Spec ***
```  /foo/{transaction_id}/bar:
    post:
      summary: |
        
       
      description: |

      consumes:
        - application/json
        - application/xml
      produces:
        - application/json
        - application/xml
      parameters:
        - in: path
          name: transaction_id
          description:
          required: true
          type: string
        - in: body
          name: account
          description: Add an account to an existing transaction. If one already exists this will error out.
          required: true
          schema:
            $ref: "#/definitions/Account"
```

wing328:
To fix it, can you try updating the following file?

[modules/swagger-codegen/src/main/resources/JavaSpring/pathParams.mustache](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/JavaSpring/pathParams.mustache)

and replace `@PathVariable("{{paramName}}")` with `@PathVariable("{{baseName}}")`

then use `-t` to specify the mustache template path.

techlogix:
I made the changes locally and tested it. Yes - the parameters are now showing up.

wing328:
@techlogix thanks for testing the fix.

@ePaul has filed https://github.com/swagger-api/swagger-codegen/pull/4875 to fix the in the template.

techlogix:
@wing328 You're welcome. 
[...]

PR #4875, opened 2017-02-28
Fix for #1219

### LINKED ISSUE #1219: springmvc: parameter names don't survive generation
**Summary**
When parameter names are sanitized, the HTTP side parameters can not be matched again with the Java side parameters.

**Details**
When generating springmvc or jaxrs Java server classes, each swagger parameter becomes an annotated parameter of the method, with both Swagger and Spring/JaxRS annotations.

Some of these annotations' arguments are meant to transport the parameter name as used in the protocol, but they are currently either empty (falling back to some default) or filled with the sanitized Java parameter name – which is wrong.

**Reproducible**

Here is my bug-trigger.yaml:

```
swagger: '2.0'
info:
  description: |
  version: 0.0.1
basePath: /api
definitions:
  example:
    type: object
    properties:
      foo:
        type: object
        properties:
          bar:
            type: string
paths:
  "/path/{foo_id}":
    parameters:
      - name: foo_id
        in: path
        type: string
    get:
      parameters:
       - name: If-None-Match
         in: header
         type: string
      responses:
        200:
          description: "blub"
          schema:
            $ref: "#/definitions/example"
```

(The important part is the Path here.)

When using the `spring-mvc` language, it generates the following file (beside others):

```
[imports snipped]

@Controller
@RequestMapping(value = "/path", produces = {APPLICATION_JSON_VALUE})
@Api(value = "/path", description = "the path API")
@javax.annotation.Generated(value = "class io.swagger.codegen.languages.SpringMVCServerCodegen", date = "2015-09-11T17:10:19.015+02:00")
public class PathApi {


  @ApiOperation(value = "", notes = "", response = Example.class)
  @ApiResponses(value = { 
    @ApiResponse(code = 200, message = "blub") })
  @RequestMapping(value = "/{foo_id}", 


    method = RequestMethod.GET)
  public ResponseEntity<Example> pathFooIdGet(
@ApiParam(value = "",required=true ) @PathVariable("fooId") String fooId
,

@ApiParam(value = ""  ) @RequestHeader(value="ifNoneMatch", required=false) String ifNoneMatch
)
      throws NotFoundException {
      // do some magic!
      return new ResponseEntity<Example>(HttpStatus.OK);
  }


}
```

The parameters `If-None-Match` and `foo_id` got transformed to `ifNoneMatch` and `fooId`, with no trace of the original names. The `@RequestHeader` and `@PathVariable` names even have the sanitized names instead of the original one.

I'm not sure about Swagger's `@ApiParam` annotations – I guess those should have the correct names, too.

When using the `jaxrs` "language", it generates this:

```
[imports snipped]

@Path("/path")


@io.swagger.annotations.Api(value = "/path", description = "the path API")
@javax.annotation.Generated(value = "class io.swagger.codegen.languages.JaxRSServerCodegen", date = "2015-09-11T17:16:59.995+02:00")
public class PathApi  {

   private final PathApiService delegate = PathApiServiceFactory.getPathApi();

    @GET
    @Path("/{foo_id}")


    @io.swagger.annotations.ApiOperation(value = "", notes = "", response = Example.class)
    @io.swagger.annotations.ApiResponses(value = { 
        @io.swagger.annotations.ApiResponse(code = 200, message = "blub", response = Example.class) })

    public Response pathFooIdGet(@ApiParam(value = "",required=true ) @PathParam("foo_id") String fooId,
    @ApiParam(value = ""  )@HeaderParam("If-None-Match") String ifNoneMatch)
    throws NotFoundException {
        return delegate.pathFooIdGet(fooId,ifNoneMatch);
    }
}
```

Here the JaxRS annotations have the correct values, the swagger annotations still have none.

I guess similar things might happen in other language generators.

**Analysis**
I guess the `*param.mustache` files need to be changed to use `baseName` instead of `paramName` at the correct places, and to have the names additionally in the @ApiParam annotation (if this is thought sensible).

ePaul:
This hopefully fixes #1219 and #4870 (and possibly other issues).

### PR checklist

- [x] Ran the shell script `./bin/spring-all-petstore.sh` to update Petstore sample so that CIs can verify the change. 
- [?] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.
    * While this changes the behavior of the generated code, this code didn't work before, and now works. I think it doesn't break anything which was working correctly.

### Description of the PR

This includes the changes of #4874 – view just the last commit in the diff, or merge that PR before this one to see the effects of the actual changes on the samples here.

As proposed by @wing328 in #4870, I tried to use `{{baseName}}` instead of `{{paramName}}` for the values of some generated Spring annotations. I used this for all parameter types: path parameters, form parameters and query parameters. Header parameters were already correct.


ePaul:
Rebased onto master to trigger CI again.

Issue #4032, opened 2016-10-18
[PHP] Can't deserialize enums


theindexer:
##### Description

--Enum here refers to an enum declared in swagger, which turns into a php class.--

My generated php client api can't get an object that includes an enum property.
I haven't checked serialization.
Didn't find any related issued when I looked.

##### Swagger-codegen version

2.2.1

##### Swagger 

Minimal json of the enclosing object (which is of course enclosed in a response)

 ```
"MyObj": {
  "type": "object",
  "properties" : {
    "MyEnum": {
      "type": "string",
      "enum": [
        "VALUE1",
        "VALUE2",
        "VALUE3"
      ]
    }
  }
}
```
##### Command line used for generation
bash, some variables omitted
```
java -jar $SWAGGERJAR generate -i $INPUT -l php -o $PHPDIR \
      --invoker-package My\\Client \
      --api-package My\\Client\\Api \
      --model-package My\\Client\\Model \
      --additional-properties packagePath=MyClient \
      || exit 1
```
##### Steps to reproduce

1) Run codegen, make sure you've got an enum in there
2) Make a get request whose response includes the object
3) Be unhappy
3) Hope that you're making a silly mistake

##### Suggest a Fix

ObjectSerializer in php looks like it assumes that all classes have a swaggerType method, but model_enum.mustache don't got one.  This causes unhapiness.

wing328:
@theindexer thanks for reporting the issue. You're right that the issue was caused by top-level enum class missing `swaggerType`.

May I know if you have time to contribute the fix?

theindexer:
Not at the moment, unfortunately :/

theindexer:
I also ran into this issue with Ruby.

A quick hack fix when all your enums are strings (which I imagine is the usual case) is as follows, basically just have a way to differentiate them from classes when deserializing.

https://gist.github.com/theindexer/5b96fadfb3dfae1bfe1b521752592a64

Kinda silly though.

jasco:
Another work-around, also for the case of string values, was to update the `swaggerTypes` in the models that reference the enum to declare the field type as `string` rather than as the enum class.

PR #4886, opened 2017-03-01
[PHP] error when deserializing enums #4032

baartosz:
### PR checklist

### Description of the PR

Fix for #4032. 
Example given in referenced ticket where Model has multiple properties is not really causing any errors on current version of SwaggerCodegen. 
This PR fixes serialization failure when we're dealing with Model containing only enum, like this one:
```
definitions:
  OuterEnum:
    type: "string"
    enum:
    - "placed"
    - "approved"
    - "delivered"
```

I added new serialization rule for this case and method `getAllowableEnumValues()` to Enum model class. Enum class is used only to validate parameters and store constants, deserialized result is just a string. Is it valid approach?

To differentiate between enum and non-enum model I am checking if model class has method `getAllowableEnumValues()`. That's dirty solution for now just to validate my idea. It would be nicer to have some EnumClassInterface instead.

Here's spec i was testing with:
```
swagger: '2.0'
info:
  version: 1.0.0
  title: Swagger TEST
host: petstore.swagger.io
basePath: /api
paths:
  /fake:
    post:
      operationId: testPostingEnum
      consumes:
        - application/json
      produces:
        - application/json
      parameters:
        - in: body
          name: body
          required: true
          schema:
            $ref: '#/definitions/Enum_Test'
      responses:
        '200':
          description: ok
    get:
      operationId: testGettingEnum
      consumes:
        - "*/*"
      produces:
        - "*/*"
      responses:
        '200':
          schema:
            $ref: '#/definitions/Enum_Test'

definitions:
  Enum_Test:
    type: object
    properties:
      enum_string:
        type: string
        enum:
          - UPPER
          - lower
          - ''
      enum_integer:
        type: integer
        format: int32
        enum:
          - 1
          - -1
      enum_number:
        type: number
        format: double
        enum:
          - 1.1
          - -1.2
      outerEnum:
        $ref: '#/definitions/OuterEnum'
  OuterEnum:
    type: "string"
    enum:
    - "placed"
    - "approved"
    - "delivered"
```

Controller class for generated silex server:
```
<?php
require_once __DIR__ . '/vendor/autoload.php';

use Symfony\Component\HttpFoundation\JsonResponse;
use Symfony\Component\HttpFoundation\Request;
use Symfony\Component\HttpFoundation\Response;
use Silex\Application;

$app = new Silex\Application();
$app->get('/api/fake', function (Application $app, Request $request) {
    return new JsonResponse([
        'enum_string' => 'UPPER',
        'enum_integer' => -1,
        'enum_number' => -1.2,
        'outerEnum' => 'placed'
    ]);
});
$app->post('/api/fake', function (Application $app, Request $request) {
    file_put_contents('/tmp/a', print_r($request->getContent(), true));
    return new Response();
});


$app->run();
```

and php code using client:
```
   $config = new \Swagger\Client\Configuration();
    $config->setHost('http://localhost/api');
    $client = new Swagger\Client\ApiClient($config);
4032
    $fakeApi = new Swagger\Client\Api\DefaultApi($client);
    $response = $fakeApi->testGettingEnum();

    var_dump($response);
    var_dump($response->getEnumInteger());
    var_dump($response->getOuterEnum());

    $fakeApi->testPostingEnum(
        new \Swagger\Client\Model\EnumTest([
            'enum_string' => 'lower',
            'enum_integer' => 'lower',
            'enum_number' => 'lower',
            'outer_enum' => \Swagger\Client\Model\OuterEnum::DELIVERED
        ])
    );
 ```
Heres what server recives through POST:
`{"enum_string":"lower","enum_integer":"lower","enum_number":"lower","outerEnum":"delivered"}`
Heres output of test php code:
```
class Swagger\Client\Model\EnumTest#8 (1) {
  protected $container =>
  array(4) {
    'enum_string' =>
    string(5) "UPPER"
    'enum_integer' =>
    int(-1)
    'enum_number' =>
    double(-1.2)
    'outer_enum' =>
    string(6) "placed"
  }
}
int(-1)
string(6) "placed"
```

### DIFF
--- a/modules/swagger-codegen/src/main/resources/php/ObjectSerializer.mustache
+++ b/modules/swagger-codegen/src/main/resources/php/ObjectSerializer.mustache
@@ -261,7 +261,8 @@ class ObjectSerializer
             return $deserialized;
         } elseif (method_exists($class, 'getAllowableEnumValues')) {
             if (!in_array($data, $class::getAllowableEnumValues())) {
-                throw new \InvalidArgumentException('invalid value provided');
+                $imploded = implode("', '", $class::getAllowableEnumValues());
+                throw new \InvalidArgumentException("Invalid value for enum '$class', must be one of: '$imploded'");
             }
             return $data;
         } else {
--- a/modules/swagger-codegen/src/main/resources/php/ObjectSerializer.mustache
+++ b/modules/swagger-codegen/src/main/resources/php/ObjectSerializer.mustache
@@ -49,10 +49,16 @@ class ObjectSerializer
             return $data;
         } elseif (is_object($data)) {
             $values = [];
-            foreach (array_keys($data::swaggerTypes()) as $property) {
+            foreach ($data::swaggerTypes() as $property => $swaggerType) {
                 $getter = $data::getters()[$property];
-                if ($data->$getter() !== null) {
-                    $values[$data::attributeMap()[$property]] = self::sanitizeForSerialization($data->$getter());
+                $value = $data->$getter();
+                if (method_exists($swaggerType, 'getAllowableEnumValues')
+                    && !in_array($value, $swaggerType::getAllowableEnumValues())) {
+                    $imploded = implode("', '", $swaggerType::getAllowableEnumValues());
+                    throw new \InvalidArgumentException("Invalid value for enum '$swaggerType', must be one of: '$imploded'");
+                }
+                if ($value !== null) {
+                    $values[$data::attributeMap()[$property]] = self::sanitizeForSerialization($value);
                 }
             }
             return (object)$values;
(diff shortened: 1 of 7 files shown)


baartosz:
What is preferred way to test when particular case is not available in petstore.swagger.io/v2 ?

baartosz:
I figured I can test just serialization in this case. Added missing tests.

One problem I have here is that when serializing just string I can't check if provided value is allowed. Its not the case when enum type is nested inside some other object like in this example:
https://github.com/swagger-api/swagger-codegen/pull/4886/commits/38b5b62fe3bb00c7c206214210430cd41b8718e0#diff-872083d684159ca361436c8f51708074R81 

Is that okay? Maybe would it be better to pass instance of OuterEnum object with value inside? Spec says that its type is supposed to be string so I am not sure if its better option.

wing328:
Yup, that's one way to test. In the C# API client, we've something similar: https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/csharp/SwaggerClient/src/IO.Swagger.Test/Api/StoreApiTests.cs#L130-L142

wing328:
cc @arnested

PR #4890, opened 2017-03-01
Make the CodegenConfigurator serializable

### LINKED ISSUE #4887: Make the CodegenConfigurator serializable so it can be used in caching for gradle tasks
Making the `CodegenConfigurator` class serializable would allow it to be used in a gradle task as an input. Task inputs in gradle get serialized and cached and if they do not change the task is marked as "UP-TO-DATE" and is skipped.

Everything inside of the `CodegenConfigurator` object is serializable so it should be a simple change.

JLLeitschuh:
### PR checklist

### Description of the PR

Makes the `CodegenConfigurator` implement `Serializable`. This is useful because it allows it to be cached with gradle which speeds up build times by allowing tasks that have already run to be skipped.

Closes #4887


JLLeitschuh:
Woot! Thanks @wing328! You just made my day!

PR #4902, opened 2017-03-02
[JavaScript] Added *WithHttpInfo methods in 'usePromises' mode

### LINKED ISSUE #3654: [JavaScript] Promise client doesn't return response headers
##### Description

I'm using Swagger with an API that allows for a request like HEAD /books which returns the number of books available as a response header.

The generated JavaScript Promise client doesn't return that information. It would help if it would at least provide a way to get access to the raw response object, so I could read the headers myself.

The swagger documentation suggests to call the method like so:

```
apiInstance.sendBooksMetadata(opts).then(function() {
  console.log('API called successfully.');
}, function(error) {
  console.error(error);
});
```
##### Swagger-codegen version

2.0

``` yaml
/books:
  head:
    summary: Get books info
    operationId: sendBooksMetadata
    parameters:
      - name: status
        in: query
        description: The book status. For example, "ACTIVE", "INACTIVE", "PENDING".
        type: string
        required: false
    tags:
      - Books
    responses:
      200:
        description: The number of books in the header
        headers:
          X-Books-Count:
            description: The number of books
            type: integer
```

```
java -jar ../../third-parties/swagger-codegen/modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -i schema/schema.yaml -l javascript -o javascript --additional-properties usePromises=true
```
##### Suggest a Fix

It looks like the exports.prototype.deserialize function of the generated ApiClient only returns the body of the response and the headers are lost in that function.

frol:
### PR checklist

### Description of the PR

This PR closes #3654 by adding *WithHttpInfo API methods, which can be used to get access to HTTP response object (e.g. headers or status code).

/cc @wing328 @amannn

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Javascript/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Javascript/ApiClient.mustache
@@ -412,7 +412,7 @@
         } else {
           try {
             var data = _this.deserialize(response, returnType);
-            resolve(data);
+            resolve({data, response});
           } catch (err) {
             reject(err);
           }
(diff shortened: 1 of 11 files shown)


wing328:
@frol the change looks good to me. For JS API client not using Promise, does it mean it's not possible to add the `WithHttpInfo` method?

frol:
@wing328 There is no need for `WithHttpInfo` methods for callback style rendering since the callback function already receives `data`, `error`, and `response`.

wing328:
@frol ok. Thanks for the explanation. PR merged.

Issue #3638, opened 2016-08-24
Generated Java APIs incorrectly encode commas in certain query parameters


michaelkourlas:
##### Description

For a query parameter of type "array" and collectionFormat "csv", the Java generated API percent-encodes the commas that separate the individual values. This is incorrect behaviour, as the commas are used as delimiters, not data, and should not be encoded.

For example, the following: 
```
/test?param=a,b,c
```
becomes:
```
/test?param=a%2Cb%2Cc
```

##### Swagger-codegen version

2.2.0

##### Swagger declaration file content or url

The following Swagger spec demonstrates this behaviour:
```json
{
  "swagger": "2.0",
  "info": {
    "title": "test spec",
    "version": "1.0"
  },
  "parameters": {
    "testParam": {
      "name": "test",
      "in": "query",
      "type": "array",
      "items": {
        "type": "string"
      },
      "collectionFormat": "csv"
    }
  },
  "paths": {
    "/test": {
      "get": {
        "description": "test get",
        "parameters": [
          {
            "$ref": "#/parameters/testParam"
          }
        ],
        "references": {
       
        }
      }
    }
  }
}
```

##### Command line used for generation

```
C:\Users\mkourlas\Downloads\swagger-codegen-2.2.0>java -jar modules\swagger-code
gen-cli\target\swagger-codegen-cli.jar generate -Dlibrary=jersey1 -i swagger.jso
n -l java -o test\java
```

##### Steps to reproduce

1. Generate and build the Java API for the above specification.
2. Create a new DefaultApi instance.
3. Call the `testGet` function with several strings a `List` passed to the `test` parameter.
4. The API will send an HTTP request with a URI looking like the following:
```
/test?test=string1%2Cstring2%2Cstring3
```
This should instead be
```
/test?test=string1,string2,string3
```

##### Suggest a Fix

This is caused by the following parts of the generated code:

```java
  public void testGet(List<String> test) throws ApiException {
    [...]
    // query params
    List<Pair> localVarQueryParams = new ArrayList<Pair>();
    [...]
    localVarQueryParams.addAll(apiClient.parameterToPairs("csv", "test", test));
    [...]
    apiClient.invokeAPI(localVarPath, "GET", localVarQueryParams, localVarPostBody, localVarHeaderParams, localVarFormParams, localVarAccept, localVarContentType, localVarAuthNames, null);
  }
```
```java
  private String buildUrl(String path, List<Pair> queryParams) {
[...]

PR #4908, opened 2017-03-03
[Java] Fix for issue #3638: Generated Java APIs incorrectly encode commas in certain query parameters

michaelkourlas:
This PR is a fix for #3638. Commas are now not percent-encoded when they are used as a delimiter in collection query parameters in the Java jersey1 and okhttp-gson APIs. A similar fix was made to the JavaScript API in #4725.

As part of this fix, a new API client method called `parameterToPair` has been added. This method converts query parameters to a list containing a single pair.

In addition, the behaviour of the API client method `parameterToPairs` has changed. The `value` parameter of this method must now always be a collection, and the `collectionFormat` parameter must now always specify a valid collection format. The values of the pairs returned are also now percent-encoded.

The Java API template now calls `parameterToPairs` when the collection format is defined for a query parameter, and `parameterToPair` if it is not. This should work because the DefaultCodegen class appears to ensure that templates receive a collection format for a query parameter ("csv" if none is explicitly specified in the spec) if and only if the parameter is of type "array" (or "object", apparently, but that type is not valid for parameters). Other templates (such as JavaScript) already seem to rely on this behaviour.

The `buildUrl` method has been updated to take into account these changes by placing collection query parameter values directly into the URL without percent encoding.

The Java unit tests for the `parameterToPairs` have been updated to take the new behaviour of the method into account. New tests have been added for the new `parameterToPair` method.

This PR was made against the 2.3.0 branch because it changed several method signatures, as in the case of PR #4725.


wing328:
@michaelkourlas when you've time, would you please rebase on the latest 2.3.0 to resolve the merge conflicts?

michaelkourlas:
@wing328 The build was failing on the Angular petstore client, which I never touched. When I tried the merge again it started failing on the Go petstore client.

wing328:
@michaelkourlas thanks for the rebase. Don't worry about the CI failures (usually intermittent). I'll take a look

samples/client/petstore/java/jersey1/src/test/java/io/swagger/client/ApiClientTest.java
```diff
@@ -203,25 +246,12 @@ public void testParameterToPairsWhenValueIsEmptyStrings() throws Exception {
...
-
-    @Test
     public void testParameterToPairsWhenValueIsCollection() throws Exception {
         Map<String, String> collectionFormatMap = new HashMap<String, String>();
         collectionFormatMap.put("csv", ",");
         collectionFormatMap.put("tsv", "\t");
         collectionFormatMap.put("ssv", " ");
-        collectionFormatMap.put("pipes", "\\|");
-        collectionFormatMap.put("", ","); // no format, must default to csv
-        collectionFormatMap.put("unknown", ","); // all other formats, must default to csv
```
wing328: @michaelkourlas is it correct to say that the default is no longer csv (",") when unknown or no format is provided?

michaelkourlas: @wing328 I think parameterToPairs still technically defaults to csv in those cases. If you want I can add the tests back.

However, I think this behaviour is wrong. I believe it is impossible to get that far without a defined collection format; if it's not defined DefaultCodegen appears to define it as "csv" for you. As such I don't think we need to handle the empty string case. Moreover, the invalid case should throw an error -- we shouldn't just silently assume csv. Don't we validate the spec before generating APIs from it first?

wing328: @michaelkourlas please put those test back when you've time.

I agree the behaviour is arguable but prefer to still keep the test cases before we change the behaviour.

> Don't we validate the spec before generating APIs from it first?

The current approach is to throw errors/warnings for invalid spec while trying to generate a functional clients. The goal is to provide a functional client that at least compiles right out of the box and the users can fix the spec or customize the setting (e.g. collection format) later.

michaelkourlas: Okay, I've gone ahead and added the tests back.

michaelkourlas:
@wing328 Not sure why the AppVeyor build is failing; I didn't touch the C# stuff.

wing328:
@michaelkourlas I'll take a look.

michaelkourlas:
@wing328 Any idea when this will be merged?

wing328:
@michaelkourlas sorry for the delay. Will take a look now...

PR #4909, opened 2017-03-03
Refactor nodejs generated code structure

fehguy:
This is a breaking change in the generated code structure for nodejs.

In all previous versions of the nodejs generated code, there was a sort of silly abstraction layer--there were controllers for receiving calls and service classes, designed to hold business logic.  The challenge with the existing structure was that the service class was interacting directly with the response object, and thus there was a mix of business logic and routing.  Way too convoluted for most nodejs developers.

This change breaks up the responsibilities.  There is a controller and service class separation of concerns still, but they are:

1) separated into different folders.  This makes it easier to ignore changes in generated code in the service folder to avoid overwriting files.

2) _Almost_ completely abstracting away the request/response cycle from the service class

3) Use of promises for better code structure

The _Almost_ is because in the service classes, there are scenarios where you need to indicate the response code that goes back over the wire.  To do this, a utility class is required, and you can then specify what code you want to send to the caller.


bin/nodejs-petstore-server.sh
```diff
@@ -26,6 +26,6 @@ fi
 
 # if you've executed sbt assembly previously it will use that instead.
 export JAVA_OPTS="${JAVA_OPTS} -XX:MaxPermSize=256M -Xmx1024M -DloggerPath=conf/log4j.properties"
-ags="$@ generate -i modules/swagger-codegen/src/test/resources/2_0/petstore.yaml -l nodejs-server -o samples/server/petstore/nodejs"
+ags="$@ generate -t modules/swagger-codegen/src/main/resources/nodejs -i modules/swagger-codegen/src/test/resources/2_0/petstore.yaml -l nodejs-server -o samples/server/petstore/nodejs"
```
fehguy: To make it easier to test template changes, the bin script is now loading templates directly from the file system instead of the compiled jar

modules/swagger-codegen/XhhGitIgnore/sdk_unit_testing_binary.json
```diff
@@ -1,67 +0,0 @@
-{
```
fehguy: @wing328 I think this file was added on accident?  Please let me know and I can restore them if needed.

modules/swagger-codegen/XhhGitIgnore/sdk_unit_testing_file_downloading.json
```diff
@@ -1,30 +0,0 @@
-{
```
fehguy: @wing328 same here...

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NodeJSServerCodegen.java
```diff
@@ -19,11 +19,12 @@
 import java.math.BigDecimal;
 import java.util.*;
 import java.util.Map.Entry;
+import java.util.regex.Pattern;
 
 public class NodeJSServerCodegen extends DefaultCodegen implements CodegenConfig {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(NodeJSServerCodegen.class);
-
+    protected String implFolder = "service";
```
fehguy: Will adjust PR to make this configurable.

modules/swagger-codegen/src/main/resources/nodejs/controller.mustache
```diff
@@ -1,13 +1,21 @@
 'use strict';
 
-var url = require('url');
+require('../utils/writer.js');
```
fehguy: pull in utility functions for `writeJson`

modules/swagger-codegen/src/main/resources/nodejs/service.mustache
```diff
@@ -2,40 +2,42 @@
...
+{{/allParams}}
+{{^returnType}}
+ * no response value expected for this operation
+{{/returnType}}
+{{#returnType}}
+ * returns {{{returnType}}}
+{{/returnType}}
+ **/
+exports.{{{operationId}}} = function({{#allParams}}{{paramName}}{{#hasMore}},{{/hasMore}}{{/allParams}}) {
+  return new Promise(function(resolve, reject) {
```
fehguy: No more `req`, `res`, only using Promise lib.

samples/server/petstore/nodejs/controllers/Pet.js
```diff
@@ -1,37 +1,97 @@
...
 module.exports.deletePet = function deletePet (req, res, next) {
-  Pet.deletePet(req.swagger.params, res, next);
+    var petId = req.swagger.params['petId'].value;
+    var api_key = req.swagger.params['api_key'].value;
+    Pet.deletePet(petId,api_key)
+        .then(function (response) {
+            writeJson(res, response);
+        })
+        .catch(function (response) {
+            writeJson(res, response);
```
fehguy: to send specific response codes, use the `respondWithCode` function:

```
writeJson(respondWithCode(500, { message: 'it broked'}));
```

samples/server/petstore/nodejs/controllers/Pet.js
```diff
@@ -1,37 +1,97 @@
...
+        });
 };
 
 module.exports.updatePetWithForm = function updatePetWithForm (req, res, next) {
-  Pet.updatePetWithForm(req.swagger.params, res, next);
+    var petId = req.swagger.params['petId'].value;
+    var name = req.swagger.params['name'].value;
+    var status = req.swagger.params['status'].value;
```
fehguy: note, all input values are extracted in generated code to make it easier to implement the service class.

fehguy:
@wing328 if you have any feedback please let me know, otherwise we can merge this.

modules/swagger-codegen/XhhGitIgnore/sdk_unit_testing_binary.json
wing328: The file was added on accident. Thanks for removing it.

modules/swagger-codegen/XhhGitIgnore/sdk_unit_testing_file_downloading.json
wing328: The file was added on accident. Thanks for removing it.

wing328:
@fehguy I did some tests and the result is good.

PR #4911, opened 2017-03-04
Update swift3 samples after 2.2.2.

ePaul:
### PR checklist

- [x] Read the [contribution guildelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change.
    → just bin/swift3-petstore-all.sh
- [x] Filed the PR against the correct branch: master for non-breaking changes.

### Description of the PR

No code changes, just regenerating the swift3 samples from master code.
This is a part of #4901 separated to see whether CI likes it or not.

Issue #4898, opened 2017-03-02
Meta issue: parameter name mangling breaks client or server functionality


ePaul:
### Description

In many languages, parameter names are mangled to conform to the conventions of the language, not be reserved words, etc. For Java, this is conversion to camelCase.

In several languages, the generated code only actually works with the protocol described in the API definition when the parameter mangling results in the same name as the original, otherwise different parameter names are actually sent to the server (for clients) or accepted from the client (for servers).

This is a meta issue collecting my research I have done.

### Swagger-codegen version

I worked on master of 2017-02-27, i.e. shortly before the release of 2.2.2.

#### spring

I first found this for spring-mvc (later renamed to just `spring`) in #1219, for path and header parameters. It was fixed for header parameters some time ago, is still wrong for path, query and form parameters.
Some days ago it was also noted by @techlogix in #4870.
I filed PR #4875 fixing this for the spring-based languages.

#### html2

Issue #4864 is mentioning the problem, it was seemingly resolved in #4602.

### Potentially affected languages

I did a search for `{{paramName}}` in the templates, especially looking for where it is used in a string which then is used for sending/accepting the parameter to the server/from the client. In those cases most probably `{{baseName}}` should be used instead.
I made a [preliminary commit](https://github.com/ePaul/swagger-codegen/commit/f754ae210ee562c6abfe88e419950a8710d6d1e2) with all the changes I found (and later a [second one](https://github.com/ePaul/swagger-codegen/commit/12058bc385655ccb57ef7bf159b1d9727392d169) with files I missed).

I'll go through the changes here, list the languages up there with details and create a pull request with just the changes for one language (and the regenerated samples). Someone with knowledge of the specific language should check them and confirm that that was really a bug and the new version is better.

Here is an overview of my findings in general (:x: = potentially broken). There might be some languages which don't have any parameter transformation, then `{{baseName}}` would be always the same as `{{paramName}}`.

Language (template directory name) | Pull request | path | query | form | header | Other comments 
------------------------------------------------ | ----------------- | ----- | -------- | ------ | --------- | --------
Groovy | # | | :x: |  | :x: | It looks like path parameters are not supported at all?
JavaJaxRS | #4918  | | | :x: * |  | It looks like only file-typed form parameters (i.e. file uploads) are affected.
JavaJaxRs/cxf | #4918 | | | :x: | |  here all kinds of form parameters.
JavaJaxRs/cxf-di | #4918 | | | :x: | |  
[...]

PR #4916, opened 2017-03-04
Fix (partially) #4898 for swift3

ePaul:
### PR checklist

   → bin/swift3-petstore-all.sh (~~once before the change, see #4911, once again~~ after the change.)

### Description of the PR

~~This includes the changes of #4911 – after merging those, it should be clearer what actually changed. (Or just have a look at the last two commits.)~~

This fixes a wrong name-mapping of path parameters, as part of solving #4898.

It looks like there is no test case in petstore which actually makes a difference here.
It might be that there actually is no problem (i.e. if parameter names are never modified for Swift3, then `baseName` is the same as `paramName`, so it doesn't matter – though then there might be problems when hitting reserved words), or that it just is not happens with our petstore example.

I need someone who knows about the swift clients to review this – cc @jaz-ah @Edubits.

_(A side remark: Some ids in the generated samples look different each time the generator is run – see commit 8e847ab. This introduces unneeded noise when looking at the diffs. Might be worth looking into, in a separate issue/PR.)_


wing328:
@ePaul thanks for the fix.

cc @hexelon @jaz-ah @Edubits @jgavris

ePaul:
Rebased onto master to get rid of conflicts.

wing328:
LGTM. Thanks for the fix.

Issue #4738, opened 2017-02-07
Add support for BeanValidation @Valid annotation on nested attributes


jmiddleton:
##### Description

When useBeanValidation flag is true, it does not add @Valid annotation to attributes of type object. For instance, I have the class JoinDetails which has a field of type MemberProfile. When I validate joinDetails instance, the validator does not validate memberProfile because there is no @Valid annotation on it. It would be better if swagger-codegen can support this use cases either using extraAnnotation (i.e. x-extraAnnotation: "@Valid") in swagger config or automatically added during code generation.

##### Swagger-codegen version

2.2.2-SNAPSHOT

##### Swagger declaration file content or url

    ```yaml
    JoinDetails:
    type: object
    required:
      - memberProfile
    properties:
      memberProfile:
        $ref: '#/definitions/MemberProfile'
     MemberProfile:
    type: object
    required:
      - firstName
      - lastName
    properties:
      firstName:
        type: string
        maxLength: 50
      lastName:
        type: string
        maxLength: 50
    ```

##### Command line used for generation

I'm using maven:

<groupId>io.swagger</groupId>
<artifactId>swagger-codegen-maven-plugin</artifactId>
<version>2.2.2-SNAPSHOT</version>

##### Suggest a Fix

Add support via x- extension or automatically detect complex object or entities from the model.

wing328:
@jmiddleton thanks for the suggestion. For this particular case, I think `@Valid` should be added automatically.

cc @jfiala

jfiala:
For spring we already have a PR: https://github.com/swagger-api/swagger-codegen/pull/4850/files

jfiala:
Overview of all the languages to be updated:

- [ ] jaxrs-spec
- [ ] jaxrs (Jersey2 + library jersey1)
- [ ] jaxrs-resteasy
- [x] jaxrs-cxf - PR #4738
- [ ] jaxrs-cxf-cdi
- [ ] spring (default library spring-boot + library spring-mvc, spring-cloud)

michaelpaesold:
This also applies to `jaxrs` (Jersey2). I could provide a patch.

olivierpaquet:
Is someone already working on the same issue for jaxrs-cxf?

jfiala:
@olivierpaquet  I did the PR in summer last year as you already seem to have discovered. But currently I'm not working on it, so feel free to catch up and get it pulled in...

fabian-braun:
[...]

PR #4922, opened 2017-03-04
[Jaxrs-cxf] Add cascaded beanvalidation @Valid for pojos #4738

jfiala:
### PR checklist

### Description of the PR

Implemented #4738 for CXF

Open issues: 
@Valid is currently also generated for properties of type (isPrimitiveType/^isContainer does not cover these):
* java.lang.BigDecimal
* javax.xml.datatype.XMLGregorianCalendar
* org.joda.time.LocalDate
* java.util.UUID

Please advise how we can best check for model-specific pojos?

### DIFF
--- a/modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf/pojo.mustache
+++ b/modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf/pojo.mustache
@@ -27,7 +27,8 @@ public class {{classname}} {{#parent}}extends {{{parent}}}{{/parent}} {
 {{#useJaxbAnnotations}}
   @XmlElement(name="{{baseName}}"{{#required}}, required = {{required}}{{/required}})
 {{/useJaxbAnnotations}}
-  @ApiModelProperty({{#example}}example = "{{example}}", {{/example}}{{#required}}required = {{required}}, {{/required}}value = "{{{description}}}")
+  @ApiModelProperty({{#example}}example = "{{example}}", {{/example}}{{#required}}required = {{required}}, {{/required}}value = "{{{description}}}"){{^isPrimitiveType}}{{^isContainer}}{{#useBeanValidation}}
+  @Valid{{/useBeanValidation}}{{/isContainer}}{{/isPrimitiveType}}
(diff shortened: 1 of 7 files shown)


jfiala:
@wing328 can you pls advise regarding the open issues above?
The question is how to best scope @Valid to be generated only for classes of the model, not for other classes.

wing328:
@jfiala My take is that {{#isPrimitive}} should cover the following as well
- java.lang.BigDecimal
- javax.xml.datatype.XMLGregorianCalendar
- org.joda.time.LocalDate
- java.util.UUID

so that the following should imply models:

```
{{^isContainer}}{{^isPrimitiveType}} ... {{/isPrimitiveType}}{{/isContainer}}
```

jfiala:
This is exactly what I am using at pojo.mustache:
``` java
{{^isPrimitiveType}}{{^isContainer}}{{#useBeanValidation}}
+  @Valid{{/useBeanValidation}}{{/isContainer}}{{/isPrimitiveType}}
``` 
PrimitiveType covers only the real primitive types and container the List/Map items.
So I think we need a new flag to check for isModelType?

fehguy:
@jfiala shouldn't `primitive` or _not_ primitive cover it?

jfiala:
@fehguy unfortunately not, see above listed mustache snippet, I actually used ^isPrimitiveType/^isContainer.
The problem is that the above datatypes are not considered primitives. 
On the other hand, probably it would be better to introduce a new flag to check if a datatype is a pojo of the model?

wing328:
> So I think we need a new flag to check for isModelType?

I would rather update "isPrimitiveType" to include the additional items (e.g. uuid) you've listed.

olivierpaquet:
@wing328 any news about this PR? We also need this feature but seems that the PR is not yet merged / final.

jfiala:
@olivierpaquet Thx for asking, but currently I'm too busy to rebase, if you like you can rebase and re-provide the PR.

mohangadm:
We were trying to use this plugin for open api 3, and the validation of cascading beans is not failing as @Valid annotation is not added on cascaded pojos.
Is this pull req merged to 3.X version for open api-3?

Issue #4932, opened 2017-03-06
[JavaScript] Do not parse null and undefined as type.


wangzw:
##### Description

Fields in models will be finally parsed to its type. But if the field is null or undefined. 
Keep it to null or undefined will benefit the caller for further process.

##### Swagger-codegen version

latest version, not a regression.

##### Related issues

#2103

wing328:
Closed via #4933. Thanks for the contribution from @wangzw

PR #4933, opened 2017-03-06
[JavaScript] Do not parse null and undefined as type.

wangzw:
### PR checklist

### Description of the PR

#4932

Fields in models will be finally parsed to its type. But if the field is null or undefined.
Keep it to null or undefined will benefit the caller for further process.


wing328:
@wangzw thanks for the PR, which looks good to me.

wing328:
@wangzw For the following line, shall we also use `===` instead of `==`?
https://github.com/wangzw/swagger-codegen/blob/0a585b7c8ccd034d4bc9232f3904ba2d942d55de/modules/swagger-codegen/src/main/resources/Javascript/ApiClient.mustache#L86

wangzw:
@wing328 I guess ```===``` is better.

wing328:
@wangzw right. Do you mind submitting a PR for the enhancement?

It would be nice if you can run [jslint](https://en.wikipedia.org/wiki/JSLint) on the Javascript API Petstore client to see if there are any other minor enhancements/bug fixes we can put into the JS API client.

PR #4946, opened 2017-03-06
Add proxy support to fix #4639

### LINKED ISSUE #4639: [Python] Proxy support in generated Python module
##### Description

Currently, the generated Python module does not support proxy usage. Is this something in the works? I've cobbled together some modifications to the code but am unsure of the best way to specify the configuration. My current setup uses an environment variable that is automatically detected and used at runtime. Using the requests library would also streamline this greatly.

##### Swagger-codegen version

2.1.0 but this appears to be the case in `master` based on https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/python/rest.mustache

##### Related issues

#4000 
#4102 

##### Suggest a Fix

Apart from using requests, I have a patch (locally, not yet submitted as a PR) to detect the environment variable `HTTPS_PROXY` at runtime. If it exists, I create an additional property for the `Configuration()` object and call `urllib3.ProxyManager()` instead of `urllib3.PoolManager()` in rest.py. I'd appreciate feedback and constructive criticism on this approach.

krmaxwell:
### PR checklist

### Description of the PR

Adds support for proxies in the Python client as discussed in #4639. Simply setting the `proxy` property of the configuration object to the proper URL should be picked up and the client will use `ProxyManager` instead of `PoolManager`. Note that this does not support authenticated proxies.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/python/rest.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/rest.mustache
@@ -76,7 +76,7 @@ class RESTClientObject(object):
         # key file
         key_file = Configuration().key_file
 
-       # proxy
+        # proxy
         proxy = Configuration().proxy
 
         # https pool manager
@@ -100,14 +100,6 @@ class RESTClientObject(object):
                 key_file=key_file
             )
 
-        # https pool manager
-        self.pool_manager = urllib3.PoolManager(
-            num_pools=pools_size,
-            cert_reqs=cert_reqs,
-            ca_certs=ca_certs,
-            cert_file=cert_file,
-            key_file=key_file
-        )
 
     def request(self, method, url, query_params=None, headers=None,
                 body=None, post_params=None, _preload_content=True, _request_timeout=None):
--- a/modules/swagger-codegen/src/main/resources/python/rest.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/rest.mustache
@@ -76,10 +76,33 @@ class RESTClientObject(object):
         # key file
         key_file = Configuration().key_file
 
+       # proxy
+        proxy = Configuration().proxy
+
+        # https pool manager
+        if proxy:
+            self.pool_manager = urllib3.ProxyManager(
+                num_pools=pools_size,
+                maxsize=maxsize,
+                cert_reqs=cert_reqs,
+                ca_certs=ca_certs,
+                cert_file=cert_file,
+                key_file=key_file,
+                proxy_url=proxy
+            )
+        else:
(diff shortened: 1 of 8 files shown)


wing328:
@krmaxwell thanks for the PR. Please run `./bin/python-petstore.sh` to update the Petstore sample for Python so that the CI can test the change.

krmaxwell:
Forgot to commit that. :blush: Also, is that the right way to specify a configuration? Or do I need getter/setter methods?

modules/swagger-codegen/src/main/resources/python/rest.mustache
```diff
@@ -76,10 +76,33 @@ class RESTClientObject(object):
         # key file
         key_file = Configuration().key_file
 
+       # proxy
```
wing328: @krmaxwell This line seems not indented properly - missing a space.

samples/client/petstore/python/petstore_api/rest.py
```diff
@@ -85,10 +85,33 @@ def __init__(self, pools_size=4, maxsize=4):
...
+                num_pools=pools_size,
+                maxsize=maxsize,
+                cert_reqs=cert_reqs,
+                ca_certs=ca_certs,
+                cert_file=cert_file,
+                key_file=key_file
+            )
+
         # https pool manager
         self.pool_manager = urllib3.PoolManager(
```
wing328: @krmaxwell Should this be removed as the code you added above already takes care of `pool_manager` initialization.

modules/swagger-codegen/src/main/resources/python/rest.mustache
```diff
@@ -76,10 +76,33 @@ class RESTClientObject(object):
...
+                num_pools=pools_size,
+                maxsize=maxsize,
+                cert_reqs=cert_reqs,
+                ca_certs=ca_certs,
+                cert_file=cert_file,
+                key_file=key_file
+            )
+
         # https pool manager
         self.pool_manager = urllib3.PoolManager(
```
wing328: @krmaxwell Should this be removed as the code you added above already takes care of `pool_manager` initialization.

wing328:
> Or do I need getter/setter methods?

I don't think so as other attributes do not have getter/setter either. If it's Python best practice to add getter/setter, then we'll reconsider.

I also left you some comments as part of the code review. Please take a look.

cc @scottrw93

scottrw93:
Looks good to me. No need for getters/setters

wing328:
@scottrw93 thanks for reviewing

@krmaxwell thanks for the contribution, which has been merged into master.

Issue #4822, opened 2017-02-18
Generate a library that can be consumed by Windows UWP app


ncksol:
I am a bit confused about the support for UWP with swagger-codegen. According to this issue here: https://github.com/swagger-api/swagger-codegen/issues/2241
the beta support for uwp has been implemented. But I've just tried to generate the code with targetFramework=uwp and the result doesn't look valid to me.

The project generated contains references to libraries like System.Data which are part of .NET Framework, but are not support by uwp apps.

If I try to build the project, I get errors around the async functionality. I believe it is because the project file is a mishmash of different incompatible options. The "TargetPlatformIdentifier" property is set to UAP, but "ProjectTypeGuids" property is not set to "A5A43C5B-DE2A-4C0C-9213-0A381AF9435A" (which indicates type Universal Windows Class Library) and therefore project type is defaulted to standard .net framework library. In that case "TargetPlatformIdentifier" is ignored completely and with no "TargetFrameworkVersion" property mentioned the .net framework is defaulted to 3.5. And that version doesn't have support for async functionality, hence the build errors.

Now, I don't have much experience with UWP apps and might be doing something completely wrong here. Because based on the issue above, @jerry1621 managed to make it work just fine with UWP.
So I am just looking for some help in understanding whether my approach is incorrect or is does UWP support still requires additional work.

wing328:
@ncksol thanks for reporting the issue.

> The project generated contains references to libraries like System.Data which are part of .NET Framework, but are not support by uwp apps.

For UWP, what should be used instead of System.Data?

> The "TargetPlatformIdentifier" property is set to UAP, but "ProjectTypeGuids" property is not set to "A5A43C5B-DE2A-4C0C-9213-0A381AF9435A" 

So when the `targetFramework=uwp` option is provided, if we change ProjectTypeGuids to "A5A43C5B-DE2A-4C0C-9213-0A381AF9435A", the issue will be addressed?

ncksol:
@wing328 

> For UWP, what should be used instead of System.Data?

Well as far as I know UWP uses .Net Core which is delivered in a single package from nugget. It doesn't have those references to specific libraries any more.

> So when the targetFramework=uwp option is provided, if we change ProjectTypeGuids to "A5A43C5B-DE2A-4C0C-9213-0A381AF9435A", the issue will be addressed?

[...]

Issue #1557, opened 2015-11-12
C# dependency that is PCL friendly


chitankensuru:
Currently the C# client export has a dependency on RestSharp. I am working on a Xamarin PCL project and unfortunately RestSharp has no PCL implementation.

Could C# client export with PCL friendly libraries be added? That would be awesome.

who:
@chitankensuru 

You may be able to use the C# .Net 2.0 generator for this.

I've used it to create clients for Unity projects.

https://github.com/swagger-api/swagger-codegen/tree/master/samples/client/petstore/csharp-dotnet2/SwaggerClientTest/Lib/SwaggerClient

chitankensuru:
The C# .Net 2.0 seems to be using RestSharp also.

wing328:
@chitankensuru have you used https://github.com/FubarDevelopment/restsharp.portable before? would it meet your requirement ?

chitankensuru:
@wing328 Tried that. Only in the generated ApiClient several functions are used from RestSharp.Extensions. Which I didn't find in RestSharp.Portable.

wing328:
@chitankensuru thanks for trying. If those functions are replaced properly, then RestSharp.Portable can be used, right ?

wing328:
@chitankensuru I've filed #1572 to remove dependency on RestSharp extension.

wing328:
@chitankensuru PR merged into master. Please pull the latest master and give it a try.

chitankensuru:
@wing328 Thanks for this. It partially solved the issue. We got it working by commenting out some parts in the ApiClient and Configuration file. And had to change RestClient.Execute and RestClient.ExecuteTaskAsync to RestClient.ExecuteAsync.

wing328:
@chitankensuru no problem. Can you please show us the updated ApiClient and Configuration file in http://gist.github.com so that we can make the change in the templates?

For RestClient.ExecuteAsync, I cannot think of a workaround for the time being. We might need to add a switch in the command line just for that.

wing328:
Closing this as there's no update.

Gronsak:
I've ran into a similar problem recently when working with a asp.net core app and I concluded that Restsharp.Portable is a good replacement, compared to Restsharp it does some stuff differently though but I've managed to get it working on one of my generated api clients with minor modifications.

would it be possible to get this issue reopened?

wing328:
[...]

PR #4955, opened 2017-03-06
[C#] Adding support for generating .NET Standard 1.3 client library

Gronsak:
### PR checklist

- [x] Ran the shell/batch script under `./bin/` to generate Petstore sample.

### Description of the PR

Should solve issue #1557 and #4822
This adds support for generating a .NET Standard 1.3 Library

The main difference to the standard C# client is that it's dependant on [Restsharp.Portable](https://github.com/FubarDevelopment/restsharp.portable) instead of [Restsharp](https://github.com/restsharp/RestSharp) with some minor tweaks for compatibility with .NET Standard, most templates are in fact copies of the standard C# templates with modifications where needed.

Being .NET Standard 1.3 makes it compatible with:

- .NET Framework 4.6+
- .NET Core 1.0+
- Mono/Xamarin Platforms vNext
- Universal Windows Platform 10.0+

[See Microsoft's post about .NET Standard Libraries for more information](https://docs.microsoft.com/en-us/dotnet/articles/standard/library)

The Petstore sample library generated and compiled without issues but unit testing isn't implemented yet.
Some extra testing was done on a Asp.net core app using [CCP's EVE Swagger Interface (ESI)](https://esi.tech.ccp.is/latest/) to verify that it was actually working and no issues has been encountered yet.


wing328:
@Gronsak thanks for the PR. I noticed that you've created a new generator: `csharp-netstandard`.

Ideally we want to keep just one `csharp` generator and use `library` (or other options) to specify which HTTP library to use. Do you think it would be too hard to do so? Is that because there're too many changes to the C# mustache templates (RestSharp)?

We can help consolidate `csharp-netstandard` into `csharp` and want to get your opinion first.

Gronsak:
It should definitely be possible to do so, I mainly made a new one since it was the least time consuming way for me to make it while at the same time learning the ins and outs of Java.

Most of the mustache templates are more or less the same so consolidating them into one shouldn't be to hard.
I have some free time today so I'll have a look at what I can accomplish. It should just be a matter of adding a switch now that I know a bit more about what I'm actually doing.

Gronsak:
Like I thought it was quite easy to fix it, I choose to put it under target frameworks as v5.0 since that's what a .NET Standard library is defined as in the project file.

If there are any other changes you want made I will do my best to fix it as soon as possible with the time I have available.

wing328:
@Gronsak I did some tests with default (v4.5) and v5.0 and the result looks good. Thanks for your contribution.

My tests with v3.5 uncover some errors (not caused by this PR) and I'll create a separate issue for tracking.

jimschubert:
@Gronsak @wing328 where does the `v5.0` in this update come from? There's currently no .NET Framework 5.0, and I think it's confusing that `v5.0` refers to .NET Standard 1.3.

Is this supposed to refer to ASP.NET 5.0 or something? I think this will only add confusing to an already confusing versioning strategy, especially since there's previously been DNX Core 5.0 and I'm assuming there will soon be an actual .NET Framework 5.0.

Gronsak:
It comes from how a PCL library is defined in a Visual Studio 2015 project file, a .NET Standard library is technically still a PCL library under VS2015 and someone made the decision that defining it as v5.0 instead of PCL or something was a good idea.
This changed for VS2017 though, where .NET standard gets it's own target framework definition.
I agree that it's somewhat confusing, naming schemes have never really been Microsofts strong suit. 

To mitigate Microsofts poor decissions it could be moved to a different property where you choose between making it a .NET framework class library or a .NET standard class library. This also opens up for a prettier way of handling differences between VS2015 and VS2017 since it's only really necessary for .NET Core/.NET Standard projects

Gronsak:
Might I suggest reworking how the target framework is selected by adding a level of abstraction?

Instead of setting the target framework property directly a better option may be to have a property for choosing what kind of project you would like to output.
I would suggest going with the same alias as Microsoft defines here: [.NET Standard Library](https://docs.microsoft.com/en-us/dotnet/articles/standard/library)

| Platform Name              | Alias       |     |     |       |     |       |       |       |       |
|----------------------------|-------------|-----|-----|-------|-----|-------|-------|-------|-------|
| .NET Standard              | netstandard | 1.0 | 1.1 | 1.2   | 1.3 | 1.4   | 1.5   | 1.6   | 2.0   |
| .NET Core                  | netcoreapp  | →   | →   | →     | →   | →     | →     | 1.0   | 2.0   |
| .NET Framework             | net         | →   | 4.5 | 4.5.1 | 4.6 | 4.6.1 | 4.6.2 | vNext | 4.6.1 |
| Mono/Xamarin Platforms     |             | →   | →   | →     | →   | →     | →     | →     | vNext |
| Universal Windows Platform | uap         | →   | →   | →     | →   | 10.0  | →     | →     | vNext |
| Windows                    | win         | →   | 8.0 | 8.1   |     |       |       |       |       |
| Windows Phone              | wpa         | →   | →   | 8.1   |     |       |       |       |       |
| Windows Phone Silverlight  | wp          | 8.0 |     |       |     |       |       |       |       |

That would mean that to generate what is now defined as `targetFramework: v3.5` you would instead input `net3.5` and to make a net standard 1.3 project you would put in `netstandard1.3`.
You could then put in all the right values in `targetFramework` and `targetFrameworkNuget` for the targeted version of the project file.

jimschubert:
I actually think .NET 2.0,  .NET Framework, and .NET Core should be different generators where the supported versions should be libraries below. Otherwise, there's a ton of changes to support 3 layers of decisions across all generators. This would also reduce confusion over the various differences, and allow for sharing code across generators (I believe).

Would you be opposed to `netstandard1.3` until we can work out a better abstraction?

Gronsak:
The changes needed for .NET Standard is so minute that I don't really think there's any need for a separate generator for it.

It seems to be Microsoft's intention that .NET Standard really should be supported by all .NET platforms going forward, which really makes it the easiest option for supporting everything later than .NET Framework 4.6 while at the same time supporting most other platforms of interest for C# developers so why put it in a separate box when there's no real differences towards .NET Framework 4.5?

The only differences not related to switching to `RestSharp.Portable` is what kind of project files they use and what is put in them, and as of VS2017 they both use the same project files and the only difference being that .NET Standard requiring a lot less to be defined.

And just to clarify, most of the functionality I pulled out when NET Standard support, like `IValidatebleObject` related stuff can go back in with an extra dependency to the `System.ComponentModel.Anotations` nuget package, which I didn't realise at the time and is why I didn't keep it in to begin with. From what I can understand Fody should work as well but I'm no actual expert there.

The only thing that would have to be changed would be what property we accept as an argument to the generator and therefore what we look at when actually setting the necessary properties for the Mustache files i.e. `supportsAsync`, `targetFramework` and `targetFrameworkNuget` and the like. From what I can make out of the top of my head there would not really be any layers of decisions more than there is now (VS2017 support is the exception but it's not really to complex to do).

jimschubert:
Your earlier comment,

> That would mean that to generate what is now defined as targetFramework: v3.5 you would instead input net3.5 and to make a net standard 1.3 project you would put in netstandard1.3.

I don't think this would work. `targetFramework` is used in project files to determine the compilation tools for the FCL differences between 3.5 and 4.5 (plus 2.0 when it gets merged into the generator).

Keeping all of these generators in one generator just means a lot of conditional branches all over the generator code and the templates. This is fine, except we will end up with changes like the one you suggested above which will impact .NET 3.5 clients. We have a base generator class for shared code already. However, the current structure of swagger-codegen only supports deciding from language+framework (e.g. C#+RestSharp). We're bastardizing "language" a bit by saying .NET means C#. Then, there is another layer below RestSharp for the .NET flavor that we're switching on in a lot of places with custom logic helpers like `supportsAsync` of the proposed `dotnet4.5` for 4.5 and higher generators. That's what I think would benefit from separate​ generators and incorporating the libraries structure. 

I would have to think more about how this would look though. Maybe we should open a new issue for community discussion around it?

wing328:
FYI. Added CI coverage (AppVeyor) for the new C# API client via https://github.com/swagger-api/swagger-codegen/pull/5249

wing328:
The new C# API client does not come with auto-generated test files. I'll create a separate task for tracking.

johnamcruz:
I tried generating it with java -jar C:\Repository\Scripts\swagger-codegen-cli.jar generate -i C:\Repository\swagger.json -l csharp -o HA.Client2 --additional-properties targetFramework=v5.0,packageGuid={3AB1F259-1769-484B-9411-84505FCCBD55}

but its generating the RestSharp version instead of the RestSharp portable version. Is there somthing i am missing @wing328

wing328:
@johnamcruz can you try using the config file instead of `--additional-properties`?

https://github.com/swagger-api/swagger-codegen/issues/1557#issuecomment-290053055

johnamcruz:
I tried this as well. @wing328 

java -jar swagger-codegen-cli.jar generate -i swagger.json -l csharp -o TEST --additional-properties packageGuid={3AB1F259-1769-484B-9411-84505FCCBD55} -c framework.json

where framework.json contains
{
    "targetFramework":"v5.0"
}

I checked the generated client and it doesnt contain RestSharp.Portable. It doesnt seem to recognize the netstandard tags. Any idea? Thanks.

wing328:
I assume you've run `mvn clean package` to rebuild the JAR, right?

wing328:
For the Petstore sample, RestSharp.portable is correctly used:
```
swagger-codegen|csharp_net10_sample⚡ ⇒ grep -R -i portable samples/client/petstore/csharp/SwaggerClient.v5/*
samples/client/petstore/csharp/SwaggerClient.v5/README.md:- FubarCoder.RestSharp.Portable.Core >=4.0.7
samples/client/petstore/csharp/SwaggerClient.v5/README.md:- FubarCoder.RestSharp.Portable.HttpClient >=4.0.7
samples/client/petstore/csharp/SwaggerClient.v5/src/IO.Swagger/Api/FakeApi.cs:using RestSharp.Portable;
samples/client/petstore/csharp/SwaggerClient.v5/src/IO.Swagger/Api/PetApi.cs:using RestSharp.Portable;
samples/client/petstore/csharp/SwaggerClient.v5/src/IO.Swagger/Api/StoreApi.cs:using RestSharp.Portable;
samples/client/petstore/csharp/SwaggerClient.v5/src/IO.Swagger/Api/UserApi.cs:using RestSharp.Portable;
samples/client/petstore/csharp/SwaggerClient.v5/src/IO.Swagger/Client/ApiClient.cs:using RestSharp.Portable;
samples/client/petstore/csharp/SwaggerClient.v5/src/IO.Swagger/Client/ApiClient.cs:using RestSharp.Portable.HttpClient;
samples/client/petstore/csharp/SwaggerClient.v5/src/IO.Swagger/Client/ExceptionFactory.cs:using RestSharp.Portable;
samples/client/petstore/csharp/SwaggerClient.v5/src/IO.Swagger/Client/IApiAccessor.cs:using RestSharp.Portable;
samples/client/petstore/csharp/SwaggerClient.v5/src/IO.Swagger/IO.Swagger.csproj:  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\Portable\$(TargetFrameworkVersion)\Microsoft.Portable.CSharp.targets" />
samples/client/petstore/csharp/SwaggerClient.v5/src/IO.Swagger/project.json:        "FubarCoder.RestSharp.Portable.Core": "4.0.7",
samples/client/petstore/csharp/SwaggerClient.v5/src/IO.Swagger/project.json:        "FubarCoder.RestSharp.Portable.HttpClient": "4.0.7",
```

johnamcruz:
Yes, I think the problem is that its not recognizing the framework 5.0 so its defaulting to 4.5. Do you know how i can install .net standard 1.3 on my machine without using nuget?

wing328:
Can you show the first few lines when generating the C# API? Here is what I got what generating C# API client for .net standard 1.3:
```
swagger-codegen|csharp_net10_sample⚡ ⇒ ./bin/csharp-petstore-v5.sh 
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=256M; support was removed in 8.0
[main] INFO io.swagger.parser.Swagger20Parser - reading from modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml
[main] INFO io.swagger.codegen.languages.CSharpClientCodegen - Generating code for .NET Framework v5.0
[main] WARN io.swagger.codegen.languages.AbstractCSharpCodegen - Return (reserved word) cannot be used as model name. Renamed to ModelReturn
[main] WARN io.swagger.codegen.languages.AbstractCSharpCodegen - Return (reserved word) cannot be used as model name. Renamed to ModelReturn
```

wing328:
Also are you generating the C# API client to the same output folder? Can you try a different folder instead?

johnamcruz:
java -jar swagger-codegen-cli.jar generate -i swagger.json
 -l csharp -o TEST --additional-properties targetFramework=v5.0,packageGuid={3AB1F259-1769-484B-9411-84505FCCBD55}
[main] INFO io.swagger.parser.Swagger20Parser - reading from swagger.json
[main] WARN io.swagger.codegen.ignore.CodegenIgnoreProcessor - Directory does not exist, or is inaccessible. No file will be evaluated.
[main] WARN io.swagger.codegen.languages.CSharpClientCodegen - Invalid .NET framework version, defaulting to v4.5
[main] INFO io.swagger.codegen.languages.CSharpClientCodegen - Generating code for .NET Framework v4.5

wing328:
What did you get when running `java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar config-help -l csharp` ? Here is what I got for `targetFramework`:
```
	targetFramework
	    The target .NET framework version. (Default: v4.5)
	        v3.5 - .NET Framework 3.5 compatible
	        v4.5 - .NET Framework 4.5+ compatible
	        v5.0 - .NET Standard 1.3 compatible
	        uwp - Universal Windows Platform (IMPORTANT: this will be decommissioned and replaced by v5.0)
```
I suspected you're not using the latest master to build the JAR locally.

johnamcruz:
`
        targetFramework
            The target .NET framework version. (Default: v4.5)
                v3.5 - .NET Framework 3.5 compatible
                v4.5 - .NET Framework 4.5+ compatible
                uwp - Universal Windows Platform - beta support`

wing328:
@johnamcruz so are you able to successfully do a `mvn clean package` to rebuild the JAR locally? (assuming you're pull the latest master)

johnamcruz:
yeah i was able to do that. @wing328 I think the problem is that I dont have v5.0 .NET standard 1.3 installed in my machine.

wing328:
No, it's not related. The option for v5.0 should still be shown no matter .NET standard 1.3 is installed locally in your machine or not. (in other words, swagger codegen won't check what .net framework/standard is installed in the machine)

I suspect you may be pulling the latest master of your folk of Swagger Codegen repo (not the official one).

Please try cloning https://github.com/swagger-api/swagger-codegen in another directory and build the JAR locally.

johnamcruz:
let me try that on another directory. thank you so much @wing328

jimschubert:
For others who may find this thread and experience similar confusion, .NET Standard isn't an installable product-- it refers to a minimal supported set of features for libraries used across .NET flavors. It is similar to  system requirements statements like "Vista or higher".

Saying something is .NET Standard 1.3 only refers to the supported feature set for the given flavor of .NET (Core, Framework, UWP). The v5.0 refers to the PCL been version

johnamcruz:
thank you @wing328 @jimschubert

wing328:
FYI. Added @Gronsak as template creator in https://github.com/swagger-api/swagger-codegen/blob/master/README.md#template-creator

Gronsak:
@wing328 haven't been able to check in for a while and come back to this. Glad I was able to contribute with something!

wing328:
@Gronsak I'm sure the community appreciates your contribution to this project. You may want to take a look at https://github.com/swagger-api/swagger-codegen/issues/5527 to see if you've any feedback.

Issue #3365, opened 2016-07-13
[All languages] Support for gzip compression


michaelkourlas:
##### Description

This is a feature request for two options enabling support for gzip compression for the generated API clients for all languages.

The first option would enable support for compression in API requests. When this option is turned on, the client would:

1. Add `Content-Encoding: gzip` to the HTTP request headers
2. Compress the HTTP request bodies using gzip

The second option would enable support for compression in API responses. When this option is turned on, the client would:

1. Add `Accept-Encoding: gzip` to the HTTP request headers
2. Decompress the bodies of HTTP responses with a response header of `Content-Encoding: gzip`

Ideally, both options would be on by default.

##### Related issues

#3279 discusses how to add gzip support to the Java API.

##### Suggest a Fix

For some languages, this won't be possible. For example, the JavaScript API presumably uses XmlHttpRequest behind the scenes when running in the browser, and the [spec won't let you set encoding-related headers](https://www.w3.org/TR/XMLHttpRequest/#the-setrequestheader()-method).

For other languages, though, this should be fairly straightforward. For example, in Java, adding support for GZIP compression for API requests and responses in the Jersey 1.x client consists of simply adding:

`client.addFilter(new GZIPContentEncodingFilter());`

to the `rebuildHttpClient` function in ApiClient.java. Passing in the parameter `false` to the constructor would enable it only for responses. Simply encapsulate this behaviour behind a couple of properties and you're done!

wing328:
I saw your proposal in https://github.com/OAI/OpenAPI-Specification/issues/738

For the time being, we can use `x-content-encoding` (in the same level as `consumes`, `produces`)

To support it, it comes down to the HTTP libraries used in the API clients (whether it has a switch/flag to easily support gzip or other content encodings).

jfiala:
added gzip support to the jaxrs-cxf client in PR #4003 + documented the supported gzip feature using interface GzipFeatures.

chenditc:
It would be very helpful for the client using mobiles. Saving mobile data and increase speed can improve UX a lot.

PR #4967, opened 2017-03-07
[Java] Fix for issue #3365: Support for gzip compression

michaelkourlas:
This PR is a fix for #3365. Request and response compression is now supported for the Java jersey1 and okhttp-gson APIs. Request compression is disabled by default; to turn it on, use `-DuseGzipFeature=true`.


wing328:
@michaelkourlas I did some tests and the result is good. Thanks for your contribution.

cc @cbornet @jfiala

Issue #4982, opened 2017-03-08
html emits 0 response code for "default" responses


DavidBiesack:
##### Description

See [petstore output](https://github.com/swagger-api/swagger-codegen/blob/master/samples/html/index.html)
```
    <h3 class="field-label">Responses</h3>
    <h4 class="field-label">0</h4>
    successful operation
```
##### Swagger-codegen version

2.2.3-SNAPSHOT

##### Swagger declaration file content or url

petstore.yaml

bin/html-petstore.sh

##### Suggest a Fix

I'll submit a patch. isDefault in CodegenResponse is insufficient because it is true for 200 responses.
Needs a new member isDefaultCode which is only true if the response key is
```
  default: 
```

wing328:
@DavidBiesack there's a similar fix (merged) for `html2`: https://github.com/swagger-api/swagger-codegen/pull/4914

DavidBiesack:
Thanks, @wing328  - I'll abandon this PR and adopt the approach used on #4914 - it is simpler and has fewer changes

DavidBiesack:
Fixed in #4991

PR #4984, opened 2017-03-08
Update html template and handle "default" response code (issue #4982)

DavidBiesack:
Add title for models to TOC and each model's section
Render model description only conditionally.
Remove extraneous Up link
Render "default" instead of 0 for response code when Swagger uses "default"
Regenerate samples/html/index.html

Addresses  #4982


DavidBiesack:
I'll abandon this PR and adopt the approach used on #4914 - it is simpler and has fewer changes. New PR coming.

PR #4991, opened 2017-03-09
Improve HTML output for model titles, default responses

DavidBiesack:
Render "default" instead of 0 for response code when Swagger uses "default" (#4982)
Add title for models to TOC and each model's section
Render model description only conditionally.
Remove extraneous Up link
Regenerate samples/html/index.html


wing328:
@DavidBiesack the change looks good to me. Thanks for the PR.

Issue #4087, opened 2016-10-27
Contract based tests


jfiala:
##### Description

Currently there are no unittests ensuring that the contracts (i.e. the outcome of the language-generators) change only when they are supposed to change. What about adding at least one unittest for each language to secure this and to track all the generated output files (better their hashes)? For different configuration options of the language maybe additional tests make sense. I think this would also track breaking changes a lot easier...

jfiala:
Automatically generating the output for all the petstore-variations and also allow for different language configurations would make the results more reliable. The current process is unreliable because if someone forgets to update the generated files they will not be validated. IMHO the expected outcome should be fixed, but the actual outcome should always get generated dynamically.

@wing328 can you please review the idea? If you want I can setup a separate module for this. It would also allow to replace the shellscripts with an integrated maven build.

wing328:
@jfiala is your suggestion similar to https://github.com/civisanalytics/swagger-diff, which shows the differences between 2 Swagger specs?

jfiala:
No, I think the manual process of running the shellscripts is tiresome and leads to outdated generated sample files (I've had this numerous times), which forces the updated to also fix other bugs that have creeped into the language. Also, the shellscripts use different contracts which adds additional complexity.

I'd suggest to setup a separate integration-test project which will  automatically generate all the languages including test variations with every build, forcing language updaters to be always consistent and otherwise the build will fail. 

Currently forgetting  to update the samples leads to no build errors and therefore bugs creep into the languages without getting noticed.

So my suggestion basically should be:
1. add a separate test module which generates all samples automatically using unittests and which will fail if the expected outcome differs from the actual (generated) outcome, forcing language updates to update the expected outcome and thus maintain the tests.

2. add additional tests for the test variations (e.g. security-contract petstore-securityy-test.yaml etc.).

3. maybe also implement cross-language consistency checks, as some features should be consistent throughout all languages

What do you think?

jfiala:
@wing328 please take a look at the PR #4721, if you like it we can migrate all the shellscripts into unittests and also add the directory layout comparison to make sure everything is equal, maybe also improve the error reporting.

jfiala:
[...]

PR #4995, opened 2017-03-09
[Jaxrs-spec] Java-based generator + unittest for jaxrs-spec

jfiala:
### PR checklist

### Description of the PR

For details see #4087 

This is the first draft:
1. Instead of using the shellscripts, define a Java class to generate the ouput with all the desired options.
first example: JaxrsSpec_Generator_Petstore.java

2. add a check for equality of all files 
(missing: compare directory structures)
example: Jaxrs_Spec_Generator_Petstore_Test.java

3. Setup a unittest to verify if the samples are equal with the generated output automatically 
This means no more outdated samples, otherwise the build will fail!

The basic idea has been taken from the swagger2markup project.

### Releated issues
#4087

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/generators/ComparePathUtil.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/generators/ComparePathUtil.java
@@ -1,0 +1,134 @@
+package io.swagger.codegen.generators;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import org.apache.commons.io.FileUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ComparePathUtil {
+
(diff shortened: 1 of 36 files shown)


jfiala:
@wing328: I implemented the first draft for jaxrs-spec to generate & compare the outcome automatically. 

Currently the shippable build fails:
``` log 
[main] INFO io.swagger.codegen.generators.util.ComparePathUtil - checking file /tmp/junit5278387355632741777/src/gen/java/io/swagger/model/Capitalization.java
[main] INFO io.swagger.codegen.generators.util.ComparePathUtil - files not equal: compare: ../../samples/server/petstore/jaxrs-spec/src/gen/java/io/swagger/model/Capitalization.java, file: Capitalization.java!
``` 

Is there an easy way to access a file from the shippable build (here: /tmp/junit5278387355632741777/src/gen/java/io/swagger/model/Capitalization.java)?
It would be necessary to compare the expected with the actual outcome during the build.
On my local pc everything runs fine, there is no difference in Capitalization.java, so I'm wondering what is different during the CI build...

wing328:
@jfiala can you create an account at Shippable CI and access the following page?

https://app.shippable.com/github/swagger-api/swagger-codegen

Issue #4986, opened 2017-03-09
Python API client refactoring


wing328:
##### Description

Currently, Python API client is singleton and we want to move away from singleton in Python to configuration per instance of API client similar to what we've done to Ruby, PHP, C#, etc.

##### Swagger-codegen version

Latest matser

##### Related issues

https://github.com/swagger-api/swagger-codegen/issues/4240

##### Suggest a Fix

If anyone from the community wants to work on this enhancement, please reply to let us know.

baartosz:
I can give it a shot but I will need someone to review it carefully as I am not doing much python besides simple cli scripts.

wing328:
Closed via #5012. Thanks for the contribution by @baartosz

taxpon:
@wing328 This issue can be closed.

wing328:
@taxpon thx for the reminder.

PR #5012, opened 2017-03-10
[Python] remove singleton

baartosz:
### PR checklist

### Description of the PR

Resolves https://github.com/swagger-api/swagger-codegen/issues/4986. Configuration class is not singleton anymore, similarly to PHP version. PR is to 2.3 branch as it changes a bit how client object is instantiated:
```
        config = petstore_api.Configuration()
        config.host = 'http://localhost/'
        api = PetApi(petstore_api.ApiClient(config))
        pet = api.get_pet_by_id(10)
```


wing328:
@baartosz thanks for the PR.

cc @scottrw93 @frol @taxpon @scop, please review when you've time.

frol:
This sounds like a great change! I am reviewing the implementation now.

@wing328 Thanks for pinging me.

samples/client/petstore/python/petstore_api/api_client.py
```diff
@@ -116,7 +113,7 @@ def __call_api(self, resource_path, method,
                                                     collection_formats)
             for k, v in path_params:
                 resource_path = resource_path.replace(
-                    '{%s}' % k, quote(str(v), safe=''))  # no safe chars, encode everything
+                    '{%s}' % k, quote(str(v)))
```
frol: @wing328 Why aren't these PRs merged into 2.3.0?

* https://github.com/swagger-api/swagger-codegen/commit/6a080d3a0c26015262120cdd9e7f4f2feeb59b23
* https://github.com/swagger-api/swagger-codegen/commit/c366f139e28b4b67f11a51807580f897afccc305

frol: This change should be reverted to the upstream version in any case since it is irrelevant to this PR (and the former variant with `safe=''` is the correct one). Hopefully, the mentioned commits will land into 2.3.0 branch soon.

frol:
@baartosz Great enhancement to the Python client! Thank you.

May I ask you to also update the README template to include the proper example of configuration?

modules/swagger-codegen/src/main/resources/python/api.mustache
```diff
@@ -24,13 +24,9 @@ class {{classname}}(object):
     """
 
     def __init__(self, api_client=None):
-        config = Configuration()
```
frol:
```python
from ..configuration import Configuration
```

This import is unnecessary now.

samples/client/petstore/python/petstore_api/api_client.py
baartosz: should I also revert other unrelated changes in regenerated samples? e.g. `fake_classname_tags123_api.py` renamed to `fake_classname_tags_123_api`

baartosz:
@frol Thanks for reviewing this PR. I added requested changes.

frol:
@wing328 This change looks great to me now.

@wing328 Please, pick the 6a080d3 and c366f13 commits into 2.3.0 line from master.

@baartosz Thank you for your work!

wing328:
> @wing328 Please, pick the 6a080d3 and c366f13 commits into 2.3.0 line from master.

I'm in the process of syncing enhancements from master to 2.3.0. Will let you guys know if I need any help. Thanks for listing out commits (enhancements, bug fixes) that need to get into 2.3.0.

baartosz:
Rebased with 2.3.0 to solve merge conflicts

wing328:
@baartosz PR merged into master. Thanks for the huge enhancement.

wing328:
## Upgrade Note ##

`Configuration` is now per `ApiClient` instead of singleton.

jdevera:
I found this when I was going to report a problem. I don't know if you have seen this, but my generated Api classes all call the ApiClient constructor without parameters, which always forces the creation of a a new configuration object per Api and leaves out the positibility of passing down any configuration from the client code.

jdevera:
Ok, just realised that what I have to do is create an ApiClient with my custom configuration and pass it to the different Api objects.

kmxillo:
When I create in the same Class the different API Client Configurations It overrides the configuration even though being called from different APIs. 

```python 
config1 = petstore_api_1.Configuration()
config2 = petstore_api_2.Configuration()
``` 
After than I do:
```python
api1 = PetApi(petstore_api_1.ApiClient(config1))
api2= PetApi(petstore_api_2.ApiClient(config2))

pet = api1.get_pet_by_id(10)
pet = api2.get_pet_by_id(10)
```

When I comment out one of the configurations it works as expected. 

Why cant I have several configurations when Configuration Class is not anymore Singleton?

wing328:
cc @taxpon @frol @mbohlool @cbornet

frol:
@kmxillo It is not clear what gets wrong on your side. There should be no global configuration anymore with Codegen 2.3.0+ (I don't believe 2.2.x include the changes). I also suggest you try the master and 3.0.0 branches just for the sake of testing, they are pretty stable.

kmxillo:
@frol  I can see that the 2.3.0+ or 3.0.0 behavior is different as 2.2.* Thanks for that. 

But as I described above, for instance, some configurations settings are overwritten.  I would really like to know why this occurs. Since I see this unexpected behavior I cannot accept that  the implementation was successfully done.

Unexpected Behaviour:
- I dont see urllib3.connectionpool - DEBUG: Starting new ... for INT when I switched debug. 

  => I have seen that petstore_api_2.Configuration() overrides config_api1.logger['urllib3_logger'].level WHY?

I cant state that calls are executed correctly if I see such wrong behaviour.

As you see in my output. I call two environments {INT,PREPROD}.  

**OUTPUT:** 
```
pythonapp - INFO: Executing Activation Services at: INT
pythonapp - INFO: {'basicAuth': {'type': 'basic'}}
DEBUG response body: Code 200
petApi.rest - DEBUG: response body: Successfully added.
pythonapp - INFO: Executing Activation Services at: PREPROD
pythonapp - INFO: {'basicAuth': {'type': 'basic'}}
DEBUG Starting new HTTPS connection (1): PREPROD.host
DEBUG Starting new HTTPS connection (1): PREPROD.host
urllib3.connectionpool - DEBUG: Starting new HTTPS connection (1): PREPROD.host
```
Here is how I initialized the **configurations**:

```
class ClientSettings(object):

    def __new__(self,system):

        #  Config for petstore_api_1
        # ########################
        config_api1 = petstore_api_1.Configuration()
        config_api1 .debug = True
        config_api1.host = 'INT.host'
        config_api1.username = username
        config_api1.password = password

        # Config for petstore_api_2
        # #######################
        config_api2 = petstore_api_2.Configuration()
        config_api2.host = 'PREPROD.host'
        config_api2.username = username
        config_api2.password = password

        configurations = {}
        configurations['api1'] = config_api1
        configurations['api2'] = config_api2

        return configuration
```

frol:
@kmxillo Well, there is a definitely weird implementation around logging:

1. Configuration.__init__ sets `self.debug = False`: https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/python/petstore_api/configuration.py#L83
2. `self.debug`, being a property, updates the level of the loggers: https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/python/petstore_api/configuration.py#L164-L176

This doesn't make Configuration to be in a shared state, but it, indeed, does unexpected things with the logging configuration. I don't know how to fix that, but remove it.

e-desouza:
I can confirm (in 2023!) that this does indeed work as expected (no singleton Configuration).

PR #5017, opened 2017-03-10
null out the string `null` in examples.  Rebuilt java clients

### LINKED ISSUE #5016: don't write string "null" for examples
in 2.2.2, we are returning an example value for model properties.  If not defined in the definition, this logic is returning the string "null".  The following check:

```
@ApiModelProperty({{#example}}example = "{{example}}", {{/example}}
```

Will then always write the example `null`.  I don't think this was the intention.

fehguy:
A small change but makes a lot of generated code changes.

Fixes #5016

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -1481,7 +1481,10 @@ public class DefaultCodegen {
         property.title = p.getTitle();
         property.getter = "get" + getterAndSetterCapitalize(name);
         property.setter = "set" + getterAndSetterCapitalize(name);
-        property.example = toExampleValue(p);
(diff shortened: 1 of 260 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
```diff
@@ -1481,7 +1481,10 @@ public CodegenProperty fromProperty(String name, Property p) {
         property.title = p.getTitle();
         property.getter = "get" + getterAndSetterCapitalize(name);
         property.setter = "set" + getterAndSetterCapitalize(name);
-        property.example = toExampleValue(p);
+        String example = toExampleValue(p);
```
fehguy: Instead of change the logic in this function (which may affect things elsewhere) just don't set a property value if the example is the string `null`.

PR #5019, opened 2017-03-10
fix (partially) #4898 for dart.

ePaul:
### PR checklist

   → ran bin/dart-petstore.sh, with no changes in the produced samples.
- [x] Filed the PR against the correct branch: master for non-breaking changes

### Description of the PR

This fixes (or hopes to fix) part of #4898 for the `dart` generator.

The problem likely occurred when a path parameter name (as declared in the OpenAPI definition) is of a form which doesn't match the form used in Dart, or is a reserved word there. In this case the parameter gets renamed for use in the code, and the algorithm for creating the actual path tries to replace the renamed parameter in the path (which will do nothing), not the original one (which would be correct).

Our Petstore API definition doesn't seem to contain a path parameter which would get modified by the sanitation, therefore there is no difference in the generated samples.

I don't know anything about Dart, so this needs review by Dart experts. /cc @yissachar @close2 @ircecho @hcwilhelm

### DIFF
--- a/modules/swagger-codegen/src/main/resources/dart/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/dart/api.mustache
@@ -25,7 +25,7 @@ class {{classname}} {
     {{/allParams}}
 
     // create path and map variables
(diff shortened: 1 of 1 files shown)


wing328:
@ePaul thanks for the fix, which looks good to me (although I'm no expert in Dart)

Issue #3436, opened 2016-07-21
Formatting not getting reflected when converting yaml file to HTML in swagger online editor


webron:
_From @Rasika3091 on July 21, 2016 14:46_
- Version: Swagger editor online
- Browser/OS:  Chrome, Windows 10

Hello,

I am converting my yaml file to HMTL through the swagger editor online. But the formatting applied in the descriptions are not getting applied. Its not looking proper as seen in the editor. I have provided on example. Please open it in Online swagger editor and covert it to HTML and take a look at all the description parts.

`````` yaml
swagger: "2.0"
info:
  version: "0.0.1"
  title: Hello World App
  description: |
               Description for test :

               ###### Types :

               ```
                  1) Standard Fields
                  2) Custom Fields
                  3) Old Fields
               ```

               ###### JSON structure :

               ```
               {"standardFields":
                   [
                     {"id":Provide field Id,
                      "value":Provide value for that standard field}
                  ]
               }
               ```

# during dev, should point to your local machine
host: localhost:10010
# basePath prefixes all resource paths 
basePath: /
# 
schemes:
  # tip: remove http to make production-grade
  - http
  - https
# format of bodies a client can send (Content-Type)
consumes:
  - application/json
# format of the responses to the client (Accepts)
produces:
  - application/json
paths:
  /hello:
      # binds a127 app logic to a route
      x-swagger-router-controller: hello_world
      get:
        description: |
                     Returns 'Hello' to the caller

                     ###### Access

                     - New user
                     - Updated User
        operationId: hello
        parameters:
          - name: name
            in: query
            description: The name of the person to whom to say hello
            required: false
            type: string
          - name: id
            in: query
            description: Id required
            required: true
            type: string
        responses:
          "200":
            description: Success
            schema:
              # a pointer to a definition
              $ref: "#/definitions/HelloWorldResponse"
          # responses may fall through to errors
          default:
            description: Error
            schema:
              $ref: "#/definitions/ErrorResponse"
  /swagger:
    x-swagger-pipe: swagger_raw
  /getName:
     get:
       description: |
                    Get name

                    ###### Access

                     - New user
                     - Updated User
       parameters:
          - name: name
            in: query
            description: The name of the person to whom to say hello
            required: false
            type: string
          - name: id
            in: query
            description: Id required
            required: true
            type: string
       responses:
          "200":
            description: Success
            schema:
              # a pointer to a definition
              $ref: "#/definitions/HelloWorldResponse"
[...]

PR #5020, opened 2017-03-10
Allow control of HTML escaping in templates (#3436)

DavidBiesack:
See #3436 for description

I tried to change the version in pom.xml from 2.2.3-SNAPSHOT to 2.2.4-SNAPSHOT but the build fails with google checkstyle (not finding the .xml file???) and I don't know how to address that, so I left the version number unchanged; I supposed that's OK since this is still a -SNAPSHOT, but really this should be a bump in the minor version (2.3.0-SNAPSHOT) because this is not really just a patch but a new feature (new API).

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConfig.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConfig.java
@@ -10,9 +10,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import com.samskivert.mustache.Escapers;
 import com.samskivert.mustache.Mustache.Compiler;
+import com.samskivert.mustache.Mustache.Escaper;
 
 public interface CodegenConfig {
+
     CodegenType getTag();
     
     String getName();
@@ -178,7 +181,7 @@ public interface CodegenConfig {
     /**
      * Library template (sub-template).
      *
-     * @return libray template
+     * @return library template
      */
     String getLibrary();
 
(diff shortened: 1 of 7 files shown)


ePaul:
I think the idea is for breaking changes to go into the 2.3.0 branch, not directly into master.

DavidBiesack:
This is *not* a breaking change. All the existing tests pass without modification - this is an enhancement/new feature (thus a dot release bump; in semantic versioning, a breaking change would have to be a 3.0.0 version instead of a dot release). 

That said, this change could go into a 2.3.0 branch instead of master if that is preferred for such enhancements.

(I think my problem with version in pom.xml is that I need to set it in the parent, but also update all the module poms.)

DavidBiesack:
Is there a release plan for 2.3.0 -- i.e. if pushed there, when might such a fix be available in Maven Central or other public Maven repo?

DavidBiesack:
Closing this pull request. Upon inspection of the code, I can see a better way to achieve the desired result. The `CodegenConfig` interface defines
```java
 Compiler processCompiler(Compiler compiler);
```
which a custom language class that extends `DefaultCodegen` can override.
I suggest the html and html2 langs do this:
```java
    /**
     * Disable HTML escaping done by the Mustache {@link Compiler} 
     * by setting the compiler to {@link Escapers#NONE Escapers.NONE}
     */
    @Override
    public Compiler processCompiler(Compiler compiler) {
        return compiler.withEscaper(Escapers.NONE);
    }
```

PR #5022, opened 2017-03-10
Fix (partially) #4898 for flash

ePaul:
### PR checklist

   → bin/flash-petstore.sh

### Description of the PR

This fixes (or hopes to fix) part of #4898 for the `flash` generator.

The problem likely occurred when a path, query or header parameter name (as declared in the OpenAPI definition) is of a form which doesn't match the form used in Flash, or is a reserved word there. In this case the parameter gets renamed for use in the code, and the algorithm for creating the actual path tries to replace the renamed parameter in the path (which will do nothing), not the original one (which would be correct). For query or header parameters, the parameter will be sent with the sanitized name instead of the original one.

Our Petstore sample API definition doesn't seem to contain any path or query parameter which would get modified by the sanitation, so the regenerated samples just have a single difference in a header parameter (which looks better now to me).

I don't know anything about Flash, so this needs review by Flash experts. /cc @glederrey

### DIFF
--- a/modules/swagger-codegen/src/main/resources/flash/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/flash/api.mustache
@@ -36,7 +36,7 @@ public class {{classname}} extends SwaggerApi {
      */
     public function {{nickname}} ({{#allParams}}{{paramName}}: {{{dataType}}}{{#hasMore}}, {{/hasMore}}{{/allParams}}): String {
         // create path and map variables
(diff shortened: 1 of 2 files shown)


wing328:
@ePaul thanks for the fix, which looks good to me (although I'm no expert in ActionScript).

glederrey:
@ePaul Sadly, I'm not an expert in ActionScript either. I just used this language for a small project during my internship. Sorry.

PR #5030, opened 2017-03-12
Fix (partially) #4898 for groovy

ePaul:
### PR checklist

   → bin/groovy-petstore.sh
- [x] Filed the PR against the correct branch: master for non-breaking changes

### Description of the PR

This fixes (or hopes to fix) part of #4898 for the `groovy` generator.

The problem likely occurred when a query or header parameter name (as declared in the OpenAPI definition) is of a form which doesn't match the form used in Groovy, or is a reserved word there. In this case the parameter gets renamed for use in the code, and the parameter will be sent with the sanitized name instead of the original one.

_**Note** that it looks like the Groovy generator has no support at all for path, form or body parameters – those are accepted by the generated code, but then ignored, and not passed to the server. I didn't try to fix this, as my Groovy knowledge is too rusty for this, but added some TODOs in the template. Related issue: #4289_

Our Petstore sample API definition doesn't seem to contain any query parameter which would get modified by the sanitation, so the regenerated samples just have a single difference in a header parameter (which looks better now to me).

I don't know anything about Groovy, so this needs review by Groovy experts.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Groovy/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/Groovy/api.mustache
@@ -18,7 +18,7 @@ class {{classname}} {
 
     {{#operation}}
     def {{operationId}} ({{#allParams}} {{{dataType}}} {{paramName}},{{/allParams}} Closure onSuccess, Closure onFailure)  {
-        // create path and map variables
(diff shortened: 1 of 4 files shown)


wing328:
@ePaul the fix looks good to me. I should be able to add the path parameter support.

PR #5049, opened 2017-03-13
Fixes ListBuffer import-mapping for scala clients

jostick:
### Description of the PR

This is a trivial change fixing a broken import. Scala client will otherwise not compile.


wing328:
@jostick thanks for the fix.

Starting from 2.3.0 (next major release), we'll consolidate scala and async-scala into a single scala generator via #4855. When you've time, please give it a try and let us know if you've any feedback.

(we'll also incorporate this fix into #4855 when we merge it into 2.3.0 later)

PR #5054, opened 2017-03-13
Update samples for sync_master_230

ePaul:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change.
  → updated all the samples (except clojure and scalatra's build.sbt change – those were fixed in master, but not yet here).
- [x] Filed the PR against the correct branch: sync_master_230 (which is intended to be later merged into 2.3.0)

### Description of the PR

This updates all the samples after the merge of master and 2.3.0 so CI (and people) can check the results.

PR #5056, opened 2017-03-13
Merge 2.3.0 into sync_master_230

ePaul:
### PR checklist

### Description of the PR

This merges the 2.3.0 branch into @wing328's sync_master_230 branch, solving some more conflicts on the way (as 2.3.0 did progress a bit more since the initial merge).

I just updated the swift3 + nodejs-server samples (which had conflicts), I guess updating some more should be done after merged.


ePaul:
Looks like Sonatype doesn't have the earlier-than-latest snapshots of swagger-parser anymore, and therefore AppVeyor build breaks. Possibly we can change to a release version, or to the latest snapshot (1.1.0-SNAPSHOT)?

ePaul:
@fehguy done ... though this seems to break at least BashCodegen.
`op.vendorExtensions.get("x-code-samples")` (line 587) now seems to return a List of LinkedHashMap instead of a list of ObjectNode, which makes BashTest.petstoreOperationTest fail (at least in this branch).

I'll try the effects of just the update in master and 2.3.0 branch.

ePaul:
Okay, no problem in master (which actually already is on 1.0.28), and not on 2.3.0 (which is on 1.0.25 currently), but already on the current sync_master_230. I'll try to identify the origin of the problem.

ePaul:
Aha, master has an if-block here. I guess I'll cherry-pick that commit for this branch too.

ePaul:
I applied the fixes on the branch for #5054, then merged into this one.

Issue #5053, opened 2017-03-13
Rails5 Server: How to generate CRUD operations? 


bluengreen:

bluengreen:
Ok. After 101 tries to figure out what the heck. It seems to be that the tags, and the paths have to match in order to have the isRestful work while generating. Which is a problem since this change 

https://github.com/apanzerj/swagger-codegen/blob/2aadbee8edc4f27d72756f99ced5d326d18c148f/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Rails5ServerCodegen.java#L284-L285

The class name and the file name need to match.

PR #5057, opened 2017-03-13
updated to fix issue #5053

bluengreen:
### PR checklist

### Description of the PR

Fix for #5053 

This fixes adding an "s" to controller names. Rails doesn't use inflection on controllers. So whatever the class name is the file name should follow otherwise you will get an missing constant error. 

Also, this fixes an issue where generating a controller to use the isRestuful in the mustache template. It requires the tag name and the path name to match in order to generate the correct default restful method names (index, update, create, show, delete). By adding the "s" automatically and trying to compensate for the additional "s" on a tag by using singular, the isRestful will not work. 

This reverts back to just using the tag name for controllers. And does not affect the model being singular when there is an "s" in the path. 

General rule for rails controllers is plural paths - "pets/"


wing328:
@bluengreen thanks for the PR. I noticed that the file `pets_controller.rb` was renamed to `pet_controller.rb`, which does to conform to Rails naming convention: https://gist.github.com/iangreenleaf/b206d09c587e8fc6399e#controllers

Would pluralizing `name` be a better solution? 

cc @apanzerj @zlx

bluengreen:
I think its just confusing to generate the server code different from the way the spec file reads. It's like trying to interpret something rather than just being explicit. To me the idea of the spec is to define what I want the paths in my server code to be. Anything else seems like trying to be too and therefore confusing. 

If I set the path to be /pets it seems obvious the controller route will be /pets. If I set it to /pet I get a route that is at /pet. Just clearer. 

I noticed this issue because I was trying to use the isRestful routes features. Which only kick in when the tag and the path match. And they didn't match because the path I wanted was /pets but the tag was pet. If I set the tag to pets then the filename is `petss_controller.rb` and the classname is Pets. If I left it the singular the filename is correct `pets_controller.rb` but the classname is incorrect 'Pet'. So just seems to make more sense to make your tags match the endpoint and then your explicitly defining what you want. Which to me, seems like the point of the spec file.

bluengreen:
Basically the easy solution to make it rails friendly with this change, is make the tags plural. But maybe that isn't agreed upon direction for the codegen code.

wing328:
@bluengreen thanks for the PR. Let me know think about this and get back to you.

aegixx:
I think you could go either way, but the current implementation does not function because of mixed usage.  Any way to move this PR forward?  

IMHO, I like singular because it matches the language (i.e. the `PetController` is the `Pet Controller`) provided, even though the RESTful route has a plural.

bluengreen:
My thinking was just make it what you are setting your route to be in the spec. Plural or otherwise. Trying to interpret the intent is too tricky and doesn't fit all situations. Just too clever. And the clever part is getting in the way. 😕

wing328:
@bluengreen thanks again for the PR and sorry for taking a bit long to review.

What I would suggest is to add an option later via another PR to allow different naming convention (e.g. singular, plural, etc).

For TypeScript, we've already offered similar option:
```
modelPropertyNaming
```

Issue #4788, opened 2017-02-13
CodegenConfigurator systemProperties are reused between code generation invokations


JLLeitschuh:
I'm using the swagger `CodegenConfigurator` in a gradle build where the swagger codegen project is loaded as a library and is being directly passed into the `DefaultGenerator`.

This usage in [this gradle plugin is very similar to my usage](https://github.com/thebignet/swagger-codegen-gradle-plugin/blob/master/src/main/groovy/org/detoeuf/SwaggerCodeGenTask.groovy#L42).

If I configure one `CodegenConfigurator` object with `systemProperties` being set, these settings seem to be being stored somewhere statically.

I'm observing that if I use the `DefaultGenerator` again in the same JVM I end up having files not get generated because of input from a previous run of the generator.

I know that the two `CodegenConfigurator` objects that I'm passing in are different because I've tested logging the `systemProperties` variable just before passing it in and for the first call it is populated and the second time it is empty.

JLLeitschuh:
It seems that these lines are the one to blame in `CodegenConfigurator `:

```
    private void setSystemProperties() {
        for (Map.Entry<String, String> entry : systemProperties.entrySet()) {
            System.setProperty(entry.getKey(), entry.getValue());
        }
    }
```
https://github.com/swagger-api/swagger-codegen/blob/029728d851e63dc2887e126e1944f61a04d4540a/modules/swagger-codegen/src/main/java/io/swagger/codegen/config/CodegenConfigurator.java#L466

otrosien:
@JLLeitschuh are you working on that subject?

ePaul:
This is a general problem of using Java system properties – a kind of global variables – for this thing.
We should try to avoid this in general, and pass the parameters over other channels.

otrosien:
@ePaul agree, do you have an idea? Beware that current users are relying on this feature.

ePaul:
>  Beware that current users are relying on this feature.

On which part of the feature do they rely? Do they actually need system properties being set (like `file.separator`), or just need a way to pass custom parameters to parts of the code?

I guess we need to clean up all locations in the code which are using `System.getProperty()` to use other ways of passing parameters instead (like the additionalProperties map, or maybe even have a "systemProperties" map being passed to the Codegens). The Configurator should put the data from the `systemProperties` parameter into the new map instead of setting system properties.

Paul

JLLeitschuh:
[...]

PR #5066, opened 2017-03-14
#4788: remove System.getProperty() from java-inflector + undertow codegens

ePaul:
Remove System.getProperty from language codegens.
It was just used for apiPackage/modelPackage, for which we have regular parameters.

### PR checklist

     → bin/java-undertow-server.sh, bin/java-inflector-petstore-server.sh (before the change, afterwards nothing changed.)
- [ ] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes. **(unclear, see below)**

### Description of the PR

This removes the usage of system properties from UndertowCodegen and JavaInflectorCodegen. It adds a warning in case the properties are still used. (This is part of a bigger effort to get rid of system properties, who have bad effects if there are have multiple (consecutive or even parallel) Codegen runs in the same JVM – see #4788.)

**I'm not sure if this counts as a breaking change** – the original effect of the system property being set is not there anymore.  – If so, I can put this against the 2.3.0 branch, and create a similar change for master which just outputs the warning, but still uses the system property. Opinions?
Also, is WARN the right level, or should this be ERROR?


ePaul:
This needs some reconsideration, I'll reopen it later again.

PR #5067, opened 2017-03-14
#4788: remove System.getProperty("noservice") in NodeJSServerCodegen

ePaul:
### PR checklist

   → bin/nodejs-petstore-google-cloud-functions.sh && bin/nodejs-petstore-server.sh. (Ran this before the change and again after the code change, with no changes, so nothing to commit.)
- [ ] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes. **(not sure, see below.)**

### Description of the PR

This introduces a CliOption `noservice` for NodeJSServerCodegen, with the same effect of the previous system property. The system property is not read anymore. This is part of a bigger effort to get rid of system properties, see #4788.)

**I'm not sure this counts as non-breaking** – if the parameter was passed via CodegenConfigurator's `systemProperties` map (e.g. via maven or `-Dnoservice` at the end of the command line in swagger-codegen-cli), it should still work now (though I didn't yet test this). If the parameter was a real system property passed from the outside (e.g. `java -D noservice ...` (i.e. before the class name, not after it), it will now not work anymore. If it is deemed to be breaking, I can move this to 2.3.0 and create a version for master which still reads the system property, but displays a warning.

(I also adapted the test so it still runs, but I'm not sure I did this correctly.)

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NodeJSServerCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NodeJSServerCodegen.java
@@ -22,7 +22,9 @@ import java.util.Map.Entry;
 
 public class NodeJSServerCodegen extends DefaultCodegen implements CodegenConfig {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(NodeJSServerCodegen.class);
+    public static final String NOSERVICE = "noservice";
+
+	private static final Logger LOGGER = LoggerFactory.getLogger(NodeJSServerCodegen.class);
 
     public static final String GOOGLE_CLOUD_FUNCTIONS = "googleCloudFunctions";
     public static final String EXPORTED_NAME = "exportedName";
@@ -91,6 +93,8 @@ public class NodeJSServerCodegen extends DefaultCodegen implements CodegenConfig
             "When the generated code will be deployed to Google Cloud Functions, this option can be "
                 + "used to update the name of the exported function. By default, it refers to the "
                 + "basePath. This does not affect normal standalone nodejs server code."));
+        cliOptions.add(new CliOption(NOSERVICE,
+        		"If this option is given, no service.js file will be generated."));
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


ePaul:
Rebased to get rid of conflicts.

ePaul:
Rebased again, now all CI jobs are okay.

wing328:
@ePaul thanks. I'll review shortly.

wing328:
>  if the parameter was passed via CodegenConfigurator's systemProperties map (e.g. via maven or -Dnoservice at the end of the command line in swagger-codegen-cli), it should still work now (though I didn't yet test this)

If that works, then I agree with you that this PR can be filed against master instead of 2.3.0

ePaul:
Hmm, actually it doesn't work with `-Dnoservice` at the end of the command line with current master, it needs `-Dnoservice=` (or any value after it). (At the start of the command line a plain `-Dnoservice` works.)

This stays the case afterwards, except that the "real system properties" at the start of the command line doesn't work anymore.

| command line | works before | works after the change |
| ------ | ------- | --------
| java -Dnoservice -jar swagger-codegen-cli.jar generate -i petstore.yaml -l nodejs-server | :white_check_mark: | :x: |
| java -jar swagger-codegen-cli.jar generate -i petstore.yaml -l nodejs-server -Dnoservice | :x: | :x: |
| java -jar swagger-codegen-cli.jar generate -i petstore.yaml -l nodejs-server -Dnoservice= | :white_check_mark: | :white_check_mark: |
| java -jar swagger-codegen-cli.jar generate -i petstore.yaml -l nodejs-server -Dnoservice=foo | :white_check_mark: | :white_check_mark: |

:white_check_mark: means that the service classes are not generated, :x: means they are generated (and the flag is ignored).

Should I add a generation script for the last case (in a separate PR)?
I've still have to check what happens with maven.

ePaul:
This `noservice` system property is not mentioned in the README, but in the help text of NodeJSServerCodegen.
Interestingly, this got copied to [some other Codegens](https://github.com/swagger-api/swagger-codegen/search?utf8=✓&q=-Dnoservice&type=), which actually don't support it.

ePaul:
This needs some reconsideration, I'll reopen it later again.

PR #5068, opened 2017-03-14
Use deterministic randomness in ExampleGenerator.

ePaul:
### PR checklist

    → bin/nodejs-petstore-google-cloud-functions.sh && bin/nodejs-petstore-server.sh (twice: before and after the update.)

### Description of the PR

This changes the randomness introduced into ExampleGenerator in #4797 to be deterministic.
This way, in each run of the generator the same output is produced when using the same input.

Previously, you got random changes each time you updated the NodeJS sample, which made it tedious to find out what actually changed. See ccc117c for an example commit containing just those changes. When merging different branches, this leads to unnecessary conflicts.

With this PR the randomness generator is totally deterministic, generating the sequence 0,6,1 for all integer values without min/max. I guess we could use the input for the example generator as a seed to the Random object, to have the result a bit more random-looking? Opinions?

(The last commit of this PR applies this on the nodejs samples, the first commit updates the samples with the old code.)

/cc @bartkummel @wing328


bartkummel:
Nive improvement, @ePaul !

wing328:
@ePaul thanks for the enhancement.

@bartkummel thanks for reviewing the change.

Issue #5075, opened 2017-03-15
[JAVA][jaxrs-spec] enumerated String values for QueryParams should render as enums instead of Strings


foxx1337:
##### Description

In Jersey it's possible to have an API like:
```Java
    enum SortingDirection {
    	ASC("asc"),
    	DESC("desc");

    	// ...
    }

    // ...

    @GET
    @Consumes({ "application/json" })
    @Produces({ "application/json" })
    public Response getSomething(@QueryParam("order") @DefaultValue("asc") SortingDirection direction) {
    	// ...
    }
```
and the runtime makes sure to convert the url `?order=asc` or `?order=desc` string to type `SortingDirection` provided some criteria are met - https://jsr311.java.net/nonav/javadoc/javax/ws/rs/QueryParam.html.
##### Swagger-codegen version

<=2.3.0-SNAPSHOT (current master)

##### Swagger declaration file content or url

```yaml
/something:
    get:
      operationId: getSomething
      parameters:
        - name: order
          description: How to sort the thing.
          in: query
          type: string
          enum: [asc, desc]
          default: asc
```
The current swagger-codegen implementation will simply generate
```Java
    @GET
    @Consumes({ "application/json" })
    @Produces({ "application/json" })
    public Response getSomething(@QueryParam("order") @DefaultValue("asc") String direction) {
    	// ...
    }
```
(please note that the `@DefaultValue("asc")` is part of pull request #5024, so it will be just missing without that one merged). In my opinion we should take advantage of the extra type safety provided by generating an enum, especially since the Java parser already has everything in memory for it.
##### Command line used for generation

`generate -l jaxrs-spec -i myapi.yaml -o jaxrs-spec-stub`

##### Steps to reproduce

It always happens, this is asking for an enhancement.

##### Suggest a Fix

I'll come by momentarily with a pull request for this.

wing328:
Closed via #5085

PR #5077, opened 2017-03-15
[JAVA][jaxrs-spec] Add support for enum as QueryParam argument type (#5075)

foxx1337:
[JAVA][jaxrs-spec] Add support for generating type-safe query parameters (#5075)

All the logic is already in place so I just make sure parameters in:query
also insert an enum in the generated API and that the type for the REST
operation parameter is that enum (and not just String as it was before).

### PR checklist

### Description of the PR

As the code was already in place, I just changed the templates a bit to also include the generated enums. For `enumClass.mustache` - https://github.com/foxx1337/swagger-codegen/blob/12a990f1/modules/swagger-codegen/src/main/resources/JavaJaxRS/spec/enumClass.mustache I indented everything by 1 extra level so that everything is properly indented in  the generated models or api classes (it was dedented in models and it looked bad).

### DIFF
--- a/modules/swagger-codegen/src/main/resources/JavaJaxRS/spec/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/JavaJaxRS/spec/api.mustache
@@ -23,6 +23,9 @@ import javax.validation.constraints.*;
 public class {{classname}}  {
 {{#operations}}
 {{#operation}}
+    {{#queryParams}}{{#datatypeWithEnum}}
+{{>enumClass}}{{/datatypeWithEnum}}{{/queryParams}}
+
 
     @{{httpMethod}}
     {{#subresourceOperation}}@Path("{{path}}"){{/subresourceOperation}}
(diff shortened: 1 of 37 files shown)


foxx1337:
Ok, i just had a look at
https://github.com/foxx1337/swagger-codegen/blob/12a990f16/samples/server/petstore/jaxrs-spec/src/gen/java/io/swagger/api/PetApi.java
and  I can see that it's broken:

```Java
 public enum List&lt;StatusEnum&gt; {

        AVAILABLE(List&lt;String&gt;.valueOf("available")), PENDING(List&lt;String&gt;.valueOf("pending")), SOLD(List&lt;String&gt;.valueOf("sold"));
```
(those &lt and &gt s are actually in the generated code).

I have to work some more on this one.

wing328:
> (those &lt and &gt s are actually in the generated code).

Thanks for the PR. You will need to use `{{{ ... }}}` instead of `{{ ... }}` to get the original value.

PR #5082, opened 2017-03-15
Modify Java/retrofit OAuth templates for #5039

### LINKED ISSUE #5039: BUG in OAuth2 OkHttpClient Interceptor: 401 is forwarded to client application
##### Description

If one uses OAuth2 to authenticate and the accessToken gets expired, the OAuth2-Interceptor correctly tries to request a new request token. But the responsible method in the generators (retrofit and others) have a bug which leads to returning the 401 response.

##### Swagger-codegen version

My current version is: 2.2.1. But it is still in master branch and 2.3.0

##### Steps to reproduce

- Set a short expiration time on an OAuth2 server
- Wait until it is expired
- look at the responsecode

##### Suggest a Fix

As the access token in the following lines is updated, it returns false if the accessToken is really updated:

https://github.com/swagger-api/swagger-codegen/blob/2.3.0/samples/client/petstore/java/retrofit2/src/main/java/io/swagger/client/auth/OAuth.java#L135

Therefore the return statement should be: `return !getAccessToken().equals(requestAccessToken);`

ThatJoeMoore:
The updateAccessToken method returns `true` when it means `false` and
vice versa. Fixes #5039 .

### PR checklist

### Description of the PR

Fixes an issue in the generated OAuth client for Java retrofit, retrofit2, and retrofit2-rx clients.  The `OAuth.updateAccessToken` method is returning true when it means false, and vice versa. This causes some authentication errors to be mistakenly forwarded to the client, even when an automatic retry is possible, and can also cause an infinite recursion loop when the client receives a 401 or 403 that is not related to the OAuth flow.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/retrofit/auth/OAuth.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/retrofit/auth/OAuth.mustache
@@ -132,7 +132,7 @@ public class OAuth implements Interceptor {
                     if (accessTokenListener != null) {
                         accessTokenListener.notify((BasicOAuthToken) accessTokenResponse.getOAuthToken());
                     }
-                    return getAccessToken().equals(requestAccessToken);
+                    return !getAccessToken().equals(requestAccessToken);
                 } else {
                     return false;
                 }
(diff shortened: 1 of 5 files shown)


wing328:
@ThatJoeMoore thanks for the PR. We'll review and let you know if we've any question.

cc @cbornet

cbornet:
LGTM

wing328:
@cbornet thanks for reviewing the change

@ThatJoeMoore thanks for your PR, which has been merged into master.

PR #5085, opened 2017-03-16
[JAVA][jaxrs-spec] Add support for enum as QueryParam argument type (#5075)

foxx1337:
[JAVA][jaxrs-spec] Add support for generating type-safe query parameters (#5075)

All the logic is already in place so I just make sure parameters in:query
also insert an enum in the generated API and that the type for the REST
operation parameter is that enum (and not just String as it was before).

The code is somewhat ugly and its functionality is still limited: for example
Container<Container<String>> param; (or member)
won't work. I think it's enough for now to use this for the regular cases
when only List<Integer> and List<String> are being passed around.

The state of the if-variable-then-variable if-notVariable-then-other
is quite ludicrous now in enumClass.mustache and queryParams.mustache.
It's probably a good idea to refactor that.
### PR checklist

### Description of the PR

I've just looked at the generated files under `samples/server/petstore/jaxrs-spec` and it looks fine to me.


foxx1337:
@jfiala - this is somehow out of the blue, but could you please have a look at what I did here and le me know your thoughts? I'm asking you because you're probably the best for it at this time :)

foxx1337:
Once this goes in and I get to it, I'll create another pull request to annotate the code thus generated with `@ApiParam(allowableValues="list,of,enum,constants"`.

jfiala:
I like the idea, but generally dislike inlined enums, please make sure to verify it with outer enums as they make much more sense (see OuterEnum in petstore-with-fake-endpoints-models-for-testing.json). Additionally, the chinese characters in the generated swagger.json file are broken (???).

foxx1337:
I sent a new commit. Basically I want this:

```yaml
paths:
  /health:
    put:
      operationId: putHealth
      tags:
        - Health
      produces:
        - text/plain
      parameters:
        - name: phoey
          in: query
          type: string
          default: foobar
          enum:
            - foobar
            - foobat
            - barbaz
          required: false
```

to turn into this

```java
    @PUT
    @Produces({ "text/plain" })
    @ApiOperation(value = "", notes = "", response = void.class, tags={ "Health" })
    @ApiResponses(value = { 
        @ApiResponse(code = 201, message = "Null response", response = void.class),
        @ApiResponse(code = 200, message = "unexpected error", response = void.class) })
    public Response putHealth(@ApiParam(value = "", allowableValues="foobar, foobat, barbaz", defaultValue="foobar") @QueryParam("phoey")  @DefaultValue("foobar") PhoeyEnum phoey) {
        return Response.ok().entity("magic!").build();
    }
```

No, regarding your comment with moving the enum outside of the the `*Api` class, I'm not sure it would work due to name clashes should different REST endpoints have identically named params. Even as it is right now, as internal class, there could be name clashes if different methods on the same endpoint (put, get, post, ...) end up defining params with the same names.

foxx1337:
How would you suggest, @jfiala, I improve this?

Please also see my above comment - I believe that it makes little sense in this context moving the enums outside of the generated *Api classes, but I do want to see this change hitting master.

iliassk:
Any news about this?

Issue #5089, opened 2017-03-16
Swagger org.springframework.core.io.Resource for 'file's 


LEgregius:
##### Description

The Spring framework support returning org.springframework.core.io.Resource interface implementations from rest api calls in order to process files in an abstract way, such as reading them from another url, database blob stream, etc, but the code gen only allows the type "file" to generate the class "java.io.File."   It would be nice to allow switch that to use the Resource interface.

##### Swagger-codegen version

Master

##### Swagger declaration file content or url

```yaml
/files/{fileId}:
    get:
      tags:
      description:
        Downloads an internal file
      parameters:
        - name: fileId
          in: path
          required: true
          type: string
      produces:
        - application/pdf
      responses:
        '200':
          description: The result
          schema:
            type: file
```

generates something like:

```java 
ResponseEntity<File> providerFilesFileIdGet(@ApiParam(value = "", required=true ) @PathVariable("fileId") String fileId);
```

But I need something like

```java
ResponseEntity<Resource> providerFilesFileIdGet(@ApiParam(value = "", required=true ) @PathVariable("fileId") String fileId);
```

Suggested Fix

I suggest adding an additional command line property to enable this feature.  I have already implemented one and will submit a PR for it.

ePaul:
I would dare to say that using java.io.File (which is an object representing a file name in the local file system) is completely wrong here ... was this ever implemented, and how?
Would this create a local file with the returned data and return its name, or for a parameter, expect the user to create this file and pass it, then have the library read from the file?

I would have expected an InputStream or something similar.
(I don't know about your Resource either.)

LEgregius:
I suppose if one were going to return a local file, as in return server file data, could make sense in some cases, but definitely not in any sort of large scale case.  The Resource is a SpringFramework thing.  It handles both opening the connection and returning the stream.

wing328:
Closed via #5181

PR #5090, opened 2017-03-16
Spring Framework resource file streams issue 5089

LEgregius:
I left a description of the issue here:  #5089

This adds a new additional command line property called `resourceStreams` that replaces `java.io.File` with `org.springframework.core.io.Resource` in Spring Framework when specifying a `file` type in the yaml.  I ran normal unit tests, and they passed, and what I added is simple, and should have no effect without specifying the command line option, but it allows me to return file data that isn't stored locally on the rest server serving the api, and it lets me stream it from the original place, through the server, and down to the client without having to store the whole file all at once.

The largest block of the code change is providing an example value that shows how to make it work exactly the same as it did by wrapping a `java.io.File` in a `FileSystemResource`, which is an implementation of the Resource interface.


wing328:

wing328:
cc @cbornet

LEgregius:
I tried the filter mechanism not realizing it would change all the commit ids.  I'll need to recreate this PR.

PR #5073, opened 2017-03-15
[Scala] update Travis CI to cover akka-scala petstore client

wing328:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [ ] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [x] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

Update pom.xml to cover akka-scala petstore client.

PR #5094, opened 2017-03-16
Fix sample script to set the correct artifact ID in the generated POM

ePaul:
### PR checklist

     → bin/akka-scala-petstore.sh (twice, once before and once after the change).

### Description of the PR

This was first just meant to update the Akka samples, there are no code changes here.
The first commit (updating samples from the script) undoes part of the change from #5073, which changed the name + artifact ID manually.
This shows that the default artifact ID is not what is used when running CI.

The second commit changes the sample update scripts to set the correct `--artifact-id` parameter.

The third commit updates the samples again, showing that everything is fine now. (There is a change in the generated reference.conf, not sure if it is relevant.)

### DIFF
--- a/bin/akka-scala-petstore.sh
+++ b/bin/akka-scala-petstore.sh
@@ -26,6 +26,6 @@ fi
 
 # if you've executed sbt assembly previously it will use that instead.
 export JAVA_OPTS="${JAVA_OPTS} -XX:MaxPermSize=256M -Xmx1024M -DloggerPath=conf/log4j.properties"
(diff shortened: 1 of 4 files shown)


wing328:
@ePaul thanks for the PR, or we can leverage .swagger-codegen-ignore to prevent the file being overwritten in the next run (other clients also use this approach)

wing328:
Actually your approach may work better as we need to update dependency in pom.xml to the latest version from time to time.

ePaul:
@wing328 I was not aware that .swagger-codegen-ignore is used for this purpose.
Though I like this way better, as it allows to actually test the generated pom.xml, not one generated ages ago.

Should I do the same change for other generators too? (Possibly in a separate PR.)

wing328:
> Should I do the same change for other generators too? (Possibly in a separate PR.)

Yup, if you've time, please do. Thanks.

(beware there might be other custom changes made to the auto-generated pom.xml)

Issue #5096, opened 2017-03-16
[Haskell] Generated Petstore sample is broken, doesn't compile


mandrean:
##### Description

The petstore sample client/server generated using `./bin/haskell-servant-petstore.sh` is broken and doesn't even compile.

It has numerous issues:

* `\` and `"` specialCharReplacements need to be escaped twice, once for Java and once for Haskell, otherwise the [specialchars list](https://github.com/swagger-api/swagger-codegen/blob/master/samples/server/petstore/haskell-servant/lib/SwaggerPetstore/Types.hs#L115) will not compile. This is a regression bug caused by the `specialCharReplacements` hashmap in `DefaultCodegen.java` being extended with more special chars in this commit: https://github.com/swagger-api/swagger-codegen/commit/7a245e31894c30a050e59cffefb97739005c0013
* The [lookupEither](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/haskell-servant/API.mustache#L58) return type (`Either Text b`) doesn't pattern-match with [fromFormUrlEncoded](https://hackage.haskell.org/package/servant-0.6/docs/Servant-API-ContentTypes.html#t:FromFormUrlEncoded) (`Either String b`)
* [Lots of dead legacy Petstore sample code left](https://github.com/swagger-api/swagger-codegen/tree/master/samples/server/petstore/haskell-servant) from the old solution before it was refactored in #2322

##### Swagger-codegen version

`v2.2.2` or latest `master`, doesn't matter.

##### Steps to reproduce

0) `git clone https://github.com/swagger-api/swagger-codegen && cd swagger-codegen`
1) `git checkout origin master && git pull origin master --rebase`
2) `mvn clean package`
3) `sh ./bin/haskell-servant-petstore.sh`
2) `cd samples/server/petstore/haskell-servant`
3) `stack install`

wing328:
Closed via #5097 

Thanks for the PR.

PR #5061, opened 2017-03-14
[Haskell] Bump dependencies

mandrean:
### PR checklist

- [X] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [X] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [X] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

Bumps the `servant` dependencies from `0.6` to `0.9.1.1` and haskell from `lts-5.11` to `lts-8.5`.

Fixes issue #4162.

PR #5097, opened 2017-03-16
[Haskell] Fix broken client/server compilation errors

mandrean:
### PR checklist

### Description of the PR

* Fixes the double Java/Haskell escapement bug mentioned in #5096.
* Removes a duplicate instance declaration that already exists in `Servant.API.Verbs`
* Bumps to latest possible dependencies without breaking anything (in contrast with #5061 which has breaking changes)
* Fixes failing pattern matching for `lookupEither`
* Cosmetic improvements consistent with Haskell best practices
* Cleans up old dead legacy Petstore sample code

Fixes all the issues in #5096.


wing328:
@mandrean thank you for the PR, which has been merged into master.

I've created https://github.com/swagger-api/swagger-codegen/issues/5100 about refactoring the Haskell generator. Please take a look to see if it makes senses to you.

(Disclaimer: no expert in Haskell)

cc @algas

mandrean:
@wing328 Thanks! Yes, I think it's a very good idea to split them up.

Also, it might be a good idea to evaluate other solutions for the client. I don't think `servant-client` is the best choice here.

I would go with [http-client](https://www.stackage.org/package/http-client), which is used by many other high-profile Haskell packages (as you can see in the `Used by 256 packages:` section on Stackage), including `PureScript`, `Yesod`, `stack` it self etc.

PR #5098, opened 2017-03-16
Ensure identifiers get sanitized & model names have the same formatting

log0ymxm:
…me formatting

### PR checklist

### Description of the PR

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AkkaScalaClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AkkaScalaClientCodegen.java
@@ -231,7 +231,7 @@ public class AkkaScalaClientCodegen extends AbstractScalaCodegen implements Code
     }
 
     private String formatIdentifier(String name, boolean capitalized) {
-        String identifier = camelize(name, true);
+        String identifier = camelize(sanitizeName(name), true);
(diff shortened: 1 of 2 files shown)


log0ymxm:
I'm not sure why the travis test fails, maybe some upstream issue.

One thing on this PR, I re-ran `bin/akka-scala-petstore.sh` and it reverted `pom.xml` changes found in https://github.com/swagger-api/swagger-codegen/commit/2da2a2631d86cf0555b9cc13c56f5d19b31ef903

@wing328 Can you comment on what that pom.xml change was, and if we need to update the shell script or something else for this.

log0ymxm:
Also there's no coverage for the specific case we saw that required the override for `toModelName`. We have a subclass typing patter where we end up with definitions of the form `Customer.Id`, when the period is in there it generated an invalid case class in scala code, so by running the model name through the format identifier method it cleans that period out.

ePaul:
See #5094 for the pom update issue. (This would also cause a problem in CI, though Travis breaks before that for an unrelated cause.)

wing328:
@log0ymxm thanks for the PR, which looks good to me.

Eventually, we might want to consolidate `scala`, `async-scala` and `akka-scala` into a single Scala API client generator. @clasnake has begun the effort with https://github.com/swagger-api/swagger-codegen/pull/5115. Please review the change when you've time.

Issue #5027, opened 2017-03-11
[General] refactor the type-mapping logic to clearly separate build-in types from API-defined types


ePaul:
Swagger Codegen has some logic which selects type names for generated models and for properties.
This handles both mapping to build-in types of the programming language (e.g. for Java int/double/String/BigDecimal/UUID/...) and name conversion (e.g. case change, escaping for reserved words, model name prefix/suffix) for types defined in the API definition file (mainly object and enum types).

Currently the code is a bit convoluted, which leads to bugs like #3678 (where the suffix/prefix is applied to a build-in type).

### Solution proposal

This logic needs to be refactored to separate both functions clearly. Any name modifications should only be applied to for API defined types, not to build-in types of the programming language.
(I guess handling of maps/lists would also be part of this.)

ePaul:
A proper fix for this would also make sure that in generated documentation (in markdown) there are links only to generated classes, not to ones included in the environment (and not documented).

Noted in https://github.com/swagger-api/swagger-codegen/pull/5118#discussion_r106815687.

PR #4739, opened 2017-02-07
[Java][Issue #1806] generate using java.util.UUID for UUIDs

MatanRubin:
Before this change if a model had a property of type=string and
format=uuid, it would be generated as a Java String.
Now, it generates the property as java.util.UUID.

PR #5118, opened 2017-03-19
Update samples for java/jersey2-java6 (and fix artifact ID)

ePaul:
### PR checklist

   → bin/java-petstore-jersey2-java6.sh

### Description of the PR

This has no generator code changes, just updates some samples so later code changes see the actual effect on the samples. I also updated the sample generation script to use `-artifact-id swagger-petstore-jersey2-java6` so the generation doesn't destroy the changed artifact name in the generated pom.xml. (There doesn't seem to be a windows script for this.)


samples/client/petstore/java/jersey2-java6/src/main/java/io/swagger/client/model/Animal.java
```diff
@@ -28,12 +16,19 @@
...
 import io.swagger.annotations.ApiModelProperty;
 
 /**
  * Animal
  */
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "className", visible = true )
+@JsonSubTypes({
+  @JsonSubTypes.Type(value = Dog.class, name = "Dog"),
+  @JsonSubTypes.Type(value = Cat.class, name = "Cat"),
+})
```
ePaul: Seems to come from #4085.

samples/client/petstore/java/jersey2-java6/src/main/java/io/swagger/client/model/AdditionalPropertiesClass.java
```diff
@@ -59,7 +47,7 @@ public AdditionalPropertiesClass putMapPropertyItem(String key, String mapProper
    * Get mapProperty
    * @return mapProperty
   **/
-  @ApiModelProperty(example = "null", value = "")
+  @ApiModelProperty(value = "")
```
ePaul: It looks like previously we got a string `"null"` instead of a real `null` for the example, which confused the null-check.
I didn't figure out which commit/PR fixed this.

samples/client/petstore/java/jersey2-java6/docs/MixedPropertiesAndAdditionalPropertiesClass.md
```diff
@@ -4,7 +4,7 @@
 ## Properties
 Name | Type | Description | Notes
 ------------ | ------------- | ------------- | -------------
-**uuid** | **String** |  |  [optional]
+**uuid** | [**UUID**](UUID.md) |  |  [optional]
```
ePaul: That looks wrong. (The `format: uuid` was in the YAML from the beginning. Possibly #4739 changing to use java.util.UUID instead of String added the link here?

Related: #5026. #5027 might possibly fix this too.

samples/client/petstore/java/jersey2-java6/src/main/java/io/swagger/client/model/AdditionalPropertiesClass.java
wing328: @ePaul I remember Tony Tam filed a PR to fix it. If you need the number, I can dig it out for you.

samples/client/petstore/java/jersey2-java6/docs/MixedPropertiesAndAdditionalPropertiesClass.md
wing328: @ePaul there's a discussion about updating `isPrimitiveType` to include UUID as well. IMO, `isPrimitiveType` should include UUID.

(the issue was caused by the change that maps `uuid` as UUID instead of String in Java)

wing328:
@ePaul thanks for updating the sample. I'll update `./bin/java-petstore-all.sh` to include `bin/java-petstore-jersey2-java6.sh` as well so that the Petstore client (jersey2, java6) won't be outdated.

PR #5128, opened 2017-03-20
[C#][aspnetcore] Fix broken Docker build option & documentation

### LINKED ISSUE #5126: [C#][aspnetcore] Docker build option broken
##### Description

When creating an ASP.NET Core server stub, the Docker build option from `README.md` doesn't work.

##### Swagger-codegen version

`v2.2.2` or latest `master`, doesn't matter.

`sh ./bin/aspnetcore-petstore-server.sh` or your own `Swagger.json`, doesn't matter.

##### Steps to reproduce

0. `git clone https://github.com/swagger-api/swagger-codegen && cd swagger-codegen`
1. `git checkout origin master && git pull origin master --rebase`
2. `mvn clean package`
3. `sh ./bin/aspnetcore-petstore-server.sh`
4. `cd samples/server/petstore/aspnetcore/src/IO.Swagger`
5. 
    a. `docker build -t IO.Swagger .`
    b. `docker build -t swagger-petstore .`
6. `docker run -p 5000:5000 swagger-petstore`

##### Issues to fix

There are two issues here:

1. The [README.md](https://github.com/swagger-api/swagger-codegen/blob/c70a2a7481b34362657546e6ca0f4a41ff1c44b3/modules/swagger-codegen/src/main/resources/aspnetcore/README.mustache) template just pulls the [packageName](https://github.com/swagger-api/swagger-codegen/blob/c70a2a7481b34362657546e6ca0f4a41ff1c44b3/modules/swagger-codegen/src/main/resources/aspnetcore/README.mustache#L25) from [AbstractCSharpCodegen](https://github.com/swagger-api/swagger-codegen/blob/c70a2a7481b34362657546e6ca0f4a41ff1c44b3/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java#L23) (or [whatever you've set in your config](https://github.com/swagger-api/swagger-codegen/blob/c70a2a7481b34362657546e6ca0f4a41ff1c44b3/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AspNetCoreServerCodegen.java#L42)) for the `docker build` example, even though docker only supports lowercase letters for `repo/tag`. Hence, **5a)** will fail on the `Petstore` (`packageName: IO.Swagger`) example and real-world .NET applications that use uppercase letters for the package name. For reference, here's the error you get when trying to run `docker build -t IO.Swagger .` (or any other `packageName` with upper-case letters):

```
invalid argument "IO.Swagger" for t: Error parsing reference: "IO.Swagger" is not a valid repository/tag: repository name must be lowercase
See 'docker build --help'.
```

2. But even if you manually try to build the project with a correct Docker build tag (**5b)**), it will still fail because the [generated Dockerfile](https://github.com/swagger-api/swagger-codegen/blob/c70a2a7481b34362657546e6ca0f4a41ff1c44b3/modules/swagger-codegen/src/main/resources/aspnetcore/Dockerfile.mustache#L1) uses `FROM microsoft/dotnet:latest` as its base image and there were some breaking changes between `dotnet:1.0.3-sdk` and `dotnet:1.0.4-sdk` (Semver? 🙄 ). [Using `:latest` is very bad practice in Docker-land](https://developers.redhat.com/blog/2016/02/24/10-things-to-avoid-in-docker-containers/); the image associated with it may unexpectedly change in the future (when you've forgotten about it) and then it breaks stuff downstream that depends on an older `:latest`. Use specific version tags instead. For reference, here's the error you get when using `dotnet >= 1.0.4-sdk` or `dotnet == latest`:

```
 ---> Running in f0bfac9cc465
/app/IO.Swagger/IO.Swagger.xproj(7,5): error MSB4019: The imported project "/usr/share/dotnet/sdk/1.0.1/Microsoft/VisualStudio/v14.0/DotNet/Microsoft.DotNet.Props" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.
The command 'dotnet restore' returned a non-zero code: 1
```

##### Suggest a Fix

Fixes:

1. Do some kind of to-lower-[kebab-case](http://stackoverflow.com/a/12273101) on the `packageName` before sending it to the template. Also, for the Petstore sample, `docker build -t petstore-sample .` is probably a better example, so the Petstore build should probably explicitly set `packageName` to `Petstore.Sample` in a `config.json` instead of relying on `AbstractCSharpCodegen.java`'s `IO.Swagger`.

2.
    a. Fix codegen so that it works with `dotnet >= 1.0.4-sdk`
    b. Or use the latest .NET Core `1.0.x` Docker image which is still compatible with the current setup, and that is `microsoft/dotnet:1.0.3-sdk-projectjson`.

mandrean:
### PR checklist

### Description of the PR

* Use lower-case packageName for README.md docker build example
* Use dotnet:1.0.3-sdk instead of just "latest"
* Regenerate Petstore server sample

Fixes #5126.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AspNetCoreServerCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AspNetCoreServerCodegen.java
@@ -86,6 +86,7 @@ public class AspNetCoreServerCodegen extends AbstractCSharpCodegen {
         super.processOpts();
 
         additionalProperties.put("packageGuid", packageGuid);
+        additionalProperties.put("dockerTag", this.packageName.toLowerCase());
 
         apiPackage = packageName + ".Controllers";
         modelPackage = packageName + ".Models";
--- a/modules/swagger-codegen/src/main/resources/aspnetcore/Dockerfile.mustache
+++ b/modules/swagger-codegen/src/main/resources/aspnetcore/Dockerfile.mustache
@@ -1,4 +1,4 @@
-FROM microsoft/dotnet:latest
+FROM microsoft/dotnet:1.0.3-sdk-projectjson
 
 ENV DOTNET_CLI_TELEMETRY_OPTOUT 1
 
--- a/modules/swagger-codegen/src/main/resources/aspnetcore/README.mustache
+++ b/modules/swagger-codegen/src/main/resources/aspnetcore/README.mustache
@@ -22,6 +22,6 @@ build.bat
 
 ```
 cd src/{{packageName}}
-docker build -t {{packageName}} .
-docker run -p 5000:5000 {{packageName}}
+docker build -t {{dockerTag}} .
+docker run -p 5000:5000 {{dockerTag}}
 ```
--- a/samples/server/petstore/aspnetcore/IO.Swagger.sln
+++ b/samples/server/petstore/aspnetcore/IO.Swagger.sln
@@ -7,7 +7,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
                 global.json = global.json
         EndProjectSection
 EndProject
-Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "IO.Swagger", "src\IO.Swagger\IO.Swagger.xproj", "{795C6B14-1C3E-45F5-AF6F-EE47B197FF1E}"
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "IO.Swagger", "src\IO.Swagger\IO.Swagger.xproj", "{3C799344-F285-4669-8FD5-7ED9B795D5C5}"
 EndProject
 Global
         GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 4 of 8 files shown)


wing328:
@mandrean thanks for the PR. We'll review and let you know if we've any question.

cc @jimschubert

jimschubert:
Looks good. I just had a question about the version change in the sample.

wing328:

I believe you're referring to the following change:
```
-FROM microsoft/dotnet:latest
 +FROM microsoft/dotnet:1.0.3-sdk-projectjson
```

Personally I prefer a hard-coded (fixed) version and the users can always pass customized templates via the `-t` option if they want to use other version or the `latest` version.

wing328:
@mandrean thanks for the PR, which has been merged into master.

mandrean:
@jimschubert You mean this?

from `samples/server/petstore/aspnetcore/global.json`:
```
-    "version": "1.0.0-preview3-003171",
+    "version": "1.0.0-preview2-003121",
```

That has to do with [this](https://github.com/dotnet/core/blob/402c83b/release-notes/download-archive.md#lts-releases):

image

jimschubert:
@mandrean Thanks for clarifying that. I was commenting on mobile while my baby was awake in the middle of the night. I just wanted to verify the explicit version (kind of a "bump") was compatible with what looked like a lower version number in global.json.

@wing328 I agree with the explicit tags in Dockerfile, but I think only `latest` was available when I wrote the initial generator.

mandrean:
@jimschubert Agree, it was confusing for me as well first.

In fact, .NET Core and .NET Core SDK versioning in general is very confusing, but that's a separate discussion.

jimschubert:
@mandrean any idea why the preview2 is revision `3121` (bottom of the chart you posted) rather than `3156` (highlighted line)?

I'm asking these questions because I think this is the version originally defined in the generator, but I seem to recall some interface changes in the version `1.0.0-preview3-003171` that were required for the DI Framework to function correctly.

PR #5144, opened 2017-03-21
Add support for Markdown in -l html

DavidBiesack:
### Description of the PR

Addresses #3436 

Support Markdown in -l html
Add https://github.com/atlassian/commonmark-java to modules/swagger-codegen to convert Markdown to HTML,
update StaticHtmlGenerator to use this (see the toHeml() method and its uses)

Add a new test case bin/html-markdown.sh and
modules/swagger-codegen/src/test/resources/2_0/markdown.yaml


wing328:
@DavidBiesack thanks for the enhancement. The change looks good to me and my test looks good as well.

DavidBiesack:
Thanks, @wing328 !
If I get the chance, I'll try to do the same for `generate -l html2`.

PR #5181, opened 2017-03-23
[JAVA-SpringFramework][5089] Added a new additional option that allows one to replace the the 'fil…

LEgregius:
**I reworked this PR from an earlier one.  All it really does is fix the commit email.**

I left a description of the issue here:  #5089

This adds a new additional command line property called `resourceStreams` that replaces `java.io.File` with `org.springframework.core.io.Resource` in Spring Framework when specifying a `file` type in the yaml.  I ran normal unit tests, and they passed, and what I added is simple, and should have no effect without specifying the command line option, but it allows me to return file data that isn't stored locally on the rest server serving the api, and it lets me stream it from the original place, through the server, and down to the client without having to store the whole file all at once.

The largest block of the code change is providing an example value that shows how to make it work exactly the same as it did by wrapping a `java.io.File` in a `FileSystemResource`, which is an implementation of the Resource interface.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SpringCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SpringCodegen.java
@@ -5,7 +5,6 @@ import io.swagger.codegen.languages.features.BeanValidationFeatures;
 import io.swagger.models.Operation;
 import io.swagger.models.Path;
 import io.swagger.models.Swagger;
-import io.swagger.models.properties.Property;
 
 import java.io.File;
 import java.util.*;
@@ -25,7 +24,6 @@ public class SpringCodegen extends AbstractJavaCodegen implements BeanValidation
     public static final String SPRING_MVC_LIBRARY = "spring-mvc";
     public static final String SPRING_CLOUD_LIBRARY = "spring-cloud";
     public static final String IMPLICIT_HEADERS = "implicitHeaders";
-    public static final String RESOURCE_FILE_STREAMS = "resourceStreams";
 
     protected String title = "swagger-petstore";
     protected String configPackage = "io.swagger.configuration";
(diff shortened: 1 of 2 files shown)


wing328:
@LEgregius thanks for the PR

cc @cbornet

cbornet:
Instead of creating yet another option, I would prefer to generate `Resource` by default as it's better than File.
@LEgregius did you test if it works fine with the spring-cloud client impl ?

LEgregius:
The petstore examples build, to say nothing of running them.  If you want to make it the default, that makes sense.  I don't know if that will break anyone.  If someone plans on uploading a file via an api, then it needs to use a org.springframework.web.multipart.MultipartFile, so a Resource isn't really the right solution there, but it doesn't seem like the generator is smart enough right now to make the distinction between incoming and outgoing types.

cbornet:
I think MultipartFile is already used for file upload (see https://github.com/swagger-api/swagger-codegen/blob/fb7599652794f7127b4f5dfaba4439d1933d3c7f/modules/swagger-codegen/src/main/resources/JavaSpring/apiController.mustache#L19) . So it should be fine.

cbornet:
The MultipartFile type is hard coded in the templates

LEgregius:
Okay, then.  What should I do?  Remove the option?  Without the option on the type, the code change is literally just 
```
    typeMapping.put("file", "Resource");
    importMapping.put("Resource", "org.springframework.core.io.Resource");```

Plus, the example value, if that's even necessary.

ePaul:
As I commented on the issue, I think using `java.util.File` never made sense here, so I'm for using Resource instead when we are in a Spring context anyways.

cbornet:
@LEgregius @ePaul yes remove the option and just change the typeMapping 👍

LEgregius:
Is there anything else I need to do for this PR before it gets merged?

wing328:
@LEgregius taking another look now...

wing328:
@LEgregius I've updated the Petstore samples for spring-related generators and the result looks good.

Thanks for the contribution 👍

Issue #5142, opened 2017-03-21
[JAVA] Add @QueryMap support for Feign API


bbdouglas:
##### Description

Feign allows gathering up of all query parameters into a single Map parameter with the [@QueryMap annotation](https://github.com/OpenFeign/feign#dynamic-query-parameters). This simplifies the API when there are a large number of parameters, especially optional ones. In cases where an API has query parameters, I suggest generating a second variant of the interface that uses the single Map parameter instead of listing out all of the parameters in the function signature. Additionally, a simple builder class could be generated that simplifies creation of that map using a fluent interface similar to how Swagger generates POST forms parameters.

Since this adds a new API method and leaves the old API methods alone, this should be backwards compatible.
 
##### Swagger-codegen version

HEAD (2.2.3-SNAPSHOT)

##### Suggest a Fix

I have a pull request with the suggested feature addition.

bbdouglas:
#5185

PR #5185, opened 2017-03-23
Fixes a bug in #5142 for multi-word parameters

bbdouglas:
In the convenience class defined for generating a Map of query parameters, the
parameter name was mistakenly being set to the Java variable name, not the
actual parameter name. These are often the same, but can be different in the
case of multi-word parameters, such as in the sample API's `enum_query_string`
vs `enumQueryString`.

### PR checklist

### Description of the PR

In the convenience class for building up a query parameter map in the Feign API, changed the value of the URL parameter name from the Java variable name to the proper parameter name. You can see how this affects the code in the pet store sample, which has a few cases of multi-word parameters where the parameter name is different from the Java variable name.


ePaul:
Related: #4898. Looks like I overlooked this when doing my analysis for `{{paramName}}`.

wing328:
@ePaul I think another way to do it is to grep the API mustache templates to ensure there's `{{baseName}}` as an API mustache template not having `{{baseName}}`is more likely to have the issue.

Issue #1482, opened 2015-10-31
[PHP] update HTTP library to use Guzzle


wing328:
We want to upgrade the HTTP library to use Guzzle, which is a very mature HTTP library with lots of features: https://github.com/guzzle/guzzle#guzzle-php-http-client

If you want to work on this and need help on where to start, please ping @wing328. Thanks!

jimmypuckett:
@wing328, I think that this is a great idea, but I would think that the effort should wait on guzzle making guzzle-service working with guzzle 6. At that point, it would just be a matter of converting the swagger specs to guzzle services.

wing328:
@jimmypuckett thanks for the tips. I'd a look at guzzle-service but couldn't find the spec it uses. I'll check with the owner to find out more about the service and its specification when the project becomes more mature.

jimmypuckett:
@wing328, ya it is not well documented.  It was very nice back in guzzle 3 days.  I am looking forward in them releasing the update for version 6.

kasperg:
How about using [HTTPlug](https://github.com/php-http/httplug) instead? This a client abstraction and thus we avoid binding the implementation to a single library.

[HTTPlug already supports a range of adapters](http://docs.php-http.org/en/latest/clients.html) including Guzzle 5 and 6 as well a simple .

@wing328: If you agree with that direction I would be interested in taking a stab at this.

I think there is some value in the fact that the current `ApiClient` does not introduce further dependencies. One way to retain this value could also be to refactor the current `ApiClient` into an abstract `ApiClient` class and a `CurlApiClient` implementation . Then third party packages could provide `HTTPlugApiClient`, `FrameworkXApiClient` implementations etc.

wing328:
@kasperg thanks for the suggestion to use HTTPlug.

As you suggested, creating an abstract class is one way to do it (which is perfectly fine with me but some developers may not want the dependency on HTTPlug)

Another way is to following Java's current approach to generate `ApiClient` file based on library specified. To apply that to PHP, you will need to create another file ApiClient.HTTPlug.mustache and use it (instead of the `ApiClient.mustache` in PHP) when users specify the `library=HTTPlug` option.

Which approach do you feel more comfortable implementing?

kasperg:
I did not know about the `library` option but I think that is the way to go.

Having thought about third party packages with implementations will not work. The `invokerPackage` configuration option means that we cannot rely on a stable class name to extend.

Babacooll:
[...]

PR #5190, opened 2017-03-24
Php client using guzzle6 & psr7 instead of curl

baartosz:
### PR checklist

### Description of the PR

Finally finished https://github.com/swagger-api/swagger-codegen/issues/1482. Looks massive but without generated samples its roughly 10 files changed plus few new test classes. Large feature anyway.
Important points:
 - decided not to follow suggestion and to remove ApiClient altogether and make generated API classes call Httplug's Client interface directly. Otherwise It would be messy abstraction (PetApi -> ApiClient -> HttpClient -> Guzzle/Curl/whatever) plus impossible to leverage fancy new features curl was missing (like streams in responses)
 - deleted generated autoload.php as it make no sense to use it when its necessary to install httplug and guzzle with composer and use vendor/autoload.php
 - tested with guzzle client (`php-http/guzzle6-adapter` + `guzzlehttp/guzzle`) and curl client (`php-http/curl-client`). Works nice with both of them. Only thing not working currently with curl client was file upload. I left it for another iteration/separate ticket. Tests are currently ran only with guzzle.
- ObjectSerializer had some methods static some not. I tried to make all methods not static and inject it to Api object but ran into issue here https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/php/SwaggerClient-php/lib/Model/Pet.php#L390. Finally decided to make it all static for consistency and rework it later
- BC breaks: 
     - new dependencies (obviously)
     - initialization of the API class is slightly changed (client adapter has to be passed as argument) (no way around that)      
     - removed options from Configuration class (client-specific options go to client now)
     - headers returned as array from `$response->getHeader($name);` (its how psr7 expects it to work)

Please let me know if its better to make it optional like `--library=httplug` or something.


wing328:
@baartosz thanks for the contribution. We'll review and let you know if we've any question.

cc @arnested @abcsum

wing328:
cc @dkarlovi , @scop as well since they also filed several PRs before to improve the PHP API client. 

```
git checkout -b baartosz-php_httplug2 2.3.0
git pull https://github.com/baartosz/swagger-codegen.git php_httplug2
```

dkarlovi:
I'm all for dropping the shipped client as it will make the generated code much more simple, focus of this projects is not to generate HTTP clients.

The problem I see is relying directly on Guzzle interface directly, the whole point of PSR-7 is NOT to do that. Generated code should rely on PSR-7 exclusively and enable me to use whatever compatible implementation. This would make integrating the SDK much harder than it should be.

For example, you might already have a PSR-7 compatible client in your app pre-configured, you should be able to just inject it and use it. This means generated code should rely on functionality provided by the interface, for example [Streams](http://www.php-fig.org/psr/psr-7/#streams).

dkarlovi:
Also: having this behind an option seems reasonable, but I think it would make supporting other libraries messy. What I feel would be a much cleaner approach is to do a "base" PSR-7 only implementation which wouldn't be able to work without injecting the client at all.

Then, `--library=httplug` would mean we generate an additional class which modifies composer.json and preconfigures everything to be ready to use (so, libraries only add the implementation to use, not modify anything).

baartosz:
@dkarlovi i see your point. So basically instead of httplug provide own interface for the client and optional --library to have one already there?

guzzle/psr7 now is used to provide helpers e.g. To build multipart requests. Are you suggesting writing it from scratch instead?

dkarlovi:
@baartosz pretty much, yes. If you have a functional PSR-7 client enabled, you can just inject it and have it work with the generated SDK. For example, your HTTP client might have a logger set up which stores all requests to Elastic search, you don't want to have another client which does the same thing, but lacks integration to your environment.

I don't suggest writing anything from scratch, no, the whole point of removing the HTTP client is avoid doing that. :) Can you point me to where exactly you're constructing multi-part messages? It might be an [XY problem](https://meta.stackexchange.com/a/66378/248583).

dkarlovi:
@baartosz BTW thank you very much for your contribution, it's a great effort on your part.

baartosz:
@dkarlovi makes sense. At this point generated client is using psr7 compatible request and response, so client itself is easy to replace with whatever is decided. 

Multipart stream from `guzzle/psr7` is used here: https://github.com/swagger-api/swagger-codegen/pull/5190/files#diff-ae9e1fe068c165f6731482ae703b427dR132. Same case with `GuzzleHttp\Psr7\Request` and `GuzzleHttp\Psr7\Response` - it's just implementation of psr interface so I don't see any problem using it if httplug will be removed.

Thanks for feedback.

modules/swagger-codegen/src/main/resources/php/Configuration.mustache
```diff
@@ -78,21 +78,7 @@ class Configuration
...
-     */
-    protected $curlTimeout = 0;
-
-    /**
-     * Timeout (second) of the HTTP connection, by default set to 0, no timeout
-     *
-     * @var string
-     */
-    protected $curlConnectTimeout = 0;
+    protected $host = '{{basePath}}';
```
wing328: Please use unescape basePath instead: {{{basePath}}}

modules/swagger-codegen/src/main/resources/php/Configuration.mustache
```diff
@@ -123,51 +109,6 @@ class Configuration
     protected $tempFolderPath;
 
     /**
-     * Indicates if SSL verification should be enabled or disabled.
```
wing328: I notice that the SSL, proxy and timeout methods have been removed. Can you share more with me how a developer will set SSL, proxy, timeout in the new client?

dkarlovi: @wing328 idea is to remove the HTTP client implementation completely and inject a [PSR-7](http://www.php-fig.org/psr/psr-7/) capable client such as [Guzzle](https://github.com/guzzle/psr7).

This means setting up these kinds of configuration would be outside of scope of generated code, it would assume the injected client is capable of being configured properly in its own right. This code would just use it as-is.

baartosz: Generated Api classes are now only creating psr compatible requests and handle responses. How this stuff is transferred is problem of the client we are trying to separate from. So ssl or proxy has to be supported and configured in provided client. For example:

guzzle client:
```
$client = new \Http\Adapter\Guzzle6\Client(['verify_ssl' => false]); //in guzzle its on by default
$apiInstance = new Swagger\Client\Api\PetApi($client)); //its enabled by default in guzzle afaik
```
curl client:
```
$client = new Http\Client\Curl\Client(null, null, ['CURLOPT_SSL_VERIFYHOST' => true]);
$apiInstance = new Swagger\Client\Api\PetApi($client)); //its enabled by default in guzzle afaik
```

wing328: @dkarlovi thanks for explaining more. To ease the migration, I think we will need to provide auto-generated documentation on how to set these values using Guzzle (the default PSR-7 http client)

baartosz: @wing328 I am not sure about priorities - is it more important to have working client out of the box, or having more flexible one? 
Should we add client or httplug+client dependencies by default?
Is enough to provide generated apis classes calling `send($request)` on `ApiClientInterface`, and expect users to choose client library and select/create adapter?
Or leave no-client version as default and add --library=httplug?

dkarlovi: Here's a suggested flow which makes sense to me:
1) generate no implementation dependency by default, this will generate a "clean" version, add additional info / sample code on how to inject own HTTP client
2) generate an additional class if `--library=httplug` is provided, this version is basically 1) with an additional item in composer.json and a HTTP client pre-injected for you (so, you can still inject another client yourself if you don't want to use the additional class, it just doesn't install a depenency you might not want by default)

We might make 2) the default to keep BC though.

wing328: > We might make 2) the default to keep BC though.

That's what I prefer - a reasonable default so that new users can simply generate a functional PHP client with the default setting (and they can later customize the output based on their requirements)

dkarlovi: @wing328 sounds good to me, documentation should note how to create a "bare" version, for example `--library=psr7`.

baartosz: Settled then. I'll update this PR soon with default version (with guzzle as its easier to use than httplug) and add `--library=psr7` in near future.

dkarlovi: @baartosz Guzzle is a better choice as you don't need an abstraction since PSR7 is your preferred abstraction, no need for an abstraction over an abstraction.

baartosz:
@wing328 

Currently readme.md says that client works with php 5.4 which is not supported anymore for 1,5 years
http://php.net/supported-versions.php

Should default version work with
guzzle 5 which is not psr7-compatible but is working with php>=5.4
or guzzle 6 psr compatible but requiring php>=5.5?

arnested:
I would say go for as new a version as needed. For this new way of doing it I don't see a need to forcefully stay behind on old PHP versions.

wing328:
Agreed with @arnested. @baartosz please feel free to drop 5.4 support.

If someone needs the PHP API client to work with PHP 5.4, they can always use Swagger Codegen v2.2.2 to generate the client.

modules/swagger-codegen/src/main/resources/php/Configuration.mustache
richhl: @baartosz want to test&adopt your httplug client + guzzle. Good time to make it a try? would I wait a little for your " I'll update this PR soon with default version.."

thx a lot for your work

baartosz: guzzle6-only version is almost ready, it will be better to wait for it.

baartosz:
Updated, rebased, tested with php 7 and 5.5, ready to review. 
Httplug removed, guzzle6 added, required php increased to 5.5.

Unfortunately python test suite is breaking build for some reason.
```
[INFO] Python Swagger Petstore Client .................... FAILURE [20.635s]
```

wing328:
@baartosz don't worry about the Python test failure. I've fixed it in the master.

richhl:
do you want us to test @baartosz ? any guidance to test it?

Thx in advance.

___________________________

Richard C. Hidalgo Lorite
Director de Sistemas
ISI Argonauta, S.L.
C/Alcalá 196, apto. 13, 28028 Madrid
Telf: +34 691 502 235
http://www.argonauta.org

2017-03-30 19:17 GMT+02:00 baartosz <notifications@github.com>:

wing328:
@richhl here are the steps to test it
```
git checkout -b baartosz-php_httplug2 2.3.0
git pull https://github.com/baartosz/swagger-codegen.git php_httplug2
mvn clean package
```
Then you can use the JAR to generate PHP API client by providing a spec.

wing328:
The Travis CI tests passed: https://travis-ci.org/swagger-api/swagger-codegen/builds/217127129

wing328:
I used [PHP CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) to inspect the code and found some enhancements (code format) we can make to the PHP API client (this PR and master). I'll create a separate task for tracking.

baartosz:
just a rebase to solve conflicts

wing328:
@baartosz PR merged into 2.3.0. Thanks for the huge enhancement!

I'm the community appreciates your effort.

dkarlovi:
@baartosz awesome work!

baartosz:
@dkarlovi @wing328 that was quick. Did you even review the code? ;)

Thanks for your time and input!

richhl:
don't worry @baartosz! hundred of eyes and hands are ready now to test and contribute changes if any.

we expect to test this week and next. will get in touch. thx 4 your work

wing328:
@baartosz I did review the code and play around with it over the weekend.

Like any other generators, I would expect it to improve over time thanks to the vibrant community.

wing328:
FYI. Updated README to include @baartosz as the template owner of PHP Guzzle

Robert430404:
Hello,

Not sure if this is the right place, but I'm wanting to use the template specified in this PR and I can't find any documentation on how I go about using it. If you guys could point me in the right direction, I would really appreciate it!

Thanks in advance,
Robert

wing328:
@Robert430404 Please open a new issue with the question instead.

Issue #5147, opened 2017-03-21
[CLI] Make -D option's syntax more similar to Java's system properties


ePaul:
### Current situation

There are currently two ways of passing system properties (which are used for language specific parameters, as well as for some options to the DefaultGenerator class) to Swagger Codegen, when using the CLI:

* Use Java's standard way, `-D` options in the command line before the class name (or `-jar`).
* Use Swagger-Codegen-CLI's `-D` option in the command line after the `generate` command.

Unfortunately those have different syntaxes, thereby confusing users and making certain uses more complicated than needed.

* Java's option can be given multiple times, each time with a pair of name and value (separated by `=`), or just a name (which then gets the empty string as a value).
* Swagger-Codegen-CLI's option can only given once (if given several times, just the last one wins), and takes a comma-separated list of name-value pairs. If you give just a name (without the `=`), it is just ignored.

#### Swagger-codegen version

2.2.2, as well as current master.

#### Command line examples

These examples are [from the Readme](https://github.com/swagger-api/swagger-codegen/blob/master/README.md#selective-generation). I omitted the values for -jar (the swagger-codegen-cli.jar) and -i (the API definition YAML or JSON) for clarity.

Generate only models and supporting files, no API:
```shell
# works already:
java -Dmodels -DsupportingFiles -jar ... generate -l java -i ...
# doesn't work yet (completely ignores both options)
java -jar ... generate -l java -i ... -Dmodels -DsupportingFiles
# doesn't work yet either (completely ignores both options)
java -jar ... generate -l java -i ... -Dmodels,supportingFiles
# needs to be written this way:
java -jar ... generate -l java -i ... -Dmodels=,supportingFiles=
```
Generate only one model and one supporting file, nothing else:
```shell
# works already
java -Dmodels=User -DsupportingFiles=StringUtil.java -jar ... generate -l java -i ...
# doesn't work (models=User is ignored)
java -jar ... generate -l java -i ... -Dmodels=User -DsupportingFiles=StringUtil.java
# needs to be written this way:
java -jar ... generate -l java -i ... -Dmodels=User,supportingFiles=StringUtil.java
```

### Proposal

Extend swagger-codegen's syntax to ...

* allow multiple `-D` options (merging their key-value pairs – in case of conflicts the last ones would win)
* allow keys without `=` and values (giving an empty string as value).

For backwords-compatibility, I would still parse commas as separators in `-D`'s values.

This would make all the CLI examples given above work.

[...]

PR #5191, opened 2017-03-25
[CLI] #5147: make -D work like system properties

ePaul:
### PR checklist

    - bin/run-all-petstore – no changes after latest rebase (2017-04-04, based on 22c80588f63629a17ae0113da11e2d9b437bd4cf).
    - bin/security/run-all-petstore-security-test → same changes as in #5325 + #5324. Not committed here, I will rebase and redo it when those are merged.
- [ ] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes. → **Not sure about this, see below** 

### Description of the PR

As detailed in #5147, this implements two changes (in separate commits, so it should be easy to separate those if needed).

* When parsing comma-separated key-value pairs we know accept empty values (and keys without value are also handled as empty values). This applies to the `-D` option as well as to `--instantiation-type`, `--type-mappings`, `--additional-properties`, `--import-mappings`.

   (Previously those keys were simply ignored, which made it impossible to pass just `-D models` (or `-D models=`) as an argument after the class/jar name (it needed to be a Java system property before the class/jar name).

* It is now possible to give several `-D` options, and have all of them apply – the key-value-lists are merged, in case of conflicting keys the last one wins.

  Previously the last `-D` option was the only one which had any effect, simply overwriting any earlier ones.

I modified some tests (and added some new ones) to reflect the new behavior.

#### About (non)breaking changes

This changes the behavior of all the key-value options to allow keys with empty values, or without values. If anyone had a script which would use e.g. `-D foo=bar,foo`, this previously would have passed `foo` with the value `bar`, now it is passing the empty string to the `bar` property. I think nobody should have done that, but who knows. **Opinions on that?**

#### Allowing several options for other plural options

It would be easy to allow the same behavior as for `-D` (allow multiple options, merge values) for some other plural options too. This applies to the key-value-list ones `--instantiation-type`, `--type-mappings`, `--additional-properties`, `--import-mappings`, but also to `--language-specific-primitives` (which is just a list of strings). **Opinions on that?**

### DIFF
--- a/bin/flaskConnexion-python2.sh
+++ b/bin/flaskConnexion-python2.sh
@@ -27,7 +27,7 @@ fi
 # if you've executed sbt assembly previously it will use that instead.
 export JAVA_OPTS="${JAVA_OPTS} -XX:MaxPermSize=256M -Xmx1024M -DloggerPath=conf/log4j.properties"
 #ags="$@ generate -i modules/swagger-codegen/src/test/resources/2_0/petstore.yaml -l python-flask -o samples/server/petstore/flaskConnexion-python2 -DsupportPython2=true"
-ags="$@ generate -t modules/swagger-codegen/src/main/resources/flaskConnexion -i modules/swagger-codegen/src/test/resources/2_0/petstore.yaml -l python-flask -o samples/server/petstore/flaskConnexion-python2 -c bin/supportPython2.json"
+ags="$@ generate -t modules/swagger-codegen/src/main/resources/flaskConnexion -i modules/swagger-codegen/src/test/resources/2_0/petstore.yaml -l python-flask -o samples/server/petstore/flaskConnexion-python2 -c bin/supportPython2.json -D service"
 
 rm -rf samples/server/petstore/flaskConnexion-python2/*
-java $JAVA_OPTS -Dservice -jar $executable $ags
+java $JAVA_OPTS -jar $executable $ags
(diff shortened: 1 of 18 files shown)


ePaul:
I also changes some of the scripts in bin/ to use the new -D arguments instead of the old ones.

ePaul:
The changed sample generation scripts show no change in behavior.

ePaul:
I rebased this onto current master. Trying to rerun the sample-generators again gives many new changes (which likely also happen when I do the same from master) ... should I do "update samples" PRs separately again?
(The bin/run-all-petstore script now tells me it doesn't need to be run. What is the way to go here?)

I still would like to get some feedback on the questions in the initial comment.

wing328:
>  I think nobody should have done that, but who knows. Opinions on that?

I'm ok with the "fix" as your way should be the correct behaviour.

> This applies to the key-value-list ones --instantiation-type, --type-mappings, --additional-properties, --import-mappings, but also to --language-specific-primitives (which is just a list of strings). Opinions on that?

Sounds good to me to make the behaviour consistent.

ePaul:
@wing328 I'll add those then in the next days (likely over the weekend). Feel free to merge this if you want – I'll open a new PR.

wing328:
@ePaul sure. I'll merge this tomorrow.

wing328:
Tests passed via https://travis-ci.org/swagger-api/swagger-codegen/builds/243586138

PR #5194, opened 2017-03-25
[Javascript] Support for ByteArray and Binary

jdupouy:
This PR replaces the String type for binary and byte array properties by a blob type which allows superagent to return bytes array instead of a String or a JSON object.


wing328:
@jdupouy Thanks for the enhancement and updating superagent to the latest version.

cc @frol @delenius

PR #5225, opened 2017-03-27
[swift] Open buildHeaders() and requestBuilderFactory property to allow extension. Even though AlamofireRequestBuilderFactory looks like is designed for extension, it is not because property has private access.

tomekc:
This change is to make possible to override AlamofireRequestBuilderFactory in order to perform
custom initialization of URLSession, for example to enable SSL pinning.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/swift3/APIs.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift3/APIs.mustache
@@ -10,7 +10,7 @@ open class {{projectName}}API {
     open static var basePath = "{{{basePath}}}"
     open static var credential: URLCredential?
     open static var customHeaders: [String:String] = [:]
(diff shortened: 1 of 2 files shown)


jaz-ah:
@tomekc I can understand opening up the factory object but why buildHeaders?  customHeaders is already exposed...

tomekc:
@jaz-ah original AlamofireRequestBuilder makes use of `buildHeaders` in createSessionManager. So subclass probably should do too.

jaz-ah:
ok - @wing328 +1 here.

wing328:
@tomekc please run `./bin/swift3-petstore-all.sh` to update the Petstore sample.

wing328:
Petstore samples updated via 56d4382

Issue #5132, opened 2017-03-21
[C#] API body parameter with Map/Dictionary Properties converting to object instead of Dictionary.


muralidharalladi:
##### Description

API body parameter with Map/Dictionary Properties converting to object instead of Dictionary.
Response  with Map/Dictionary Properties converting to Dictionary. we are expecting same behavior in parameters  also. below spec generates "ResourceApi.cs" file

Methods:
Dictionary<string, Resource> GetResource (Object resourceParameters);
ApiResponse<Dictionary<string, Resource>> GetResourceWithHttpInfo (Object resourceParameters);
System.Threading.Tasks.Task<Dictionary<string, Resource>> GetResourceAsync (Object resourceParameters);
System.Threading.Tasks.Task<ApiResponse<Dictionary<string, Resource>>> GetResourceAsyncWithHttpInfo (Object resourceParameters);

In above methods instead of Object we are expecting Dictionary<string, ResourceParameters>

##### Swagger-codegen version

Swagger-codegen version

Swagger Editor

##### Swagger declaration file content or url

{
  "swagger": "2.0",
  "info": {
    "description": "sample",
    "version": "1.0.0",
    "title": "Swagger sample",
    "termsOfService": "http://swagger.io/terms/",
    "contact": {
      "email": "test@swagger.io"
    },
    "license": {
      "name": "Apache 2.0",
      "url": "http://www.apache.org/licenses/LICENSE-2.0.html"
    }
  },
  "host": "test.swagger.io",
  "basePath": "/v2",
  "tags": [
    {
      "name": "Resource",
      "description": "Everything about resource"
    }
  ],
  "schemes": [
    "http"
  ],
  "paths": {
    "/Resources": {
      "get": {
        "tags": [
          "Resource"
        ],
        "summary": "Find Resource",
        "description": "",
        "operationId": "getResource",
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "ResourceParameters",
            "in": "body",
            "description": "ResourceParameters object that needs to be added",
            "required": true,
            "schema": {
              "type": "object",
              "additionalProperties": {
                "$ref": "#/definitions/ResourceParameters"
              }
            }
          }
        ],
        "responses": {
          "200": {
            "description": "successful operation",
            "schema": {
              "type": "object",
              "additionalProperties": {
                "$ref": "#/definitions/Resource"
              }
            }
          },
          "400": {
            "description": "Invalid status value"
          }
        }
      }
[...]

PR #5228, opened 2017-03-27
[all] Map additionalParameters.$ref into a RefModel

jimschubert:
### PR checklist

### Description of the PR

Possible fix for #5132 in which a RefModel defined as additionalProperties on a body parameter was being ignored.

I'm not entirely sure this is an appropriate fix because this assumes additionalProperties.$ref refers to *the* model definition, rather than a mixin of properties to be applied to some dictionary of key/values.

The way `additionalProperties` is meant to work is to validate additional allowed values for a JSON structure defined outside of `patternProperties` or `properties` (although Swagger 2.0 specification doesn't mention `patternProperties`). Ideally, I'd assume this would mean all properties from both are merged into a final object model for generation. This PR *does not* do that. This PR assumes the existence of `additionalProperties.$ref` means the definition of the RefModel (as opposed to common RefModel processing). This may not be correct, but I'm opening the PR to open discussion.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -1381,10 +1381,25 @@ public class DefaultCodegen {
                 m.allowableValues = new HashMap<String, Object>();
                 m.allowableValues.put("values", impl.getEnum());
             }
+
+            addVars(m, impl.getProperties(), impl.getRequired());
+
             if (impl.getAdditionalProperties() != null) {
+                final Property additionalProperties = impl.getAdditionalProperties();
+                if(additionalProperties instanceof RefProperty){
+                    RefProperty refProperty = ((RefProperty)additionalProperties);
+                    m.dataType = getSwaggerType(refProperty);
+                    final String modelName = refProperty.getSimpleRef();
+                    if(allDefinitions != null) {
+                        final Model child = allDefinitions.get(modelName);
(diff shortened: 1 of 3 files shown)


ePaul:
I didn't yet try to understand what your code does, but about the explanation:

> I'm not entirely sure this is an appropriate fix because this assumes additionalProperties.$ref refers to the model definition, rather than a mixin of properties to be applied to some dictionary of key/values.

> The way additionalProperties is meant to work is to validate additional allowed values for a JSON structure defined outside of patternProperties or properties (although Swagger 2.0 specification doesn't mention patternProperties).

As I understand it (disregarding patternProperties for now), it means that for the validated/described JSON object:
* all properties with names mentioned in `properties` comply to their respective schemas
* all properties with names not mentioned in `properties` comply to the schema given by `additionalProperties`. (There is no restrictions on which names there could be for those, other than that the ones above are excluded).

And if any of those "schemas" are `$ref`s, the properties comply to the referred schema instead.

> Ideally, I'd assume this would mean all properties from both are merged into a final object model for generation.

This is for the on-the-wire format, so it doesn't really give restrictions on how it is mapped to objects in any language.

In most statically-typed languages you can't really define "this object can have properties of any name, all with the same type X, except properties a, b, c which have other types" for an object.

> This PR does not do that. This PR assumes the existence of additionalProperties.$ref means the definition of the RefModel (as opposed to common RefModel processing). This may not be correct, but I'm opening the PR to open discussion.

Sorry, I don't really understand your sentence. (Maybe I need to first dig a bit into Codegen's internals to see how all the models are related.)

PR #5232, opened 2017-03-28
Add port 80 to swagger spec

wing328:
### PR checklist

### Description of the PR

Add port 80 to Swagger spec so as to ensure the port is shown correctly in the auto-generated code/doc.


ePaul:
Shouldn't port 80 be the default for HTTP, i.e. shouldn't mentioning it be superfluous?

wing328:
@ePaul I remember there was an issue the port number was not captured in {{host}} so that's why I put 80 there to ensure the port number is there.

Ideally we should write a unit test to ensure `:80` is part of the host.

ePaul:
I updated the samples (in #5243), the diff might show the relevant places (potentially where we have 8080 → 80?).

wing328:
@ePaul thanks! Later I'l add a Java test case as well.

> (potentially where we have 8080 → 80?).

Not something I would expect.

PR #5234, opened 2017-03-28
Fixes in escaping when generating Java from swagger.json.

spr3nk3ls:
### Description of the PR

Some valid json expressions within a swagger.json file result in uncompilable Java code.

How to test:
Use [this pom.xml](https://gist.github.com/spr3nk3ls/d7021930cc42941d413ef120b710beb9) and [this swagger.json](https://gist.github.com/spr3nk3ls/f165f0f19656a9e1ef8bf0f900f5067a) to generate an example swagger API client in Java. 

Expected: the code generated results in compilable Java.

Result: the resulting Java does not compile, because expressions like '"example": "[\"RABO23124NL\",\"RABO12345NL\"]",' and '"example": "\\\"[A-Z.]{1,10}\\",' are escaped incorrectly.


wing328:

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java
```diff
@@ -705,6 +706,20 @@ public void setParameterExampleValue(CodegenParameter p) {
...
+        if(p.getExample() != null) {
+            return escapeText(p.getExample().toString());
+        } else {
+        	return super.toExampleValue(p);
+        }
+    }
+    
+    @Override
+    public Mustache.Compiler processCompiler(Mustache.Compiler compiler) {
+    	return compiler.escapeHTML(false);
```
wing328: Instead of turning it off, what about using {{{ ... }}} instead to use the unescape value?

The change will have a lot of impacts to mustache templates used by Java AP clients as well as server stubs (spring. jaxrs, play, etc)

spr3nk3ls:
I was unable to fix this PR, so I made a new one with the suggested changes. See https://github.com/swagger-api/swagger-codegen/pull/5247

wing328:
@spr3nk3ls np. You may find https://github.com/swagger-api/swagger-codegen/wiki/FAQ#how-can-i-rebase-my-pr-on-the-latest-master useful.

PR #5243, opened 2017-03-28
Updating samples after #5232.

ePaul:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change.
  I ran bin/run-all-petstore, but before committing I omitted some changes listed below.
- [x] Filed the PR against the correct branch: master for non-breaking changes.

### Description of the PR

No code changes, just updating the samples.

All the changes here come from #5232, which forgot to update the samples.
I omitted the changes covered by #5242, #5241, #5212, as well as changes in typescript-angular2/npm and typescript-jquery/npm, which have the date in their artifact ID (see #3084).

Interesting in the diff might be the several cases which replaced 8080 → 80.

PR #5246, opened 2017-03-28
[JaxRS Server][Jersey1/2] Pass server port by cli

ePaul:
### PR checklist

  → bin/jaxrs-jersey1-petstore-server.sh, bin/jaxrs-petstore-server.sh

### Description of the PR

We are introducing a CLI argument `serverPort` to the generators which overrides the port given in the API definition's `host` or the default of `8080`.
This we use from the generating scripts, so the CI tests won't try to use `80` (which is usually reserved to privileged processes).

(Triggered by the change in #5232 and the failing tests in #5243.)


wing328:
@ePaul thanks for the enhancement, which makes it much easier to set the server port.

PR #5212, opened 2017-03-26
[Java][JaxRS][Spring-MVC] Update samples for jaxrs with Jersey 1/2 and for spring-mvc

ePaul:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change.
  → bin/spring-mvc-petstore-j8-async-server.sh, bin/spring-mvc-petstore-server.sh, bin/jaxrs-petstore-server.sh, bin/jaxrs-jersey1-petstore-server.sh
- [x] Filed the PR against the correct branch: master for non-breaking changes.

### Description of the PR

No code changes, just updating some samples so later updates show the actual changes.
I grouped those 4 samples together because the changes were quite similar.

* FakeApi.testEnumParameters has some additional annotation parameters for enums in parameters.
  Some of them look wrong.
* In some model files line breaks for `@Min`/`@Max` annotations changed.

PR #5247, opened 2017-03-29
Fixes in escaping when generating Java from swagger.json (2)

spr3nk3ls:
Some valid json expressions within a swagger.json file result in uncompilable Java code.

How to test:
Use [this pom.xml](https://gist.github.com/spr3nk3ls/d7021930cc42941d413ef120b710beb9) and [this swagger.json](https://gist.github.com/spr3nk3ls/f165f0f19656a9e1ef8bf0f900f5067a) to generate an example swagger API client in Java.

Expected: the code generated results in compilable Java.

Result: the resulting Java does not compile, because expressions like `"example": "[\"RABO23124NL\",\"RABO12345NL\"]",` and `"example": "\\\"[A-Z.]{1,10}\\",` are escaped incorrectly.

I was unable to fix my original PR (https://github.com/swagger-api/swagger-codegen/pull/5234), so I created this one instead. Suggestions made on the original PR have been processed in this one.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java
@@ -705,6 +705,15 @@ public abstract class AbstractJavaCodegen extends DefaultCodegen implements Code
     }
 
     @Override
+    public String toExampleValue(Property p) {
+        if(p.getExample() != null) {
+            return escapeText(p.getExample().toString());
+        } else {
+        	return super.toExampleValue(p);
+        }
+    }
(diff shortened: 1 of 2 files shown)


ePaul:
There are no sample changes caused by this – regenerating the samples for the java projects just gives the changes which are included in my update-samples PRs #5243 and #5212.

Maybe we should have an example in one of the test YAMLs which shows the effect?

wing328:
@ePaul yup, we can add examples to the fake petstore spec so as to cover these cases

wing328:
@spr3nk3ls thanks for the PR, which has been merged into master.

I'll create an issue to track updating fake petstore with examples you provided in your spec.

PR #5263, opened 2017-03-30
Add File support + some fixes to Play generator

JFCote:
### PR checklist

### Description of the PR

@wing328 
I need help to make the java import for FileInputStream works. But right now, it import java.io.File instead.

Please add this endpoint in your original yaml to test:

```yaml
/devices/{deviceId}/usb-config-update-zip:
  post:
    tags:
      - Device
    summary: Create the USB config update for the specified device
    description: ""
    operationId: createUsbConfigUpdateZip
    produces:
      - application/zip
    parameters:
      - in: path
        name: deviceId
        required: true
        type: string
    responses:
      200:
        description: OK
        schema:
          type: file
```

As you can see in the output, everything is ok except the import is not good. 
If you can help me find the problem, I will commit a fix and it will be a very good addition for the Play generator.


ePaul:
Why `FileInputStream` and not just `InputStream`?

The `type: file` in the API definition doesn't mean that the data comes from an actual file (or goes into one), it just means it is some binary stream which is not specified more in detail (and is not JSON/XML).

For responses, there is no reason the server implementation should have to store anything first into a file before returning a FileInputStream. For a parameter, there is no reason the framework should have to store  the received data into a file before passing it into your server implementation.

JFCote:
@ePaul : Thanks for the feedback. You are right! I will replace it with InputStream

Do you understand why the import is not working? I didn't hear from @wing328 so if you have an idea, let me know :) 

I'll send an update for the InputStream maybe on monday because right I'm pretty busy.

ePaul:
I didn't look into it in detail, hope to come to it tomorrow.

wing328:
@JFCote I didn't have a chance to look into the issue yet. I'll do it on Monday if no one has time to  suggest a fix.

ePaul:
DefaultCodegen has a type-mapping `file` → `File`, and an import-mapping `File` → `java.io.File`. I'm a bit dubious about this ... shouldn't DefaultCodegen be language-agnostic?
(And also, I doubt java.io.File is ever a valid parameter or result type.)

ePaul:
@JFCote I didn't figure out why your inserted `if` doesn't work, but adding a

```java
        typeMapping.put("file", "FileInputStream");
```
(after line 188) seems to work, even without your added if-block in lines 293-297. (Of course, as discussed above, please use `InputStream` instead.)

JFCote:
@ePaul : I will do the fix this morning and test it out. I'll submit an update later! Thanks for your time.

JFCote:
@wing328 and @ePaul : Here is a "working" version but something is not right:

- Why does it work without the `if` ?
- Why does it import java.io.InputStream even if the file is not using it? Check the updated sample. You can see that PetApiController.java does import java.io.InputStream but it is not used...

Any help to "clean" this would be appreciated!
Thanks!

ePaul:
@JFCote When using typeMapping `file` → `InputStream`, then the return type of your method won't be `File` anymore, but `InputStream`, so your if block has no effect now.
It had no visible effect even before, because `postProcessOperations` is called from `DefaultGenerator.processOperations` after the imports map is build (actually just some lines).

`CodegenOperation.imports` will be build from all the types which were used in the response schemas and the parameters (DefaultCodegen.fromOperation collects those in the `imports` local variable, with help of `fromParameter`). So if there is a Swagger type `file` in either parameter or response, the `InputStream` class is imported.

For your PetApiControllerImp, that class contains a method generated from the uploadFile operation:

```json
    "/pet/{petId}/uploadImage": {
      "post": {
        "tags": [
          "pet"
        ],
        "summary": "uploads an image",
        "description": "",
        "operationId": "uploadFile",
        "consumes": [
          "multipart/form-data"
        ],
        "produces": [
          "application/json",
          "application/xml"
        ],
        "parameters": [
          {
            "name": "petId",
            "in": "path",
            "description": "ID of pet to update",
            "required": true,
            "type": "integer",
            "format": "int64"
          },
          {
            "name": "additionalMetadata",
            "in": "formData",
            "description": "Additional data to pass to server",
            "required": false,
            "type": "string"
          },
          {
            "name": "file",
            "in": "formData",
            "description": "file to upload",
            "required": false,
            "type": "file"
          }
        ],
        "responses": {
          "default": {
            "description": "successful operation"
          }
        },
        "security": [
          {
            "petstore_auth": [
              "write:pets",
              "read:pets"
            ]
          }
        ]
      }
    },
```
The generated method doesn't get an InputStream as a parameter, though:
```
    @Override
    public void uploadFile(Long petId, String additionalMetadata, Http.MultipartFormData.FilePart file) throws Exception {
        //Do your magic!!!
        
    }
```

The file parameter got (by special logic in the template, see formParams.mustache) the type `Http.MultipartFormData.FilePart` instead of `InputStream`, which it would get if it hadn't the special-handling here.

I'm not sure how that would be done better ... maybe this special-handling of the parameter should be done in the generator class (JavaPlayFrameworkCodegen), possibly overriding fromParameter to add `FilePart` instead of `InputStream` to the imports (and use it as the parameter type) – this way the template file could be a lot simpler.

JFCote:
@ePaul : Thank you very much for this explanation, it completely make sense now! I will take a quick look tomorrow morning to see if I can implement the simplification you suggest at the end of your comment. Will let you know!

JFCote:
@ePaul and @wing328 : Here you go. It's not perfect but it's the best I can do based on your comments and my knowledge of the app. The templates now doesn't contains the code specific for the InputStream and instead it is done in the JavaPlayFramework processing. I hope it is ok, let me know if it needs something before being merged!

Thanks

wing328:
@JFCote I may have missed it. What's the issue with `isFile` mustache tag to check if the parameter is a file?

JFCote:
@wing328 : The issue was simply that since it's in the mustache, we could not do dynamic management of the import so I needed to do unused import in most file because of that. I was based on @ePaul comment on how I could do it better.

Now it's handled in the postProcessing and the mustache is much more simple and clean.

JFCote:
@wing328 : Do you think this will be merged soon? Thanks!

wing328:
@JFCote just taken another look and the change looks good. Thanks for the enhancement.

PR #5267, opened 2017-03-30
[Swift3] Fix bug where non-camel-case path params didn't work.

GriffinSchneider:
### PR checklist

### Description of the PR

Since https://github.com/swagger-api/swagger-codegen/pull/4916, path parameters that aren't in camel case have been broken.

When there are path parameters, code is generated like this:

```
        var path = "/{accountId}/thing"
        path = path.replacingOccurrences(of: "{account_id}", with: "\(accountId)", options: .literal, range: nil)
```

note that the path contains `{accountId}`, but we're trying to replace `{account_id}`, so it will never work. This is because the `normalizePath` function was camel-casing path parameters in the path.

This pullreq removes `normalizePath`, so the generated code now looks like this:

```
        var path = "/{account_id}/thing"
        path = path.replacingOccurrences(of: "{account_id}", with: "\(accountId)", options: .literal, range: nil)
```

and the replacement will work.


jaz-ah:
@GriffinSchneider I'm wondering what cases that code was originally added to solve - I think it would be worth adding in some tests for your cases and the original cases this code was meant to add to make sure we cover everything here.

GriffinSchneider:
Here's what I think happened:

The problem area in the swift3 templates is here: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/swift3/api.mustache#L107-L108

Before https://github.com/swagger-api/swagger-codegen/pull/4916, we used to be calling `replacingOccurrences` with  `paramName`. According to https://github.com/swagger-api/swagger-codegen/issues/4898, using `paramName` in this way was causing bugs in other clients, because `paramName` is the 'normalized' (i.e. camel-cased) value, but the path we're trying to replace in is not camel cased. However, the swift3 client didn't have this bug, because someone added `normalizePath` as a workaround that would normalize all the parameters in the path so that they matched the normalized value in `paramName`. 

Then, to fix https://github.com/swagger-api/swagger-codegen/issues/4898, `paramName` was replaced with `baseName` in the templates of many clients. In the swit3 client, this caused a bug, since the path was still being normalized but `baseName` contains the _non-normalized_ parameter name. There are no non-camel-case path parameters in the petstore example specs, so the problem wasn't noticed and got merged.

So, my fix is simply to remove `normalizePath`, which was a bad workaround to begin with. Now the path contains non-normalized parameter names as expected, and we use the non-normalized `baseName` to replace them.

I will attempt to add a test that would have caught this issue, but including a non-camel-case path parameter in the petstore specs.

GriffinSchneider:
While trying to create a test for my original bug, I noticed a separate unrelated bug that was already breaking the swift3 tests. 

The bug was in https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/swift3/_param.mustache 

For enums of integer and long, that template was generating code that would work for integers or longs, but did not compile for _enums_ of integer or long. c61f0ccf1b3b042488a0d3efccd0e2e25dc0c0b1 fixes it. The swift3 tests now pass.

GriffinSchneider:
OK, I've changed the tests so that they would have failed due to path parameter issues before this pull request.

jaz-ah:
@GriffinSchneider thx for the extra effort here!  @wing328 +1

wing328:
@GriffinSchneider thanks for fixing the tests.

PR #5268, opened 2017-03-30
[Java,Jersey2] Files.copy in generated code throws FileAlreadyExistsException

zszugyi:
The code generated for Jersey2 is unable to dowload file responses.
For an endpoint with the file response type, the API call with the generated client results in a FileAlreadyExistsException.

A sample response:
      responses:
        200:
          description: The template workbook
          schema:
            type: file

The exception:
xyz.client.ApiException: java.nio.file.FileAlreadyExistsException: e:\temp-dir\download-4619092686092804097
    at xyz.client.ApiClient.downloadFileFromResponse(ApiClient.java:576)
    at xyz.client.ApiClient.deserialize(ApiClient.java:550)
    at xyz.client.ApiClient.invokeAPI(ApiClient.java:693)
    at xyz.client.api.DefaultApi.getTemplateFile(DefaultApi.java:79)

In order for Files.copy to not fail, Files.copy needs to have the StandardCopyOption.REPLACE_EXISTING option specified (prepareDownloadFile creates an empty temp file, it doesn't just create a path)

### DIFF
--- a/modules/swagger-codegen/src/test/resources/2_0/templates/Java/libraries/jersey2/ApiClient.mustache
+++ b/modules/swagger-codegen/src/test/resources/2_0/templates/Java/libraries/jersey2/ApiClient.mustache
@@ -27,6 +27,7 @@ import java.io.InputStream;
 
 {{^supportJava6}}
 import java.nio.file.Files;
+import java.nio.file.StandardCopyOption;
 {{/supportJava6}}
 {{#supportJava6}}
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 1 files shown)


wing328:

wing328:
@zszugyi should the file be moved to another folder right after download instead? Otherwise, there may be a chance for data loss.

In other words, I think the exception is the expected behaviour as the file should not be overwritten.

zszugyi:
@wing328 there is no data loss. The file is an empty file that has just been created by createTempFile. It's been created so that the payload can be written to it (which is what Files.copy would do). It's the same thing that happens a few lines below in the supportJava6 branch.

wing328:
@zszugyi yes, you're right that `copyToFile` is already doing that: https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FileUtils.html#copyToFile(java.io.InputStream,%20java.io.File)

wing328:
For your upcoming PRs, I would suggest creating a new branch as per git best practices.

HaraldWalker:
@wing328 I've been wondering why I don't see it. Pull request changed only a test file, not the actual ApiClient.mustache file that is being used.
Only file changed was: /src/test/resources/2_0/templates/Java/libraries/jersey2/ApiClient.mustache

And I noticed that the option supportJava6 is supposed to be for Jersey1 only according to the help.

PR #5271, opened 2017-03-31
Merged changes from Sean11s PR#2790 to latest master.

markebbers:
### PR checklist

### Description of the PR

Merged the work of @sean11 to allow query collections in C#. See https://github.com/swagger-api/swagger-codegen/pull/2790

### DIFF
--- a/modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
@@ -116,7 +116,7 @@ namespace {{packageName}}.Client
 
         // Creates and sets up a RestRequest prior to a call.
         private RestRequest PrepareRequest(
-            String path, {{^netStandard}}RestSharp.{{/netStandard}}Method method, Dictionary<String, String> queryParams, Object postBody,
+            String path, RestSharp.Method method, List<KeyValuePair<String, String>> queryParams, Object postBody,
             Dictionary<String, String> headerParams, Dictionary<String, String> formParams,
             Dictionary<String, FileParameter> fileParams, Dictionary<String, String> pathParams,
             String contentType)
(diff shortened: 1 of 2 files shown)


modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -116,7 +116,7 @@ namespace {{packageName}}.Client
 
         // Creates and sets up a RestRequest prior to a call.
         private RestRequest PrepareRequest(
-            String path, {{^netStandard}}RestSharp.{{/netStandard}}Method method, Dictionary<String, String> queryParams, Object postBody,
+            String path, RestSharp.Method method, List<KeyValuePair<String, String>> queryParams, Object postBody,
```
jimschubert: @wing328 we'll need to document this breaking change.

wing328: @jimschubert yes, you're right. I'll take note of that. Should the {{netStandard}} tags be removed?

jimschubert: @wing328 I'm pretty sure the `{{^netStandard}}` is required because of the differences between RestSharp and RestSharp.Portable.  iirc, we were holding off on #2790 until version 2.3.0 because of the breaking change in the method signature.

Do you think it would make sense to provide an overloaded signature with the older Dictionary type to avoid breaking consumers that rely on this method?

wing328: >  we were holding off on #2790 until version 2.3.0 because of the breaking change in the method signature

Yup, I think that's why @markebbers filed this change against 2.3.0 branch.

> Do you think it would make sense to provide an overloaded signature with the older Dictionary type to avoid breaking consumers that rely on this method?

Yup, but I would suggest we consider it if there are demands from our users.

wing328:
Closed via https://github.com/swagger-api/swagger-codegen/pull/5307

Issue #5296, opened 2017-04-01
[Java][feign] snake_case parameter names break feign client


ePaul:
##### Description

Updating the samples for java/feign in current master and then running the tests of the generated client leads to test failures:

```
Tests in error: 
  testDeleteOrder(io.swagger.client.api.StoreApiTest): status 400 reading StoreApi#getOrderById(Long); content:(..)
  testPlaceOrder(io.swagger.client.api.StoreApiTest): status 400 reading StoreApi#getOrderById(Long); content:(..)
```

##### Swagger-codegen version

Yesterday's master (64c7e23a04d68db9ad), not sure if this applies to any released version.

##### Swagger declaration file content or url

[petstore-with-fake-endpoints-models-for-testing.yaml](https://github.com/swagger-api/swagger-codegen/blob/64c7e23a04d68db9ad27f11295141d8df4b413b9/modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml)

##### Command line used for generation

    bin/java-petstore-feign.sh

##### Steps to reproduce

```bash
bin/java-petstore-feign.sh
mvn verify -P samples -pl :swagger-petstore-feign
```

##### Related issues

I think this might be another instance of #4898, which was uncovered by the parameter name change in #5267. (I noticed this by the failing tests in #5281.)

##### Suggest a Fix

This is now generated:
```java
  /**
   * Delete purchase order by ID
   * For valid response try integer IDs with value &lt; 1000. Anything above 1000 or nonintegers will generate API errors
    * @param orderId ID of the order that needs to be deleted (required)
   */
  @RequestLine("DELETE /store/order/{order_id}")
  @Headers({
    "Content-Type: application/json",
    "Accept: application/json",
  })
  void deleteOrder(@Param("orderId") String orderId);
```
The `@RequestLine` changed from previously `{orderId}` to now `{order_id}`, while the `@Param` didn't change, and now doesn't match anymore.

ePaul:
I did actually have a look at the `@Param({{paramName}})` in the template when doing the analysis for #4898. For query and header parameters, the sanitized name is actually wanted, because the annotations refer to them, mapping the on-the-wire name to the sanitized names. (That is why I said "don't change this".)

For path parameters this does not happen, we just get the whole path (including the templates with their specification-defined path parameter names) pasted in, as this is already the format wanted by Feign.

I see two approaches here:
[...]

PR #5301, opened 2017-04-02
[Java] add logic to camelize path variables for Feign client

wing328:
### PR checklist

### Description of the PR

For https://github.com/swagger-api/swagger-codegen/issues/5296

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
@@ -242,14 +242,14 @@ public class JavaClientCodegen extends AbstractJavaCodegen
     @Override
     public Map<String, Object> postProcessOperations(Map<String, Object> objs) {
         super.postProcessOperations(objs);
-        if(usesAnyRetrofitLibrary()) {
+        if (usesAnyRetrofitLibrary()) {
             Map<String, Object> operations = (Map<String, Object>) objs.get("operations");
             if (operations != null) {
                 List<CodegenOperation> ops = (List<CodegenOperation>) operations.get("operation");
                 for (CodegenOperation operation : ops) {
                     if (operation.hasConsumes == Boolean.TRUE) {
 
-                        if ( isMultipartType(operation.consumes) ) { 
+                        if (isMultipartType(operation.consumes)) {
                             operation.isMultipart = Boolean.TRUE;
                         }
                         else {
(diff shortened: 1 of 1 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
```diff
@@ -264,6 +264,27 @@ private boolean usesRetrofit2Library() {
...
             }
         }
+
+        // camelize path variables for Feign client
+        if ("feign".equals(getLibrary())) {
+            Map<String, Object> operations = (Map<String, Object>) objs.get("operations");
+            List<CodegenOperation> operationList = (List<CodegenOperation>) operations.get("operation");
+            for (CodegenOperation op : operationList) {
+                String path = new String(op.path);
+                String[] items = path.split("/", -1);
```
ePaul: Hmm, not sure if this splitting is right ... by this, we can't match e.g. `/foo{bla_id}bar/`. Not sure what the specification says about this, though.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
```diff
@@ -264,6 +264,27 @@ private boolean usesRetrofit2Library() {
                 }
             }
         }
+
+        // camelize path variables for Feign client
+        if ("feign".equals(getLibrary())) {
+            Map<String, Object> operations = (Map<String, Object>) objs.get("operations");
+            List<CodegenOperation> operationList = (List<CodegenOperation>) operations.get("operation");
+            for (CodegenOperation op : operationList) {
+                String path = new String(op.path);
```
ePaul: Why `new String(...)` here? op.path is already a string.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
```diff
@@ -264,6 +264,27 @@ private boolean usesRetrofit2Library() {
...
+                String[] items = path.split("/", -1);
+                String opsPath = "";
+                int pathParamIndex = 0;
+
+                for (int i = 0; i < items.length; ++i) {
+                    if (items[i].matches("^\\{(.*)\\}$")) { // wrap in {}
+                        // camelize path variable
+                        items[i] = "{" + camelize(items[i].substring(1, items[i].length()-1), true) + "}";
```
ePaul: I guess instead of `camelize` we should use `toParamName` here.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
ePaul: Could we use something like this instead?

```java
            for (CodegenOperation op : operationList) {
                String path = op.path;
                for(CodegenParameter param : op.pathParams) {
                    path = path.replace("{"+param.baseName+"}", "{"+param.paramName+"}");
                }
                op.path = path;
            }
```

ePaul: Both the [2.0 spec](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/2.0.md#pathTemplating) and the [upcoming 3.0 spec](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#pathTemplating) are very vague about how path templating works:

> Path templating refers to the usage of curly braces ({}) to mark a section of a URL path as replaceable using path parameters.

I would argue this does not say it is restricted to `/`-separated segments.

wing328: Yup your solution is better.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
wing328: Right, `toParamName` should be used as it's the method for naming parameter

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
wing328: Sure. Will remove it.

Issue #4235, opened 2016-11-22
[Python] No way to handle exceptions in asyncronous code (with callback)


masterandrey:
##### Description

Produced by codegen Python code does not support handling of exceptions in anync mode (with callback).
If swagger client code raises exeption, callback will never be called.

##### Swagger-codegen version

5

##### Command line used for generation

-l python

##### Suggest a Fix

Catch all exceptions in generated swagger client and return special type of result in case of exceptions.

wing328:
@masterandrey can you elaborate a bit on "return special type of result"? In your case, what exception did you get?

masterandrey:
with callback I have this unhandled exception:

``
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/home/sorokin/anaconda3/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/home/sorokin/anaconda3/lib/python3.5/threading.py", line 862, in run
    self._target(*self._args, **self._kwargs)
  File "/home/sorokin/cottrep/api/stand/swagger/agentticketsapi_client/api_client.py", line 157, in __call_api
    post_params=post_params, body=body)
  File "/home/sorokin/cottrep/api/stand/swagger/agentticketsapi_client/api_client.py", line 370, in request
    body=body)
  File "/home/sorokin/cottrep/api/stand/rest.py", line 242, in POST
    body=body)
  File "/home/sorokin/cottrep/api/stand/rest.py", line 210, in request
    raise ApiException(http_resp=r)
rest.ApiException: (404)
Reason: Handler not registered for  corresponding URI: /api/agentticketsapi/request_pay
HTTP response headers: HTTPHeaderDict({'Content-Type': 'text/html; charset=iso-8859-1', 'Connection': 'keep-alive', 'Content-Length': '0', 'Date': 'Wed, 23 Nov 2016 07:51:48 GMT', 'Server': 'nginx/1.11.1'})
``

without callback I can handle the exception in main thread:
``
Traceback (most recent call last):
  File "/home/sorokin/cottrep/api/stand/call_api.py", line 107, in call_api
    response = api_call(**args)
  File "/home/sorokin/cottrep/api/stand/swagger/agentticketsapi_client/apis/default_api.py", line 392, in controllers_agent_tickets_request_pay
    (data) = self.controllers_agent_tickets_request_pay_with_http_info(card_num, **kwargs)
[...]

PR #5308, opened 2017-04-03
[Python] async error handling fix

baartosz:
### PR checklist

### Description of the PR
Fixes issue https://github.com/swagger-api/swagger-codegen/issues/4235. Implementation of solution suggested here: https://github.com/swagger-api/swagger-codegen/issues/4235#issuecomment-291086921. See test cases for examples: https://github.com/swagger-api/swagger-codegen/pull/5308/files#diff-c5c849b26d5134175812cdb46ddbe39cR143

In next iteration implementation can be improved by adding separate methods for async stuff like `api.get_pet_by_id_async()` and `api.get_pet_by_id_async_with_http_info()`.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/python/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/api.mustache
@@ -37,20 +37,16 @@ class {{classname}}(object):
         {{{notes}}}
 {{/notes}}
         This method makes a synchronous HTTP request by default. To make an
-        asynchronous HTTP request, please define a `callback` function
-        to be invoked when receiving the response.
-        >>> def callback_function(response):
-        >>>     pprint(response)
-        >>>
+        asynchronous HTTP request, please pass async=True
 {{#sortParamsByRequiredFlag}}
-        >>> thread = api.{{operationId}}({{#allParams}}{{#required}}{{paramName}}, {{/required}}{{/allParams}}callback=callback_function)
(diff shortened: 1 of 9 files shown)


wing328:
@baartosz thanks for the PR. 

@scottrw93 @frol @taxpon @scop please take a look when you've time.

frol:
Looks good to me!

@baartosz Thank you for the PR!

wing328:
@baartosz I noticed that the parameter `callback` has been renamed to `async`. Would it be ok to keep using `callback` as I want to minimize the change required for developers to use the enhancement?

wing328:
@baartosz sorry please ignore my comment above. After rereading the code I think I now have a better understanding of how it works.

wing328:
@frol thanks for reviewing the PR.

@baartosz thanks for fix 👍 

It would be nice to update the auto-generated documentation to provide an example on how to make an async call with the Python SDK.

arcward:
I'm a little late on this, but it is possible to handle exceptions via the callback, though it's not pretty. I ended up, similar to the suggestion in #4235, creating a wrapper around the callback that accepts ``threading.Event()`` , which is added to a list. Then I ``wait()`` on all of the events, and ``set()`` is called from inside the callback. So when an ApiException gets thrown, it doesn't happen in an untouchable background thread. There's a better example here: https://stackoverflow.com/questions/38499261/how-to-block-and-wait-for-async-callback-based-python-function-calls

PR #5337, opened 2017-04-06
Add option to skip Maven plugin execution

sadimusi:
### PR checklist

### Description of the PR

The execution is skipped if either the `codegen.skip` property or the `<skip>`
configuration parameter is set. This is consistent with how many other Maven
plugins, such as maven-exec-plugin and maven-clean-plugin, handle this.

### DIFF
--- a/modules/swagger-codegen-maven-plugin/src/main/java/io/swagger/codegen/plugin/CodeGenMojo.java
+++ b/modules/swagger-codegen-maven-plugin/src/main/java/io/swagger/codegen/plugin/CodeGenMojo.java
@@ -228,7 +228,11 @@ public class CodeGenMojo extends AbstractMojo {
     @Parameter(name = "generateApiDocumentation", required = false)
     private Boolean generateApiDocumentation = true;
 
-
+    /**
+     * Skip the execution.
+     */
+    @Parameter(name = "skip", property = "codegen.skip", required = false, defaultValue = "false")
+    private Boolean skip;
 
     /**
      * Add the output directory to the project as a source root, so that the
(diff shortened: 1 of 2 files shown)


wing328:
@sadimusi thanks for the PR.

cc @ePaul

modules/swagger-codegen-maven-plugin/src/main/java/io/swagger/codegen/plugin/CodeGenMojo.java
```diff
@@ -228,7 +228,11 @@
     @Parameter(name = "generateApiDocumentation", required = false)
     private Boolean generateApiDocumentation = true;
 
-
+    /**
+     * Skip the execution.
+     */
+    @Parameter(name = "skip", property = "codegen.skip", required = false, defaultValue = "false")
```
ePaul: I'm not sure about this name ... is `codegen.skip` a name used by other code generators too (where it might be sensible to skip all code generation in a project)?
Otherwise maybe `swagger-codegen.skip` would be more clear.

ePaul:
Not sure about the property name (see comment), but in general it looks good.
(I didn't try it yet, will try to do later.)

sadimusi:
Good point, I wasn't quite sure about the name myself. I've looked at a few other generators, but none of them offered such a property. I finally settled on `codegen.skip` mainly because of maven-exec-plugin's `exec.skip`. Another option would be to name it `swagger.codegen.skip`, after maven-clean-plugin's `maven.clean.skip`.

wing328:
I think we can go with `codegen.skip` and see if developers using that have any feedback.

Issue #5240, opened 2017-03-28
[Java] How can we set List and Map to null by default in models?


egormasalitin:
##### Description

For example we have model in `crud-service`:
```
ModelDto {
 private List<String> list = null;
 private SomeOtherModelDto object = null;
 //get set
}
```

in swagger spec it will appear as:

```
"list": {
  "type": "array",
  "items": {
    "type": "string"
  }
}
```

and then `swagger-codegen` will generate such dto:
```
ModelDto {
 private List<String> list = new ArrayList<>();
 private SomeOtherModelDto object = null;

 //get set
}
```

How can we prevent this?

It is important for us because there is a difference between `null` list and empty list.

Thank you.

##### Swagger-codegen version

`io.swagger:swagger-codegen-cli:2.2.2`

##### Command line used for generation

`generate -i http://SOME_URL -l java -o bin/sdk --artifact-id sdk-name`

ePaul:
The OpenAPI Specification 2.0 (= Swagger Specification 2.0) doesn't allow defining anything as JSON `null`.
What it allows is to not make it mandatory to include a property in the containing object. This would be done by not listing it in the `required` list of the containing object.

When mapping this to Java, I would expect a receiver (client for response, server for request) to have the property as `null` when the value is not passed. A sender would simply not include the null value in the sent object.

I guess the initialization of the list was done to prevent null pointer exceptions when trying to access the object. I would argue that allowing a nullable list property (for non-required properties) would be the better way to go.

egormasalitin:
@ePaul Yes, it is main point, we don't need nulls in json, but this list(in my example) is not required, and json for deserialization just do not contain this field (with list) at all, but after deserialization it is (because of default empty array value) and for us it would be perfect to have nullable list (and Map) property.

egormasalitin:
Any suggestions? @wing328 maybe you can help me?

egormasalitin:
@wing328 I created solution for this: https://github.com/swagger-api/swagger-codegen/pull/5341

egormasalitin:
@fehguy Can you please review my pull?

cbornet:
I think it' not a good idea to add yet another option. I would set null by default.
An empty list could be generated when the field is required though.

egormasalitin:
@wing328 @fehguy @cbornet @ePaul 
Hello, I’ll try to explain why this patch is so important for us.

[...]

PR #5341, opened 2017-04-06
Add option to set default model values to null (Like `List`, `Map`)

egormasalitin:
### PR checklist

### Description of the PR

Issue: #5240
I added new additional param for java client coden, it will allow us to set all Model fields to null. This change is important for us, because `null` List and empty list is two different things.

### Why it is important?

For example you have api with partial update support (our case). It means that if user send json with just one field from model, API will update just this field. But current version of codegen set List and Map default values to empty list and empty map, it means that when user will send update request to our api using SDK, it will override list options as well.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConstants.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConstants.java
@@ -186,4 +186,6 @@ public class CodegenConstants {
     public static final String IGNORE_FILE_OVERRIDE = "ignoreFileOverride";
     public static final String IGNORE_FILE_OVERRIDE_DESC = "Specifies an override location for the .swagger-codegen-ignore file. Most useful on initial generation.";
 
+    public static final String MODEL_DEFAULT_NULL = "modelDefaultNull";
+    public static final String MODEL_DEFAULT_NULL_DESC = "Generate models with null as default value";
 }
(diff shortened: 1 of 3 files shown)


wing328:
Closed via https://github.com/swagger-api/swagger-codegen/pull/5363

jmiddleton:
Hi @wing328, is this PR coming in 2.2.3 ? is there any way I can get the snapshot version including this fix?

inthegarage:
@jmiddleton It is not in 2.2.3 and I don't see it in the snapshot either. We're stuck on 2.2.2 until it is incorporated - without avoiding major rework.

wing328:
The PR was not merged as we added the fix in the templates directly (without introducing additional option) via #5363

If you still encounter issue with the default value, please kindly open a new issue with the details (e.g. spec to reproduce the issue) and link back to this one.

inthegarage:
@wing328 I'm a little confused now, I agree the templates have been modified. However I ideally need this option described in this ticket (5341)  - IE the option to not set default value to null, in other words the functionality present in 2.2.2.  Using 2.2.3 means in effect there is regression as code which rightly assumes maps to be not null are now breaking.  How should we use this "defaultValue"? I don't see any reference to its usage, is a quick example possible? Obviously with Maps and Lists the default value for various classes could be all manner of things, hence my confusion.

wing328:
@inthegarage sorry if the change may have confused you. May I know if you've read the discussion in https://github.com/swagger-api/swagger-codegen/issues/5240?

Again I would suggest opening a new issue with an example to illustrate the issue so that other contributors can also more easily look into the issue and suggest better fixes.

inthegarage:
@wing328 That's explained it clearly, thank you. We can simply use the following construct in our yaml:

```
    required:
      - languages
```
As appropriate per model. This is a now a simple upgrade. We had already used it, however the code relied upon the auto-default (non null). This is a better, more controlled solution.

PR #5369, opened 2017-04-11
fix(swift3): decoders call parents decoders too

julienfouilhe:
### PR checklist

### Description of the PR

I was having problems with classes inheriting from other classes, fields defined in parent classes were `nil`. I fixed it by calling the parent's decoder in the child's decoder and passing the child's instance.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
@@ -197,7 +197,7 @@ open class AlamofireRequestBuilder<T>: RequestBuilder<T> {
                     return
                 }
                 if let json: Any = response.result.value {
-                    let body = Decoders.decode(clazz: T.self, source: json as AnyObject)
+                    let body = Decoders.decode(clazz: T.self, source: json as AnyObject, instance: nil)
                     completion(Response(response: response.response!, body: body), nil)
                     return
                 } else if "" is T {
--- a/modules/swagger-codegen/src/main/resources/swift3/Models.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift3/Models.mustache
@@ -37,17 +37,17 @@ open class Response<T> {
 
 private var once = Int()
 class Decoders {
-    static fileprivate var decoders = Dictionary<String, ((AnyObject) -> AnyObject)>()
+    static fileprivate var decoders = Dictionary<String, ((AnyObject, AnyObject?) -> AnyObject)>()
 
-    static func addDecoder<T>(clazz: T.Type, decoder: @escaping ((AnyObject) -> T)) {
+    static func addDecoder<T>(clazz: T.Type, decoder: @escaping ((AnyObject, AnyObject?) -> T)) {
         let key = "\(T.self)"
-        decoders[key] = { decoder($0) as AnyObject }
+        decoders[key] = { decoder($0, $1) as AnyObject }
     }
 
     static func decode<T>(clazz: T.Type, discriminator: String, source: AnyObject) -> T {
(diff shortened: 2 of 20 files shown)


wing328:
@julienfouilhe thanks for the PR.

cc @jaz-ah @Edubits @jgavris  @hexelon

jaz-ah:
@julienfouilhe could you regenerate the sample files and put in here as well?  I wanted to see what the output looks like now...

julienfouilhe:
Hi jaz-ah, it should be in the commit. Most of it is in Models.swift, and the diff is not displayed on github unless you click because it's too large. :)

jaz-ah:
@julienfouilhe ah I meant the actual sample output -i.e. the sample/test project checked into the tree...

julienfouilhe:
@jaz-ah I ran swift3-petstore-all.sh. Tell me if that's not what you needed

jaz-ah:
@julienfouilhe I must have missed it - I see it now - looks good - thx!  +1 @wing328

julienfouilhe:
@jaz-ah Yes I just ran them and amended my commit. ;) 

I don't think there's a test for that though. If you want to see where the bug was, you would need to add an endpoint to the Petstore API `/cats` returning `Cat` objects and not `Pet`. Without this commit, it wouldn't go through the `Pet` decoder, and thus it wouldn't fill fields described in `Pet`

wing328:
@julienfouilhe thanks for the PR.

wuf810:
I am double checking  but the changes in this PR have broken my swagger generated code .

This line in the template 

`_ = Decoders.decode(clazz: {{parent}}.self, source: source, instance: result) `

generates the following code for classes in the Models.swift file :

`_ = Decoders.decode(clazz: String.self, source: source, instance: result)`

As a result the decode function is looking for a class decoder named "String". 

NB The classes this fails on do use AdditionalProperties (which @danwoodel and I are currently fixing)…and I think this might be the issue as the current implementation of AdditionalProperties handling in the Swift3 codeine is completely wrong in that it tries to make the class a subclass of a string collection. This could be what {{parent}} is mistakenly returning.

Comments/thoughts anyone?

dwoodel:
I think it might be a good idea to add some models in the PetStore example with some additional properties.

julienfouilhe:
We should check if `decoders["\({{parent}}.self)"]` exists too. I should have done that.

wing328:
> I think it might be a good idea to add some models in the PetStore example with some additional properties.

We do have that in [petstore-with-fake-endpoints-models-for-testing.yaml](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml#L1084), which covers a lot more edge cases.

Currently, the Swift Petstore is using petstore.yaml (the basic one) and I'll create a task to track updating Swift Petstore to use [petstore-with-fake-endpoints-models-for-testing.yaml](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml#L1084)

wing328:
Created https://github.com/swagger-api/swagger-codegen/issues/5419 for tracking

wing328:
@dwoodel actually swift3 petstore is already using the fake petstore. Please feel free to submit a PR to add additional test cases to fake petstore so as to cover more edge cases.

wing328:
@julienfouilhe when you've time I wonder if you can make the same change against 2.3.0 branch instead as I got merge conflict when syncing the changes from master to 2.3.0. Thanks!

julienfouilhe:
@wing328 Done!

Issue #5385, opened 2017-04-13
Unsafe add/put calls in model classes


bbdouglas:
##### Description

PR #5363 (fix for issue #5240) introduced unsafe add/put methods on list/map fields. It used to be possible to call add/put on a list/map field without first setting it to an empty list. Now, if you do that, it throws a NullPointerException.

I find it unintuitive that you must call `set` before you can call `add`/`put`. It also breaks backwards compatibility with clients that before were safe just calling `add` or `put` without `set`.

##### Swagger-codegen version
Current
##### Swagger declaration file content or url
petstore
##### Command line used for generation
-l java
##### Steps to reproduce

If you call, for example, `Pet.addTagsItem(item)` in the petstore API without first calling `Pet.setTagsItem(new ArrayList<Tag>())` then you get a NPE.

##### Related issues

#5240

##### Suggest a Fix

I suggest that we add a null pointer check to add/put methods and instantiate the list/map if it is null. This maintains the spirit of issue #5240 and makes what I believe is a more intuitive and safe interface for building up lists/maps.

wing328:
@bbdouglas agree a null check can help avoid the NPE. I'll later review C# API client to see if we need similar fix.

wing328:
UPDATE: C# models do not have something similar that requires the same enhancement.

PR #5386, opened 2017-04-13
Add null pointer check to add/put methods (#5385)

bbdouglas:
Commit e3d04ee01 (issue #5240) introduced unsafe add/put methods for optional
list/map parameters. This change maintains the spirit of issue #5240 (optional
containers are null by default) while still making add/put calls safe. It does
this by checking for null first and, if so, initializing it with an empty
container.

Also updated the affected samples using the various scripts in bin/.

### PR checklist

### Description of the PR

A minor change to #5363 to check for nulls before adding/putting values into optional lists/maps.


bbdouglas:
Fix for #5385

wing328:
@bbdouglas thanks for the PR, which looks good to me.

PR #5390, opened 2017-04-13
Added code to templates to ensure correct generation of code for Swift3

vincent-coetzee:
Added code to templates to ensure correct generation of code for Swift3 so that it builds on both Apple platforms and Linux. Fixed a bug in the Model decoders that prevented code from building for Linux.

Currently the templates for AlamofireImplementations.swift, Extensions.swift and DefaultAPI.swift assume that the files are being compiled in the context of an iOS, macOS, watchOS or tvOS application, and consequently that Foundation has already been imported. This is not true in the case of applications being developed for Swift on Linux. Adding 
~~~~
import Foundation
~~~~
to the top of these files has no effect on Apple platforms, but enables the files to compile without errors in applications targeted for Linux. Additionally when files are compiled for Linux there is a limitation in the Swift compiler that causes the following code to be marked as an error

~~~~
    static func decode<T>(clazz: T.Type, source: AnyObject) -> T {
        initialize()
        if T.self is Int32.Type && source is NSNumber {
            return source.int32Value as! T;
        }
...
~~~~ 
In this case it is safe to force cast the source object to be an NSNumber since we know from the test in the line above that source is in fact an NSNumber. Therefore changing the code as shown below fixes the problem ( safely ).
~~~~
    static func decode<T>(clazz: T.Type, source: AnyObject) -> T {
        initialize()
        if T.self is Int32.Type && source is NSNumber {
            return (source as! NSNumber).int32Value as! T;
        }
...
~~~~ 

### PR checklist

### Description of the PR

### DIFF
--- a/modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
@@ -4,6 +4,7 @@
 // https://github.com/swagger-api/swagger-codegen
 //
 
+import Foundation
 import Alamofire
 
 class AlamofireRequestBuilderFactory: RequestBuilderFactory {
--- a/modules/swagger-codegen/src/main/resources/swift3/Extensions.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift3/Extensions.mustache
@@ -4,6 +4,7 @@
 // https://github.com/swagger-api/swagger-codegen
 //
 
+import Foundation
 import Alamofire{{#usePromiseKit}}
 import PromiseKit{{/usePromiseKit}}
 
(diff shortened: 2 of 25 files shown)


wing328:

wing328:
Please also run `./bin/swift3-petstore-all.sh` to update Swfit3 Petstore samples.

wing328:
cc @jaz-ah @Edubits @jgavris @hexelon for review

jgavris:
👍  LGTM

wing328:
@jgavris thanks for reviewing the change.

@vincent-coetzee thanks for the PR.

vincent-coetzee:
Glad to help.

Take care.

PR #5391, opened 2017-04-13
Feature: Enable validation

kenjones-cisco:
When useBeanValidation is enabled make sure the annotation @Valid
is added to the attributes of the model.

### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [x] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

(details of the change, additional tests that have been done, reference to the issue for tracking, etc)

Issue #5360, opened 2017-04-10
[JavaSpring] cascaded bean validation not implemented.


ebelanger:
##### Description

When activated, Hibernate Bean Validation only applies the first level properties.

##### Swagger-codegen version

2.2.0

##### Suggest a Fix

@Valid needs to be added on properties which contains RefProperty, Map of RefProperty and List of RefProperty.

wing328:
@ebelanger thanks for the suggestion. May I know if you've time to contribute the enhancement? If yes, I can show you some good starting points.

kenjones-cisco:
If we could just modify #4850 and simply make the following changes:

modules/swagger-codegen/src/main/resources/JavaSpring/beanValidation.mustache
append the following to the end of the file:
```
   @Valid
```

modules/swagger-codegen/src/main/resources/JavaSpring/model.mustache
insert after `{{#useBeanValidation}}`
```
import javax.validation.Valid;
```

I've done this previously (recently rebased) at https://github.com/kenjones-cisco/swagger-codegen/tree/features/validation

But since the existing PR has additional testing, the 2 extra lines being added to those templates should handle validation from API through Model.

wing328:
@kenjones-cisco I've merged #4850 into master. Please submit a PR on top of the latest master so that we can review the enhancement.

ebelanger:
@wing328 I just opened a PR with the fix.  Like I mentioned it only contains template modifications.  If you need more testing, I am open to suggestions.

PR #5393, opened 2017-04-13
[JavaSpring] add support for cascading bean validation

ebelanger:
add @Valid annotation when to model getter when type is a RefProperty or a collection containing RefProperty.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/JavaSpring/beanValidation.mustache
+++ b/modules/swagger-codegen/src/main/resources/JavaSpring/beanValidation.mustache
@@ -1,5 +1,6 @@
 {{#required}}
   @NotNull
-{{/required}}
+{{/required}}{{#isContainer}}{{^isPrimitiveType}}{{^isEnum}}
+  @Valid{{/isEnum}}{{/isPrimitiveType}}{{/isContainer}}{{#isNotContainer}}{{^isPrimitiveType}}
+  @Valid{{/isPrimitiveType}}{{/isNotContainer}}
 {{>beanValidationCore}}
-  @Valid


ebelanger:
#5360 

This should fix the problem without causing problems with primitive types since Hibernate would have runtime errors if an @Valid is added to an array for exemple.

There are no unit tests since it's only template modification.  I will modify the PetStore example if deemed required.

wing328:
@kenjones-cisco can you please review this PR when you've time? It seems like this PR will cover https://github.com/swagger-api/swagger-codegen/issues/5391 you filed.

kenjones-cisco:
@wing328 I'm trying to determine the difference between without the samples being regenerated.
With out digging further into the code about the impact of each of the conditionals added within the template, I'm not exactly sure if it will cover what was done within #5391.

wing328:
@kenjones-cisco thanks for taking a look. I think what I'll do is to merge both to ensure enhancements in both PRs are merged into master.

wing328:
@ebelanger when you've time, I wonder if you can rebase on the latest master to resolve the merge conflicts. Thanks!

ebelanger:
@wing328 ask and you shall receive :)  I pretty much add to undo what was done in master though so please notify me if it broke anything.

modules/swagger-codegen/src/main/resources/JavaSpring/beanValidation.mustache
```diff
@@ -1,5 +1,6 @@
 {{#required}}
   @NotNull
-{{/required}}
+{{/required}}{{#isContainer}}{{^isPrimitiveType}}{{^isEnum}}
```
kenjones-cisco: Can you explain the difference? Is there a reason not to include `@Valid` for any attribute on the bean?

I've been using it on all attributes in production for 3+ months now with no issues and validation has been enforced successfully.

So understand the difference and what attributes are not being validated and why would be helpful.

ebelanger: Sure.  I tried this approach first myself.  The problem comes from a property which returns an array.  @Valid validates objets, so Hibernate Validation casts to object during validation and a ClassCastException occurs.

This also removes some load on the Hibernate Validation process I am guessing since it doesn't need to visit all properties.

kenjones-cisco: There is the rub, the Hibernate part. I'd have to retest all my beans again to determine if there will be impact with this approach. I do not have any Hibernate backed beans.

Thanks for the information.

ebelanger: It's Hibernate Validator, which is an implementation of the Bean Validation.  Spring uses it for validation.  It is not to be mixed with Hibernate ORM, which is the persistence library.

wing328: @ebelanger how did the testing go? If the test result is good, I'll merge this PR into master.

ebelanger: @wing328 on my part I can safely say that it works as intended since we have deployed this fix in production for a while now :)

wing328:
@ebelanger PR merged into master. Thanks for the contribution.

modules/swagger-codegen/src/main/resources/JavaSpring/beanValidation.mustache
magictortoise: @ebelanger I try to use {{#isContainer}}{{^isPrimitiveType}}{{^isEnum}} in the Resteasy beanValidation.mustache template but it doesn't seem to be working (No @Valid adds on the string type of values). Add you saying array is not an object so that @Valid can't validate it?

ebelanger: @magictortoise like mentionned in a previous comment, validating an array with @Valid causes a ClassCastException in Hibernate Validation when it tries to cast it to an Object.

magictortoise: @ebelanger Do you know where I can learn to use tags like {{#isContainer}} and how can I debug if it is not working?

Issue #4307, opened 2016-12-02
Support maxProperties (and/or consider for oneOf semantics)


alechenninger:
##### Description

maxProperties does not appear to be used in swagger codegen.

##### Swagger-codegen version

commit: fdb6d30 (2.2.2-SNAPSHOT)

##### Suggest a Fix

maxProperties could be used to indicate only N fields should be non-null. Where n=1, this is "oneOf" semantics.

For example:

```
MyObject:
  type: object
  properties:
    thing1:
      $ref: '#/definitions/Thing1'
    thing2:
      $ref: '#/definitions/Thing2'
  maxProperties: 1
```

For objects with maxProperties, setters could validate at runtime whether the maxProperties count has been breached. In fact, this is similar to how gRPC generates code for protobuf oneof definitions, which unsets other fields if another among the "oneof" has been set.

wing328:
@alechenninger yup, it's not supported at the moment. We'll try to support it after the 2.2.2 release.

We also welcome PR to support `maxProperties`, `minProperties`

ebelanger:
@wing328 Here is another small contribution for minProperties and maxProperties support by reusing minItems and maxItems so no template change is required.

wing328:
Closed via #5395

PR #5395, opened 2017-04-13
[JavaSpring] add support for map minProperties and maxProperties bean…

ebelanger:
#4307

Add support for minProperties and maxProperties.  The change is deliberately re-using the minItems and maxItems of CodegenProperty so that all templates using those properties requires no changes.  This is considering JavaSpring only where bean validation is the same annotation.


wing328:
@ebelanger thanks for the enhancement, which looks good and consistent with `array` support for minProperties and maxProperties.

Issue #5375, opened 2017-04-12
Top-level enum definition (e.g. OuterEnum) is not processed properly in html doc and python client code generation


easz:
**test input:** 
   [test_enum.yaml](https://gist.github.com/easz/56ceba60bd06276c9891544e8b93dc7d)
   parameter is defined with a referenced type which references another string enum type.

**test with swagger-codegen 2.2.2:**
   - $ java -jar swagger-codegen-cli-2.2.2.jar generate -i [test_enum.yaml](https://gist.github.com/easz/56ceba60bd06276c9891544e8b93dc7d) -l python -o python-client
    and
   - $ java -jar swagger-codegen-cli-2.2.2.jar generate -i [test_enum.yaml](https://gist.github.com/easz/56ceba60bd06276c9891544e8b93dc7d) -l html -o doc-html

**test result:**
  - Without FOO, BAR enum information as expected
    - python-client/swagger_client/models/[an_enum_type.py](https://gist.github.com/easz/5dc0eb12d5113a29b388f1448cf3d72d)
    - doc-html/[index.html](https://gist.github.com/easz/e1a9d793c32e73532ea52452fa0fd3d0)

**note:**
   - java, javascript, html2 generation are OK

wing328:
@easz Python API client lacks support for enum at the moment. May I know if you've time to contribute the enhancement? I can show you some good starting points as we've done it in other languages: PHP, C#, Java, JS, etc

easz:
Sure,

easz:
@wing328  do you know what kind of template variables (e.g. isEnum, allowableValues ..etc) when codegen parses **outerEnum** **property** inside the **model** **Enum_Test**.
Apparently there is **!isEnum and !hasValidation**

It would be easier to fix it on the model property level as in top-level enum itself.

```
 definitions:
    Enum_Test:
      type: object
      properties:
        outerEnum:
          $ref: '#/definitions/OuterEnum'
    OuterEnum:
      type: "string"
      enum:
        - "placed"
        - "approved"
```

easz:
The actual model information of a referenced model is not used for template generation.

Namely the **RefModel** is at the moment not processed in **DefaultCodegen**.**fromModel** and will also not be processed properly during **DefaultGenerator.generateModels** 

The intuitive solution would be just to assign the actual Model for RefModel where the TODO is...
@wing328 does it fit to swagger design?


```
[...]

PR #5403, opened 2017-04-16
Issue 5375

easz:
### PR checklist

### Description of the PR

 - fix some existing unit test which was wrong
 - the validity of a property of a model will be checked in the constructor
 - add allowable enum values in python client code generation which can be further improved.
 - detail see #5375


modules/swagger-codegen/src/main/resources/python/model.mustache
```diff
@@ -34,7 +41,15 @@ class {{classname}}(object):
         }
 
 {{#vars}}
-        self._{{name}} = {{name}}
+        self._{{name}} = None
+{{/vars}}
+
+        # TODO: let required properties as mandatory parameter in the constructor.
+        #       - to check if required property is not None (e.g. by calling setter)
+        #       - ApiClient.__deserialize_model has to be adapted as well
```
wing328: @easz Thanks for the "TODO" note. 

Will we better off replacing ApiClient.__deserialize_model with [kim](https://github.com/mikeywaites/kim) or similar Python modules?

easz: I think the usual reflection technique is enough for that.

wing328: @easz thanks for the feedback. To be clear, I'm not saying we need to do it as part of this PR.  This is more for discussion on the future of the Python API client.

I've created https://github.com/swagger-api/swagger-codegen/issues/5421 for the discussion.

wing328:
@easz thanks for the enhancements (including validity of a property of a model in the constructor).

Issue #4907, opened 2017-03-03
[Swift3] Additional Properties in Swagger specs generating non compiling code


dwoodel:
##### Description
[Swift3] Additional Properties in Swagger specs generating non compiling code.  When using additionalProperties in the swagger specification. 

We are using String additonal properties on a model to add generic string fields.  This works fine in java where the model use inheritance to extend a HashMap.  In swift the primitive dictionary [String : String] is used.  This code will not compile, this cannot be done like this.  An NSDictionary could be used like this.  Unfortunately this is not the only problem.

##### Swagger-codegen version

Milestone release: 2.2.2 

Building the latest 2.2 branch.  Commit in question: 
6963bf874877aa8a009079a89845c5a3a06ebea7

Pull Request that seems to have broken the generator.
https://github.com/swagger-api/swagger-codegen/pull/4052

Already merged into master.

##### Swagger declaration file content or url

Example Yaml which generates the problem: 

```yaml
UserResponse:
    type: object
    properties:
      email:
        type: string
        example: jdoe@mail.com
        description: "User's email address"
      firstName:
        type: string
        example: John
        description: First name
      surname:
        type: string
        example: Doe
        description: Surname
      secondName:
        type: string
        description: Second name
      secondSurname:
        type: string
        description: Second surname
      birthday:
        type: string
        example: '1982-05-26'
        description: "User's birthday"
      username:
        type: string
        example: jdoe
        description: Name of user
      phoneNumber:
        type: string
        example: '+375291112233'
        description: "User's phone number"
      status:
        type: string
        example: REGISTERED
        description: "User's status"
    description: Response user model
    additionalProperties:
      type: string
```
 
##### Command line used for generation

generate --input-spec XXXX --lang swift3 --output ./ --artifact-id biid-ios-sdk-public-api-client --group-id com.biid --invoker-package /com.biid.ios.sdk.core.publicapi.client.gen --api-package /com.biid.ios.sdk.core.publicapi.client.api --model-package /com.biid.ios.sdk.core.publicapi.client.model --artifact-version 0.7.48

##### Steps to reproduce

Make a swagger file which contains additonal properties.  Generate the swift3 code.  Try to compile!!  Collect prize! 

##### Related issues

No releated

##### Suggest a Fix

wing328:
@dwoodel thanks for reporting the issue. Do you mind providing a screen shot of the generated code that has compilation errors?

cc @jaz-ah @Edubits @jgavris @hexelon

[...]

PR #5407, opened 2017-04-17
[swift3] Swift3 code generator & templates fixed to support additiona…

dwoodel:
…l properties

Fix for issue #4907 

Correct swift3 code generation when handling additional properties.  

### PR checklist

### Description of the PR

Modified templates
Fixed swift3 code generator
Added support for additional properties

### DIFF
--- a/modules/swagger-codegen/src/main/resources/swift3/Models.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift3/Models.mustache
@@ -183,7 +183,10 @@ class Decoders {
                 return Decoders.decode(clazz: {{classname}}.self, discriminator: discriminator, source: source)
             }
             {{/discriminator}}
-
+            {{#additionalPropertiesType}}
+            var propsDictionary = sourceDictionary
+            let keys : [AnyHashable] = [{{#allVars}}{{^-last}}"{{baseName}}", {{/-last}}{{#-last}}"{{baseName}}"{{/-last}}{{/allVars}}]
+            {{/additionalPropertiesType}}
             {{#unwrapRequired}}
             let result = {{classname}}({{#requiredVars}}{{^-first}}, {{/-first}}{{#isEnum}}{{name}}: {{classname}}.{{datatypeWithEnum}}(rawValue: (sourceDictionary["{{baseName}}"] as! {{datatype}}))! {{/isEnum}}{{^isEnum}}{{name}}: Decoders.decode(clazz: {{{baseType}}}.self, source: sourceDictionary["{{baseName}}"]! as AnyObject, instance: nil){{/isEnum}}{{/requiredVars}})
             {{#optionalVars}}{{#isEnum}}
@@ -208,6 +211,18 @@ class Decoders {
             }{{/isEnum}}
             {{^isEnum}}result.{{name}} = Decoders.decodeOptional(clazz: {{{baseType}}}.self, source: sourceDictionary["{{baseName}}"] as AnyObject?){{/isEnum}}{{/allVars}}
             {{/unwrapRequired}}
+            {{#additionalPropertiesType}}
+
+            for key in keys {
+                propsDictionary.removeValue(forKey: key)
+            }
+
+            for key in propsDictionary.keys {
+                if let decodedValue = Decoders.decodeOptional(clazz: String.self, source: propsDictionary[key] as AnyObject?) {
+                    result[key] = decodedValue
+                }
+            }
+            {{/additionalPropertiesType}}
             return result
 {{/allVars.isEmpty}}
 {{/isEnum}}
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Swift3Codegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Swift3Codegen.java
@@ -5,6 +5,7 @@ import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 import io.swagger.codegen.*;
 import io.swagger.models.Model;
+import io.swagger.models.ModelImpl;
 import io.swagger.models.Operation;
 import io.swagger.models.Swagger;
 import io.swagger.models.parameters.HeaderParameter;
@@ -60,6 +61,16 @@ public class Swift3Codegen extends DefaultCodegen implements CodegenConfig {
     @Override
     public String getHelp() {
         return "Generates a swift client library.";
+    }
+
+    @Override
+    protected void addAdditionPropertiesToCodeGenModel(CodegenModel codegenModel, ModelImpl swaggerModel) {
+
+        final Property additionalProperties = swaggerModel.getAdditionalProperties();
+
+        if(additionalProperties != null) {
+            codegenModel.additionalPropertiesType = getSwaggerType(additionalProperties);
+        }
     }
 
     public Swift3Codegen() {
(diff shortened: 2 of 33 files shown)


modules/swagger-codegen/src/main/resources/swift3/Models.mustache
```diff
@@ -202,6 +205,20 @@ class Decoders {
...
+            {{#parent}}
+            for key in keys {
+                propsDictionary.removeValue(forKey: key)
+            }
+
+            for key in propsDictionary.keys {
+                if let decodedValue = Decoders.decodeOptional(clazz: String.self, source: propsDictionary[key] as AnyObject?) {
+                    instance[key as! String] = decodedValue
+                } else {
+                    instance[key as! String] = nil
```
jgavris: Can you avoid using the force cast operator here?

dwoodel: Actually yes good point.  Im just headed out, ill update and run a test when I get back.

dwoodel: We wanted to keep propsDictionary the same as sourceDictionary (Its just a copy).  sourceDictionary uses AnyHashable.  So we need to cast to a string we could add more protection here but its not really needed.

jgavris: You really never want to use the force operators in Swift. How about `if let key = key as? String` ?

dwoodel: Hey @jgavris im happy to change it if you like. But.. 

The properies array is constructed from properies we already know that we have.  The Keys dictionary is a subset of the propertiesDictonary, there will never be a case where we have a key that is nil.    I think we can agree that the key is always going to be of base type string as its a dictonary.  However if your worried about nils.  A nil check here wont be the correct solution as the keys array is an array of none-optional keys.  Ref. line 188

`let keys : [String] = [{{#allVars}}{{^-last}}"{{baseName}}", {{/-last}}{{#-last}}"{{baseName}}"{{/-last}}{{/allVars}}]`

If your worried about nil's we should change this to be an optional string array.  

Again all this I think is unnecessary. 

However its your call.  Let me know.

jgavris: In my organization, we don't approve pull requests that use any of the force operators in Swift (`!` or `as!`). In general, why try to fight the type system? It's your friend. In this case, you're force casting the key to a string, which may get overlooked in future refactorings. It's not hard to make a change and I would be most appreciative :)

dwoodel:
Hang on I got some issue with merge conflicts let me just fix this.

dwoodel:
Looks like the code has move on since my fork.  Ill update that now.

dwoodel:
Looks like we got some breaking changes I have to update from.  Ill check this tomorrow. @jgavris shall we close this review and re-open another one?

jgavris:
I'd prefer if you could just rebase / force-push this review, I'm ok with that. Opening a new one loses all the discussion we've been having so far.

dwoodel:
Ok so me and @wuf810 have updated the PR and update the petstore for the CI to test.  Everything should be working now with the latest changes.

dwoodel:
@jgavris Looks good to me :)

wing328:
@dwoodel if no one has further feedback/comment on this PR, I'll merge later today (Monday).

cc @jaz-ah @Edubits @hexelon

dwoodel:
Im just going to remove some unnecessary line spaces in the templates (2 mins).

dwoodel:
Hey @wing328 yeah im happy now.  Diffs look much better 👍.  If it passes I think its good to go!

dwoodel:
Just looking at the latest inheritance changes I think this may break it.  Let me have a quick look.

dwoodel:
@jgavris @wing328 Dont merge this, we have discovered a potential issue.  We need 2 properties in the mustache templates along the lines of {{hasAdditionalProperties}} and {{additionalPropertyType}}.  We were using parent before you guys fixed it, because it was being used as the determination of whether or not you have additional properties. Now {{parent}} actually means inheritance, we need a separate qualifier.

dwoodel:
Just had a quick look in the Codegen model.  There is a property:

`public String additionalPropertiesType ` Which seems not to be set.  Iv tried setting it via the interface as it is in the java implementations.   Let me do some debugging.

dwoodel:
Hey @wing328 @jgavris, We have done as I said previously, we are now using the `additionalPropertiesType` in the mustache files. Inheritance and additionalProperties are now not coupled.  We have run a few tests with different YMAL files and everything seems to be fine.   I think you guys are ok to review now.

dwoodel:
👍 :1st_place_medal: 👍

wuf810:
Hey @wing328 any idea when this might be reviewed. Sorry to chase…but just interested in time scales.

thanks

jaz-ah:
+1 @wing328 lgtm

wing328:
@dwoodel thanks for the PR.

Also thanks @jaz-ah @jgavris for reviewing the enhancement.

wing328:
@dwoodel when you've time, I wonder if you can help resolve the following merge conflicts when sync'ing your enhancement to `2.3.0` branch (which contains breaking changes):
```
313                 {{#isEnum}}{{#isMapContainer}}*/ default: break //TODO: handle enum map scenario{{/isMapContainer}}{{/isEnum}}
314                 }
315             {{/allVars}}
316             {{/unwrapRequired}}
317 <<<<<<< HEAD
318                 return .success(result)
319             } else {
320                 return .failure(.typeMismatch(expected: "{{classname}}", actual: "\(source)"))
321             }
322 =======
323             {{#additionalPropertiesType}}
324 
325             for key in keys {
326                 propsDictionary.removeValue(forKey: key)
327             }
328 
329             for key in propsDictionary.keys {
330                 if let decodedValue = Decoders.decodeOptional(clazz: String.self, source: propsDictionary[key] as AnyObject?) {
331                     result[key] = decodedValue
332                 }
333             }
334             {{/additionalPropertiesType}}
335             return result
336 >>>>>>> origin/master
337 {{/allVars.isEmpty}}
338 {{/isEnum}}
339         }{{/isArrayModel}}{{/model}}
340         {{/models}}
341     }()
342 
343     static fileprivate func initialize() {
344         _ = Decoders.__once
345     }
```
modules/swagger-codegen/src/main/resources/swift3/Models.mustache

wing328:
UPDATE: I've tried to resolve the conflicts via 413a00e. Please review and let me know if you see any problems with the merge.

dwoodel:
@wing328 Hey sorry about the delay.  Ill solve these tonight.

wing328:
@dwoodel thanks for getting back to me. I believe I've resolved it and my merge should work. Please pull the latest 2.3.0 and give it a try.

dwoodel:
@wing328 Our contract that we have been using is not 2.3 compliant, so we need to update our model.  Also the callback architecture is different now, so I cant test end to end with our API.  However I have manually been through the edits that we made and I have looked at the generated models.  All the gen code that is to do with additional properties look good to me.

wuf810:
@wing328 just a follow up from what @dwoodel said. Does 2.3 require changes to the contract file structure or should we be seeing  valid code even with our existing contract file?

wuf810:
Hey @wing328 Just wondering if you could confirm the above ^. 

If this is not the case i.e. our exiting json contract should still work, then we are seeing some invalid generated swift code.

If this is the case I'll happily investigate but want to be sure before I do.

Thanks, M

wing328:
> Does 2.3 require changes to the json contract file or should we be seeing valid code generated even with our existing contract file?

@dwoodel Sorry for late reply. You should not need to change your swagger spec. We just merged https://github.com/swagger-api/swagger-codegen/pull/5609 into 2.3.0 to fix the model issue. Please pull the latest 2.3.0 to give it another try.

We're also working on setting up CIs for Swift, ObjC API clients to catch issues moving forward.

Issue #5382, opened 2017-04-13
Definitions Of Non-Aggegate Types Generate Empty Classes


bbdouglas:
##### Description

When you have a definition that is a simple type (such as "string", "number", "boolean") the class generated for it is empty. This makes the class unusable for serializing/deserializing since it holds no data.

##### Swagger-codegen version

Current

##### Swagger declaration file content or url

    definitions:
      JustABoolean:
        type: boolean

Becomes

    public class JustABoolean {
      // Note, no member variables, getters, setters, etc
      @Override
      public boolean equals(java.lang.Object o) {...}
      @Override
      public int hashCode() {...}
      @Override
      public String toString() {...}
      private String toIndentedString(java.lang.Object o) {...}  
    }

##### Command line used for generation

    -l java

##### Steps to reproduce

Any definition that isn't declared as type "object" or "array" will exhibit this behavior.

##### Related issues

I couldn't find a related issue.

##### Suggest a Fix

I can imagine two possible fixes.

* Create wrappers around all the non aggregate types. This might get a little tedious for each language/library, but since the number of types is small it is doable. The only downside I see is extra boilerplate.

    request.setMyJustABoolean(new JustABoolean(true))

* Replace the definition name with the underlying class name. That is, instead of creating a new JustABoolean class, use Boolean everywhere you would have used JustABoolean. This makes the client code a little cleaner, at the expense of making swagger-codegen potentially more complicated. It also obscures a bit of what's going on.

    request.setMyJustABoolean(true);

I thought about extending the simple types (Boolean, String, BigDecimal, etc), but that has its own problems. String, for example, is declared `final` and so can't be extended and BigDecimal can't be constructed directly from another BigDecimal :/

wing328:
Closed via #5408

Issue #5304, opened 2017-04-02
[CI][Typescript] nondeterministic failure on Travis CI of typescript builds with NPM


ePaul:
##### Description

The CI runs on Travis CI fail sometimes (non-deterministically) when doing typescript builds.

Example log extract (from the log of [a build of #5147](https://travis-ci.org/swagger-api/swagger-codegen/jobs/217776446)):
```
npm ERR! Linux 4.4.0-51-generic
npm ERR! argv "/home/travis/.nvm/versions/node/v4.1.2/bin/node" "/home/travis/.nvm/versions/node/v4.1.2/bin/npm" "install"
npm ERR! node v4.1.2
npm ERR! npm  v2.14.4

npm ERR! Callback called more than once.
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /home/travis/build/swagger-api/swagger-codegen/samples/client/petstore/typescript-angular/npm-debug.log
```

This time it was in the build of the TS Angular Petstore client, often it also happens with one of the other TypeScript builds. (You'll need to "view raw log" in order to see the actual error messages, but I'm linking to the pages of the builds because those link to the pull requests.)

* TS Fetch Petstore Test Client ([1](https://travis-ci.org/swagger-api/swagger-codegen/jobs/217575218), [2](https://travis-ci.org/swagger-api/swagger-codegen/jobs/217564077) [3](https://travis-ci.org/swagger-api/swagger-codegen/jobs/213653620))
* TS Angular Petstore Client ([1](https://travis-ci.org/swagger-api/swagger-codegen/jobs/215684567), [2](https://travis-ci.org/swagger-api/swagger-codegen/jobs/214915052))
* TS Fetch Default Petstore Client ([1](https://travis-ci.org/swagger-api/swagger-codegen/jobs/215676368), [2](https://travis-ci.org/swagger-api/swagger-codegen/jobs/215653286), [3](https://travis-ci.org/swagger-api/swagger-codegen/jobs/215299302), [4](https://travis-ci.org/swagger-api/swagger-codegen/jobs/215207107), [5](https://travis-ci.org/swagger-api/swagger-codegen/jobs/214958727), [6](https://travis-ci.org/swagger-api/swagger-codegen/jobs/214354344))
   * this one is the first of the TS builds, so I guess it is natural that it happens more often.

##### Swagger-codegen version

[...]

PR #5408, opened 2017-04-17
Add alias type definitions for Java

bbdouglas:
When a spec defines a Model at the top level that is a non-aggretate type (such
as string, number or boolean), it essentially represents an alias for the simple
type. For example, the following spec snippet creates an alias of the boolean
type that for all intents and purposes acts just like a regular boolean.

    definitions:
      JustABoolean:
        type: boolean

This can be modeled in some languages through built-in mechanisms, such as
typedefs in C++. Java, however, just not have a clean way of representing this.

This change introduces an internal mechanism for representing aliases. It
maintains a map in DefaultCodegen that tracks these types of definitions, and
wherever it sees the "JustABoolean" type in the spec, it generates code that
uses the built-in "Boolean" instead.

This functionality currenlty only applies to Java, but could be extended to
other languages later.

The change adds a few examples of this to the fake endpoint spec for testing,
which means all of the samples change as well.

### PR checklist

### Description of the PR

This is a fix to #5382. It creates an internal mapping between top-level, simple type definitions and their corresponding simple types. When code is generated, only the simple types are emitted, essentially erasing the type from the API.

I also investigated creating wrapper classes instead, to reduce confusion of the class names disappearing. However, the Java language makes this awkward, for example the fact that subclasses do not inherit constructors from their parents and lack of operator overloading. This makes the classes difficult to write in a generic way and bulky to use.

The change looks large, but actually the code change inside modules/ is relatively small and the rest is changes to the samples. The samples changed because I added a few testing endpoints to the sample yaml spec. The java feign client contains the built-out tests. The generated code is not affected if there are no top-level model definitions of simple types (which is by far the most common case). I re-generated the samples by running bin/run-all-petstore.


bbdouglas:
The latest CI issue seems to be a transient one. One of the two jobs passed, and the other one failed with this error:

    npm ERR! Linux 4.4.0-51-generic
    npm ERR! argv "/home/travis/.nvm/versions/node/v4.1.2/bin/node" 
    "/home/travis/.nvm/versions/node/v4.1.2/bin/npm" "install"
    npm ERR! node v4.1.2
    npm ERR! npm  v2.14.4
    npm ERR! code ECONNRESET
    npm ERR! errno ECONNRESET
    npm ERR! syscall read

    npm ERR! network read ECONNRESET
    npm ERR! network This is most likely not a problem with npm itself
    npm ERR! network and is related to network connectivity.
    npm ERR! network In most cases you are behind a proxy or have bad network settings.
    npm ERR! network 
    npm ERR! network If you are behind a proxy, please make sure that the
    npm ERR! network 'proxy' config is set properly.  See: 'npm help config'

wing328:
@bbdouglas thanks for the PR. The CI issue is tracked via https://github.com/swagger-api/swagger-codegen/issues/5304

cc @cbornet @ePaul for review.

cbornet:
It's not common to alias a model to a native type ! Is there really an interest of doing so ?

ePaul:
I didn't look at the code yet, but to answer @cbornet, I think there are two main use cases for a named model definition (i.e. under `definitions`) with a non object/array schema:

* Describing an enum – often strings, sometimes with numbers as values. I think this is handled by Swagger-Codegen, in the Java case by generating an `enum` class.
   (OpenAPI/JSON schema even allows mixed enums and enums with object values, I think this is not supported by Swagger Codegen yet.)
* Adding meta data (e.g. title, description) or restrictions (minimum, maximum, pattern, minLength, maxLength) to the primitive type. Of course there is not strictly a need to have this as a separate definition, you could as well inline it, but it is useful to help with the DRY principle if you have the same type of value in several places.

I didn't yet have a look if the latter is actually supported by this pull request, or if it just supports a simple aliasing.

bbdouglas:
Thank you for the review. I agree that the two main use cases are defining enums and adding metadata/validation to a primitive type. The enum case is already supported and this PR adds support for the second. It is this second case that I ran into in my work.

One could imagine scenarios such as numbers for person height and weight aliased as "HeightInCm" and "WeightInKg" with restrictions about being within a certain range. Or a string value such as "SKU" for product numbers, which are validated against a certain regex pattern. Once defined, they can be referenced in multiple places in the API.

bbdouglas:
It strikes me that this giant PR is 1% code change and 99% additional generated code for testing. Perhaps it would be easier to review if I split it into two PR's: one for the code change and the other for the samples generated from the new testing endpoint.

What do you think?

ePaul:
@bbdouglas I think changes of samples caused by code changes should be in the same PR.
On the other hand, adding new samples could be in a separate PR – if so, I would add those first with the old generation code, so we later can see the differences.

bbdouglas:
@ePaul That sounds good. I'll work on a separate PR for generating the new test endpoints with the old code, then update this PR to show the changes.

wing328:
If no one has further question or feedback, I'll take another look tomorrow (Thur) and merge accordingly.

Issue #5310, opened 2017-04-03
[Spring] Not able to provide examples in response-body


elizabetht:
##### Description

I am using the swagger-codegen in my gradle file to generate all the controllers & models using the petstore JSON file as input. Is there a way to include pre-defined responses that can be served up when running the generated server code in tomcat? When running the server code in tomcat locally, I only get an empty response body. I am looking to include predefined example/sample response body like a mock server.

##### Swagger-codegen version

2.2.1. 

##### Swagger declaration file content or url

https://gist.github.com/elizabetht/3cd961dae529089eaff789d04d382ad5

##### Command line used for generation

    language = 'spring'
    inputFile = file('src/main/resources/swagger.json')
    configFile = file('src/main/resources/config.json')

wing328:
@elizabetht looks like the Spring templates do not include any response example at the moment.
```
swagger-codegen|master⚡ ⇒ grep -R -i "example}}" modules/swagger-codegen/src/main/resources/JavaSpring/ 
modules/swagger-codegen/src/main/resources/JavaSpring//pojo.mustache:  @ApiModelProperty({{#example}}example = "{{example}}", {{/example}}{{#required}}required = {{required}}, {{/required}}{{#isReadOnly}}readOnly = {{{isReadOnly}}}, {{/isReadOnly}}value = "{{{description}}}")
```
(only POJO template contains examples defined in the spec)

May I know if you've time to contribute the enhancement? I can show you some good starting points.

elizabetht:
@wing328: yes please let me know. I can follow your help.

elizabetht:
To clarify, what I am looking for is to start up a tomcat with the war created using spring swagger code gen and get a predefined response when someone hits the endpoint from swagger. Now, I am able to startup tomcat with the war compiled using the generated files but getting an empty response when I hit the endpoint from swagger. The idea is for UI developers to go on with their development using this predefined response while the API services team is working on the actual implementation.

wing328:
@elizabetht thanks for offering help to add the example support.

We can use the the following mustache tags to add `examples` in the auto-generated code:
```
{{#examples}}
{{contentType}} // there should be a if-block to check content type
{{example}} // the example (string) can then be used as a response
{{/examples}}
```
You can add `-DdebugOperations=true` to inspect the examples defined in the spec. These values will be used by the mustache templates, e.g.
```
      "imports" : [ "Pet" ],
      "examples" : [ {
        "contentType" : "application/xml",
[...]

PR #5411, opened 2017-04-18
[Java][Spring] Add examples defined in the spec to Spring MVC server generator

wing328:
### PR checklist

### Description of the PR

For https://github.com/swagger-api/swagger-codegen/issues/5310.

Credits also go to @elizabett, @cbornet


wing328:
@kenjones-cisco because we need to determine the "Accept" in the HTTP request so as to return the response based on the format specified in "Accept".

Is it causing issue to your application?

kenjones-cisco:
@wing328 Thanks for the explanation.

wing328:
@kenjones-cisco I think about this a bit more and this is breaking change that should go to 2.3.0 (a branch with breaking changes) instead as the method signature has changed. (we want to avoid breaking changes to the current master)

I'm sorry for the inconvenience caused by this.

I'll file a PR to revert the change in the master and submit the change against 2.3.0 instead by coming Thursday.

cc @elizabett, @cbornet

wing328:
> I'll file a PR to revert the change in the master and submit the change against 2.3.0 instead by coming Thursday.

Due to busy schedule, I'll revert the change tomorrow instead.

wing328:
@kenjones-cisco the change has been rolled back via #5599 

I'll file a separate PR to make the enhancement in 2.3.0 branch.

wing328:
https://github.com/swagger-api/swagger-codegen/pull/5606 merged into 2.3.0 to provide the same enhancment to Spring generators.

@elizabetht @kenjones-cisco please kindly pull the latest master and 2.3.0 to give it a try.

wingsofovnia:
Hi @wing328 !
In my application, where I'm using codegen for generating api intefaces (only) dynamically during build, `@RequestHeader("Accept") String accept` is a breaking change. As far as I can see, it's only used in a Controller stub. Can we consider disabling this param for interfaceOnly config enabled?

wing328:
>  Can we consider disabling this param for interfaceOnly config enabled?

@wingsofovnia sounds reasonable to me. Please submit a PR so that we can review the change.

cbornet:
I think there is a better way : instead of using RequestHeader, the header can be retreived from ((ServletRequestAttributes)RequestContextHolder.getRequestAttributes()).getRequest().getHeader("Accept"). This way the API would be the same for all implementations and cleaner IMO.

cbornet:
And we can do it on the master since the change is non-breaking

PR #5416, opened 2017-04-18
fix(swift3): fix infinite loop with inheritance and check if parent decoder exists

julienfouilhe:
### PR checklist

### Description of the PR
My last commit (https://github.com/swagger-api/swagger-codegen/commit/b1a39ac820fa1e3920b1ff826797f6fd354921fe) had introduced an infinite loop if you were using `discriminator` *and* returning a class that has a parent in one of your routes. By ensuring the instance is `nil` before "going down" in the tree, we ensure that we're not already "going up", thus removing the infinite loop.

#### Steps to reproduce

In the petstore, add a route `/cats` returning `Cat` elements only. It will first go through the `Cat` decoder, that will call the `Animal` decoder, which has a discriminator on `className` and thus will call the `Cat` decoder again...

### DIFF
--- a/modules/swagger-codegen/src/main/resources/swift3/Models.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift3/Models.mustache
@@ -179,7 +179,7 @@ class Decoders {
             let sourceDictionary = source as! [AnyHashable: Any]
             {{#discriminator}}
             // Check discriminator to support inheritance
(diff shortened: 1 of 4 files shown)


julienfouilhe:
cc @wing328 @jaz-ah @Edubits @jgavris @hexelon

Issue #3069, opened 2016-06-07
[JavaScript] Invalid date parsing for all browsers. 


Danielku15:
##### Description

We are using the JavaScript code generator in combination with browserify. It brings us some benefits to have some "static code"-generation for JavaScript. The problem is that the date parsing function is not working for all browsers. Not all browsers have proper ISO-8601 parsing capabilities. Only Chrome parses the dates correctly. 

https://github.com/swagger-api/swagger-codegen/blob/8f258b9a485076333c2dcf358675ed113ecdb906/modules/swagger-codegen/src/main/resources/Javascript/ApiClient.mustache#L413

##### Swagger-codegen version

2.1.6

##### Swagger declaration file content or url

Any date-time parameter. e.g. http://petstore.swagger.io/#!/store/getOrderById

##### Command line used for generation

    java -jar swagger-codegen.jar generate -i http://petstore.swagger.io/v2/swagger.json -l JavaScript - o jsclient 
    cd jsclient
    npm install
    cd ..
    browserify jsclient\src\index.js -o client.js --standalone PetStoreClient

https://github.com/swagger-api/swagger-codegen/blob/8f258b9a485076333c2dcf358675ed113ecdb906/samples/client/petstore/javascript/src/ApiClient.js#L397

##### Steps to reproduce

1. Generate the client 
2. Use the client.js in the browser to make a call to `getOrderById` 
3. Inspect the shipDate

##### Related issues

No

##### Suggest a Fix

One solution could be to use moment.js to parse the date. An other solution might be to parse the ISO date manually with some regex. 
https://www.safaribooksonline.com/library/view/javascript-cookbook/9781449390211/ch03s05.html

wing328:
@Danielku15 thanks for reporting the issue with details and suggesting a few solutions to tackle the problem

From what I read (http://stackoverflow.com/questions/5802461/javascript-which-browsers-support-parsing-of-iso-8601-date-string-with-date-par), seems like the implementation is not consistent across browsers and therefore I prefer moment.js.

May I know if you've cycle to contribute the enhancements?

Ref: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

Danielku15:
[...]

PR #5417, opened 2017-04-19
typescript-angular2: treating dates as strings (datetimes are still Dates)

ibwhite:
### PR checklist

### Description of the PR

The current 2.3.0 typescript-angular2 treats both Swagger dates and datetimes as Javascript `Date`s and sends their `toISOString()` over the wire. This is reasonable behavior for datetimes, but for dates, which don't contain times, the generated client will send across meaningless times, in violation of the Swagger spec and breaking APIs that are expecting to receive dates.

Javascript/TypeScript does not have a native way to store dates (the native `Date` object only stores datetimes), so an ISO string (e.g. `"2017-04-19"` is the best way to represent this data.

This PR changes the type to `string` for dates, but leaves datetimes as `Date`s.

Feedback welcome and let me know if there's anything I can do to help.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngular2ClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngular2ClientCodegen.java
@@ -42,7 +42,7 @@ public class TypeScriptAngular2ClientCodegen extends AbstractTypeScriptClientCod
         modelTemplateFiles.put("model.mustache", ".ts");
         apiTemplateFiles.put("api.service.mustache", ".ts");
         languageSpecificPrimitives.add("Blob");
-        typeMapping.put("Date","Date");
+        typeMapping.put("date","string");
         typeMapping.put("file","Blob");
         apiPackage = "api";
         modelPackage = "model";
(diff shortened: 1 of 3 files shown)


wing328:
@ibwhite thanks for the PR. Would using [moment.js](https://momentjs.com/) help? Here is a discussion related to datetime in JS API client: https://github.com/swagger-api/swagger-codegen/issues/3069

ibwhite:
I like moment a lot, but I think it would be a shame to force users of the typescript-angular2 generator to use any specific library besides Angular itself. moment is very heavyweight and adds a significant overhead in performance and bundle size. A lot of people prefer not to use it for that reason, in favor of lighter/modular libraries like date-fns.

wing328:
@ibwhite I've added your suggestion (date-fns) to the discussion in https://github.com/swagger-api/swagger-codegen/issues/3069

I'm ok with the change. Shall we do the same for other TS generators (query, fetch, etc) in a separate PR?

ibwhite:
Thanks for the merge, @wing328. I think it would be great to make the other TS generators consistent with each other. If I get a chance this week, I can take a look at putting together a PR.

wing328:
@ibwhite thank you 👍

Issue #5355, opened 2017-04-09
[csharp-client] logging of requests


iamkarlson:
Add logger to the client: to the configuration, api client, etc.

##### Swagger-codegen version

last from the masteer

##### Suggest a Fix

create simple wrapper which always calls partial methods with same signature

wing328:
@iamkarlson sounds good to me. May I know if you've time to contribute the enhancement?

Please note that the latest C# generator in master supports API client with RestSharp or RestSharp.Portable.

iamkarlson:
@wing328 please review PR

PR #5424, opened 2017-04-19
Api client loggin

iamkarlson:
Add logs for most common operations. To implement logging, it's required to implement partial methods of Logger.cs
#5355 suggestion issue

### DIFF
--- a/modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
@@ -28,8 +28,7 @@ namespace {{packageName}}.Client {
     /// </summary>
     {{>visibility}} partial class ApiClient {
 
-        private static Logger log = new Logger(nameof(ApiClient));
-        private const string SERVER_DEFAULT_ADDRESS = "{{basePath}}";
+        private static Logger log = new Logger("ApiClient");
 
         private JsonSerializerSettings serializerSettings = new JsonSerializerSettings {
             ConstructorHandling = ConstructorHandling.AllowNonPublicDefaultConstructor
@@ -50,12 +49,13 @@ namespace {{packageName}}.Client {
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ApiClient" /> class
-        /// with default configuration and base path (<see cref="SERVER_DEFAULT_ADDRESS"/>).
+        /// with default configuration and base path (<see cref="{{basePath}}"/>).
         /// </summary>
-        public ApiClient() {
+        public ApiClient(
+        {
             log.Trace("start");
             Configuration = Configuration.Default;
-            RestClient = new RestClient(SERVER_DEFAULT_ADDRESS);
+            RestClient = new RestClient("{{basePath}}");
             {{#netStandard}}
             RestClient.IgnoreResponseStatusCode = true;
             {{/netStandard}}
@@ -64,10 +64,11 @@ namespace {{packageName}}.Client {
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ApiClient" /> class
-        /// with default base path (<see cref="SERVER_DEFAULT_ADDRESS"/>).
+        /// with default base path (<see cref="{{basePath}}"/>).
         /// </summary>
         /// <param name="config">An instance of Configuration.</param>
-        public ApiClient(Configuration config = null) {
+        public ApiClient(Configuration config = null
+        {
             log.Trace("start");
             if (config == null){
                 Configuration = Configuration.Default;
@@ -75,7 +76,7 @@ namespace {{packageName}}.Client {
                 Configuration = config;
             }
 
-            RestClient = new RestClient(SERVER_DEFAULT_ADDRESS);
+            RestClient = new RestClient("{{basePath}}");
             {{#netStandard}}
             RestClient.IgnoreResponseStatusCode = true;
             {{/netStandard}}
@@ -86,9 +87,11 @@ namespace {{packageName}}.Client {
         /// with default configuration.
         /// </summary>
         /// <param name="basePath">The base path.</param>
-        public ApiClient(string basePath = SERVER_DEFAULT_ADDRESS) {
+        public ApiClient(string basePath = "{{basePath}}"
+        {
             log.Trace("start");
-            if(string.IsNullOrEmpty(basePath)) {
+            if (string.IsNullOrEmpty(basePath))
+            {
                 throw new ArgumentException("basePath cannot be empty");
             } 
 
@@ -123,7 +126,8 @@ namespace {{packageName}}.Client {
             String path, {{^netStandard}}RestSharp.{{/netStandard}}Method method, Dictionary<String, String> queryParams, Object postBody,
             Dictionary<String, String> headerParams, Dictionary<String, String> formParams,
             Dictionary<String, FileParameter> fileParams, Dictionary<String, String> pathParams,
-            String contentType="application/json") {
+            String contentType="application/json")
+            {
             log.Trace("start");
             var request = new RestRequest(path, method);
 
@@ -144,7 +148,8 @@ namespace {{packageName}}.Client {
                 request.AddParameter(param.Key, param.Value);
 
             // add file parameter, if any
-            foreach(var param in fileParams) {
+            foreach(var param in fileParams)
+            {
                 {{#netStandard}}
                 request.AddFile(param.Value);
                 {{/netStandard}}
@@ -160,8 +165,16 @@ namespace {{packageName}}.Client {
                 {{/netStandard}}
             }
 
-            if (postBody != null) {
-                request.AddParameter(contentType, postBody, ParameterType.RequestBody);
+            if (postBody != null) // http body (model or byte[]) parameter
+            {
+                if (postBody.GetType() == typeof(String))
+                {
+                    request.AddParameter("application/json", postBody, ParameterType.RequestBody);
+                }
+                else if (postBody.GetType() == typeof(byte[]))
+                {
+                    request.AddParameter(contentType, postBody, ParameterType.RequestBody);
+                }
             }
             log.Trace("end");
             return request;
@@ -189,7 +202,8 @@ namespace {{packageName}}.Client {
             Dictionary<string, string> formParams,
             Dictionary<string, FileParameter> fileParams,
             Dictionary<string, string> pathParams,
-            string contentType) {
+            string contentType)
+            {
             log.Trace("start");
             var request = PrepareRequest(
                 path, method, queryParams, postBody, headerParams, formParams, fileParams,
@@ -242,7 +256,8 @@ namespace {{packageName}}.Client {
             Dictionary<string, string> formParams,
             Dictionary<string, FileParameter> fileParams,
             Dictionary<string, string> pathParams,
-            string contentType) {
+            string contentType)
+            {
             log.Trace("start");
             var request = PrepareRequest(
                 path, method, queryParams, postBody, headerParams, formParams, fileParams,
@@ -260,7 +275,8 @@ namespace {{packageName}}.Client {
         /// </summary>
         /// <param name="str">String to be escaped.</param>
         /// <returns>Escaped string.</returns>
-        public string EscapeString(string str) {
+        public string EscapeString(string str)
+        {
             return UrlEncode(str);
         }
 
@@ -270,9 +286,11 @@ namespace {{packageName}}.Client {
         /// <param name="name">Parameter name.</param>
         /// <param name="stream">Input stream.</param>
         /// <returns>FileParameter.</returns>
-        public FileParameter ParameterToFile(string name, Stream stream) {
+        public FileParameter ParameterToFile(string name, Stream stream)
+        {
             log.Trace("start");
-            if (stream is FileStream) {
+            if (stream is FileStream)
+            {
                 log.Debug("FileStream");
                 return FileParameter.Create(name, ReadAsBytes(stream), Path.GetFileName(((FileStream)stream).Name));
             } else {
@@ -287,23 +305,28 @@ namespace {{packageName}}.Client {
         /// </summary>
         /// <param name="obj">The parameter (header, path, query, form).</param>
         /// <returns>Formatted string.</returns>
-        public string ParameterToString(object obj) {
+        public string ParameterToString(object obj)
+        {
             log.Trace("start");
-            if (obj is DateTime) {
+            if (obj is DateTime)
+            {
                 // Return a formatted date string - Can be customized with Configuration.DateTimeFormat
                 // Defaults to an ISO 8601, using the known as a Round-trip date/time pattern ("o")
                 // https://msdn.microsoft.com/en-us/library/az4se3k1(v=vs.110).aspx#Anchor_8
                 // For example: 2009-06-15T13:45:30.0000000
-                log.Debug("obj is "+nameof(DateTime));
+                log.Debug("obj is DateTime");
                 return ((DateTime)obj).ToString (Configuration.DateTimeFormat);
-            } else if (obj is DateTimeOffset) {
+            }
+            else if (obj is DateTimeOffset)
+            {
                 // Return a formatted date string - Can be customized with Configuration.DateTimeFormat
                 // Defaults to an ISO 8601, using the known as a Round-trip date/time pattern ("o")
                 // https://msdn.microsoft.com/en-us/library/az4se3k1(v=vs.110).aspx#Anchor_8
                 // For example: 2009-06-15T13:45:30.0000000
-                log.Debug("obj is "+nameof(DateTimeOffset));
+                log.Debug("obj is DateTimeOffset");
                 return ((DateTimeOffset)obj).ToString (Configuration.DateTimeFormat);
-            } else if (obj is IList) {
+            } else if (obj is IList)
+            {
                 var flattenedString = new StringBuilder();
                 foreach (var param in (IList)obj)
                 {
@@ -311,7 +334,7 @@ namespace {{packageName}}.Client {
                         flattenedString.Append(",");
                     flattenedString.Append(param);
                 }
-                log.Debug("obj is "+nameof(IList));
+                log.Debug("obj is IList");
                 return flattenedString.ToString();
             } else {
                 log.Trace("end");
@@ -325,24 +348,30 @@ namespace {{packageName}}.Client {
         /// <param name="response">The HTTP response.</param>
         /// <param name="type">Object type.</param>
         /// <returns>Object representation of the JSON string.</returns>
-        public object Deserialize(IRestResponse response, Type type) {
+        public object Deserialize(IRestResponse response, Type type
+        {
             log.Trace("start");
             {{^netStandard}}IList<Parameter>{{/netStandard}}{{#netStandard}}IHttpHeaders{{/netStandard}} headers = response.Headers;
-            if (type == typeof(byte[])) {
+            if (type == typeof(byte[])
+            {
                 // return byte array
                 log.Trace("end");
                 return response.RawBytes;
             }
 
-            if (type == typeof(Stream)) {
-                if (headers != null) {
+            if (type == typeof(Stream)
+            {
+                if (headers != null
+                {
                     var filePath = String.IsNullOrEmpty(Configuration.TempFolderPath)
                         ? Path.GetTempPath()
                         : Configuration.TempFolderPath;
                     var regex = new Regex(@"Content-Disposition=.*filename=['""]?([^'""\s]+)['""]?$");
-                    foreach (var header in headers) {
+                    foreach (var header in headers
+                    {
                         var match = regex.Match(header.ToString());
-                        if (match.Success) {
+                        if (match.Success
+                        {
                             string fileName = filePath + SanitizeFilename(match.Groups[1].Value.Replace("\"", "").Replace("'", ""));
                             File.WriteAllBytes(fileName, response.RawBytes);
                             return new FileStream(fileName, FileMode.Open);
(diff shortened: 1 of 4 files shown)


modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -22,15 +22,16 @@ using RestSharp.Portable.HttpClient;
...
     /// API client is mainly responsible for making the HTTP call to the API backend.
     /// </summary>
-    {{>visibility}} partial class ApiClient
-    {
-        private JsonSerializerSettings serializerSettings = new JsonSerializerSettings
-        {
+    {{>visibility}} partial class ApiClient {
+
+        private static Logger log = new Logger(nameof(ApiClient));
+        private const string SERVER_DEFAULT_ADDRESS = "{{basePath}}";
```
wing328: @iamkarlson for `basePath`, please use {{{basePath}}} instead so as to use the unescaped (original) value.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -49,30 +50,32 @@ namespace {{packageName}}.Client
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ApiClient" /> class
-        /// with default configuration and base path ({{{basePath}}}).
+        /// with default configuration and base path (<see cref="SERVER_DEFAULT_ADDRESS"/>).
         /// </summary>
-        public ApiClient()
-        {
+        public ApiClient() {
```
wing328: @iamkarlson `{` should start with a new line so as to conform to C# style guide.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -83,10 +86,11 @@ namespace {{packageName}}.Client
         /// with default configuration.
         /// </summary>
         /// <param name="basePath">The base path.</param>
-        public ApiClient(String basePath = "{{{basePath}}}")
-        {
-           if (String.IsNullOrEmpty(basePath))
+        public ApiClient(string basePath = SERVER_DEFAULT_ADDRESS) {
+            log.Trace("start");
+            if(string.IsNullOrEmpty(basePath)) {
```
wing328: @iamkarlson please add a space after `if`

wing328:
@iamkarlson thanks for the PR. I've left some comments. Please review when you've time.

Ref: C# style guide - https://msdn.microsoft.com/en-us/library/ff926074.aspx

cc @jimschubert

wing328:
@iamkarlson I've added some batch files for updating C# Petstore samples.

Please rebase on the latest master and run `bin\windows\csharp-petstore-all.bat` to update the Petstore samples so that CI (Appveyor) can verify the change.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -22,15 +22,16 @@ using RestSharp.Portable.HttpClient;
...
     /// <summary>
     /// API client is mainly responsible for making the HTTP call to the API backend.
     /// </summary>
-    {{>visibility}} partial class ApiClient
-    {
-        private JsonSerializerSettings serializerSettings = new JsonSerializerSettings
-        {
+    {{>visibility}} partial class ApiClient {
+
+        private static Logger log = new Logger(nameof(ApiClient));
```
jimschubert: `nameof` is not compatible with all supported .NET versions here. This should just hardcode the `ApiClient` string as that will never change, and if it does it's a breaking change.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -49,30 +50,32 @@ namespace {{packageName}}.Client
...
             {{#netStandard}}
             RestClient.IgnoreResponseStatusCode = true;
             {{/netStandard}}
+            log.Trace("end");
         }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ApiClient" /> class
-        /// with default base path ({{{basePath}}}).
+        /// with default base path (<see cref="SERVER_DEFAULT_ADDRESS"/>).
```
jimschubert: `SERVER_DEFAULT_ADDRESS` is private, so there's nothing to *see* if the client generates docs.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -49,30 +50,32 @@ namespace {{packageName}}.Client
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ApiClient" /> class
-        /// with default configuration and base path ({{{basePath}}}).
+        /// with default configuration and base path (<see cref="SERVER_DEFAULT_ADDRESS"/>).
         /// </summary>
-        public ApiClient()
-        {
+        public ApiClient() {
+            log.Trace("start");
```
jimschubert: Don't these `start` and `end` trace messages make assumptions about how the Logger class is implemented? For example, if I were to implement logger to write message directly to a file in the format `[LEVEL] MSG`, wouldn't this just log `[TRACE] start`? If this is forcing the implementation to pull the method name off the stack, I think that would be a negative performance hit if logging is enabled.

I think it would be more helpful to embed the method name being wrapped here.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -157,18 +160,10 @@ namespace {{packageName}}.Client
                 {{/netStandard}}
             }
 
-            if (postBody != null) // http body (model or byte[]) parameter
-            {
-                if (postBody.GetType() == typeof(String))
-                {
-                    request.AddParameter("application/json", postBody, ParameterType.RequestBody);
-                }
-                else if (postBody.GetType() == typeof(byte[]))
```
jimschubert: Removing this will break support for File inputs.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -262,12 +270,14 @@ namespace {{packageName}}.Client
         /// <param name="name">Parameter name.</param>
         /// <param name="stream">Input stream.</param>
         /// <returns>FileParameter.</returns>
-        public FileParameter ParameterToFile(string name, Stream stream)
-        {
-            if (stream is FileStream)
+        public FileParameter ParameterToFile(string name, Stream stream) {
+            log.Trace("start");
+            if (stream is FileStream) {
+                log.Debug("FileStream");
```
jimschubert: The word `FileStream` here wouldn't be useful in logs unless it says what method it is called from, and more about what the log message means.

Because log messages exist in asynchronously executed methods and can be called asynchronously by a client, you could end up with a log output like:

```
…
start
end
start
end
end
FileStream
…
```

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -312,54 +325,73 @@ namespace {{packageName}}.Client
...
+            if(!(deserialize(response, type) is T)) {
+                return default(T);
             }
-            catch (Exception e)
-            {
+
+            return (T) deserialize(response, type);
+        }
+
+        private object deserialize(IRestResponse response,Type type) {
```
jimschubert: `deserialize` should be capitalized to follow C# coding standards. Since this would conflict with the existing `Deserialize` method, I'd recommend calling it `DeserializeObject` to match the `JsonConvert` method being abstracted.

modules/swagger-codegen/src/main/resources/csharp/Logger.mustache
```diff
@@ -0,0 +1,55 @@
...
+            Initialize(loggerName);
+        }
+
+        partial void Initialize(string loggerName);
+
+        public void Trace(string message) {
+            TracePartial(message);
+        }
+
+        partial void TracePartial(string message);
```
jimschubert: I don't understand why public method are passed through to partial methods of the same name suffixed with the method being "partialed".  Why not just use partials:

```
partial void Trace(string message);
```

This has the added benefit of allowing the compiler to optimize away unimplemented trace method calls if the implementor hasn't provided an implementation. The way this class is currently implemented, I believe all calls are executed as noops if the implementor hasn't provided the *Partial implementations.

I don't think the Logger class needs to be public, if that's a factor into the design here.

modules/swagger-codegen/src/main/resources/csharp/api.mustache
```diff
@@ -80,6 +80,8 @@ namespace {{packageName}}.{{apiPackage}}
     /// </summary>
     {{>visibility}} partial class {{classname}} : {{interfacePrefix}}{{classname}}
     {
+
+        private static Logger log = new Logger(nameof({{classname}}));
```
jimschubert: `Logger` needs to be fully qualified here to avoid conflicts with models named `Logger`. Also, I don't know if we'd want to use `nameof` as this is limiting to the compiler the clients may use. I'd consider this a pretty big breaking change.

jimschubert:
I've added comments, trying to keep it to a single comment for repeated instances of the same thing.

There is at least one breaking change (removing `byte[]` response support), and multiple usages of `nameof` which as far as I know isn't available in .NET 3.5 or .NET 4 (both are supported by these same templates).

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
iamkarlson: yep, you're right. but if you goin' to rename this class... doesn't matter, I'll change.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
iamkarlson: Actually, I don't think so. As far as I know this is a common practice to implement your logging with pulling method name from the logger static object, not from the stack. For example, NLog specification with static declaration https://github.com/NLog/NLog/wiki/Tutorial#creating-loggers

iamkarlson:
@jimschubert _nameof_ is a feature of C#, not of .NET.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
jimschubert: @iamkarlson I'm not sure how that link was meant to support your comment. In NLog, this is the relevant source for pulling the method name from stack: https://github.com/NLog/NLog/blob/31fdc39f3ecee901ed359f7cc157244cf0a1f538/src/NLog/LoggerImpl.cs#L77

What you're saying is that implementors should implement this same logic to get the method name from the stack.

The problem is that if I were to implement your logging class as a pass-through to NLog, for example:

```
public partial class Logger 
{
    private static NLog.Logger logger = NLog.LogManager.GetCurrentClassLogger();
   
    partial void TracePartial(string message)
    {
        logger.Trace(message);
    }
}
```

The method name seen by NLog would be TracePartial, not whatever method is calling it.

jimschubert:
@iamkarlson Let me restate the last comment.

`nameof` is a feature of C# 6.0 and higher. When a user installs the .NET Framework 3.5, 4.0, or 4.5, they haven't installed the C# 6.0 compiler. I'm fairly certain the compiler installed for each is C# 3.0, C# 4.0, and C# 5.0, respectively. I'm not sure if Microsoft has released updates for those installs to use the C# 6.0 compiler by default.

Because this generator supports generating code to target the .NET Framework 3.5 and 4.5, we can't assume the users with only those framweorks installed (and thus older versions of the C# compiler) to have the C# 6.0 compiler available. 

For example, it's possible we have users generating these clients through the web interface and compiling on a clean Windows 7 image. Using `nameof` here could be considered a breaking change for such a user.

I've tested this in Mono 3.10 and Mono 4.0 docker images... `nameof` seems to be supported there. Unfortunately, my Windows VM won't start at the moment so I can't verify on a clean .NET 3.5 or .NET 4.5 install.

I'm personally fine with using `nameof`. I'd actually prefer it to hardcoded strings. I'm just not sure it'll be supported by our user base.

@wing328 thoughts?

jimschubert:
@iamkarlson on another note. Is there a driving factor to create a `Logger` abstraction rather than using something like `Trace.WriteLine` or `Debug.WriteLine`? I understand these rely on compiler flags for conditional generation which could be annoying, but they also allow for multiple listeners without the need for an extra level of logging abstraction.

I haven't done any investigation into whether those are supported by uwp or how Debug.WriteLine listeners could be used in .NET Core (see http://stackoverflow.com/a/37897902/151445).

I was just wondering if there was an argument against those for application level logging within the client.

wing328:
> Because this generator supports generating code to target the .NET Framework 3.5 and 4.5, we can't assume the users with only those frameworks installed (and thus older versions of the C# compiler) to have the C# 6.0 compiler available.

@jimschubert good catch. Definitely valid point.

> I'd actually prefer it to hardcoded strings

Me too

jimschubert:
@wing328 would you be cool with using nameof as a breaking change and require C# 6.0 going forward?

wing328:
@jimschubert I believe there're still lots of developers (especially those working on enterprise such as banks, insurance companies, etc) using older version of C# so I think we should avoid using `nameof`.

This is not to say `nameof` nor C# 6.0 is bad but just that we want the API clients to have better compatibility of with different version of C#.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -499,18 +532,16 @@ namespace {{packageName}}.Client
...
-            {
+            if (match.Success) {
                 return match.Groups[1].Value;
-            }
-            else
-            {
+            } else {
                 return filename;
             }
+            log.Trace("end");
```
jimschubert: This is dead code. The trace needs to be duplicated in each block above, or the if/else would need to be rewritten to store in a variable and return after this statement.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
iamkarlson: Could you explain why? I thought that is completely normal to use only one place to store strings such that. In case my URL is changed, I should go step by step and replace it manually.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
iamkarlson: But you have a log time usually which allows you to sort out all of this mess and get the proper order, don't you?

modules/swagger-codegen/src/main/resources/csharp/Logger.mustache
iamkarlson: Definitely, it was something that I tried to workaround :)  I will try to find why I did so and let you know.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
jimschubert: Again, you're making assumptions about the logging implementation.

To put my comment more into perspective, logging implementations (log4net, for example) grab the executing method name by getting a reference to the execution stack, going back one entry, and getting the method name. I haven't looked at your recent changes, but with the `Debug` method deferring to a `DebugPartial` someone using a common implementation of getting a method name would end up getting "Debug". So the message would be something like:

```
Debug: FileStream
```

That's easy enough to fix, obviously.

But, if I just wanted to implement `DebugPartial` as:

```
Console.WriteLine(message);
```

There's no timestamp here. If the message was `ParameterToFile: FileStream`, I would think that'd be less of an issue. But with just `FileStream`, it's not telling the logger or the human reading the logs anything about the log message and this makes the message less useful.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
jimschubert: @iamkarlson He's not saying to get rid of the constant. In mustache, `{{{basePath}}}` with three braces tells the code to use the raw input. `{{basePath}}` with two braces tells mustache to HTML-escape the string, and this can cause issues with URIs.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -312,26 +348,29 @@ namespace {{packageName}}.Client
...
             {
-                if (headers != null)
+                if (headers != null
                 {
                     var filePath = String.IsNullOrEmpty(Configuration.TempFolderPath)
                         ? Path.GetTempPath()
                         : Configuration.TempFolderPath;
                     var regex = new Regex(@"Content-Disposition=.*filename=['""]?([^'""\s]+)['""]?$");
-                    foreach (var header in headers)
+                    foreach (var header in headers
```
jimschubert: You're missing a closing parens here and on line 373 below this.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -312,26 +348,29 @@ namespace {{packageName}}.Client
...
+                // return byte array
+                log.Trace("end");
                 return response.RawBytes;
             }
 
-            if (type == typeof(Stream))
+            if (type == typeof(Stream)
             {
-                if (headers != null)
+                if (headers != null
```
jimschubert: Missing closing parens.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -312,26 +348,29 @@ namespace {{packageName}}.Client
         /// <param name="response">The HTTP response.</param>
         /// <param name="type">Object type.</param>
         /// <returns>Object representation of the JSON string.</returns>
-        public object Deserialize(IRestResponse response, Type type)
+        public object Deserialize(IRestResponse response, Type type
```
jimschubert: Missing closing parens.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -340,26 +379,55 @@ namespace {{packageName}}.Client
...
+        public T Deserialize<T>(IRestResponse response
+        {
+            log.Trace("start");
+            var type = typeof(T);
             // at this point, it must be a model (json)
-            try
+            if (!(DeserializeObject(response, type) is T)
             {
-                return JsonConvert.DeserializeObject(response.Content, type, serializerSettings);
+                return default(T);
```
jimschubert: This is not standard behavior, and would be a breaking change.

Why not just make `DeserializeObject` generic, and perform the cast in that method? Then, you'll automatically send a 500, and you can catch the cast exception in that method if you want to expose some 4xx message.

I don't feel like we're really gaining much by changing the simple try/catch to a reusable method. Am I missing an additional usage in your PR or something?

jimschubert: To be clear, this isn't standard behavior, because if my code is supposed to deserialize `Cat`, but it attempts to deserialize `Dog` then your new code is swallowing an obvious issue. We'd want to either return a 500 message here (meaning the server screwed up somewhere), or possibly a 400 Bad Request.  I don't think we can easily assume response code 400 for this scenario, so we should probably stick with 500.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -340,26 +379,55 @@ namespace {{packageName}}.Client
...
             }
-            catch (Exception e)
+
+            return (T) DeserializeObject(response, type);
+        }
+
+        private object DeserializeObject(IRestResponse response,Type type
```
jimschubert: missing closing parens.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -340,26 +379,55 @@ namespace {{packageName}}.Client
...
+        {
+            try {
+                return JsonConvert.DeserializeObject(response.Content,type, serializerSettings);
+            }
+            catch(Exception e
```
jimschubert: missing closing parens.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -369,14 +437,18 @@ namespace {{packageName}}.Client
         /// </summary>
         /// <param name="obj">Object.</param>
         /// <returns>JSON string.</returns>
-        public String Serialize(object obj)
+        public String Serialize(object obj
```
jimschubert: missing closing parens.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -388,13 +460,18 @@ namespace {{packageName}}.Client
         /// </summary>
         /// <param name="contentTypes">The Content-Type array to select from.</param>
         /// <returns>The Content-Type header to use.</returns>
-        public String SelectHeaderContentType(String[] contentTypes)
+        public String SelectHeaderContentType(String[] contentTypes
```
jimschubert: missing closing parens.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -406,13 +483,17 @@ namespace {{packageName}}.Client
         /// </summary>
         /// <param name="accepts">The accepts array to select from.</param>
         /// <returns>The Accept header to use.</returns>
-        public String SelectHeaderAccept(String[] accepts)
+        public String SelectHeaderAccept(String[] accepts
```
jimschubert: missing closing parens.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -422,7 +503,7 @@ namespace {{packageName}}.Client
         /// </summary>
         /// <param name="text">String to be encoded.</param>
         /// <returns>Encoded string.</returns>
-        public static string Base64Encode(string text)
+        public static string Base64Encode(string text
```
jimschubert: missing closing parens.

modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
```diff
@@ -434,8 +515,7 @@ namespace {{packageName}}.Client
         /// <param name="source">Object to be casted</param>
         /// <param name="dest">Target type</param>
         /// <returns>Casted object</returns>
-        {{#supportsAsync}}public static dynamic ConvertType(dynamic source, Type dest){{/supportsAsync}}{{^supportsAsync}}public static object ConvertType<T>(T source, Type dest) where T : class{{/supportsAsync}}
-        {
+        {{#supportsAsync}}public static dynamic ConvertType(dynamic source, Type dest){{/supportsAsync}}{{^supportsAsync}}public static object ConvertType<T>(T source, Type dest) where T : class{{/supportsAsync}} {
```
jimschubert: Curly brace at the end of this line should be on the following line.

iamkarlson:
I suppose this is better to close this PR, isn't it?

wing328:
Closing this one as it's been replaced by https://github.com/swagger-api/swagger-codegen/issues/5689

PR #5439, opened 2017-04-20
The new .NET project file format support

nikitozz:
### Description of the PR

Adding support for the new .NET project files format (introduced in VS 2017, https://docs.microsoft.com/en-us/dotnet/articles/core/tools/project-json-to-csproj) for generated C# clients.

The feature is disabled by default and can be enabled by using `netCoreProjectFile` flag when generating a C# client. For example: `java -jar ./modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -i modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml -l csharp -o samples/client/petstore/csharp/SwaggerClientNetCoreProject --additional-properties netCoreProjectFile=true`.

New sample scripts are added and a new sample is generated.


wing328:
@nikitozz thanks for the enhancement. We'll review and let you know if we've any question.

cc @jimschubert

bin/csharp-petstore-netcore-project.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
```
jimschubert: This file isn't executable.

Need to run:

```
chmod u+x bin/csharp-petstore-netcore-project.sh
```

jimschubert:
I don't have VS 2017, so I can't verify this in the IDE right now. I did build and generate locally, then ran `dotnet restore` followed by `dotnet build` and things seem to be as expected.

bin/csharp-petstore-netcore-project.sh
wing328: I'll take care of that.

wing328:

@nikitozz thanks for the contribution.

wing328:
FYI. Made 4 shell scripts executable via 1cd8341

Issue #5074, opened 2017-03-15
Tag names, descriptions are not available in Mustache templates


DavidBiesack:
##### Description

The actual name and description of the tags in an OpenAPI document is not preserved in the variables available to templates. Only the *identifier* form is carried over. If a tag contains blanks, they are lost. So it is impossible to use swagger-codegen to create labels for operation tags for -l html such that the tags are rendered as in swagger-ui

Further, swagger-codegen organizes operations by tags ( `{{apis}}` ) so that if an operation has multiple tags, the operation is repeated in the output.

##### Swagger-codegen version

2.2.2-SNAPSHOT

##### Swagger declaration file content or url

[modified-petstore-swagger.json](https://gist.githubusercontent.com/DavidBiesack/70d2910322d5c68d730878f0d8c36587/raw/6b93d6450a74ea08bd31ef0250240e4dcd11a083/modified-petstore-swagger.json?_sm_au_=iVVDq7n0kPL7s5tR) is a simplified sub of petstore, but with the tags renamed to "pet" -> "Personal Pet", "store" -> "Pet Store" and "user" -> "Pet Owner" 

The operation 
```
    "/store/inventory": {
      "get": {
        "tags": [
            "Pet Store", "Personal Pets"
        ],

```

is tagged with two tags (line 271 of the gist), which is meant to indicate that the operation should be groups with both "Pet Store" and "Personal Pets". 

This yields labels for the goups in the html2 navigation (unfortunately ALLCAPS style used  in html2)

API METHODS - PERSONALPET
API METHODS - PETSTORE
API METHODS - PETOWNERS

and the above operation (getInventory) has **duplicate** HTML output: 

file:///tmp/petstore/index.html#api-PersonalPets-getInventory
file:///tmp/petstore/index.html#api-PetStore-getInventory

See [the -debugOperations output](https://gist.githubusercontent.com/DavidBiesack/f635d3504320cba1701f2e35ad9379d3/raw/a391d420ff15f0134fc1f36cb21e13463e066a83/debugOperations.txt?_sm_au_=iVVDq7n0kPL7s5tR) ; note that the original tag text "Personal Pet" or "Pet Store" or "Pet Owner" is not available, nor are the tag descriptions  "Everything about your Pets" or externalDocs

In summary, tags are meant to be used to organize operations and the Open API speci specifically allows multiple tags per operation.

##### Command line used for generation

java -DdebugOperations -jar swagger-codegen-cli.jar  generate \
[...]

PR #5455, opened 2017-04-23
Enable Tag details to be used in templates

### LINKED ISSUE #1073: Please add support for the description of Tag objects
The name of a tag object is used to group all operations marked with this tag under the same file. I would like to be able to output the description of the tag object as well in this file. For example:

Swagger:

```
"tags": [{ "name": "other", "description": "my description" }]
```

C# library:

```
/// <summary>
/// my description
/// </summary
public class OtherApi { }
```

nbruno:
Adds the ability to reference the tag name, description, external docs, and vendor extensions in Mustache templates during codegen for tags.

Filed against the 2.3.0 branch since it is a breaking change.

I did not use the method Swagger#getTag that was introduced as part of #1073 since it requires updating to swagger core version 1.5.13 which causes other regressions. 

Closes #1073
Related #5074

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
@@ -1,6 +1,7 @@
 package io.swagger.codegen;
 
 import io.swagger.models.ExternalDocs;
+import io.swagger.models.Tag;
 
 import java.util.ArrayList;
 import java.util.HashSet;
@@ -28,7 +29,7 @@ public class CodegenOperation {
     public List<CodegenParameter> headerParams = new ArrayList<CodegenParameter>();
     public List<CodegenParameter> formParams = new ArrayList<CodegenParameter>();
     public List<CodegenSecurity> authMethods;
-    public List<String> tags;
+    public List<Tag> tags;
     public List<CodegenResponse> responses = new ArrayList<CodegenResponse>();
     public Set<String> imports = new HashSet<String>();
     public List<Map<String, String>> examples;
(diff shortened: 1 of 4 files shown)


wing328:
@nbruno thanks for the PR. When I regenerated PHP petstore, I got the following change:
```
-$api_instance = new Swagger\Client\Api\FakeApi();
+$api_instance = new Swagger\Client\Api\DefaultApi();
```
FakeApi was renamed to DefaultApi.

I did another test with Ruby Petstore and got similar result:
```
-api_instance = Petstore::FakeApi.new
+api_instance = Petstore::DefaultApi.new
```

I reviewed the 'fake' tag and it seems to be defined correctly: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml#L675

I need to dig further to find out why this occurs.

nbruno:
@wing328 I think I see the problem and added a fix and test case for it - It's possible for a tag listed on an operation to not be included in the Swagger tag definitions at the root of the spec, which is what the 'fake' tag is doing in the sample spec. Would you mind running your tests again?

wing328:
@nbruno I did some tests again and no longer found the behaviour I reported earlier.

Thanks for the enhancement 👍

ari:
Can anyone assist with pointing to the documentation on how to use the tag description in mustache? I'm tying to add it to the top of operation.mustache for the dynamic-html templates but nothing is being rendered. {{ classname }} displays the name of the tag just fine, but {{description}} doesn't work for the description under that.

wing328:
@ari would the following help?

https://github.com/swagger-api/swagger-codegen/wiki/Mustache-Template-Variables

ari:
Thanks. That's very helpful and it looks like tag descriptions are not available in mustache, at least for the dynamic-html language. I'm not sure what this pull request did, but perhaps it needs to be applied for different languages.

PR #5457, opened 2017-04-23
[TypeScript][Angular2] Add "toString()" to form parameters

wing328:
### PR checklist

### Description of the PR

toString() seems to be removed when syncing changes from master to 2.3.0. Adding it back via this PR.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/typescript-angular2/api.service.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular2/api.service.mustache
@@ -241,7 +241,7 @@ export class {{classname}} {
             body: {{paramName}} == null ? '' : JSON.stringify({{paramName}}), // https://github.com/angular/angular/issues/10612
 {{/bodyParam}}
 {{#hasFormParams}}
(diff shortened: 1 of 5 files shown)


wing328:
cc @sebastianhaas @Vrolijkx

sebastianhaas:
I have just taken a look, everything seems fine to me.

wing328:
@sebastianhaas thanks for reviewing.

Shall we do the same to other typescript generators (e.g. jquery, node, etc)

sebastianhaas:
Don't think this is applicable in the cases you mentioned, `formParams` seems to be of Angular-specific class `URLSearchParams`, that's why the `toString()` is implemented and suitable for the present situation.

wing328:
@sebastianhaas got it. Thanks for the explanation.

sebastianhaas:
Just for the record, in case someone wishes to understand these changes later: `URLSearchParams` is some sort of dictionary, which is then flattened by calling `toString()`.
https://angular.io/docs/ts/latest/api/http/index/URLSearchParams-class.html

PR #5461, opened 2017-04-23
Set name for item property from 'x-item-name' annotation if present.

mkistler:
### PR checklist

### Description of the PR

The Swagger/OpenAPI spec (2.0 at least) does not have a way to give a "name" to an element in an array property.  Currently, the generator confers the name of the array property onto the property for an array item.  This can lead to awkward/unnatural method names for adding or removing an item from an array property.

Consider:
```
properties:
  children:
    type: array
    items:
      $ref: "#/definitions/Child"
```

Currently, the CodegenProperty describing an item in the `children` array will also have the name `children`.  For some languages it is common to provide methods to access individual elements of an array property, like add<element>.  For the `children` property above, the natural signature (in Java) for such a method would be:
```
  public void addChild(Child child) {
```
But since the item name is actually `children`, the method produced would have the signature:
```
  public void addChildren(Child children) {
```
This is awkward and unnatural and potentially misleading.

This PR adds support for an `x-item-name` vendor extension on array properties that specifies the name for an item of the array.  Using this extension, the swagger above would be modified as follows:
```
properties:
  children:
    type: array
    items:
      $ref: "#/definitions/Child"
    x-item-name: child
```
This will set the `name` and `basename` in the CodegenProperty describing the array item to `child` so that element accessors can be generated with the desired signatures.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -1736,7 +1736,11 @@ public class DefaultCodegen {
             ArrayProperty ap = (ArrayProperty) p;
             property.maxItems = ap.getMaxItems();
             property.minItems = ap.getMinItems();
-            CodegenProperty cp = fromProperty(property.name, ap.getItems());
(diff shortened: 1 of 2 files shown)


wing328:
@mkistler thanks for the enhancement, which looks good to me.

PR #5462, opened 2017-04-24
Change info log to debug log as it print info message each time we generate code generator even if we dont need ignore file.

sanjeewa-malalgoda:
### PR checklist

### Description of the PR


wing328:
The Travis CI tests failed. Please review the result here: https://travis-ci.org/swagger-api/swagger-codegen/builds/225089539

> even if we dont need ignore file.

Understood that you don't need it but what about still keeping it in the output directory just in case there's a need in the future.

Instead of removing it, what about revising the wording instead? Personally I want to keep the message to make users (especially new ones) aware of ".swagger-codegen-ignore".

sanjeewa-malalgoda:
Yes i agree with you about informing new users about this. If we do not allow to override codegen ignore behavior and its not required for some use case should we still log this. What i though was "if some user sees issue in ignore file generation then they will enable debug logs and then they will see this message and realize reason was missing that directory". What do you think. Please apologize if  i have understood anything in wrong way here. Thank you very much for your attention on this issue.

Issue #5467, opened 2017-04-25
Swagger Codegen for Javascript throws NullPointerException for missing version in swagger json


erdarun:
##### 
version field not included in swagger json/yaml causing issue for javascript codegen but java/ruby works.

Exception in thread "main" java.lang.NullPointerException
	at io.swagger.codegen.languages.JavascriptClientCodegen.escapeQuotationMark(JavascriptClientCodegen.java:1045)
	at io.swagger.codegen.languages.JavascriptClientCodegen.preprocessSwagger(JavascriptClientCodegen.java:247)
	at io.swagger.codegen.DefaultGenerator.generate(DefaultGenerator.java:139)
	at io.swagger.codegen.cmd.Generate.run(Generate.java:223)
	at io.swagger.codegen.SwaggerCodegen.main(SwaggerCodegen.java:36)

##### Swagger-codegen version

swagger-codegen 2.2.2

##### Command line used for generation

swagger-codegen generate -i ./swagger.json -l javascript -o javascript/.
##### Steps to reproduce

Remove info version from swagger json/yaml

##### Suggest a Fix

Null check for info.version should be implemented

wing328:
@erdarun can you try the latest master to see if you still encounter the same issue?

erdarun:
Yes still the same issue in master branch too

wing328:
For the master branch, can you help show us the stack trace?

wing328:
Btw, version is a required field: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#info-object

erdarun:
@wing328 
swagger-codegen/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java line no. 259

erdarun:
Yes after adding version info it works. But null check should be added for escapeUnsafeCharacters, escapeQuotationMark methods.

As I tested Java/Ruby client codegen doesn't report any error if there is no info.version

wing328:
@erdarun good point. Do you mind submitting a PR to avoid NPE in those methods?

erdarun:
@wing328 Raised PR https://github.com/swagger-api/swagger-codegen/pull/5471

wing328:
Thanks for the contribution from @erdarun 

Closing this issue.

PR #5471, opened 2017-04-25
Fix NullPointerException for info.version on Javascript client codegen

erdarun:
### PR checklist

### Description of the PR

Issue Reference #5467

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java
@@ -152,7 +152,7 @@ public class DefaultGenerator extends AbstractGenerator implements Generator {
         if (info.getVersion() != null) {
             config.additionalProperties().put("appVersion", config.escapeText(info.getVersion()));
         } else {
-            LOGGER.error("Missing required field info version. Default appVersion set to 1.0.0")
+            LOGGER.error("Missing required field info version. Default appVersion set to 1.0.0");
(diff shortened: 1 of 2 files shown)


wing328:
@erdarun the function is used by other operations as well so I don't think it's a good idea to default to "1.0.0" if it's set to null.

I can show you where to set the default for "version" if it's missing.

wing328:
You will need to add an else-clause to set a default version/appVersion:

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java#L152-L154

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java#L181-L183

wing328:
Please also use LOGGER.error to report the missing required fields.

erdarun:
@wing328 
Can we report LOGGER.WARN for version missing and
Can we return default value as 1.0.0 at,
https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java#L152-L154

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java#L181-L183

erdarun:
@wing328 PR updated as per your comments

wing328:
@erdarun also looks good to me.

PR #5478, opened 2017-04-25

wy-z:
### PR checklist

### Description of the PR

- [go][model] Use pointers to avoid circular references

### DIFF
--- a/modules/swagger-codegen/src/main/resources/go/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/go/model.mustache
@@ -9,7 +9,7 @@ import ({{/imports}}{{#imports}}
 type {{classname}} struct {
 {{#vars}}{{#description}}
 	// {{{description}}}{{/description}}
-	{{name}} {{{datatype}}} `json:"{{baseName}}{{^required}},omitempty{{/required}}"`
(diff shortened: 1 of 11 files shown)


wing328:
@wy-z thanks for the PR.

cc @antihax @guohuang @neilotoole

antihax:
Will check tonight.

Is there a way to write a test case for this and some of the other fixed issues?

Perhaps we could use versioned `.yaml` spec with appropriate test cases for new clients going forward? This could give a good way to incrementally set the bar for the stability of the client code (all languages).

wing328:
> Is there a way to write a test case for this and some of the other fixed issues?

@antihax I would suggest adding new models to test the scenarios in https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml to start with.

antihax:
I had done that with #4417, but it causes a lot of other clients with the same issue to break. Hence versioning may be a way to go?

wing328:
Right. That happened before.

Let's give versioning a try to see how that goes.

antihax:
This looks good.

I've starting writing test cases in the spec including a recursive model. Should have a PR tonight for that.

wy-z:
Thanks for your fruitful work~
@wing328 @antihax

Issue #5374, opened 2017-04-12
withCredentials:true  for Typescript-Angular 2 Generated code


xmeng1:
##### Description

When I used the generated code of Typescript-Angular 2, and try to CORS request, the session information is always losing. After search the issue, I found that I need to add an option of `withCredentials:true ` when sending HTTP request. But the generated code does not include this option.

see: http://stackoverflow.com/questions/13734686/communication-between-angularjs-and-a-jersey-webservice-which-are-on-a-different/14111039#14111039

http://stackoverflow.com/questions/16882245/http-request-doesnt-send-cookies-cross-domain-in-angular-cors

So I hope the option `withCredentials:true` can be added to the HTTP request in 

```
        let requestOptions: RequestOptionsArgs = new RequestOptions({
            method: RequestMethod.Post,
            headers: headers,
            body: formParams.toString(),
            search: queryParameters
        });
```
##### Swagger-codegen version

git clone at April 3rd.
##### Suggest a Fix

The option `withCredentials:true` can be added to the HTTP request in 

```
        let requestOptions: RequestOptionsArgs = new RequestOptions({
            method: RequestMethod.Post,
            headers: headers,
            body: formParams.toString(),
            search: queryParameters
        });
```

wing328:
@xmeng1 thanks for reporting the issue. Your suggestion sounds reasonable to me as we've done something similar to JS API client: https://github.com/swagger-api/swagger-codegen/search?q=withCredentials&type=Issues&utf8=%E2%9C%93

Do you mind submitting a PR so that we can review the fix?

Please also checkout the `typescript-angular2` client in `2.3.0`, which contains enhancements and breaking changes.

xmeng1:
@wing328  OK, let me try the latest firstly.

xmeng1:
I have modified the template and submit a PR. thanks
https://github.com/swagger-api/swagger-codegen/pull/5380

kernwig:
I don't use cookies for authentication. Adding withCredentials:true broke everything for me. This should not be default behavior.

`XMLHttpRequest cannot load https://--server--/api/User/GetUser. Response to preflight request doesn't pass access control check: The value of the 'Access-Control-Allow-Credentials' header in the response is '' which must be 'true' when the request's credentials mode is 'include'. Origin 'http://localhost:3000' is therefore not allowed access. The credentials mode of requests initiated by the XMLHttpRequest is controlled by the withCredentials attribute.`

[...]

PR #5484, opened 2017-04-25
add withCredentials to the configuration

xmeng1:
#5374 
1. `withCredentials` in Request is `false` default
2. if want to set the `withCredentials` as `true` (for CORS request),
 instance `Configuration` object and set the `withCredentials` as `true`.

### PR checklist

### Description of the PR

### DIFF
--- a/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
@@ -195,6 +195,10 @@ export class {{classname}} {
         {{/isListContainer}}
 
 {{/formParams}}
+        let isWithCredentials: boolean = false;
+        if (this.configuration.withCredentials) {
+            isWithCredentials = this.configuration.withCredentials;
+        }
         let requestOptions: RequestOptionsArgs = new RequestOptions({
             method: {{httpMethod}},
             headers: headers,
@@ -205,8 +209,12 @@ export class {{classname}} {
             body: formParams.toString(),
 {{/hasFormParams}}
             search: queryParameters,
-            withCredentials:true
+            withCredentials:isWithCredentials
         });
+
+        if (this.configuration.apiKey) {
+        headers.set('{{keyParamName}}', this.configuration.apiKey);
+        }
 
         // https://github.com/swagger-api/swagger-codegen/issues/4037
         if (extraHttpRequestParams) {
--- a/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
@@ -195,10 +195,6 @@ export class {{classname}} {
         {{/isListContainer}}
 
 {{/formParams}}
-        let isWithCredentials: boolean = false;
-        if (this.configuration.withCredentials) {
-            isWithCredentials = this.configuration.withCredentials;
-        }
         let requestOptions: RequestOptionsArgs = new RequestOptions({
             method: {{httpMethod}},
             headers: headers,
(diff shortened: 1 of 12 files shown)


wing328:
@xmeng1 thanks for the quick turnaround.

@kernwig would you please take when you've time?

modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
```diff
@@ -195,6 +195,10 @@ export class {{classname}} {
         {{/isListContainer}}
 
 {{/formParams}}
+        let isWithCredentials: boolean = false;
```
kernwig: Simplify?

`let isWithCredentials: boolean = (this.configuration.withCredentials === true);`

modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
```diff
@@ -205,9 +209,13 @@ export class {{classname}} {
             body: formParams.toString(),
 {{/hasFormParams}}
             search: queryParameters,
-            withCredentials:true
+            withCredentials:isWithCredentials
```
kernwig: Or instead of above, it why not right here?

`    withCredentials: this.configuration.withCredentials`

modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
xmeng1: yes, this expression is more concise thanks

modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
xmeng1: In the case of the `withCredentials` is not defined in the configuration object, I just define a local var to set the default value to `false`.

kernwig: If it's not defined, then you will set withCredentials to undefined, which exactly replicates what the old code did by not setting it at all. :-)

xmeng1: Got it!  :)
So I could remove all code about the local var:  
```
let isWithCredentials: boolean = false;
if (this.configuration.withCredentials) {
  isWithCredentials = this.configuration.withCredentials;
}
```
and just use this expression to pass the value of configuration to the option. is that right?

sorry, I am Java developer and habitually `null` checker.

kernwig: Yes.
Javascript has confusing ideas about "truthy" and "falsy". https://developer.mozilla.org/en-US/docs/Glossary/Falsy

xmeng1: thanks for your explanation :+1:   Need I make new PR?

wing328: @xmeng1 Please update this PR with new commits instead of creating a new one

xmeng1: No problem, I will update it tonight! @wing328

wing328:
@xmeng1 PR merged into master. Thanks again for the quick turnaround.

@kernwig please pull the latest to give it a try.

xmeng1:
@wing328  cheers
@kernwig Thanks for your code review

kernwig:
It was merged into master, not 2.3.0.

wing328:
@kernwig FYI. Enhancements in the master will later be sync'd to 2.3.0

Issue #5486, opened 2017-04-26
[JAVA] Wrong subtype class name if lower case


bbdouglas:
##### Description

When a model is a subclass of another model, and the subclass  model name is in lower case, the parent class annotates the wrong name in its annotations.

Here is an example. Note that the "dog" model is in lower case.

    definitions:
      Animal:
        type: object
        discriminator: className
        ...
      dog:
        allOf:
          - $ref: '#/definitions/Animal'
          ...

A class named `Dog` is generated (following Java code style), but it is referred to as `dog` in the Animal class.

    @JsonSubTypes({
      @JsonSubTypes.Type(value = dog.class, name = "dog"),
    })
    public class Animal {

Since there is no `dog` class, this code doesn't compile. Here is what the output should be.

    @JsonSubTypes({
      @JsonSubTypes.Type(value = Dog.class, name = "dog"),
    })
    public class Animal {

This looks like just a minor typo in the mustache file.

##### Swagger-codegen version

Current (2.2.3)

##### Swagger declaration file content or url

See above. It happens for any model that is a subtype of another type and begins with a lower case letter.

##### Command line used for generation

-l java

##### Related issues

I have not seen any similar issues.

##### Suggest a Fix

I have a one-line fix coming up.

wing328:
Closed via #5487

PR #5487, opened 2017-04-26
Fix class name in discriminator subtypes

bbdouglas:
### PR checklist

### Description of the PR

This fixes a typo in the mustache file for discriminator types. It is generating a class name, so it should use the {{classname}} field, not the {{name}} field. Most of the time, these two are the same, but in cases where the model name in the spec file is lower cased, they are different. This edge case doesn't show up in the existing testing endpoint, so the samples are unaffected.


bbdouglas:
This is a fix for #5486.

PR #5491, opened 2017-04-26
Remove unused generation script android-petstore.sh

bbdouglas:
The old android-petstore.sh script is writing output to the android volley
directory. There is a new script called android-petstore-volley.sh that also
writes to that same directory. They overwrite each other, which makes the
run-all-petstore script output dependent on which one gets run last. This change
removes the old script.

### PR checklist

### Description of the PR

I noticed this issue when re-generating the samples with `run-all-petstore` for #5408. It looks like a new script was created for android volley without deleting the old one.


wing328:
@bbdouglas you're right. My bad forgetting to remove the old one.

I just check the Window batch files and it's missing those batch files for Android (httpclient, volley). I'll create a separate PR to add back those.

Issue #5432, opened 2017-04-19
[Java] Spring Cloud Client is broken due to template change


wing328:
##### Description

Broken due to https://github.com/swagger-api/swagger-codegen/pull/5411. We will need to add tag to output code that are API client only.

##### Swagger-codegen version

Latest master

##### Related issues

https://github.com/swagger-api/swagger-codegen/pull/5411

##### Suggest a Fix

Add a mustache tag (e.g. {{#isApiClient}} to remove code not needed for API client.

PR #5497, opened 2017-04-27
[Java][Spring] Fix spring cloud client due to enhancement in adding server example responses.

wing328:
### PR checklist

### Description of the PR

To fix https://github.com/swagger-api/swagger-codegen/issues/5432

### DIFF
--- a/modules/swagger-codegen/src/main/resources/JavaSpring/apiController.mustache
+++ b/modules/swagger-codegen/src/main/resources/JavaSpring/apiController.mustache
@@ -92,11 +92,11 @@ public class {{classname}}Controller implements {{classname}} {
             }
         };
         {{/async}}
-{{/useSpringCloudClient}}
         {{/isDelegate}}
         {{#isDelegate}}
         return delegate.{{operationId}}({{#allParams}}{{paramName}}{{#hasMore}}, {{/hasMore}}{{/allParams}});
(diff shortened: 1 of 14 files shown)


cbornet:
Wouldn't it be easier to have some "returnExamples" additionalProperty that is false for spring-cloud and that equals "examples" for the other ?
That way, we wouldn't have duplication in the template.

wing328:
@cbornet thanks for the suggestion. I think the challenge lies in the fact that both client and server share the same templates and I believe we still need the `useSpringCloudClient` tag to avoid unnecessary Java library import.

Anyway feel free to submit a PR to further enhance the template after this PR is merged into master.

modules/swagger-codegen/src/main/resources/JavaSpring/apiController.mustache
```diff
@@ -17,8 +17,10 @@ import org.springframework.web.bind.annotation.RequestHeader;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RequestPart;
 import org.springframework.web.multipart.MultipartFile;
+{{^useSpringCloudClient}}
```
cbornet: Spring cloud client doesn't generate the apiController so I guess this is not needed...

Issue #5421, opened 2017-04-19
[Python] Better JSON Serialization and Marshaling


wing328:
##### Description

Currently, the Python API client relies on our own deserialiser to convert JSON string back into Python objects with support of attribute mapping. For other generators (e.g. Java, C#), we're leveraging 3rd-party libraries/modules to handle serialization/deserialization between objects and JSON strings

We would like to evaluate different Python modules to perform similar tasks instead of maintaining our own serialization/deserialisation logic:

- [Kim: JSON Serialization and Marshaling framework](https://github.com/mikeywaites/kim)

If anyone has comments/suggestions or want to contribute to this enhancement, please reply to let us know. Thank you.

wing328:
Kim is just one of the Python module for serialization/deserialization. If anyone has recommendations for other Python modules to perform similar tasks, please let us know.

frol:
[Marshmallow](https://github.com/marshmallow-code/marshmallow/) is a mature and well-maintained project for serializing/deserializing, encoding, and validation. Marshmallow has quite rich ecosystem providing a number of extensions: Marshmallow-SQLAlchemy, Webargs, Apispec, Flask-Marshmallow, Flask-Apispec, ...

I have no experience with Kim, so I cannot compare it to Marshmallow, but Marshmallow has never failed me.

frol:
There is a benchmark of various Python serializers: https://voidfiles.github.io/python-serialization-benchmark/

Given the results of the benchmark, [`lima`](https://lima.readthedocs.io/en/latest/) looks like a clear winner, and there are at least 3 more alternatives: [serpy](http://serpy.readthedocs.io/en/latest/), [Strainer](https://strainer.readthedocs.io/en/latest/), [Lollipop](http://lollipop.readthedocs.io/en/latest/).

Issue #5277, opened 2017-04-01
[PHP] Better code format


wing328:
##### Description

When using [PHP CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer), we found potential enhancemnets for auto-generated PHP code.

##### Swagger-codegen version

- [ ] Current master
- [ ] 2.3.0

##### Steps to reproduce

```
composer install
./vendor/bin/phpcs --extensions=php ./lib
```

##### Related issues

https://github.com/swagger-api/swagger-codegen/pull/5190#issuecomment-290909884

##### Suggest a Fix

If anyone wants to work on the enhancement, please reply to let us know.

dkarlovi:
This was actually being worked on in #3898, I never got around to resubmitting the changes. Would need to take into account changes from #5190, though.

wing328:
@dkarlovi right. I remember your PRs. 

#5190 is submitted against 2.3.0 so your work on the `master` version of PHP API client is still valid.

After merging #5190 into 2.3.0, I'll publish a report by PHP CodeSniffer so that everyone knows what needs to be enhanced.

dkarlovi:
In #3863 I've added support for [PHP CS Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) which seems to becoming PHP CS style tool of choice, not only can it find CS violations, it can fix them.

Would need to update the config file for a new version. Also, I'm using full code inspection from PHPStorm which also catches typos and such (some of which are also fixed in the PR linked).

wing328:
For the time being, we can leverage PHP-CS-Fixer to update the code in terms of style and format.

Ideally I want to update the templates for PHP API client without the need to run php-cs-fixer manually.

PR #5512, opened 2017-04-29
Python improved

easz:
### PR checklist

### Description of the PR

try to improve python client generation.
 - model construction with mandatory attributes
 - improve class reflection in deserialization
 - test (top-level) enum class deserialization

See 
#5375 
#5421


wing328:
Thanks for the PR.

cc @frol @scottrw93 @taxpon @baartosz

@easz btw, have you checked out the 2.3.0 branch in which Python API client has a lot of enhancements (with breaking changes)?

easz:
$ cd swagger-codegen/

$ git checkout remotes/swagger-api/2.3.0

// the last commit c01cb9e of this PR is not part of changes
$ git cherry-pick  ca45e6f  5b5c2f6  5bb87a2  89d8bd4 c40b7dc 8b584d2  c64c7ad

// delete 'test' for new code 
// the 'test' folder will not be overwritten if already exist. Desired behavior? 
$ rm -rf samples/client/petstore/python/test

$ ./bin/python-petstore.sh

$ cd samples/client/petstore/python/
$ ./test_python2_and_3.sh

Ran 100 tests in 11.110s

OK (SKIP=1)
________________________________________________________ summary ________________________________________________________
  py27: commands succeeded
  py3: commands succeeded
  congratulations :)

wing328:
> // the 'test' folder will not be overwritten if already exist. Desired behavior?

Yes, that's the expected behaviour because we do not want to overwrite the test files as the users might have manually added a few test cases in the test files.

easz:
that means 
./bin/python-petstore.sh will never update 'test' folder even 'model_test.mustache' is changed?

actually I did assume 'test' folder is automated generated tests
and 'tests'  folder has user defined tests.

wing328:
> ./bin/python-petstore.sh will never update 'test' folder even 'model_test.mustache' is changed?

Yes, that's the current behaviour.

> actually I did assume 'test' folder is automated generated tests
and 'tests' folder has user defined tests.

That's one way to do it.

We plan to leverage .swagger-codegen-ignore to control which files not be overwritten during code generation.

wing328:
@easz if I understand correctly, these changes are backward-compatible with the existing auto-genreated Python API using the latest master.

I'll merge this PR into master on coming Wed if no one has any further feedback.

.gitignore
```diff
@@ -23,6 +23,7 @@ packages/
 .pub
 .packages
 .vagrant/
+.vscode/
```
frol: I think, this should be added locally on your own machine to the global git ignore file: https://davidwalsh.name/global-gitignore

easz:
yes. the change (e.g. required attributes need to be defined when calling model constructor) is backward-compatible with the existing auto-genreated Python API.

but some **test cases** has been changed to compatible with the change.

and any **client code** by any user may break if they did not assign required attributes in the constructor.

frol:
The change looks great to me.

P.S. I have noticed that there are some places where 2 spaces are used for indentation while 4 spaces indentation is recommended. Is this worth to update in the scope of this PR?

wing328:
> Is this worth to update in the scope of this PR?

I don't think so. I'll create another issue to track the code format of Python API client similar to what we've done for the PHP API client (refactored): https://github.com/swagger-api/swagger-codegen/issues/5277

wing328:
> and any existing client code by any user may break if they did not assign required attributes in the constructor.

@easz It's definitely the correct behaviour and I would imagine some applications using the upgraded Python API client will throw exception.

May I know what would be the exception message looks like? If it's clear, I think it's fine for this enhancement to go into the current master.

easz:
a ValueError, if a required attribute is not given in a model constructor

Exception is thrown from files generated by model.mustache

```
  if {{name}} is None:
    raise ValueError("Invalid value for `{{name}}`, must not be `None`")
```

before the change, the test cases under test/ just construct every possible model with null content

wing328:
@easz the error message looks good to me. We might revise it later to make it consistent with the error message shown by other API clients.

Issue #5515, opened 2017-04-30
hasMore field in supporting files authMethods incorrect


bbdouglas:
##### Description

The hasMore field of the global authMethods dictionary in the supporting files is incorrect for the petstore.yaml sample.

```
  "authMethods" : [ {
    "name" : "api_key",
    "type" : "apiKey",
    "hasMore" : false,  <--- HERE
    "isBasic" : false,
    "isOAuth" : false,
    "isApiKey" : true,
    "keyParamName" : "api_key",
    "isKeyInQuery" : false,
    "isKeyInHeader" : true,
    "isCode" : false,
    "isPassword" : false,
    "isApplication" : false,
    "isImplicit" : false
  }, {
    "name" : "petstore_auth",
    "type" : "oauth2",
    "hasMore" : true,  <--- HERE
    "isBasic" : false,
    "isOAuth" : true,
    "isApiKey" : false,
    "isKeyInQuery" : false,
    "isKeyInHeader" : false,
    "flow" : "implicit",
    "authorizationUrl" : "http://petstore.swagger.io/api/oauth/dialog",
    "scopes" : [ {
      "scope" : "write:pets",
      "hasMore" : "true",
      "description" : "modify pets in your account"
    }, {
      "scope" : "read:pets",
      "hasMore" : null,
      "description" : "read your pets"
    } ],
    "isCode" : false,
    "isPassword" : false,
    "isApplication" : false,
    "isImplicit" : true
  } ],
```

I noticed this when trying to generate the clojure samples, which produced a strange-looking line in `core.clj`:

    (def auth-definitions
      {"api_key" {:type :api-key :in :header :param-name "api_key"}"petstore_auth" {:type :oauth2}
       })
##### Swagger-codegen version

Current

##### Swagger declaration file content or url

petstore.yaml

##### Command line used for generation

JAVA_OPTS=-DdebugSupportingFiles bin/clojure-petstore.sh

##### Suggest a Fix

It looks like this is being caused by sorting the list of auth methods after the hasMore field is set. If the order changes due to the sort (as it does in petstore.yaml) then the hasMore fields are not accurate any more. I have a small PR to fix this coming up.

Oddly, the current `core.clj` file is correct. I don't know how it was generated before, because using the current version consistently produces the bad file.

wing328:
Closed via #5516

PR #5516, opened 2017-04-30
Fix hasMore field and order consistency of authMethods

bbdouglas:
The hasMore field of the global authMethods was broken due to sorting after
setting that field. The sort order for per-operation authMethods was not
guaranteed to be consistent across runs because it didn't do sorting. This
change fixes those issues by always sorting and setting the hasMore field
post-sort.

### PR checklist

### Description of the PR

Adopted code from DefaultGenerator and AkkaScalaClientCodegen to consistently sort the authMethods dictionaries and set the hasMore field.

Fix for #5515


wing328:
@bbdouglas thanks for the enhancement 👍

PR #5519, opened 2017-05-01
Implement generating interfaces option for api resources, closes #5453.

### LINKED ISSUE #5453: [typescript/angular2] Implement interfaces for api resources
##### Description

Currently, generated api files consist of bare classes for typescript / angular2. We are happy to have them as is but would like to make implemented alongside with their interfaces so we can create our own behaviours.

To extend a bit, depending on development scenario we sometimes want api resources to use local storage or json fixtures rather than querying our api endpoint. Therefore, would like to be able to implement its interface and overwrite the generated class.

##### Swagger-codegen version

2.2.2

```yaml
paths:
  /user:
    get:
      operationId: getUsers
      summary: Returns all users.
```
Currently generated:
```typescript
export class UsersApi {
   // snipped..
}
```
Proposed:
```typescript
export interface IUsersApi {
  // snipped.
}

export class UsersApi implements IUsersApi {
  // snipped.
}
```

##### Related issues

Improve TypeScript Template #1334

##### Suggest a Fix

I am happy to spend my time, keeping in mind that I'm new to the java and typescript world.

kirpit:
### PR checklist

### Description of the PR

This commit introduces another _system property_ for typescript-angular2, called `withInterfaces`
that is default to `false`. You can activate this flag by:

```
  -i http://petstore.swagger.io/v2/swagger.json \
  -l typescript-angular2 -o samples/server/petstore/springboot/typescript-angular2 \
  -D withInterfaces=true
```

If set to `true`, generated api resource classes will be implemented with their interfaces in separate files.

This change should not break and change the behaviour even though `withInterfaces` flag set to `true`.
You are also welcome to create your own custom resource class by implementing those generated
interfaces as they support defining new attributes with `[others: string]: any;` signature.


kirpit:
Most of the new files are freshly generated samples (with interfaces option) for petstore..

kirpit:
And apparently travis is complaining about a failed test:
```
VERBOSE MODE: ON. Additional debug options are injected

 - [debugSwagger] prints the swagger specification as interpreted by the codegen

 - [debugModels] prints models passed to the template engine

 - [debugOperations] prints operations passed to the template engine

 - [debugSupportingFiles] prints additional data passed to the template engine

Tests run: 643, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 17.943 sec <<< FAILURE! - in TestSuite

checkOptionsHelp(io.swagger.codegen.typescript.typescriptangular2.TypeScriptAngular2ClientOptionsTest)  Time elapsed: 0.167 sec  <<< FAILURE!

java.lang.AssertionError: These options weren't checked: withInterfaces.

Results :

Failed tests: 

  TypeScriptAngular2ClientOptionsTest>AbstractOptionsTest.checkOptionsHelp:44 These options weren't checked: withInterfaces.
```

Will fix that asap.

kirpit:
I have fixed the failing test but now there is a new one travis complaining about:

```
[ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:1.2.1:exec (npm-install) on project TypeScriptAngularPestoreClientTests: Command execution failed. Process exited with an error: 1 (Exit value: 1) -> [Help 1]
[ERROR] 
[ERROR] 
[ERROR] 
[ERROR]   mvn <goals> -rf :TypeScriptAngularPestoreClientTests
```

Could it be a network issue as it says failed on `npm install`?

kirpit:
The last error was a network issue apparently as the recent dummy commit passed every CI tool. 

Good to go?

wing328:
@kirpit thanks for the PR.

@Vrolijkx @taxpon @TiFu, please take a look when you've time. Thanks!

modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
```diff
@@ -10,6 +10,9 @@ import 'rxjs/add/operator/map';
 import * as models                                           from '../model/models';
 import { BASE_PATH, COLLECTION_FORMATS }                     from '../variables';
 import { Configuration }                                     from '../configuration';
+{{#withInterfaces}}
+import { I{{classname}} }                                    from './{{classname}}Interface';
```
Vrolijkx: If we look at the [coding guidlines from typescript](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines) the I prefix should be avoided.
If not to much work it would be beter to rename the api class that implements this interface with something like default.

kirpit: It's embarrassing I missed that. How do we want the new naming to be exactly? Let's say in the `UserApi` case:

Class name: `UserApi`
Interface: `UserApiDefault` ?

If we can clear that out, I can make necessary changes and also fix these merge conflicts by this weekend.

taxpon: Since the interface name convention depends on each project policy (MS is saying ["Feel free to adopt them for your own team."](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines)), how about providing an option to specify the prefix/postfix word? Example, default is "I" prefix (like `IUserApi`) but if user specify "Interfece" postfix, the interface name would be `UserApiInterface` as well as interface file name. What do you think?

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngular2ClientCodegen.java
```diff
@@ -26,6 +26,7 @@
     public static final String NPM_VERSION = "npmVersion";
```
taxpon: As you may know, this file is conflicting my previous [commit](https://github.com/swagger-api/swagger-codegen/pull/5569).

kirpit: No worries, will fix that.

modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
kirpit: Giving an option to specify interface suffix and also the filename might be indeed useful for some projects. However in most cases, they will not care and just go with the default interface name whatever it is. Even here where we needed heavily custom usage of those interfaces, didn't need any specific naming.

This will also introduce unnecessary complexity for our code base for the 99% of use cases IMHO.

TiFu: I agree with @Vrolijkx . The interface should have neither a suffix nor a prefix. The TS Guidelines only specify the 'I' prefix, but I think the idea behind it is, that prefixes & suffixes are unnecessary. 

I would propose to prefix the api implementation with `Default`.

For the `UserApi` the change would look like this:
class name: `DefaulUserApi`
Interface: `UserApi`

kirpit: Hi @TiFu,

Sure we can rename both the interface and the class name, if @Vrolijkx and everyone else is happy with that. But this will virtually change the current class names and their file names, so if someone has imported and using `UserApi` resource _class_ at the moment, this implementation will replace it with the new interface, therefore will break their code.

kirpit:
One thing I would like to learn before merging this PR for my mental health; where is the documentation / wiki page for all these _system dependency_ parameters list that we pass with `-D`? My PR literally introduces one more and I'd like to update it and also like to have a peak at others..

taxpon:
@kirpit You can find all options as follows:

```shell
$ java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar config-help -l typescript-angular2

CONFIG OPTIONS
        sortParamsByRequiredFlag
            Sort method arguments to place required parameters before optional parameters. (Default: true)

        ensureUniqueParams
            Whether to ensure parameter names are unique in an operation (rename parameters that are not). (Default: true)

        allowUnicodeIdentifiers
            boolean, toggles whether unicode identifiers are allowed in names or not, default is false (Default: false)

        modelPropertyNaming

        supportsES6
            Generate code that conforms to ES6. (Default: false)

        npmName
            The name under which you want to publish generated npm package

        npmVersion
            The version of your npm package

        npmRepository
            Use this property to set an url your private npmRepo in the package.json

        snapshot
            When setting this property to true the version will be suffixed with -SNAPSHOT.yyyyMMddHHmm (Default: false)

        useOpaqueToken
            When setting this property to true, OpaqueToken is used instead of InjectionToken (Default: false)
```

modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
TiFu: That's a good point. I'm okay with both options but have a slight preference for the clearer naming convention. Let's see what the others think about it.

kirpit:
Okay, here is the current situation, to get things going forward:

None of the changes so far is breaking anything, even though the newly introduced `withInterfaces` system config is enabled (defaults to false that is disabled).

Once enabled by `-D withInterfaces=true`, it creates interfaces for each API resource, in its own file with `Interface` suffix, both to the default class implementation and also to the filename. 

For example, for the `UserApi` resource class that is generated within `UserApi.ts` file, there will be an interface implementation named `UsersApiInterface` within the `UsersApiInterface.ts` file. `UserApi` class definition will be eventually changed to:

`export class UsersApi implements UsersApiInterface {}`

If the common ground requires renaming the default class implementation, it will probably make more sense to do that in version `2.3.0` as it will be a hard breaking change.

wing328:
Thanks @taxpon @Vrolijkx for reviewing this PR.

Thanks @kirpit for the enhancement.

Issue #5542, opened 2017-05-03
[Spring] Improve CI, maven integration: regenerate all files depending on swagger specs at every build


nicusX:
##### Description

As a best practice for code generation with a CI pipeline driven by maven, it should be able to regenerate every file that depends on Swagger spec file.

Speaking about Spring Boot, these files are:
- API Interfaces
- Model classes
- `SwaggerDocumentationConfig.java` (defines API info based on the Swagger specs)

Using Maven, these files should be regenerated at every build, into `./target/generated-sources` (usually ignored in VCS).

Any other file might be optionally generated, on demand (as opposed as automatically at every build), only on the developer's machine and into `./src`.

Currently, this only partially works using the `swagger-codegen-maven-plugin` and `spring` language.

Specifyng `interfaceOnly=true`,  the following files are generated:
- API Interfaces
- Model classes
- `pom.xml`
- `README.md`
- `.swagger-codegen-ignore`

By default, `swagger-codegen-maven-plugin` correctly generates into `./target/generated-sources`

I see two distinct problems here:

1. `pom.xml`, `README.md`, `.swagger-codegen-ignore` should not be generated when `interfaceOnly=true` (and it also doen't make sense generating them into `./target/generated-sources`)
2. It should be possible to regnerate also `SwaggerDocumentationConfig.java`, but not 

The `.swagger-codegen-ignore` file is not useful, as you only may ignore files, not selectively activate their generation.

##### Swagger-codegen version

Verified using Swagger Codegen Maven plugin 2.2.2 (release) and CLI 2.2.2

##### Swagger declaration file content or url

It does not depend on Swagger specs. I used this one for testing: https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v2.0/json/petstore-minimal.json

##### Command line used for generation

*pom..xml*
```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>
    <groupId>com.example</groupId>
    <artifactId>example</artifactId>
    <version>0.0.1-SNAPSHOT</version>
    <packaging>jar</packaging>
    <name>sample</name>

    <parent>
[...]

PR #5545, opened 2017-05-03
[spring] improve build-time code generation, for CI and maven integration

nicusX:
### PR checklist

### Description of the PR

Behaviour changes for the `spring` language only, to allow generating code depending on the swagger specs at build time, driven by the maven plugin. As described by #5542

- Do not generate pom.xml and README.md when `interfaceOnly`=`true`

- New configOption `swaggerDocketConfig` (default: `false`) to allow forging generation of SwaggerDocumentationConfig.java when `interfaceOnly`=`true`. Ignored when `interfaceOnly`=`false` or `library` = `spring-cloud`

The goal of these changes it having a proper build-time code generation with Spring projects: files depending on the Swagger specs should change automatically at every build, but no file modified by the developer should never be touched. The ignore approach is prone to error.


nicusX:
I requested merging to 2.3.0 as this changes modify the behaviour when `interfaceOnly=true`
Actually, I think the previous behaviour (generating pom.xml and README.md) when `interfaceOnly=true` was incorrect.

Potentially, the new config option `swaggerDocketConfig` might be moved to a separate PR, but I think it is an integral part of making build-time code generation working properly

wing328:
@nicusX thanks for the PR.

cc @cbornet @jfiala

wing328:
@nicusX thanks for the contribution, which has been merged into master. (FYI. I've updated the samples via 5d03905)

Issue #5138, opened 2017-03-21
[C#] Consolidate C# 2.0 generator into  C# generator


wing328:
##### Description

Currently, C# (`csharp`) generator supports the following .net framework/standard:
```
	targetFramework
	    The target .NET framework version. (Default: v4.5)
	        v3.5 - .NET Framework 3.5 compatible
	        v4.5 - .NET Framework 4.5+ compatible
	        v5.0 - .NET Standard 1.3 compatible
	        uwp - Universal Windows Platform - beta support
```
and there's a separate generator for C# 2.0.

We want to consolidate/merge C# 2.0 generator into `csharp` generator for easier maintenance and support.

##### Swagger-codegen version

Latest master

If anyone wants to work on the enhancement, please reply to let us know. Thanks!

jimschubert:
@wing328 I would disagree that merging the C# 2.0 generator into the others will lead to easier maintenance.

The other generators currently use syntax which is unavailable in C# 2.0 (object initialization, default/named parameters, implicit types). As well, there are additional keywords in later versions which are not keywords in C# 2.0. It's been nearly ten years since I last used C# 2.0, but I seem to recall there were issues with the migration of delegates and enumerations to C# 3.0 (really 3.5 since we don't have a 3.0 generator).

Merging them all will create many unnecessary conditional blocks which I think will become difficult to reason about.

One path I think we could take is to merge the generator code in Java, but keep the templates physically separate. However, this could be a breaking change if anyone currently relies on the existing template structure for custom templates.

wing328:
@jimschubert fair enough. My goal is to keep the C# 2.0 generator up-to-date with the enhancements, bug fixes we applied to the C# (3.5, 4.5, UWP) generator as I believe C# 2.0 generator is still popular to Unity developers.

jimschubert:
@wing328 I agree with your points. I was just trying to say I think it'd be more maintainable as a separate generator that extends the other C# generator. However, if you want them all to be executed via `-l csharp`, I don't think my suggestion is possible (easily) with the current code.

Do you think it would make more sense to push most of the bug fixes into the abstract class? I think this would solve most of the issues.

jimschubert:
@wing328 I opened a PR to remove a bit of the duplicated code in the C# 2.0 generator in favor of the abstract base class.  I explained in the PR why it doesn't consolidate templates, which basically reiterates what I stated above.

[...]

PR #5552, opened 2017-05-04
[csharp2] Extend generator from AbstractCSharpCodegen

jimschubert:

see #5138

This is the start of consolidation efforts of C# 2.0 to the greater C# client generator code.

There is further discussion in the issue related to maintainability of combining C# 2.0 code into the newer code. For example, current users of both generators have the option to extend built-in templates to enhance the generator for their use case. This means users of `CsharpDotNet2` as well as `csharp` may have a file called `api.mustache`. The `CsharpDotNet2` version of the file is extremely basic because C# 2.0 doesn't support many of the advanced features that C# 3.0 and higher support. One option is to introduce extensive conditions into the logicless mustache template, which I think would reduce maintainability. Another option is to rename the file from `api.mustache` to `api_2_0.mustache` or something similar, which would be a breaking change. I don't feel that either option is a maintenance gain. This goes for all template files in the `CsharpDotNet2` generator.

For the above reason, this PR simply offloads some of the duplicated logic to the abstract generator class.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CsharpDotNet2ClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CsharpDotNet2ClientCodegen.java
@@ -17,15 +17,11 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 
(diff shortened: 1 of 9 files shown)


wing328:
@jimschubert I think this is definitely a good start.

For the templates, let's keep them separate for now and we'll reconsider if there are more demands (enhancements, bug fixes, etc) to the C# 2.0 generators.

PR #5571, opened 2017-05-05
[PHP] Add default value to method signature if present in Swagger spec

corleonis:
### PR checklist

### Description of the PR

Currently the default value from the spec is ignored in the generated code. This change makes sure that if there is existing default value for non-required arguments it's added to the signature of the method.


wing328:
@corleonis thanks for the PR, which looks good to me.

cc @arnested @baartosz

wing328:
cc @@dkarlovi as well

Issue #5577, opened 2017-05-06
CSharp language generator results in code with ambigious references


mminns:
##### Description

The CSharp api..mustache template creates API classes that reference Model objects. However if the Model contains classes with the same name as core C# classes the code will not compile without the references being explicitly specified.

##### Swagger-codegen version

Built from master @ commit ea16da8

##### Swagger declaration file content or url

https://api.bitbucket.org/swagger.json

##### Command line used for generation

java -jar C:\Users\mminn\Source\github.com\mminns\swagger-codegen\modules\swagger-codegen-cli\target\swagger-codegen-cli.jar generate -i swagger.json -l csharp -o client-mminns/csharp

##### Steps to reproduce

1. Run the generator
2. Open the generated sln in Visual Studio
3. Try and compile 

Results:

Compilation fails
Open Issue_trackerApi.cs

Listed with multiple errors, e.g.

Severity	Code	Description	Project	File	Line	Suppression State
Error	CS0104	'Version' is an ambiguous reference between 'IO.Swagger.Model.Version' and 'System.Version'	IO.Swagger	C:\Users\mminn\Source\bitbucket.org\mminns\bitbuckit.net.swagger\client-mminnns\csharp-debug\src\IO.Swagger\Api\Issue_trackerApi.cs	562	Active

 
##### Related issues

https://github.com/swagger-api/swagger-codegen/issues/5576

##### Suggest a Fix

In https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java#L310

Add the following method:

```
+    private void processUsings(Map<String, Object> objs) {
+        List<Map<String, String>> imports = (List<Map<String, String>>)objs.get("imports");
+        Map<String,String> usings = new HashMap<String,String>();
+        for(Map<String, String> importMap: imports) {
+            for(Map.Entry<String, String> entry: importMap.entrySet()) {
+                String fullModelName = entry.getValue();
+                String[] parts = fullModelName.split("\\.");
+                usings.put(parts[1], fullModelName);
+            }
+        }
+        objs.put("usings", usings.entrySet());
+    }


```
And using it in:

```
    @Override
    public Map<String, Object> postProcessModels(Map<String, Object> objs) {
        List<Object> models = (List<Object>) objs.get("models");
        for (Object _mo : models) {
            Map<String, Object> mo = (Map<String, Object>) _mo;
[...]

Issue #5576, opened 2017-05-06
CSharp language generator fails for Bitbuckets swagger JSON


mminns:
##### Description
1. Using codegen built from the latest master the CSharp language generator fails when trying to parse https://api.bitbucket.org/swagger.json.

##### Swagger-codegen version

Built from master @ commit ea16da813

##### Swagger declaration file content or url

https://api.bitbucket.org/swagger.json

##### Command line used for generation

java -jar C:\Users\mminn\Source\github.com\mminns\swagger-codegen\modules\swagger-codegen-cli\target\swagger-codegen-cli.jar generate -i swagger.json -l csharp -o client-mminns/csharp

Results:

"C:\Program Files (x86)\Java\jdk1.7.0_55\bin\java" -Dfile.encoding=windows-1252 -jar C:\Users\mminn\Source\github.com\mminns\swagger-codegen\modules\swagger-codegen-cli\target\swagger-codegen-cli.jar generate -i swagger.json -l csharp -o client-mminnns/csharp-debug
[main] INFO io.swagger.parser.Swagger20Parser - reading from swagger.json
[main] INFO io.swagger.codegen.languages.CSharpClientCodegen - Generating code for .NET Framework v4.5
Exception in thread "main" java.lang.RuntimeException: Could not process model 'pipeline_ref_target'.Please make sure that your schema is correct!
	at io.swagger.codegen.DefaultGenerator.generateModels(DefaultGenerator.java:315)
	at io.swagger.codegen.DefaultGenerator.generate(DefaultGenerator.java:665)
	at io.swagger.codegen.cmd.Generate.run(Generate.java:234)
	at io.swagger.codegen.SwaggerCodegen.main(SwaggerCodegen.java:43)
Caused by: java.lang.NullPointerException
	at io.swagger.codegen.DefaultCodegen.fromModel(DefaultCodegen.java:1264)
	at io.swagger.codegen.languages.CSharpClientCodegen.fromModel(CSharpClientCodegen.java:428)
	at io.swagger.codegen.DefaultCodegen.fromModel(DefaultCodegen.java:1246)
	at io.swagger.codegen.languages.CSharpClientCodegen.fromModel(CSharpClientCodegen.java:431)
	at io.swagger.codegen.DefaultGenerator.processModels(DefaultGenerator.java:907)
	at io.swagger.codegen.DefaultGenerator.generateModels(DefaultGenerator.java:310)
	... 3 more

Process finished with exit code 1

[...]

PR #5581, opened 2017-05-07
Fix for issues 5576 and 5577 related to generating CSharp clients

mminns:
### PR checklist

Ran the Petstore tests for CSharp and they seem to run correctly. However the existing tests do not test the part of the code that has changed here.

### Description of the PR

Corrects the code for finding parents of Models and removes ambiguous references in generated API class by explicitly referencing Model classes


wing328:
@mminns thanks for the PR. We'll review and let you know if we've any question.

Linked to #5576 and #5577

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java
```diff
@@ -307,10 +307,26 @@ public void processOpts() {
...
             }
         }
+
+        processUsings(objs);
+
         // process enum in models
         return postProcessModelsEnum(objs);
     }
 
+    private void processUsings(Map<String, Object> objs) {
```
wing328: @mminns it would be nice to add a 1-liner or proper Java docstring to explain what this function does.

wing328:
cc @jimschubert

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java
jimschubert: this method is bound to cause lots of conflicts. For instance, add System.IO to imports and create a model called File.

I think the only solution to this concern is fully qualifying all usages of generated code.

mminns: Hi @jimschubert I did wonder about that solution, fully qualifying all objects, however as far as I can see all if the imports, at least running when my tests against https://api.bitbucket.org/swagger.json have to be in the form "Model.ClassName" unless there is a configuration overriding the import mapping. It doesn't appear possible to have an import in the form System.IO.

To be fair I am new to Swagger so I may be missing something

jimschubert: @mminns 

Sorry, I should have been more clear.

Swagger Codegen allows users to specify their own modifications to templates using the following switches passed to the cli:

```
[(-t <template directory> | --template-dir <template directory>)]
```

So, for example, a user could copy the contents of `modules/swagger-codegen/src/main/resources/csharp` to a local directory such as `/src/csharp-template`. Any of the templates may be modified as long as the template names remain the same and the mustache syntax is valid.

As an example, if I've copied out the templates to `/src/csharp-template`, I can change `api.mustache` to include `System.IO`:

```
using System;
using System.Collections.Generic;
using System.Collections.ObjectModel;
using System.Linq;
using System.IO;
{{#netStandard}}
using RestSharp.Portable;
{{/netStandard}}
{{^netStandard}}
using RestSharp;
{{/netStandard}}
using {{packageName}}.Client;
{{#hasImport}}using {{packageName}}.{{modelPackage}};
{{/hasImport}}
```

I provided `File` as an example, because it's not uncommon for APIs to have models named as common API inputs/outputs. To demonstrate the conflict with `File`, you'd have to perform this hypothetical adjustment to the usings in your custom template.

However, the problem exists for every model of the existing usings. We had issues with users having conflicts on `Object` with the exiting `using System;`, which lead to the decision to fully qualify all model and infrastructure in the generated C# clients.

Without modifying the templates, you could experience the same conflicts using these not very far-fetched model names:

* Environment
* Type
* Uri
* Version
* A handful of types from RestSharp and RestSharp.Portable.

So, for a better example using the built-in templates, if I had a model called `DateTime` your change would yield:

```
using System;
…
using DateTime = IO.Swagger.Models.DateTime;
``` 

And in this scenario, it would be confusing to the consumers when `DateTime` refers to an aliased type and we're forcing a fully qualified `System.DateTime` anytime someone wants to use a *real* DateTime.

The answer to the above could just be "don't name your models like that". But there may be business needs we're unaware of, or C# could be one of many client SDKs being generated to which enforcing a rule like this could affect many clients of our consumer's generated codebase just to rename a model. It is unobtrusive, however, if every instance in the generated code that needs access to the swagger generated DateTime instance is fully qualified as `IO.Swagger.Models.DateTime` and we don't use aliases as they're done above. Couple that with our inability to know how the templates may have been modified by users, and aliases aren't really viable for us.

I hope that helps. Sorry for the previous short and confusing comment.

mminns: Funnily enough Version is exactly the class name that led me here :)

mminns: So in principal it sounds like the fix should be in code, to ensure the references to Model objects are always fully qualified?

If the fix was in the templates, prepending the namespace to model references, then it would still be at risk of the same issue where someone copied the templates, modified them and removed the namespace. Also the templates would potentially need to either understand the difference between Model references and core references to it didn't prefix the wrong ones.

If that makes sense :)

jimschubert: I think I get what you're saying.

It's not really possible to prevent developers from misusing the templates (removing the fully qualified model references). Your last comment about the templates knowing the difference between Model references and core references can easily be done in the generator's code when parameters are being mapped, because we know which parameters come from the swagger definition (they point to a `ModelRef`, if I remember correct). I know @wing328 was working on a solution to fully qualify models but I don't know the state of that work.

wing328: > I know @wing328 was working on a solution to fully qualify models but I don't know the state of that work.

Last time when I tried, it's not that straightforward due to issue with Array/Map of models (with fully-qualified names).

mminns: Hi. I've had a bit of rethink. I 'think' I have a solution, i.e. it works for me :),  but I'm going to pull this PR for now and resubmit after a bit more testing.

mminns:
Closed while I re-think my approach.s

PR #5582, opened 2017-05-08
Fix#5510

### LINKED ISSUE #5510: [C++][QT5] Enhancement : error management from API
##### Description

Current generated code for QT5 does litlle in term of error management.
The generated code looks like (from pet store sample):
```
void SWGPetApi::getPetByIdCallback(HttpRequestWorker * worker) {
    QString msg;
    if (worker->error_type == QNetworkReply::NoError) {
        msg = QString("Success! %1 bytes").arg(worker->response.length());
    }
    else {
        msg = "Error: " + worker->error_str;
    }
    
    QString json(worker->response);
    SWGPet* output = static_cast<SWGPet*>(create(json, QString("SWGPet")));
    
    worker->deleteLater();
    emit getPetByIdSignal(output);
}
```

from this snippet of code, we see that, no matter the error_type, the signal is emitted without any indication of an error. In case the signal has a parameter, the parameter is null when an error occurs.

This does not allow us to really determine error when calling the api.

##### Swagger-codegen version

2.2.2

##### Suggest a Fix

I have multiple proposal :

1. emit an errorSignal containing the error_type and the error_str and do not emit the success signal.
   This change breaks the flow of signal, since in case of error the success signal is not emitted anymore.

2. extend the success signal to contain the error_type and the error_msg. When error_type is different than NoError, the parameters is empty (as today)
    This change breaks the contract

3. emit an errorSignal containing the error_type and the error_str and still emit the success signal  as today.
     This is a conservative move to keep contract and flow. However, this makes the code less understandable for the future (Why do we emit the success signal ?).

4. emit the success signal with a QVariant containing either the succes object (or Void ?), either an Error Object. (if it is possible. I'm not confortable with QVariant). This would obfuscate the contract of the signal though. (coming C++17 std::variant would be better in that matter)
    This change breaks the contract.

5. a mix of 2 and 3. Keep the success signal as-is and add another success signal extended with error_type and error_str.
    This change is conservative for contract and flow. The one interested to errors just connect to the extended signal instead of the old signal.
    Drawback : all signals are doubled.
    Advantage : As it does not break contract and flow, it can go in a 2.2.3. In the 2.3.0, we may deprecate and remove the signal without the error.
Generated code will look like 
```
void SWGPetApi::getPetByIdCallback(HttpRequestWorker * worker) {
    auto errorType = worker->error_type;
    QString errorMsg = worker->error_str;

    QString msg;
    if (errorType == QNetworkReply::NoError) {
        msg = QString("Success! %1 bytes").arg(worker->response.length());
    }
    else {
        msg = "Error: " + worker->error_str;
    }
    
    QString json(worker->response);
    SWGPet* output = static_cast<SWGPet*>(create(json, QString("SWGPet")));
    
    worker->deleteLater();
    emit getPetByIdSignal(output);
    emit getPetByIdSignal(output,errorType,errorMsg);
}
```

sabras75:
### Description of the PR

fix #5510 

Add signals that support error information in addition to previous signal.
The new signal has the same name as previous message with a 'E' suffix. Eg:
```
void addPetSignal();
void getPetByIdSignal(SWGPet* summary);

void addPetSignalE(QNetworkReply::NetworkError error_type, QString& error_str);
void getPetByIdSignalE(SWGPet* summary, QNetworkReply::NetworkError error_type, QString& error_str);
```
This is because signal overloading usage is cumbersome for consumer code and it breaks compatibility.

The patch will not break compatibility with client of the previously generated code.  The new error management system is an addition and it is then optional to use it.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/qt5cpp/api-body.mustache
+++ b/modules/swagger-codegen/src/main/resources/qt5cpp/api-body.mustache
@@ -180,14 +180,8 @@ void
     {{/returnType}}
     worker->deleteLater();
 
-    {{#returnType}}
-    emit {{nickname}}Signal(output);
-    emit {{nickname}}Signal(output,error_type,error_str);
-    {{/returnType}}
-    {{^returnType}}
-    emit {{nickname}}Signal();
-    emit {{nickname}}Signal(error_type,error_str);
-    {{/returnType}}
+    emit {{nickname}}Signal({{#returnType}}output{{/returnType}});
+    emit {{nickname}}SignalE({{#returnType}}output, {{/returnType}}error_type, error_str);
 }
 
 {{/operation}}
--- a/modules/swagger-codegen/src/main/resources/qt5cpp/api-body.mustache
+++ b/modules/swagger-codegen/src/main/resources/qt5cpp/api-body.mustache
@@ -90,7 +90,8 @@ void
     HttpRequestWorker *worker = new HttpRequestWorker();
     HttpRequestInput input(fullPath, "{{httpMethod}}");
 
-    {{#formParams}}if ({{paramName}} != nullptr) {
+    {{#formParams}}
+    if ({{paramName}} != nullptr) {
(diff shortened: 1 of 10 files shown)


wing328:
@sabras75 thanks for the PR, which has been merged into master.

wing328:
@sabras75 when you've time, I wonder if you can contact me via my personal email address in my [Github profile page](https://github.com/wing328) as I've some questions for you regarding the C++ QT5 API client.

Issue #4962, opened 2017-03-07
Empty response body handling in typescript-angular2


aacanakin:
##### Description

Empty response 200 OK codes are not handled by auto generated api code.

##### Swagger-codegen version
2.2.2

##### Command line used for generation

language: typescript-angular2

##### Steps to reproduce

Build an api response just give 200 as http code but no response body

##### Suggest a Fix
Here's the current `api.mustache` at line 45;
```typescript
return response.json()
```
It should be;
```typescript
return response.json() || {};
```

wing328:
@aacanakin thanks for reporting the issue. Please submit a PR so that we can review the fix.

kenisteward:
@wing328 The change is running and is non breaking.  Check the PR @5585

wing328:
Thanks for the fix by @kenisteward 

@aacanakin please pull the latest master to give it a try.

macgyver2k:
please re-open; this does not work for me, and probaply for no one. when the .json() method is called on an empty body, an exception is thrown. checking for existing Content first should solve the issue, and works with my Code. 

`return Response
    .arrayBuffer().byteLength > 0 
        ? response.json() 
        : {};`

wing328:
@macgyver2k do you mind filing a PR with the suggested fix?

macgyver2k:
@wing328 yes i will do it.

kenisteward:
@wing328 @macgyver2k So I have done more research into this.  The way it works for angular 2 is based on this code:
https://github.com/angular/angular/blob/4.0.0/packages/http/src/body.ts#L13-L36

what this does is if there is data in the body and that data is a string or an array buffer it attempts to parse it as json.  so that means doing if _body is null / undefined or even {} no errors will be thrown!

In the case that _body is set to a string that is NOT parseable into a json format (e.g. the headers for the request are not set to accept json or the server is not set to serve json OR you are requesting a source that isn't json) then .json() will cause an error!

Because of this, even your solutions @macgyver2k won't work if the data isn't parsable to json!  We need another solution for this or we need  to make the restriction that you can only use the API to parse JSON.

We could do a try/catch and in case there is a failure parsing just return response.text() instead but that doesn't handle making requests for blobs and such.  Let me know what you guys think before I work on a full PR for it. 

cc @Vrolijkx Would like your opinion as well.

Vrolijkx:
Can't we just check on response.text().lenght and only call .json() if greater as 0?

[...]

PR #5585, opened 2017-05-08
#4962 Empty Response body handling in typescript-angular2 - 

kenisteward:
### PR checklist

- [ X] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [X ] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [ X] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

Followed directly what the suggestion said in #4962 as it is transparent and well handled that way.


kenisteward:
@wing328 This should be a quick check.

wing328:
@kenisteward agreed. The change looks good to me.

wing328:
@kenisteward thanks for the PR. Do we need to do the same for other Typescript generators?

e.g. typescript-fetch: https://github.com/kenisteward/swagger-codegen/blob/cdb456ef01cac6c39e5a7112433fa10d7777bb36/modules/swagger-codegen/src/main/resources/TypeScript-Fetch/api.mustache#L150

kenisteward:
@wing328 upon further investigation we don't have to change the fetch API unless we want to bring it on par with the angular API. When the response body won't be able to parse they just return the full response for fetch rather than the parsed body. That's fine if the users are okay with using the app differently for not parsable  response bodies (empty) In my honest opinion I think it is bad practice when you could just return an empty response which is what the problem is.

wing328:
@kenisteward thanks for the investigation. I'll leave those generators intact.

Issue #5589, opened 2017-05-09
[Csharp] Code using Netstandard "RestSharp.Portable" serializes JSON twice


voydz:
##### Description

Using the netstandard version of csharp to generate code uses restsharp.portable in the background. As the implementation differs from RestSharp.Net, the body in "application/json" requests is serialized twice.

##### Swagger-codegen version

current master branch @ fa2e575762dc71939cfcf8ae8bc1d058bce5cdab

##### Swagger declaration file content or url

Not exactly caused by a specific declaration, just any normal JSON model as request body.

##### Command line used for generation

```bash
java -jar swagger-codegen-cli.jar generate \
  -i "api.json" \
  -c "config.json" \
  -l "csharp" \
  -o "generated"
```

```js
// ... config.json
{
    "packageName": "My.Api",
    "targetFramework": "v5.0",
    "packageGuid": "<insert random package guid here>"
}
```

##### Steps to reproduce

1. Generate project with above command and config
2. Attach HTTP proxy debugger like Charles or Fiddler
3. Make random POST/PUT request with JSON body
4. Inspect request in HTTP proxy debugger
5. See double serialized JSON

##### Related issues

Introduced in #1557 ?

##### Suggest a Fix

Disabling the restsharp.portable serialization by respecting the slightly different api. PR is incoming.

**EDIT:** see #5590 for details

wing328:
Closed via #5590. Thanks for the fix from @voydz

PR #5590, opened 2017-05-09
[csharp] fixes for netstandard (json serializing and project files)

voydz:
### PR checklist

### Description of the PR

This is a non-breaking change.

see #5589 

As a side node: due to including `**/*.cs` as files to compile, there have always been some files of the `obj\` directory which where included.

This is fixed due to only include all files of the generated root directories, thus excluding the `obj\` directory.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/csharp/Project.mustache
+++ b/modules/swagger-codegen/src/main/resources/csharp/Project.mustache
@@ -88,7 +88,10 @@
     {{/netStandard}}
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="**\*.cs"/>
+    <Compile Include="Api\**\*.cs"/>
+    <Compile Include="Client\**\*.cs"/>
+    <Compile Include="Model\**\*.cs"/>
+    <Compile Include="Properties\**\*.cs"/>
   </ItemGroup>
   {{^netStandard}}
   <ItemGroup>
--- a/modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/csharp/ApiClient.mustache
@@ -122,6 +122,10 @@ namespace {{packageName}}.Client
             String contentType)
         {
             var request = new RestRequest(path, method);
+            {{#netStandard}}
+            // disable ResetSharp.Portable built-in serialization
+            request.Serializer = null;
+            {{/netStandard}}
 
             // add path parameter, if any
             foreach(var param in pathParams)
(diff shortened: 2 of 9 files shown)


wing328:
@voydz thanks for the fix. I did some tests and the result is good.

cc @jimschubert @Gronsak

modules/swagger-codegen/src/main/resources/csharp/Project.mustache
```diff
@@ -88,7 +88,10 @@
     {{/netStandard}}
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="**\*.cs"/>
+    <Compile Include="Api\**\*.cs"/>
```
voydz: @jimschubert Yes, i know the namespace can be changed (as I did, if you have a look at the related issue). However, as I can tell, the namespace has no influence on folder structure. 

I am using both, a default namespaced ("IO.Swagger") and a custom namespaced ("MyProject.Api") generated project. Both got the same folder structure for me.

jimschubert: I'm not sure I understand. Are you just setting `packageName` to "IO.Swagger" and "MyProject.Api"? Or are you changing `apiPackage` and `modelPackage`?

You'd set the api and model packages/namespaces with an additional switch, for example:

```
--additional-properties=apiPackage=PublicApi
```

And in this case, the modified glob patterns will cause the API classes to not be compiled in the new code.

wing328: I also did a test and didn't experience any issue after the change related to how files are included.

wing328: Btw, I'm using Xamarin Studio 6.2, MacOS.

voydz: @jimschubert Sorry, just recognized I answered not all of your questions. The issue with `**/*.cs` was, that `*.cs` files from the `obj/` dir were included, causing in some duplication compile errors.

We are using swagger-codegen in a Xamarin.Forms project. If you need further information, just mail me.

jimschubert: @wing328 did you load the code in an IDE or run via `compile-mono.sh`? The script performs the same recursive glob that was removed from the project file, so the script would still work as expected.

wing328: > did you load the code in an IDE or run via compile-mono.sh? 

IDE only. Will given compile-mono.sh a try later as I'm in the middle of something.

jimschubert: @wing328 sounds good. I'm going to check in Rider and Visual Studio 2015.

@voydz It sounds like you're experiencing a bug in Xamarin. I know prior to 6.0, there were issues with Build and Rebuild not executing cleanup tasks. I'll see if we can exclude `obj` from the glob in a way that works in Xamarin.

jimschubert: @voydz @wing328 I verified the change breaks solutions in Rider and Visual Studio. I opened #5605 to resolve the issue, but I don't have Xamarin installed and no space on my drive to install it. Could one of you verify the change works as expected in Xamarin.

Screenshots below to clarify the issue, where the solution doesn't display namespaces specified with apiPackage or modelPackage.

Visual Studio solution on master:
windows_10

Rider solution on master:
rider_1_0_eap

wing328: Let me run some tests...

voydz: @jimschubert @wing328 Just ran it in VS2017, rider and XS. It seems to work nicely.

jimschubert: @voydz Thank you, sir! And thanks for fixing the double serialization as well as exposing the `obj\*.cs` issue. Sorry for all the comments today, I just realized it is mid-week and it probably seems weird. I'm on paternity leave and had free time this morning for open source.

wing328: My test results also look good (I'd to reinstall mono in my Mac as the commands couldn't be found)

Issue #5596, opened 2017-05-10
[Java] Don't throw exception with unknown polymorphic type


bbdouglas:
##### Description

If a client receives an unknown discriminator label in a polymorphic type, it currently throws an exception deep in the feign stack. This makes it difficult to manage upgrades in a system of microservices, as it is difficult to co-ordinate every service upgrading simultaneously. To allow clients to gracefully degrade in the case unknown subtypes, it would be helpful for the stack not to throw an exception.

##### Swagger-codegen version

current (2.3.0)

##### Swagger declaration file content or url

Any polymorphic type, such as the Animal/Dog/Cat example in the sample spec.

##### Command line used for generation

N/A

##### Steps to reproduce

Imagine with the petstore example that the service started publishing a new animal called "Mouse" before the client upgraded its version of the spec.

##### Related issues

There are a number of issues generally related to polymorphism, such as #358, #1686 and #3904. But they touch on general support of the feature, while this is more on the topic of exceptional cases.

The one most closely related to this is #5436, which proposes an alternative solution to this problem. In all transparency, I work with @gstarnberger and we discovered together an issue with his approach. If you declare a default implementation as a subclass of the base class (such as UnknownAnimal as a subclass of Animal) then it makes it impossible to pass any responses with a subclass. That is, you can never return a Cat response, because Jackson would back off to an UnknownAnimal when it doesn't recognize the discriminator, and UnknownAnimal is not a subclass of Cat. Jackson recognizes this as an error at runtime and refuses (with yet another exception) to even attempt to parse the Cat.

##### Suggest a Fix

There is a feature of Jackson where it returns a null pointer instead of throwing an exception when it doesn't recognize a discriminator value. One achieves this by disabling the `FAIL_ON_INVALID_SUBTYPE` deserialization option. This is a nice approach because it gives the client the ability to degrade gracefully if it doesn't understand part of the response. It also exhibits parity with the feign client's handling of unknown enum values, which also returns null.

I have a small PR that makes this change.

wing328:
Closed via #5597

PR #5597, opened 2017-05-10
Disable Jackson FAIL_ON_INVALID_SUBTYPE in feign

bbdouglas:
With this change, Jackson does not fail even if it doesn't recognize a
discriminator type name. This is helpful when upgrading a microservice with
a new subtype while keeping compatibility with its old clients. The instance
is returned as null instead of throwing an exception deep in the feign
framework, allowing clients to gracefully degrade.

### PR checklist

### Description of the PR

This is a fix for #5596. It disables the Jackson FAIL_ON_INVALID_SUBTYPE deserialization feature so that unknown discriminator subtypes return null instead of throwing an exception.


cbornet:
Can you also modify other jackson based libraries (jersey, ...) ?

bbdouglas:
I went ahead and modified all of the projects that I could find that set properties on Jackson ObjectMappers. Ran bin/java-petstore-all.sh to update relevant samples.

wing328:
@cbornet thanks for reviewing the change

@bbdouglas thanks for the enhancement, which has been merged into master.

PR #5008, opened 2017-03-10
fix #2159 dataformatter error for iOS Swift

fumitoito:
### PR checklist

- [x] Read the [contribution guildelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [x] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

When trying to parse 24-hour time to 12-hour time, date formatter throws fatal error.
It's caused by lack of settings locale for date formatter.

see also #2159 .

Issue #2159, opened 2016-02-17
NSDateFormatter FatalError on iOS Swift


CMasterson:
Setting the device to use 12-hour time format then trying to parse a 24-hour time from JSON causes the formatter to fail and the Models.swift initialize method to throw a FatalError.

Example time string: 2016-02-17T20:00:00.000+00:00

Setting the formatters locale fixes this issue:

`let formatter = NSDateFormatter() //Models.swift Line 106

formatter.locale = NSLocale(localeIdentifier:"en_US_POSIX") //Set the formatter locale

formatter.dateFormat = format

return formatter`

wing328:
@CMasterson thanks for reporting the issue and suggesting a fix. Would you please file a PR so that we can review the change?

Ref: https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

CMasterson:
@wing328 The fix I suggested is in the generated code, I honestly have no idea where in the Swagger source code this would need to be implemented.

wing328:
A good starting point is https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/swift/Models.mustache#L16

After making the change, please run the following to ensure all test cases passed:
```
mvn clean && ./bin/swift-petstore.sh
cd samples/client/petstore/swift/SwaggerClientTests
mvn integration-test
```

PR #5602, opened 2017-05-10
fix(swift3): add en_US_POSIX locale to date formatters

julienfouilhe:
### PR checklist

### Description of the PR

The same thing has been done for the swift generator, it resolves the same issue, described there: https://github.com/swagger-api/swagger-codegen/issues/2159
https://github.com/swagger-api/swagger-codegen/pull/5008

The only file edited is this one: `modules/swagger-codegen/src/main/resources/swift3/Models.mustache`


wing328:
@julienfouilhe thanks for the PR.

cc @Edubits @jaz-ah @jgavris for review.

jgavris:
👍  possible / easy to isolate a unit test in the example suite?

julienfouilhe:
@jgavris Where are they located? I don't see any swift tests.

wing328:
@julienfouilhe you can find the test cases for Swift3 Petstore client under https://github.com/swagger-api/swagger-codegen/tree/master/samples/client/petstore/swift3

julienfouilhe:
@wing328 @jgavris After giving it some thought I'm not sure how to test that as it depends on the user's device locale. But here is a link to Apple's recommendations: https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/DataFormatting/Articles/dfDateFormatting10_4.html

> If you're working with fixed-format dates, you should first set the locale of the date formatter to something appropriate for your fixed format. In most cases the best locale to choose is en_US_POSIX, a locale that's specifically designed to yield US English results regardless of both user and system preferences. en_US_POSIX is also invariant in time (if the US, at some point in the future, changes the way it formats dates, en_US will change to reflect the new behavior, but en_US_POSIX will not), and between platforms (en_US_POSIX works the same on iPhone OS as it does on OS X, and as it does on other platforms).

jgavris:
I've had success swizzling `+[NSLocale currentLocale]` in tests (not in this project though). You could maybe do one or two examples.

wing328:
@jgavris @jaz-ah @Edubits shall we add the tests in a separate PR instead and merge this PR into master if the PR looks good and has been tested manually?

jaz-ah:
i'm ok w/ that @wing328

jgavris:
👍  as well, sorry for the delay

PR #5606, opened 2017-05-10
[Java][Spring] Add examples defined in the spec to Spring MVC server generator

wing328:
### PR checklist

### Description of the PR

Redo https://github.com/swagger-api/swagger-codegen/pull/5411 for 2.3.0 (not master) as it's a breaking change

### DIFF
--- a/modules/swagger-codegen/src/main/resources/JavaSpring/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/JavaSpring/api.mustache
@@ -16,6 +16,9 @@ import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RequestPart;
 import org.springframework.web.multipart.MultipartFile;
+{{^useSpringCloudClient}}
+import java.io.IOException;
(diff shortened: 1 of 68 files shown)


wing328:
The CI error will be fixed via https://github.com/swagger-api/swagger-codegen/commit/3e95f464e3763fcda4bb220c3c174c994377e223

wing328:
### Upgrade Note ###

An additional parameter `@RequestHeader("Accept") String accept` is added to the end of the method signature, e.g.
```
    ResponseEntity<Void> addPet(@ApiParam(value = "Pet object that needs to be added to the store" ,required=true )  @Valid @RequestBody Pet body, @RequestHeader("Accept") String accept);
```

The parameter contains the HTTP "Accept", which is needed for determining the format of the response (e.g. application/JSON, application/XML, etc)

pcbue:
@wing328 In my case i use the delegate pattern, which leads to fail with this modification.
Delegate pattern add a final delegate to the controller.
Your modification add a constructor for an ObjectMapper parameter, but no delegate.

Compilation failed, complaining for the field to not being initialized in every case, which is true.

Is there any way to skip the addition of the ObjectMapper ?

PR #5655, opened 2017-05-17
Fix #5653 so body parameter schema link matches model anchor

### LINKED ISSUE #5653: Link and label for body parameter uses TitleCase, not source case in models
##### Description

The link to the model for a body parameter uses TitleCase, but the anchor in models uses the exact case of the model. 
```
      <div class="param">thing <a href="#CreateThing">CreateThing</a> (required)</div>
```
In the Models section the schema named `createThing` is has an HTML anchor `createThing` not `CreateThing`:

```
    <h3 class="field-label"><a name="createThing">createThing - Create a Thing</a> <a class="up" href="#__Models">Up</a></h3>
```

The anchor name and href differ so the link does not work.

##### Swagger-codegen version

2.2.3-SNAPSHOT (latest)

https://gist.github.com/DavidBiesack/4f3008fb7a035b2aebf7068ebcd4aeca

mvn package
java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -l html -i /tmp/openapi.yaml -o /tmp

##### Suggest a Fix

In modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/StaticHtmlGenerator.java
add:

```java
  @Override
    public String toModelName(final String name) {
        return name;
    }
```
(I'll submit a PR)

DavidBiesack:
### Description of the PR

Override getModelName() in StaticApigenerator to return the name unmodified rather than apply TitleCase to it.


wing328:
@DavidBiesack thanks for the fix.

wing328:
Linked to #5653

PR #5662, opened 2017-05-18
Allow overriding the service delegate

mhvelplund:
### PR checklist

### Description of the PR

With this update it is now possible to choose an implementation class for the service delegate. By default the generated service will behave as normal, generating a stub implementation. But now it is also possible to add an init parameter to the the Jersey servlet in web.xml with a class name, eg.:

```
<servlet>
    <servlet-name>Swagger Servlet</servlet-name>
    <servlet-class>org.glassfish.jersey.servlet.ServletContainer</servlet-class>
... snip ...
    <init-param>
        <param-name>PetApi.implementation</param-name>
        <param-value>org.example.petshop.CustomPetApiService</param-value>
    </init-param>
    <load-on-startup>1</load-on-startup>
</servlet>
```
The generated PetApi class will now use the ``org.example.petshop.CustomPetApiService`` instead of using ``PetApiServiceFactory`` to make an instance of ``io.swagger.api.impl.PetApiServiceImpl``.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/JavaJaxRS/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/JavaJaxRS/api.mustache
@@ -18,6 +18,7 @@ import java.io.InputStream;
 import org.glassfish.jersey.media.multipart.FormDataContentDisposition;
 import org.glassfish.jersey.media.multipart.FormDataParam;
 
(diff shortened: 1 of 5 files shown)


wing328:
@mhvelplund thanks for the enhancement, which looks good to me.

PR #5682, opened 2017-05-22
Fix for issues 5576 and 5577 related to generating CSharp clients . Attempt 2

mminns:
### PR checklist

Follow up PR for https://github.com/swagger-api/swagger-codegen/pull/5581

Rebased onto Master and setup Travis to run the build, it is currently failing, but so is master. see
https://travis-ci.org/mminns/swagger-codegen/branches

**Update**
The PR builds all pass, despite the travis failures shown above.

### Description of the PR

Corrects the code for finding parents of Models and removes ambiguous references in generated API class by explicitly referencing Model classes

This time rather than adding "usings" explicitly reference Model classes in the APIs

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
@@ -17,7 +17,7 @@ public class CodegenOperation {
             isResponseBinary = false, isResponseFile = false, hasReference = false,
             isRestfulIndex, isRestfulShow, isRestfulCreate, isRestfulUpdate, isRestfulDestroy,
             isRestful;
-    public String path, operationId, returnType, httpMethod, returnBaseType,
+    public String path, operationId, returnType, returnTypePackageName, httpMethod, returnBaseType,
             returnContainer, summary, unescapedNotes, notes, baseName, defaultResponse, discriminator;
     public List<Map<String, String>> consumes, produces, prioritizedContentTypes;
     public CodegenParameter bodyParam;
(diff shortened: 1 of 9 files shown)


wing328:
@mminns thanks for the PR.

cc @jimschubert

mminns:
This seems to have gotten a bit stuck in the mud, so I'm going to close it. 

The bug is real, the CSharp generator doesn't cope will with snake_case names in swagger.json, found using the Bitbucket swagger.json, but this may not be the right fix. 

Without this fix, I couldn't move on to create [Refit ](https://github.com/paulcbetts/refit) based generator within this project so rather than pursue  this PR I created a new generator here: https://bitbucket.org/mminns-itofinity/swagger-csharp-refit

wing328:
@mminns I'm sincerely sorry for the delay.

We've just formed the [technical committee](https://github.com/swagger-api/swagger-codegen#members-of-technical-committee) to help review PRs and issues based on langauges as I can't review so many PRs and don't want to become the bottleneck.

@mandrean please review when you've time (reopening for the time being)

jimschubert:
@mminns @wing328 @mandrean Sorry, for some some reason I never received notification from the previous tag.

I've had numerous refactoring issues with partially qualified names such as is in this PR. This may not be an issue if we assume users are only ever regenerating code as needed. But, if a team does spec-first design as my teams have done in the past, a need to rename the namespace for Models is no longer a find & replace and Visual Studio has never been smart enough to update partial namespace types (I haven't tried 2017, could be different).

Is there a constraint that prevents us from using fully qualified types I'm parameter lists?

Also, I apologize if my responses are slow. My mother is in the hospital with a spinal tumor, and I'll have limited free time in the coming months as my family deals with these health issues.

PR #5695, opened 2017-05-23
PR-1334 Improved angular 2 - typescript generation

iain17:
A bunch of fixes for the angular 2 typescript client generation. I had to make these changes to make use of this project. Large problems fixed, such as making it typescript compliant (my app just refused to compile without these fixes). I've updated the tests as well :)

Other problems that have now been fixed:
- Models being generated as interfaces. They are now classes.
- naming conflicts when it comes to enums.
- array of a definition would result in a generated typescript syntax error.

Issue #1334 .


wing328:
@iain17 thanks for the PR. 

It looks to me it's a breaking change so it should be filed against 2.3.0 branch instead.

Also you may want to checkout 2.3.0 to generate TS Angular2 client, which has a lot of enhancements and bug fixes.

wing328:
@Vrolijkx @taxpon @TiFu @kenisteward , please take a look when you've time. Thanks!

modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
```diff
@@ -46,8 +46,11 @@ export class {{classname}} {
             .map((response: Response) => {
                 if (response.status === 204) {
                     return undefined;
-                } else {
+                }
+                try {
                     return response.json() || {};
+                }catch(e){
```
TiFu: There should be a space between `}` and `catch`.

modules/swagger-codegen/src/main/resources/typescript-angular2/variables.mustache
```diff
@@ -1,9 +1,9 @@
 import { {{{injectionToken}}} } from '@angular/core';
 
-export const BASE_PATH = new {{{injectionToken}}}('basePath');
+export const BASE_PATH = new {{{injectionToken}}}<String>('basePath');
```
TiFu: Can you elaborate on this change?  Shouldn't it be `string` like before?

samples/client/petstore/typescript-angular2/default/api/PetApi.ts
```diff
@@ -1,9 +1,9 @@
 /**
  * Swagger Petstore
- * This is a sample server Petstore server.  You can find out more about Swagger at [http://swagger.io](http://swagger.io) or on [irc.freenode.net, #swagger](http://swagger.io/irc/).  For this sample, you can use the api key `special-key` to test the authorization filters.
+ * This is a sample server Petstore server.  You can find out more about Swagger at <a href=\"http://swagger.io\">http://swagger.io</a> or on irc.freenode.net, #swagger.  For this sample, you can use the api key \"special-key\" to test the authorization filters
  *
  * OpenAPI spec version: 1.0.0
- * Contact: apiteam@swagger.io
+ * Contact: apiteam@wordnik.com
```
TiFu: This change was probably not intended. Could you please revert it?

wing328: The latest master is using `petstore.yaml` instead of `petstore.json` (which still references wordnik.com) so rebasing the PR on the latest master should avoid the above change.

kenisteward:
Hello @iain17 please look at my comment for #1334 and provide an explanation of why your code doesn't compile using the latest changes. I have had no issues using interfaces with my application so I worry you will be making everyone rewrite code when they upgrade for a reason that Is lovalized and not broad.

modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
iain17: sure thing.

modules/swagger-codegen/src/main/resources/typescript-angular2/variables.mustache
iain17: Please see: https://angular.io/docs/ts/latest/api/core/index/InjectionToken-class.html
I changed it because my app just wouldn't compile otherwise.

iain17:
@wing328 Thanks for your comment. I'll look into it and apply the necessary changes.

iain17:
@kenisteward The main issue I had with using interfaces was the fact that I couldn't create an new instance of that model. The other was that in most languages it is bad practice to use interfaces to describe data types. Usually, you have structs or classes that do that job.

However, I understand your point and reading more into this, I see that it's common practice in Typescript to use interfaces to describe data types (see for instance: https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-aliases).

I'll make the necessary changes and try and apply them to the current project I'm doing, see how it works out for me. Right now it feels more natural to me to say: ```new User();``` which cannot be done with an interface.

kenisteward:
It's much more succinct to just do

Const u : IUser = {
}

As a is developer I can tell you more and more you get used to not using
new cause honestly it mostly isn't needed unless you want to actually
define custom logic or functions

modules/swagger-codegen/src/main/resources/typescript-angular2/variables.mustache
TiFu: That's weird. Can you post a minimum working example which doesn't compile?
`string` is the TypeScript string type and String is the default JavaScript string type, which is why I am wondering if the underlying issue is something else.
See e.g. [here](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html)

kirpit:
> honestly it mostly isn't needed unless you want to actually define custom logic or functions

That's the thing, we sometimes do need to define different behaviours and alternative implementations based on inheritance. To do that in the right way, I believe there should be classes and sometimes both the classes and interfaces needed, whether it creates some upgrading work for some or not. 

I was actually considering to make this change and file a PR, as it looks like I'm a bit late. 

@wing328 you marked the milestone to `v2.2.3` but you also mentioned this should go against `v2.3.0`. Is it intentional?

@kenisteward Should the new model classes work consistently with the new `withInterfaces` parameter I introduced recently here in #5519?

kenisteward:
If you want to make classes just have them implement the generated ones.
You'll get the properties that are defined in the interface ( that's a TS
thing ) with no need to change the generation model

wing328:
> @wing328 you marked the milestone to v2.2.3 but you also mentioned this should go against v2.3.0. Is it intentional?

I'll update it once @iain17 changes this PR for 2.3.0 (there's an "Edit" button next to the subject line)

sebastianhaas:
@wing328  I too would like to vote against this change. I don't see any reason to move from interfaces to classes, and as @kenisteward points out
> If you want to make classes just have them implement the generated 

@Vrolijkx what's your opinion on this? 

> The other was that in most languages it is bad practice to use interfaces to describe data types.
Would you care to elaborate on this @iain17 ? While I don't think this is true for any of the programming languages I know, I sincerely have my doubts on this when it comes to Typescript. An Interface in Typescript will just disappear after compiling, why throw that away? Also, you don't even have to have a class implement an Interface to create an object of that type.

> The main issue we had with using interfaces was the fact that we couldn't create a new instance of a model. 

I feel like this is not true. Whats wrong with this code?
```typescript
interface A {
    abc: string;
}

let test: A = {
    abc: "def"
};
```

wing328:
@iain17 when you've time, I wonder if you can file another PR to just do the following:
- naming conflicts when it comes to enums.
- array of a definition would result in a generated typescript syntax error.

Please pull the latest master as there're some enhancements to TS generators (and see if the above issues have been addressed)

PR #5700, opened 2017-05-24
[docker] Run as non-root user, fixes ownership of generated artefacts

caseyfw:
### PR checklist

### Description of the PR

This change uses the [official maven docker image's method](https://github.com/carlossg/docker-maven#running-as-non-root) to set the user in the docker container to the same user who invokes the `run-in-docker.sh` script. The outcome of this is that all the artefacts generated by the container (m2 cache in home dir, target directories, output client and server code, etc) are owned by the expected user instead of root.


caseyfw:
Closing this to re-roll for master branch.

PR #5701, opened 2017-05-24
[php] 2.3 - fix form params json encoding

baartosz:
### PR checklist

### Description of the PR
Issue introduced when client was reworked to use guzzle instead of curl in https://github.com/swagger-api/swagger-codegen/pull/5190.

Apparently I missed this case earlier. Content of request is not json-encoded if endpoint `consumes: application/json` and params are in `formData`. This PR fixes this issue.
I added example of spec causing this issue to FakeApi definition so I could add test for it.

```
  /fake/jsonFormData:
    get:
      tags:
        - fake
      summary: test json serialization of form data
      description: ''
      operationId: testJsonFormData
      consumes:
        - application/json
      parameters:
        - name: param
          in: formData
          description: field1
          required: true
          type: string
        - name: param2
          in: formData
          description: field2
          required: true
          type: string
      responses:
        '200':
```


modules/swagger-codegen/src/main/resources/php/api.mustache
```diff
@@ -237,22 +248,14 @@ use {{invokerPackage}}\ObjectSerializer;
                 }
                 $httpBody = new MultipartStream($multipartContents); // for HTTP post (form)
 
+            } elseif ($headers['Content-Type'] === 'application/json') {
```
wing328: @baartosz do you mind adding a new method (e.g. isJsonMime) to determine if a MIME string is "JSON"?

For Java, we've https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/java/okhttp-gson/src/main/java/io/swagger/client/ApiClient.java#L752-L754

and here are some tests:

https://github.com/swagger-api/swagger-codegen/blob/fc01457d77315afdd023c45587d6f48a26a2d131/samples/client/petstore/java/okhttp-gson/src/test/java/io/swagger/client/ApiClientTest.java#L81-L102

wing328:
@baartosz thanks for the PR. I left you a feedback in the code.

cc @arnested

modules/swagger-codegen/src/main/resources/php/api.mustache
wing328: I'll create a separate task to track this.

wing328:
@baartosz when you've time, I wonder if you can contact me via the email address in my Github profile as I've some questions for you (not related to this PR). Thanks.

Issue #5558, opened 2017-05-04
[HTML2] Curl example copies HTTP verb(GET/POST) in lower case in Firefox


svarogg:
##### Description

The curl example shows i.e. `curl -X GET "http://localhost:8000/greetings/hello/{subject}"`.
However, if you copy it in firefox (In chromium it does not reproduce) you get `curl -X get "http://localhost:8000/greetings/hello/{subject}"` which fails if run from the command line.

It seems like the verb is capitalized through `text-transform: uppercase;` which does not get represented when copied from firefox.

##### Swagger-codegen version
v2.2.2

##### Swagger declaration file content or url
This bug reproduces for any swagger file.
For example, you can use the hello-world sample from https://github.com/OAI/OpenAPI-Specification/wiki/Hello-World-Sample

##### Command line used for generation
java -jar /tmp/swagger-codegen-cli.jar generate -i swagger.json -l html2

##### Steps to reproduce

1. Create any json or yaml swagger specification
2. Run the above command line to generate an index.html file
3. Open said html file with Firefox
4. Copy the curl sample to the clipboard, paste in terminal and run.

##### Related issues
Firefox bug to fix this issue upstream:
https://bugzilla.mozilla.org/show_bug.cgi?id=35148
Doesn't look like they are planning to fix it anytime soon.

##### Suggest a Fix
Actually capitalize the http-verb instead of just the way it's displayed.
It seems like this is not possible through mustache, so maybe use a small javascript snippet to do that.

wing328:
@svarogg I agree that we should do it in mustache instead of with Javascript.

May I know if you've time to contribute the fix (which should be pretty straightforward)? I can show you some good starting points.

svarogg:
I've began looking into a solution...
It seems like the problem is in swagger-codegen/src/main/resources/htmlDocs2/index.mustache line 227., specifically: `<span style="text-transform: uppercase;">{{httpMethod}}</span>`

It would be nice if we could tell mustache to capitalize the thing, but apparently it's not possible by-design (See http://stackoverflow.com/questions/8460720/mustache-jquery-javascript-how-to-perform-method-on-mustache-variable).

So, unless I misunderstand something, we have two options I can think of:
1) Capitalize with Javascript. Would work, but very ugly.
2) Capitalize the source. I guess the source goes to all other templates, and maybe used in places where capitalization is not desirable.
A solution could be the introduction of another value (say httpMethodCapitalized or something).

[...]

PR #5705, opened 2017-05-25
Made httpMethod in curl example actually upper case in html2 to support firefox (#5558)

svarogg:
### PR checklist

I am not sure if this is considered a breaking change or not, so pushed to master, if there's a need, let me know, and will change PR to 2.3.0

### Description of the PR
Added x-codegen-httpMethodUpperCase variable to vendorExtensions, and using it instead of httpMethod in the CURL example.

This is in reference to #5558


wing328:
@svarogg the change looks good to me. Thanks for your contribution.

PR #5716, opened 2017-05-26
[cpprest] fix cpprest model source mustache template

apopiak:
which was generating code that doesn't compile because of a wrong variable name

I have run the `cpprest-petstore.sh` and everything seems fine. I did not get changes in the generated files which I suspect is the reason this was overlooked in the first place (the relevant part of the template is not used for any generation in the petstore example).


wing328:
@apopiak thanks for the PR. Let me know if you need help updating the Petstore sample for CppRest API client.

What we want to do next is to setup some testings in CI (travis) so as to make sure the CppRest API client at least compiles without issue. If you've any idea on how to do that, please let us know.

apopiak:
AFAICT this is everything that needs to be changed? Can you tell me how I would need to modify anything? As I wrote: it does not change the output of the example.

wing328:
@apopiak you're right. I did another test and confirm no change in the C++ (cpprest) Petstore sample. PR has been merged into master. Thanks for your contribution.

PR #5722, opened 2017-05-29
StaticHtml2Gen | Speed up appDescription rendering

JuH33:
Pre-format Html for speed up the render on page loading.


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/StaticHtml2Generator.java
```diff
@@ -195,6 +199,15 @@ public CodegenOperation fromOperation(String path, String httpMethod, Operation
         return op;
     }
 
+    private void preparHtmlForGlobalDescription(Swagger swagger) {
```
wing328: @JuH33 thanks for the PR. It would be nice to add a 1-liner or a proper docstring explaining what this function does.

JuH33: My pleasure. Of course i can do that :)

Issue #5071, opened 2017-03-15
[Kotlin] Add Kotlin API client generator


wing328:
##### Description

[Kotlin](https://kotlinlang.org/) is a statically-typed programming language that runs on the Java Virtual Machine and also can be compiled to JavaScript source code. Its primary development is from a team of JetBrains programmers based in Saint Petersburg, Russia (the name comes from the Kotlin Island, near St. Petersburg).[3] Kotlin was named Language of the Month in the January 2012 issue of Dr. Dobb's Journal.[4] While not syntax compatible with Java, Kotlin is designed to interoperate with Java code and is reliant on Java code from the existing Java Class Library, such as the collections framework.

Ref: https://en.wikipedia.org/wiki/Kotlin_(programming_language)

If anyone from the community wants to contribute and need help to start working on the generator, please reply to let us know.

cbornet:
Note : Jetbrain's IDEA can transform Java code to Kotlin. The produced code is generally almost direclty usable. You just have to tweak some issues with optionals and visibility.
So it shouldn't be too much work to transform the templates from Java to Kotlin.

jimschubert:
@wing328 I'd like to take this up within the next month or so if nobody else wants to do it. I have an existing Kotlin client project which should be fairly easy to convert over.

@cbornet while the Java -> Kotlin conversion often "works", it is neither idiomatic or ideal. In many cases, the generated code will have horrible performance.

michalharakal:
I'm really curious about your implementation @jimschubert, looking forward to see it. What is a branch/repo you are working on?

jimschubert:
I haven't had a chance to start work on this yet. The past few weeks have been pretty hectic. The existing code I mentioned is private, so it can't be shared.

I'll try to find time for this over the next couple of weeks.

michalharakal:
@jimschubert I'm just starting to learn Kotlin. As a starting point I have found this blog very useful [Retrofiting on Android with Kotlin](https://segunfamisa.com/posts/using-retrofit-on-android-with-kotlin)
Lets see if I can contribute to project too ...

jimschubert:
@michaelkourlas If you'd like to work on the Kotlin generator, I can leave it to you. If you'd like help, we could work together on it.

Here's a script to get all the pieces in place for a new generator. I was going to include this in my PR for the kotlin client:

https://gist.github.com/jimschubert/d9548527b8fce41b3922d2a12abb01e7

[...]

PR #5727, opened 2017-05-29
Kotlin client

jimschubert:
### PR checklist

### Description of the PR

Initial implementation of a kotlin client generator. See #5071

There are currently some missing pieces to this, such as enum support (see findPetsByStatus) and correctly handling default response codes (see addPet, returns Unit rather than Pet). This should be a good starting point, at least, for community contributions to further flesh it out.

This generator fully qualifies kotlin and java types to reduce conflicts with swagger defined models.

I'll be going on vacation later this week, so I may not be able to revisit the enum generation and default responses for a couple weeks.

\cc @wing328 @michalharakal


jimschubert:
reference prototype project: https://github.com/jimschubert/kotlin-petstore-example/

jimschubert:
An example test suite:

```
package io.swagger.petstore.functional

import io.kotlintest.matchers.shouldBe
import io.kotlintest.matchers.should
import io.kotlintest.matchers.beGreaterThan
import io.kotlintest.specs.ShouldSpec
import io.swagger.client.apis.PetApi
import io.swagger.client.models.Category
import io.swagger.client.models.Pet
import io.swagger.client.models.Tag

class EvaluateTest : ShouldSpec() {
    init {
        should("query against pet statuses") {
            val api = PetApi()
            val results = api.findPetsByStatus(arrayOf("available", "pending"))

            results.size should beGreaterThan(1)
        }

// TODO: Handle default (200) response
/*
        should("post data (new pet)") {
            val api = PetApi()
            val pet = Pet(
                    id = 0,
                    name = "kotlin client test",
                    category = Category(0, "string"),
                    tags = listOf(Tag(0, "string"))
            )
            val result = api.addPet(pet)

            result.name shouldBe(pet.name)
            result.category shouldBe(pet.category)
            result.tags shouldBe(pet.tags)
        }
*/
    }
}
```

The `query against pet statuses` test fails because the generator generates `Pet` as:

```
data class Pet (
  val id: kotlin.Long?,
  val category: Category?,
  val name: kotlin.String,
  val photoUrls: kotlin.Array<kotlin.String>,
  val tags: kotlin.Array<Tag>?,
  /* pet status in the store */
  val status: kotlin.String?
)
```

Where `name` and `photoUrls` are the required properties defined in `petstore.yaml`. However, when querying `findPetsByStatus`, Pets exist in the array without a `name` property. I haven't dug into why this is, or any settings available in moshi to loosen this behavior. The test fails, and I believe that's to be expected with the given data.

wing328:
> There are currently some missing pieces to this, such as enum support (see findPetsByStatus) and correctly handling default response codes (see addPet, returns Unit rather than Pet). This should be a good starting point, at least, for community contributions to further flesh it out.

Yeah. Definitely a very good starting point.

> I'll be going on vacation later this week, so I may not be able to revisit the enum generation and default responses for a couple weeks.

Enjoy your vacation 👍 

And thanks for the "new.sh" to make it much easier in creating new generators 👏

cbornet:
:clap: now we definitely need the retrofit library !

jimschubert:
I tried to structure ApiClient to be agnostic of the underlying libraries. Hopefully this makes it easier to include retrofit. I tried to make it as small of a library as possible for the base generator.

Issue #5730, opened 2017-05-29
[Kotlin] Improve client generator functionality


jimschubert:
##### Description

This is a tracking issue for adding functionality to the Kotlin client generator.

- [ ] Support "default" responses. Currently 200 OK isn't handled, for example, in `PetApi#addPet`.
- [ ] Support enum type generation
- [ ] Support inline/embedded enums
- [ ] Support `collectionFormat` (multi)
- [ ] Test/Implement arrays of maps, arrays, objects
- [ ] Test/Implement maps of arrays/maps, objects
- [ ] Support allOf via composition
- [ ] Support allOf via inheritance
- [ ] (optional?) Per-request modification of headers,authorization
  - NOTE: ApiClient.defaultHeaders can be modified *once* to globally set headers for auth, user-agent, etc.
  - [ ] Test, or better implement basic auth support
   - [ ] Test, or better implement OAuth support
  - [ ] Test, or better implement apiKey auth
- [ ] Generate Api tests
- [ ] Generate Model tests
- [ ] Generate Api docs
- [ ] Generate Model docs
- [ ] Update README.md to be more project-specific rather than generic information about the generator/template

##### Related issues

* #5071

ALander26:
+1 set defaultHeaders

onionhammer:
Any plans for using [async](https://github.com/codepath/android_guides/wiki/Using-OkHttp#asynchronous-network-calls) ?

olegivo:
> Add support for Retrofit library

+1

8secz-johndpope:
supposedly this branch has successfully spat out kotlin + retrofit 
https://github.com/kroegerama/kotlin-retrofit-generated/blob/master/petstore/src/main/java/com/kroegerama/petstore/RetrofitHolder.kt

using this branch https://github.com/kroegerama/openapi-generator/tree/kotlin-retrofit
@kroegerama


```shell
java -jar modules/openapi-generator-cli/target/openapi-generator-cli.jar generate  \
   -i openapi.json \
   -g kotlin  --library=retrofit \
   -o kotlin_client
Exception in thread "main" java.lang.RuntimeException: Unknown library: retrofit
Available libraries:
  NONE
	at org.openapitools.codegen.DefaultCodegen.setLibrary(DefaultCodegen.java:3802)
	at org.openapitools.codegen.config.CodegenConfigurator.toClientOptInput(CodegenConfigurator.java:566)
[...]

PR #5731, opened 2017-05-30
[kotlin] api/model docs

jimschubert:
### PR checklist

### Description of the PR

This commit adds Api/Model/Auth documentation to the generated
README.md. Because auth support is not yet fully implemented (users can
manually set default headers globally), there aren't examples for helper
auth methods.

Models with inline enums document allowed values rather than pointing to
a generated enum class.

Two global additionalProperties were added (`generateApiDocs`,
`generateModelDocs`) to allow templates to conditionally display
documentation depending on these mutually exclusive settings. All
current generators supporting docs will attempt to link to generated
models when only api docs are specified.

This also moves the `$@` bash parameter in `bin/kotlin-client-petstore.sh`
to the end of the args variable. This is because `$@` can only be used to
pass System properties like `-DdebugModels`, can can already be passed as:

```bash
JAVA_OPTS="$JAVA_OPTS -DdebugModels" ./bin/kotlin-client-petstore.sh
```

By moving the `$@` to the end of the args, it allows us to pass additional
properties and other switches directly to the script.

**NOTE** I didn't regenerate samples because I've moved the kotlin output from `samples/client/kotlin` to `samples/client/petstore/kotlin` as part of PR #5729. I wanted to avoid too many merge conflicts and figured docs would get generated eventually anyway. Here are some samples:

---

# Pet

## Properties
**id** | **kotlin.Long** |  |  [optional]
**category** | [**Category**](Category.md) |  |  [optional]
**name** | **kotlin.String** |  | 
**photoUrls** | **kotlin.Array&lt;kotlin.String&gt;** |  | 
**tags** | [**kotlin.Array&lt;Tag&gt;**](Tag.md) |  |  [optional]
**status** | [**inline**](#StatusEnum) | pet status in the store |  [optional]

<a name="StatusEnum"></a>
## Enum: status
Name | Value
---- | -----
status | available, pending, sold

---

# StoreApi

All URIs are relative to *http://petstore.swagger.io/v2*

Method | HTTP request | Description
------------- | ------------- | -------------
[**deleteOrder**](StoreApi.md#deleteOrder) | **DELETE** /store/order/{orderId} | Delete purchase order by ID
[**getInventory**](StoreApi.md#getInventory) | **GET** /store/inventory | Returns pet inventories by status
[**getOrderById**](StoreApi.md#getOrderById) | **GET** /store/order/{orderId} | Find purchase order by ID
[**placeOrder**](StoreApi.md#placeOrder) | **POST** /store/order | Place an order for a pet

<a name="getInventory"></a>
# **getInventory**
> kotlin.collections.Map&lt;kotlin.String, kotlin.Int&gt; getInventory()

Returns pet inventories by status

Returns a map of status codes to quantities

### Example
```kotlin
// Import classes:
//import io.swagger.client.infrastructure.*
//import io.swagger.client.models.*

val apiInstance = StoreApi()
try {
    val result : kotlin.collections.Map<kotlin.String, kotlin.Int> = apiInstance.getInventory()
    println(result)
} catch (e: ClientException) {
    println("4xx response calling StoreApi#getInventory")
    e.printStackTrace()
} catch (e: ServerException) {
    println("5xx response calling StoreApi#getInventory")
    e.printStackTrace()
}
```

### Parameters
This endpoint does not need any parameter.

### Return type

**kotlin.collections.Map&lt;kotlin.String, kotlin.Int&gt;**

### Authorization

[api_key](../README.md#api_key)

### HTTP request headers

 - **Content-Type**: Not defined
 - **Accept**: application/json


jimschubert:
This PR addresses a few todos from #5730

Issue #5531, opened 2017-05-02
Review serialization and deserialization of `date` property 


wing328:
##### Description

Review serialization and deserialization of `date` property as it might not be (de)serialized correctly to `full-date` in RFC3339

##### Swagger-codegen version

Latest master

##### Related issues

https://github.com/swagger-api/swagger-codegen/issues/5513

##### Suggest a Fix

We'll need to write test cases in Petstore sample client to cover this.

Issue #5513, opened 2017-04-29
[C#] swagger-codegen, date datatype in swagger, and services that expect/validate date only


sumo300:
##### Description

We currently use a service bus platform called Tibco to develop RESTful servics for our enterprise.  Tibco supports both the `date` and the `datetime` data types and validates them as such in requests to the services.  `date` expects a strict date with no time element.  However, when these services are used from C#, which swagger-codegen generates properties as `DateTime?`, the service throws an exception because the value submitted always comes with a time element.

The issue in C# is discussed at length [here](http://stackoverflow.com/questions/5314309/a-type-for-date-only-in-c-sharp-why-is-there-no-date-type).

Is there a way to solve this problem without forcing our service developers to always use `datetime` and ignore the time element?  We've tried just making it a `string`, but that feels wrong as well.

##### Swagger-codegen version

`2.2.2`

##### Command line used for generation

`generate -i "swagger.json" -l csharp -c "config.json" -o ".\SomeProject" --model-name-prefix Ent`

Contents of `config.json`

```
{
  "packageName": "SomePackageName"
}
```

##### Steps to reproduce

1. Define swagger that uses a `date` data type for an input parameter
2. Generate C# API using swagger-codegen
3. Call service using generated API, passing a C# date value

wing328:
@sumo300 thanks for reporting the issue. Seems like we need to provide a converter function as described in http://stackoverflow.com/questions/18635599/specifying-a-custom-datetime-format-when-serializing-with-json-net

May I know if you've time to contribute the enhancement? A good starting point is https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/csharp/model.mustache and the {{#isDate}} mustache tag.

jimschubert:
@wing328 this should be considered a breaking change. I can pick it up within the next week or so if @sumo300 doesn't call dibs.

jimschubert:
Also, proper formats for `date` and `date-time` are documented [here](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types). This may affect more generators than just the C# ones.

wing328:
> Also, proper formats for date and date-time are documented here. This may affect more generators than just the C# ones.

Yup, I've created https://github.com/swagger-api/swagger-codegen/issues/5531 for tracking.

wing328:
>  this should be considered a breaking change.

[...]

PR #5734, opened 2017-05-30
[csharp] date proper format

jimschubert:
### PR checklist

### Description of the PR

Per spec (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types), DateTime instances defined as 'date' datatype need to be serialized as defined by full-date - RFC3339, which has the format:

```
full-date       = date-fullyear "-" date-month "-" date-mday
```

ref: https://xml2rfc.tools.ietf.org/public/rfc/html/rfc3339.html#anchor14

I've also added a unit test to FormatTestTests.cs which verifies the date-only format when serialized to JSON. I believe this test runs as part of CI, but it can be run manually with:

```
cd samples/client/petstore/csharp/SwaggerClient
sh mono_nunit_test.sh
```

When I run this locally, I have two failing tests but they're related to FindPetsByStatusTest and FindPetsByTagsTest neither of which are touched by this change (Pet, Tag, Category don't have any `date` properties, either).

see #5513
see #5531


jimschubert:
@wing328 per convo in linked issue, while this is a bug in general in that we're not properly handling "date" in most (all?) generators, this is a pretty big breaking change. I opened against master, but this may be more appropriate against the 2.3.0 branch because of the client+server impact of the change for existing consumers.

wing328:
I know some generators already have it fixed or worked around but totally agree with you that we need to review all the other generators to have the issue covered.

To me, it's ok for this fix to get into master as this is correcting the incorrect behaviour/handling of 'date'

wing328:
FYI. Fixed minor warning due to missing XML doc via 4fd52d1

jimschubert:
@wing328 thanks for the heads up. oddly, I'm not getting warnings about missing pubic XML docs. I'll have to see if I have a global setting somehow that is preventing those warnings, so you don't have to keep fixing them ☘️

PR #5742, opened 2017-05-31
[C++] Restbed Server Stub Code Generator

stkrwork:
### PR checklist

### Description of the PR

This is the implementation for the Restbed Code Generator. Its for C++.


modules/swagger-codegen/src/main/resources/META-INF/services/io.swagger.codegen.CodegenConfig
```diff
@@ -9,6 +9,7 @@ io.swagger.codegen.languages.CSharpClientCodegen
 io.swagger.codegen.languages.ClojureClientCodegen
 io.swagger.codegen.languages.ConfluenceWikiGenerator
 io.swagger.codegen.languages.CppRestClientCodegen
+io.swagger.codegen.languages.RestbedCodegen
```
wing328: @stkrwork please follow the alphabetical order in that file.

stkrwork: Done

stkrwork:
I added the changes.

stkrwork:
I replaced the string.join function with a manual way.

wing328:
@stkrwork thanks for the contribution, which has been merged  into master.

wing328:
Updated README.md with @stkrwork as the owner for C++ Restbed template: https://github.com/swagger-api/swagger-codegen/blob/master/README.md#template-creator

wing328:
I'll send out a tweet about this new generator later today: http://twitter.com/wing328

wing328:
@stkrwork for your upcoming contributions, I would suggest you to create a new branch instead of using master.

wing328:
@stkrwork one question about setter for model properties that are "array" if you don't mind. 

Is it correct to say that the `tags` property in the Pet model is missing the setter? https://github.com/swagger-api/swagger-codegen/blob/master/samples/server/petstore/restbed/model/Pet.cpp#L92-L95

If yes, what would the correct setter look like for array of something (string, model, etc)?

Related issue: https://github.com/swagger-api/swagger-codegen/issues/5862#issuecomment-309322587

stkrwork:
the mustache template for the models is actually copied from the cpprest model template, but I changed the json functions, and removed the model base template. Might be that the original cpprest model template was false.

If this is the getter:
`std::vector<std::shared_ptr<Tag>>& Pet::getTags()
{
    return m_Tags;
}`

Then the setter should be something like this:
`void Pet::setTags(std::vector<std::shared_ptr<Tag>>const& tags)
{
    m_Tags = tags;
}`

I also just realized, if the model is actually used, it will produce a compile error as described in #5862. I guess I should have checked the model code much closer from the cpprest template. I'll try to look into this, and fix it by tomorrow hopefully.

stkrwork:
#5873 is a pull request offering a solution

solvingJ:
@stkrwork Could you possibly try generating this spec with your latest PR and see if it compiles all the setters properly?  

https://raw.githubusercontent.com/Azure/azure-rest-api-specs/master/arm-iothub/2017-01-19/swagger/iothub.json

This comes from:  #5862

stkrwork:
It compiles with restbed, but cpprest might we a completely different story because restbed is only a server stub, cpprest generates a fully functional client. i will apply my changes to the cpprest template,  but i cant promise anything, because i dont know the cppres templates that well.

stkrwork:
I added my changes to the cpprest template, which should fix the settag problem, but there is still the issue with the toMultipart, toJson, and fromJson function. The cpprest template will require some repairs.

solvingJ:
Thank you so much @stkrwork . I don't understand how it got this far with those functions being broken. Still, fixing setTag is a good start!

stkrwork:
#5877 maybe someone else knows the template better. this just contains the first step

PR #5759, opened 2017-06-02
[javascript] Fix empty response body when schema type is string

cliffano:
### PR checklist

- [v] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [v] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [v] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

This fixes the bug with empty response body when response schema type is a string.

Here's an example spec snippet:
```
      responses:
        200:
          schema:
            type: string
```

I've checked its ruby equivalent to confirm that string schema type should have response body.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Javascript/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Javascript/ApiClient.mustache
@@ -423,6 +423,8 @@
 
     if (returnType === 'Blob') {
       request.responseType('blob');
+    } else if (returnType === 'String') {
+      request.responseType('string');
     }
 
     // Attach previously saved cookies, if enabled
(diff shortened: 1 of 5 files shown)


wing328:
@cliffano thanks for the PR, which has been merged into master.

Do we need similar fix for ES6 version of ApiClient.mustache?

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/Javascript-es6/ApiClient.mustache#L432

cliffano:
@wing328 My bad, I missed out ES6. Handled in #5819 .

wing328:
@cliffano not at all. Thanks for fixing it for ES6 as well. I'll check out your PR later.

FYI. I'll move the javascript ES6 template under a subfolder (e.g. es6) in "Javascript" later.

PR #5189, opened 2017-03-24
[JAVA][#5172] Allow vendor json media types

davidharrigan:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [x] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR
Adds support for vendor media type JSON (ends with `+json`) as specified in [RFC-3839](https://tools.ietf.org/html/rfc6839#section-3.1). 

PR made against master since the change is non-breaking and backwards compatible (still has the same support for `application/json`). 

More details in the issue here: #5172

PR #5764, opened 2017-06-02
Add JSON-patch as a JSON mime-type

brendandburns:
### Description of the PR
Add support for marshalling JSON-Patch based objects (https://tools.ietf.org/html/rfc6902)

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache
@@ -744,7 +744,7 @@ public class ApiClient {
      */
     public boolean isJsonMime(String mime) {
       String jsonMime = "(?i)^(application/json|[^;/ \t]+/[^;/ \t]+[+]json)[ \t]*(;.*)?$";
-      return mime != null && mime.matches(jsonMime);
+      return mime != null && (mime.matches(jsonMime) || mime.equalsIgnoreCase("application/json-patch+json"));
     }
 
     /**
--- a/samples/client/petstore-security-test/java/okhttp-gson/src/main/java/io/swagger/client/ApiClient.java
+++ b/samples/client/petstore-security-test/java/okhttp-gson/src/main/java/io/swagger/client/ApiClient.java
@@ -750,7 +750,7 @@ public class ApiClient {
      */
     public boolean isJsonMime(String mime) {
       String jsonMime = "(?i)^(application/json|[^;/ \t]+/[^;/ \t]+[+]json)[ \t]*(;.*)?$";
-      return mime != null && mime.matches(jsonMime);
+      return mime != null && (mime.matches(jsonMime) || mime.equalsIgnoreCase("application/json-patch+json"));
     }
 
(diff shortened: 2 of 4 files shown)


modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache
```diff
@@ -744,7 +744,7 @@ public class ApiClient {
      */
     public boolean isJsonMime(String mime) {
       String jsonMime = "(?i)^(application/json|[^;/ \t]+/[^;/ \t]+[+]json)[ \t]*(;.*)?$";
-      return mime != null && mime.matches(jsonMime);
+      return mime != null && (mime.matches(jsonMime) || mime.equals("application/json-patch+json"));
```
wing328: @brendandburns shall we use a case insensitive comparison with "application/json-patch+json"?

Ref: https://stackoverflow.com/a/12869287/677735

wing328: It would be nice to add a test case too: https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/java/okhttp-gson/src/test/java/io/swagger/client/ApiClientTest.java#L81

brendandburns: done for both.

brendandburns:
Comments addresed, please take another look.

Thanks!

wing328:
Tests passed via https://travis-ci.org/swagger-api/swagger-codegen/builds/241095920

@brendandburns thanks for the contribution.

I'll file a PR to apply a similar fix to other Java API client (jersey2, retrofit2, feign, etc)

modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache
```diff
@@ -744,7 +744,7 @@ public class ApiClient {
      */
     public boolean isJsonMime(String mime) {
       String jsonMime = "(?i)^(application/json|[^;/ \t]+/[^;/ \t]+[+]json)[ \t]*(;.*)?$";
-      return mime != null && mime.matches(jsonMime);
+      return mime != null && (mime.matches(jsonMime) || mime.equalsIgnoreCase("application/json-patch+json"));
```
ePaul: I know I'm a bit late here, but as far as I can see the regular expression above already matches "application/json-patch+json". Did your new test case fail before this change?

brendandburns: Ah, I think that I used a generator that was released that didn't have:

https://github.com/swagger-api/swagger-codegen/pull/5189

We should revert this change. I will send a PR...

Thanks for catching this @ePaul

brendandburns: Or rather, I will leave the test, but revert the extra check...

PR #5765, opened 2017-06-02
[javascript] Fixed `fs` files support for Electron

frol:
### PR checklist

### Description of the PR

[Electron](https://electron.atom.io/) has `fs` module with correctly working `ReadStream`, but the check `typeof window === 'undefined'` forced it to skip the `ReadStream` check in Electron, while it is completely valid.

P.S. Other than this, the generated client works perfectly in Electron! (Well, I had to workaround an issue in formidable (https://github.com/felixge/node-formidable/issues/337#issuecomment-153408479), which is a dependency of superagent)

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Javascript-es6/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Javascript-es6/ApiClient.mustache
@@ -160,9 +160,15 @@ export default class ApiClient {
     */
     {{/emitJSDoc}}
     isFileParam(param) {
-        // fs.ReadStream in Node.js (but not in runtime like browserify)
-        if (typeof window === 'undefined' && typeof require === 'function' && require('fs') && param instanceof require('fs').ReadStream) {
-            return true;
+        // fs.ReadStream in Node.js and Electron (but not in runtime like browserify)
+        if (typeof require === 'function') {
+            let fs;
+            try {
+                fs = require('fs');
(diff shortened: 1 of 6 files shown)


modules/swagger-codegen/src/main/resources/Javascript-es6/ApiClient.mustache
```diff
@@ -160,9 +160,15 @@ export default class ApiClient {
...
     {{/emitJSDoc}}
     isFileParam(param) {
-        // fs.ReadStream in Node.js (but not in runtime like browserify)
-        if (typeof window === 'undefined' && typeof require === 'function' && require('fs') && param instanceof require('fs').ReadStream) {
-            return true;
+        // fs.ReadStream in Node.js and Electron (but not in runtime like browserify)
+        if (typeof require === 'function') {
+            let fs
+            try {
+                fs = require('fs')
```
kemurphy: Would be nice to be consistent with semicolons.

frol: I completely agree! I will update it tomorrow.

frol:
@kemurphy I have added the missing semicolons.

@wing328 I think, it is ready to merge.

wing328:
@kemurphy thanks for reviewing the change.

@frol thanks for your contribution.

PR #5767, opened 2017-06-02
[javascript] Added README section for Webpack configuration

frol:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [x] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

This PR closes #3466.

Issue #3466, opened 2016-07-27
[Javascript][2.2.0] AMD define module paths fails in Webpack


darylteo:
##### Description

Version 2.1.6 produced AMD paths with `./` prefixes. For example:

```
define(['./ApiClient', './model/Article', './model/ArticleResourceSingle', ... , './api/DefaultApi'], factory);
```

Version 2.2.0 produces AMD paths without `./` prefixes. For example: 

```
define(['ApiClient', 'model/Article', 'model/ArticleResourceSingle', ..., 'api/DefaultApi'], factory);
```

This has broken my webpack build.

```
ERROR in ../api/build/swagger/javascript/src/index.js
Module not found: Error: Cannot resolve module 'model/Article' in ../api/build/swagger/javascript/src
 @ ../api/build/swagger/javascript/src/index.js 28:4-701

ERROR in ../api/build/swagger/javascript/src/index.js
Module not found: Error: Cannot resolve module 'model/ArticleResourceSingle' in ../api/build/swagger/javascript/src
 @ ../api/build/swagger/javascript/src/index.js 28:4-701

```
##### Swagger-codegen version

2.2.0

darylteo:
Have tried similar exercise by copying codegen output into the node_modules directory to see if it is some sort of module paths issue, no avail. Those don't look like valid paths no matter how I skin it. 

If anyone else is facing similar issue in Webpack, I have disabled AMD loading using the imports-loader.

zsimo:
Hello!
Exactly the same here!

zsimo:
Solved:
```
npm install --save-dev imports-loader
```
and then add this loader to the webpack configs
```
{ test: /\.js/, loader: 'imports?define=>false'}
```

hasangilak:
i wrote a template for webpack build. may help
https://github.com/hasangilak/webpack-swagger-template

wing328:
> If anyone else is facing similar issue in Webpack, I have disabled AMD loading using the imports-loader, which works.

Shall we add an option to disable ADM loading in the auto-generated JS code?

frol:
Webpack 2 requires using `imports-loader` loader full name instead of a short name (`imports`). Thus, here is what works for me for Webpack 2:

```javascript
module: {
  rules: [
    {
      test: /my_client\/.*\.js$/,
      use: 'imports-loader?define=>false'
    }
  ]
},
node: {
  fs: 'empty'
}
```

akeemphilbert:
[...]

PR #5768, opened 2017-06-02
[javascript] Mark "fs" package as absent in browser in package.mustache

kemurphy:
### PR checklist

### Description of the PR

WIthout this, webpack and react-native both choke trying to resolve dependencies for packages generated by `swagger-codegen`.


frol:
How will Electron handle this? (Electron has fs support) It seems that using package.json config option might make it uncontrollable from the user side.

The fix in #5765 might also solve the issue in browser as it catches the error of require('fs').

kemurphy:
Electron doesn't seem to be affected -- per electron/electron#1574, the "browser" field in package.json is ignored since the author recognized Electron is more node-like than browser-like.

#5765 isn't sufficient because the breakage in both webpack and react-native isn't caused by running the code.  Rather, it's caused by static analysis of all `require` calls in the code.  Dynamic conditions don't affect it, nor can it be caught like an exception.

frol:
@kemurphy Well, #5765 in combination with #5767 works for me in SSR mode (the chunk of code is not executed, but the module can be imported fine), in browser (Google Chrome and Firefox, but I will test it more thoroughly, yet I use Next.js so it may do some magic), and in Electron. I haven't tried react-native. It would be great if you can give #5765 a try and report the exact error you get before and after the fix (without your fix).

kemurphy:
For #5767, the `node: { fs: 'empty' }` webpack configuration is equivalent to the `"browser": { "fs": false }` in package.json, and webpack can actually override the keys it looks for in package.json to pick alternate files and mark dependencies missing (https://webpack.js.org/configuration/resolve/#resolve-aliasfields).  (I actually suspect that AMD is not an issue and `imports-loader` isn't needed there.)  For react-native, however, there's no equivalent configuration -- it's either the "browser" field or the "react-native" field in package.json, and "browser" kills 2 birds with one stone.  #5765 and #5767 shouldn't be necessary with this one.

frol:
@kemurphy OK, I will give it a try and report my results back tomorrow.

kemurphy:
As expected, react-native blows up with #5765 applied instead of this one:

image

wing328:

kemurphy:
Aha, sorry, I'll fix that now.

wing328:
@frol if the test result looks good, I'll merge the enhancement into master.

frol:
@wing328 I have tested it in browsers, in server-rendering mode, and in Electron. The fix works well and there is no need for the webpack config (`node: { fs: 'empty' }`) I mentioned in my PR #5767!

@kemurphy Thank you!

wing328:
@frol thanks for testing it in different scenarios.

@kemurphy thanks again for the enhancement.

frol:
@kemurphy You have said:

> (I actually suspect that AMD is not an issue and `imports-loader` isn't needed there.)

I would love to hear how to avoid `imports-loader`! Currently, if I remove the `imports-loader` config, I get the following errors:

On the server side:

```
These modules were not found:

* ApiClient in ./~/cloudsml-client/src/index.js
* model/BaseDataSource in ./~/cloudsml-client/src/index.js
* model/BaseDataTransformation in ./~/cloudsml-client/src/index.js
...
```

On the client side:

```
ModuleNotFoundError: Module not found: Error: Can't resolve 'ApiClient' in '/project/node_modules/cloudsml-client/src'
    at factoryCallback (/project/node_modules/webpack/lib/Compilation.js:264:39)
    at factory (/project/node_modules/webpack/lib/NormalModuleFactory.js:247:20)
    at resolver (/project/node_modules/webpack/lib/NormalModuleFactory.js:65:21)
    at asyncLib.parallel (/project/node_modules/webpack/lib/NormalModuleFactory.js:138:21)
    at /project/node_modules/async/dist/async.js:3838:9
    at /project/node_modules/async/dist/async.js:421:16
    at iteratorCallback (/project/node_modules/async/dist/async.js:996:13)
    at /project/node_modules/async/dist/async.js:906:16
    at /project/node_modules/async/dist/async.js:3835:13
    at resolvers.normal.resolve (/project/node_modules/webpack/lib/NormalModuleFactory.js:130:23)
```

frol:
@kemurphy I am sorry to bother you, but I am still curious about the proper fix for the loader... Could you please shed some light on the issue? I am using Next.js and I am not sure what exactly they do with Webpack, but given the responses in #3466, it is not specific to Next.js usage.

PR #5773, opened 2017-06-04
Fix typo in Ruby examples/mustache files

musicmage4114:
Changed all appearances of "instact" to "instance". Appears this was limited to Ruby files.


wing328:
@musicmage4114 thanks for the fix. 

I did a grep on the "resources" folder and confirmed only Ruby templates contain the typo

Issue #5423, opened 2017-04-19
Add endpoint examples defined in the spec to API servers


wing328:
##### Description

Similar to what we've done for SpringMVC generator, we want to add support of endpoint examples defined in the spec to the auto-generated server stubs.

Ref: https://github.com/swagger-api/swagger-codegen/pull/5411

Endpoint example support in the server generators:

- [ ] C# (NancyFx, ASP.net Core)
- [ ] Elixir
- [ ] Erlang
- [ ] Go
- [ ] Haskell Servant
- [ ] JavaScript/NodeJS
- [ ] Java (MSF4J, Spring:white_check_mark:, Undertow, JAX-RS: CDI, CXF, Inflector, RestEasy)
- [ ] PHP (Lumen, Slim, Silex, Zend Expressive)
- [ ] Python Flask
- [ ] Ruby (Sinatra, Rails)
- [ ] Scala Finch
- [ ] Scala Scalatra

##### Suggest a Fix

#5411 is a good reference. If anyone wants to contribute to this task, please reply to let us know. Thank you.

thure:
Just a heads-up: I'm looking into doing a Lumen branch based on #5411. Thanks for organizing everything here 😄

wing328:
Thanks @thure.

FYI. The lambda functions have been renamed to `lambdaEscapeDoubleQuote` and `lambdaRemoveLineBreak`

PR #5775, opened 2017-06-05
[JavaPlayFramework] Add support for endpoint examples. HELP NEEDED

JFCote:
### PR checklist

### Description of the PR

Here is an attempt to add endpoint examples to the JavaPlayFramework as discussed here: https://github.com/swagger-api/swagger-codegen/issues/5423

I was asked by @wing328 to add the support but I'm not very sure what is the use case of this.

**Unfortunatly, it doesn't work and I need help from the community to finish this PR. Right now, it seems that `{{{example}}}` is always empty, even if I use the "fake endpoint yaml".**

Any ideas and fix would be appreciated.


wing328:
@JFCote Thanks for the PR. We'll add some examples to the fake petstore spec.

Issue #5460, opened 2017-04-23
[All?] examples for string definitions are over-quoted


reece:
##### Description

In a `definition` section, examples for string objects are over-quoted. For example, this stanza from an input yaml

```yaml
# input yaml
definitions:
  Id:
    type: string
    example: example from def
```

is rewritten by swagger-codegen as

```yaml
# resulting yaml from swagger-codegen
  Id:
    type: "string"
    example: "\"example from def\""
```

Explicitly quoting the input `example` text has no effect.

The above example is for python-flask and nodejs-server. I suspect other servers generate the same yaml.

This bug does *not* occur for examples within `object` definitions. For example, the following stanza works as expected:

```yaml
  Allele:
    type: "object"
    properties:
      id:
        $ref: "#/definitions/Id"
      location_id:
        $ref: "#/definitions/Id"
      replacement:
        $ref: "#/definitions/Sequence"
    example:
      id: "Allele example"
      location_id: "Allele example"
      replacement: "Allele example"
```

##### Swagger-codegen version

2.2.2

##### Swagger declaration file content or url

https://github.com/reece/vmc-openapi/blob/master/spec/vmc.yaml

##### Command line used for generation

`swagger-codegen generate -l python-flask -i ../spec/vmc.yaml -o python-flask`

##### Steps to reproduce

* Install swagger-codegen 2.2.2 (other versions untested)
* Run command line as above.

##### Related issues

The following issues are vaguely related.

* #5458 
* #1641
* #4611

wing328:
@reece thanks for reporting the issue. When I clicked on https://github.com/reece/vmc-openapi/blob/master/spec/vmc.yaml, I got a 404.

If you've time, I would suggest you to try the latest master as well to see if there's by any chance the issue may have been addressed already.

reece:
@wing328: Yes, this affects latest master (42a5a13fe) as well.

I worked up an minimal example at https://gist.github.com/reece/7d343afd3638fc530c3d3607b57ababd. 

```yaml
definitions:
  Def1:
    type: "string"
    example: word
  Def2:
    type: "string"
    example: two words
  Def3:
    type: "string"
    example: "quotedword"
  Def4:
    type: "string"
    example: "two quoted words"
  Def5:
    type: "object"
    properties:
      namespace:
        type: "string"
        example: "NCBI"
      accession:
        type: "string"
        example: "ENST000001234.5"
```

[...]

PR #5778, opened 2017-06-05
Fix for issue #5460

JFCote:
### PR checklist

### Description of the PR

This is quick fix for https://github.com/swagger-api/swagger-codegen/issues/5460

I don't know if it's a "good" way to deal with this but my understanding is that a "string" model is the exception to the rule because the example is not "json" but a simple string, so there is no need to double quote it.

I didn't ran the shell/batch script because I didn't know if there were a quick way to run them all at once... @wing328 , if the fix is accepted, can you run the appropriate script? Thanks!


wing328:
@JFCote thanks for the fix. Am I correct in saying that the fix won't work with examples that contain double quote?

modules/swagger-codegen/src/main/java/io/swagger/codegen/InlineModelResolver.java
```diff
@@ -191,6 +191,12 @@ public void flatten(Swagger swagger) {
         }
     }
 
+    private void fixStringModel(ModelImpl m) {
+        if (m.getType() != null && m.getType().equals("string") && m.getExample() != null) {
+            m.setExample(m.getExample().toString().replace("\"", ""));
```
ePaul: As @wing328 mentioned, this removes all the `"` from the example string, which likely is not what we want.
Could we restrict this to the first/last character?

Unfortunately, Swagger-models (or at least the classes/interfaces in use here) have no javadocs, so it is not really clear what should be in this `example` property.

JFCote: Yes you are right, I didn't think about this possible scenario. I will submit something today to fix this. Thanks for the review @wing328 and @ePaul

JFCote:
Will take a little bit more time for my fix, I'm pretty busy right now! Just wanted to let you know I did not forget about this

wing328:
@JFCote I'll take a look tomorrow to see if there're other ways to address it

JFCote:
There you go. It will only remote the double quote if they are at the beginning and the end of the example, when it's a string. Like I said, I did not run all of the sample .sh, I don't know how to do it all at once.

ePaul:
@JFCote you could use `bin/run-all-petstore` ... it gives a warning telling that it is meant to be used just from CI, but it works (at least for me). (It will give some unrelated changes due to other people not having run it before.)

Do we have a test case in the examples where it formerly was wrong?

modules/swagger-codegen/src/main/java/io/swagger/codegen/InlineModelResolver.java
```diff
@@ -191,6 +191,16 @@ public void flatten(Swagger swagger) {
         }
     }
 
+    private void fixStringModel(ModelImpl m) {
+        if (m.getType() != null && m.getType().equals("string") && m.getExample() != null) {
+            String example = m.getExample().toString();
+            if (example.substring(0, 2).equals("\"") &&
+                    example.substring(example.length() - 2).equals("\"")) {
+                m.setExample(example.substring(2, example.length() - 2));
```
ePaul: I think your count is wrong here. The Java string `"\"Hello\""` has length 7, not 9, and you need to cut off one character at each end, not two. Though please test this first.

JFCote:
@ePaul : For an example, check https://github.com/swagger-api/swagger-codegen/issues/5460
You just have to add this to your list of definitions:

```yaml
ReturnStatus:
  type: string
  enum:
    - NotReturned
    - Recycled
    - Trashed
    - InTransit
    - InvestigationPending
    - Lost
  description: "Device's return status"
  example: InTransit
```

modules/swagger-codegen/src/main/java/io/swagger/codegen/InlineModelResolver.java
JFCote: I've test it and it works. And we want to cut "\"", which is 2 chars.

For example:
```
0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8
\ | " | H | E | L | L | O | \ | "
```

First condition will return` \"` which is true
Second condition will return 9 - 2 = 7 -> `\"`
Then we simply remove everything between both which returns `HELLO`

I've test it and it works :) Let me know if it doesn't work in your case and I'll look into it!

ePaul: Hmm, maybe I misunderstood what you are doing here ... I thought we have here a pair of `"` too much, which needs to be removed. (This is what the old `replace` code was doing, assuming there are no more quotation marks in the string.)
I thought the escaping (`"` → `\"`) was only done after this step (while outputting).

Anyway, if you want to remove not just a single `"`, but the combination `\"`, then you'll need to write it `"\\\""` as a Java string literal (the single backslash is escaping the `"`).

Though I'll try it out instead of just using my Java knowledge to discuss things theoretically.

ePaul: @JFCote I added the example snippet to src/main/resources/petstore.yaml, and regenerated the code (from your branch). The [bug still occurs there](https://github.com/swagger-api/swagger-codegen/commit/947303bd7adab7d458919972357bd4accd082bd9#diff-7e2ada6ae234106c8b7784808b18326fR633). [Using 1 instead of 2 fixes it.](https://github.com/swagger-api/swagger-codegen/commit/c4be016d212dea0f63f380ccdd6f9d82cbec1e5a).

(Feel free to pick those commits [from my fork](https://github.com/StingrayDigital/swagger-codegen/compare/fix-5460...ePaul:5460-5778-test), or copy just the needed parts.)

JFCote: You are right... I **really** don't understand how it worked for me (tested it a couple of times...) but now, it doesn't work.
I will submit your fix in a couple of minutes..

JFCote:
@ePaul : I added you proposed changes to the PR. Thanks for the help and sorry again, I really don't know what happened during my testing!

JFCote:
@ePaul and @wing328 : Is there something blocking this PR to be merged? I have tried to run the `bin/run-all-petstore` but there is tons of changes not related to my own changes, just because it's not everybody that runs the `.sh` after each changes. I'm not at ease to commit all of this in the PR unless you ask me :)

Let me know what I can do to help the merge of this PR. Thanks!

ePaul:
@JFCote Looking at the code, I don't see anything blocking this.

I can't build the project at my work computer currently (I'm stuck with an old maven installation), will try it later at home to see which changes actually come from this PR and which are already in master.

wing328:
@JFCote I'll take a look tomorrow.

ePaul:
I just tried the sample-updating once on master (see #6004), and once again on the merge commit of your branch. Same result (apart from two markdown/txt files containing the build date).

A question would be, could we have some examples so that there is actually a difference?

JFCote:
@ePaul : Should I create a special petstore.yaml for this or modify the main yaml file that every generator use?

wing328:
@JFCote I think we can add a few to the existing petstore.yaml. I'll try to do it later and run some tests with your PR.

modules/swagger-codegen/src/main/java/io/swagger/codegen/InlineModelResolver.java
```diff
@@ -191,6 +191,16 @@ public void flatten(Swagger swagger) {
         }
     }
 
+    private void fixStringModel(ModelImpl m) {
```
wing328: @JFCote it would be nice to add a 1-liner or a proper docstring explaining what this function does.

JFCote:
Ok, I have added the comment + example in the `petstore.yaml` file.
I have only updated the sample for the Play Framework. Please feel free to run the `bin/run-all-petstore` file later to update all the other generators. If you want to see the fix itself in action, using the new `petstore.yaml`, remove the content of the `fixStringModel` function and you will see the double quoting reappear proving that the fix works.

ePaul:
I tried updating the other generators, no changes (which were not already changes before). (I guess they don't use the same YAML, or don't use the examples.)

JFCote:
@wing328 Can you merge this when you have time? Thanks!

wing328:
@JFCote I'll do it tomorrow (sorry for the delay. too busy these days..)

JFCote:
@wing328 No problem man! I completely understand!

modules/swagger-codegen/src/test/resources/2_0/petstore.yaml
```diff
@@ -701,6 +701,15 @@ definitions:
         type: string
       message:
         type: string
+  PetStatus:
```
wing328: @JFCote I'll remove this tomorrow add the enum example to https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml instead.

When the Play generator, we should consider using petstore-with-fake-endpoints-models-for-testing.yaml, which covers a lot more edge cases.

wing328:
Removed "PetStatus" via 943b47d

Issue #5777, opened 2017-06-05
[JAVA] java/jersey2 ApiClient.invokeAPI doesn´t close the Response object


Moscagus:
##### Description

The code does not guarantee the correct closing of the Response object, since it does not always reach a response.readEntity or call a response.readEntity(InputStream.class) that does not close the connection either.
See:
https://jersey.github.io/documentation/latest/client.html#d0e5222

##### Swagger-codegen version

swagger-codegen-cli-2.2.2.jar

##### Command line used for generation

-l java --library jersey2
##### Suggest a Fix

  ...    
  Response response **= null**;
    
  **try {**
      if ("GET".equals(method)) {
        response = invocationBuilder.get();
  ... 
  **} finally {
      try {
        response.close();      
      } catch (Exception e) {}
  }**
}

wing328:
@Moscagus thanks for reporting the issue. Please file a PR so that we can more easily review the fix.

PR #5782, opened 2017-06-05
Fix for issue #5777 - java/jersey2 ApiClient.invokeAPI doesn´t close the Response object

Moscagus:
Fix for issue #5777

### PR checklist

### Description of the PR

Fix for: java/jersey2 ApiClient.invokeAPI doesn´t close the Response object

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
@@ -729,7 +729,9 @@ public class ApiClient {
     } finally {
       try {
         response.close();
-      } catch (Exception e) {}
+      } catch (Exception e) {
+        // it's not critical, since the response object is local in method invokeAPI; that's fine, just continue
+      }
     }
   }
 
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
@@ -685,47 +685,47 @@ public class ApiClient {
     Response response = null;
 
     try {
-    if ("GET".equals(method)) {
-      response = invocationBuilder.get();
-    } else if ("POST".equals(method)) {
-      response = invocationBuilder.post(entity);
-    } else if ("PUT".equals(method)) {
-      response = invocationBuilder.put(entity);
-    } else if ("DELETE".equals(method)) {
-      response = invocationBuilder.delete();
-    } else if ("PATCH".equals(method)) {
-      response = invocationBuilder.header("X-HTTP-Method-Override", "PATCH").post(entity);
-    } else {
-      throw new ApiException(500, "unknown method type " + method);
-    }
-
-    statusCode = response.getStatusInfo().getStatusCode();
-    responseHeaders = buildResponseHeaders(response);
-
-    if (response.getStatus() == Status.NO_CONTENT.getStatusCode()) {
-      return null;
-    } else if (response.getStatusInfo().getFamily() == Status.Family.SUCCESSFUL) {
-      if (returnType == null)
-        return null;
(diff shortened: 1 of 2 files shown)


modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
```diff
@@ -682,8 +682,9 @@ public class ApiClient {
 
     Entity<?> entity = serialize(body, formParams, contentType);
 
-    Response response;
+    Response response = null;
```
wing328: @Moscagus I think it's null by default so it's not necessary to initialize with null

modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
```diff
@@ -682,8 +682,9 @@ public class ApiClient {
...
 
+    try {
     if ("GET".equals(method)) {
```
wing328: @Moscagus Seems like the code within the try block is missing 2-space for indentation.

Please also run `./bin/java-all-petstore.sh` to update Java Petstore samples so that CIs can verify the change.

modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
```diff
@@ -725,6 +726,11 @@ public class ApiClient {
         buildResponseHeaders(response),
         respBody);
     }
+    } finally {
+      try {
+        response.close();
+      } catch (Exception e) {}
```
wing328: @Moscagus What about throwing a runtime exception if the response.close fails?
```
      throw new RuntimeException(e);
```

modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
Moscagus: @wing328 Local variables are slightly different; the compiler never assigns a default value to an uninitialized local variable. If I do not initialize the response variable, in compilation I get error "variable response might not have been initialized"

modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
Moscagus: You are right. I have no experience with GitHub, but I will try to do what you mention me

modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
Moscagus: Can be. In general I usually handle the subject of resources in that way, so I avoid even the exception Java NullPointerException.
Could be:
    } finally {
      try {
        if (response != null) response.close();      
      } catch (Exception e) {
        throw new RuntimeException(e);
      }
    }

What do you think ?

modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache
Moscagus: @wing328 Do I have to change the file again to add the 2 spaces ?

Moscagus: @wing328 I think I already changed the file correctly. Now I see running the ./bin/java-all-petstore.sh

wing328:
Please also run `./bin/java-petstore-jersey2.sh` to update the Petstore sample for Java (jersey2) API client so that the CIs can verify the change.

Moscagus:
I already ran ./bin/java-petstore-jersey2.sh

wing328:
@Moscagus I did some tests locally and the result is good. Thanks for your contribution.

PR #5783, opened 2017-06-05
[javascript] Fix documentation for ES6 ApiClient

kemurphy:
### PR checklist

### Description of the PR

ES6 docs point people at using `ApiClient.default` when it really meant `ApiClient.instance`.

The issue makes me wonder if the intention was to change the way this was done from ES5 to ES6.  Something more idiomatic in ES6 might be to have `ApiClient` use named exports (`ApiClient` for the class and `defaultClient` for the instance) instead of a default export.  Then consume it as `import { defaultClient } from 'foo-api/ApiClient';`.  Thoughts?  Happy to do that instead of this PR.


wing328:
@kemurphy thanks for the fix.

> Something more idiomatic in ES6 might be to have ApiClient use named exports (ApiClient for the class and defaultClient for the instance) instead of a default export.

I would suggest you to create a new issue to start the discussion.

cc @dinukadesilva

Issue #4320, opened 2016-12-05
PowerShell module


wing328:
##### Description

To add a code generator to generate PowerShell module for accessing REST API.

##### Swagger-codegen version

Future

##### Details

If anyone is interested in or has already started working on a code generator for PowerShell module, please reply to let us know.

beatcracker:
Just FIY: since you already have a C# generator, you could just make a thin PowerShell wrapper around generated C# code. This could save some time.

wing328:
@beatcracker thanks for the suggestion. May I know if you've time to contribute the PowerShell generator?

beatcracker:
@wing328 Depends on what you need. I can help with PowerShell code, but unfortunately I can't do Java (which is actually used to write generators, right?).

wing328:
@beatcracker We can work together on that 😃 

I think a good starting point is to create a PowerShell wrapper for [PetStore](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.yaml) by focusing on a few endpoints.
- [GET](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.yaml#L156-L183)
- [POST (form)](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.yaml#L184-L218)
- [POST (JSON body)](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.yaml#L31-L56)

The PowerShell code can be put in the folder `samples/client/petstore/powershell`.

Then we can reverse-engineer the PowerShell code to create the templates and the generator in Java.

What do you think?

beatcracker:
I had a look at C# generators, and it's dawned on me, what a can of worms I just opened 😰. 

I'll try summarize my thoughts here, at least for my own sake, before I try to do some kind of MVP PowerShell module example.

---

## How do we generate a PowerShell code?

### Using PowerShell wrappers for C#

#### Pros:

[...]

PR #5789, opened 2017-06-06
[PowerShell] Add PowerShell API client generator (WIP)

wing328:
### PR checklist

### Description of the PR

Closes https://github.com/swagger-api/swagger-codegen/issues/4320

It's based on the PowerShell petstore sample provided by @beatcracker

It's still work-in-progress.


wing328:
@beatcracker I've a couple of questions if you don't mind.

1. Do we need to have 1 function per 1 PS file? Currently we generate functions for different endpoints within the same PS file.

2. Is it ok to put the API-related files under the "API" subfolder and model files under the "Model" subfolder? I can put all of them under "Public" but just want to see if we can use subfolders to organize these files.

beatcracker:
@wing328 

> Do we need to have 1 function per 1 PS file? Currently we generate functions for different endpoints within the same PS file.

No, absolutely not. It's OK to bundle them together.

> Is it ok to put the API-related files under the "API" subfolder and model files under the "Model" subfolder? I can put all of them under "Public" but just want to see if we can use subfolders to organize these files.

The same as above. Private/Public folder structure is just an unofficial convention which helps keep things logically separated when writing code manually. We can organize code the way we see fit.

---

I'm currently testing your PR locally and I'll give you more feedback soon.

beatcracker:
Ok, here is today's results:

1. I've fixed some essential issues with mustache templates. I'm not sure what's the best way to merge them with this PR, so for now I've just created [beatcracker/powershell_generator_fix](https://github.com/beatcracker/swagger-codegen/tree/powershell_generator_fix) branch in my fork.

    There are still more issues, which currently I don't know how to fix. I'll try to point to them in this PR's code review later.

2. Regarding one-function-per-file question. I've said, that there is no hard requirement for functions to be in separate files, but I've forgot that there is a code in the manifest generator that generates `FunctionsToExport` property using these filenames. So we need either to fallback to separate files or figure out some other way.

wing328:
1. I've merged your fix into my branch via 4850b83

> So we need either to fallback to separate files or figure out some other way.

I think we need to figure some other ways as it's not easy to generate an endpoint function per file.

samples/client/petstore/powershell_test/src/IO.Swagger/API/PetApi.ps1
```diff
@@ -0,0 +1,164 @@
...
+        )
+    }
+}
+
+function PetApi-deletePet {
+    [CmdletBinding()]
+    Param (
+        [Parameter(Position = , ValueFromPipeline = $true, ValueFromPipelineByPropertyName = $true, Mandatory = $true)]
+        [IO.Swagger.Model.Pet]
+        ${ petId }
```
beatcracker: Parameter definitions should be separated by comma, e.g.: `${petId}, ${apiKey}`

Extended example (note the comma after ``${petId}`):

```posh
        [Parameter(ValueFromPipeline = $true, ValueFromPipelineByPropertyName = $true, Mandatory = $true)]
        [IO.Swagger.Model.Pet]
        ${petId},

        [Parameter(ValueFromPipeline = $true, ValueFromPipelineByPropertyName = $true, Mandatory = $false)]
        [IO.Swagger.Model.Pet]
        ${apiKey}
```

samples/client/petstore/powershell_test/src/IO.Swagger/API/PetApi.ps1
```diff
@@ -0,0 +1,164 @@
...
+
+function PetApi-updatePetWithForm {
+    [CmdletBinding()]
+    Param (
+        [Parameter(Position = , ValueFromPipeline = $true, ValueFromPipelineByPropertyName = $true, Mandatory = $true)]
+        [IO.Swagger.Model.Pet]
+        ${ petId }
+        [Parameter(Position = , ValueFromPipeline = $true, ValueFromPipelineByPropertyName = $true, Mandatory = $false)]
+        [IO.Swagger.Model.Pet]
+        ${ name }
```
beatcracker: `$name` should have type of `string`. In PoweShell, you can use either [accelerator](https://learn-powershell.net/2013/10/10/quick-hits-finding-creating-and-removing-powershell-type-accelerators/) type, such as `[string]` or `[xml]` or .NET type, such as `[System.String]` or `[System.Int64]`.

samples/client/petstore/powershell_test/src/IO.Swagger/IO.Swagger.psm1
```diff
@@ -0,0 +1,94 @@
...
+
+#endregion
+
+
+#region Initialize APIs
+
+'Creating object: IO.Swagger.Api.PetApi' | Write-Verbose
+$Script:PetApi= New-Object -TypeName IO.Swagger.Api.PetApi -ArgumentList @($null)
+
+'Creating object: IO.Swagger.Api.PetApi' | Write-Verbose
```
beatcracker: Look like something gone wrong with mustache template, so it repeats the same line.

samples/client/petstore/powershell_test/src/IO.Swagger/Model/New-Order.ps1
```diff
@@ -0,0 +1,36 @@
...
+        [System.Nullable[Int64]]
+        ${ id },
+        [Parameter(Position = 0, ValueFromPipelineByPropertyName = $true)]
+        [System.Nullable[Int64]]
+        ${ petId },
+        [Parameter(Position = 0, ValueFromPipelineByPropertyName = $true)]
+        [System.Nullable[Int32]]
+        ${ quantity },
+        [Parameter(Position = 0, ValueFromPipelineByPropertyName = $true)]
+        [System.Nullable[JavaTimeLocalDateTime]]
```
beatcracker: Wrong type mapping slipped in. This should probably be `[System.Nullable[System.DateTime]]`.

samples/client/petstore/powershell_test/src/IO.Swagger/Model/New-Pet.ps1
```diff
@@ -0,0 +1,36 @@
+function New-Pet {
+    [CmdletBinding()]
+    Param (
+        [Parameter(Position = 0, ValueFromPipelineByPropertyName = $true)]
+        [System.Nullable[Int64]]
+        ${ id },
+        [Parameter(Position = 0, ValueFromPipelineByPropertyName = $true)]
+        [System.Nullable[Category]]
```
beatcracker: This, and others like this, should use full type name, e.g.: `[System.Nullable[IO.Swagger.Model.Category]]`

samples/client/petstore/powershell_test/src/IO.Swagger/Model/New-ApiResponse.ps1
```diff
@@ -0,0 +1,24 @@
+function New-ApiResponse {
+    [CmdletBinding()]
+    Param (
+        [Parameter(Position = 0, ValueFromPipelineByPropertyName = $true)]
+        [System.Nullable[Int32]]
+        ${ code },
+        [Parameter(Position = 0, ValueFromPipelineByPropertyName = $true)]
+        [System.Nullable[String]]
```
beatcracker: `String` can't be nullable, got this exception in the PowerShell: 

```none
Unable to find type [System.Nullable[System.String]].
Details: GenericArguments[0], 'System.String', on 'System.Nullable`1[T]' violates the constraint of type 'T'.
```

beatcracker:
> I think we need to figure some other ways as it's not easy to generate an endpoint function per file.

Can we obtain this info from the generator? `operationId` and `classname` tags in mustache look promising.

If not, I can always write some code to parse `*.ps1` files for function names.

modules/swagger-codegen/src/main/resources/powershell/Build.ps1.mustache
```diff
@@ -0,0 +1,49 @@
...
+
+$Manifest = @{
+    Path = "$ScriptDir\src\{{{packageName}}}\{{{packageName}}}.psd1"
+
+    Author = 'Swagger Codegen Team'
+    CompanyName = 'swagger.io'
+    Description = '{{{packageName}}} - the PowerShell module for {{{appName}}}'
+
+    RootModule = '{{{packageName}}}.psm1'
+    Guid = 'a27b908d-2a20-467f-bc32-af6f3a654ac5' # Has to be static, otherwise each new build will be considered different module
```
beatcracker: While this has to be static between module builds, it has to be unique for a given API. I.e, we need deterministic GUIDs. May we could use something like this in the build script: [How can I generate a GUID for a string?](https://stackoverflow.com/a/28608877/4424236).

beatcracker:
We should also consider naming scheme for the functions. For example, `PetApi-addPet` is not a vaild PowerShell function name, which should follow `Verb-Noun` pattern and use [approved verbs](https://msdn.microsoft.com/en-us/library/ms714428.aspx). Invalid function names for exported functions will generate warnings on module import.

Perhaps, we could use something like `Invoke-AddPet`.

samples/client/petstore/powershell_test/src/IO.Swagger/API/PetApi.ps1
```diff
@@ -0,0 +1,164 @@
+function PetApi-addPet {
+    [CmdletBinding()]
+    Param (
+        [Parameter(Position = , ValueFromPipeline = $true, ValueFromPipelineByPropertyName = $true, Mandatory = $true)]
```
beatcracker: Template failed to generate `Position` attribute correctly (missing position number after `=`)

wing328:
@beatcracker thanks for all the feedback 👍  Working on it...

samples/client/petstore/powershell_test/src/IO.Swagger/API/PetApi.ps1
wing328: Fixed via 3902199

samples/client/petstore/powershell_test/src/IO.Swagger/API/PetApi.ps1
wing328: I used the primitive type defined in https://github.com/swagger-api/swagger-codegen/pull/5789/files#diff-f766d24f8851685d039700dd876127ceR44. Is that ok?

samples/client/petstore/powershell_test/src/IO.Swagger/IO.Swagger.psm1
wing328: Fixed via 3902199

samples/client/petstore/powershell_test/src/IO.Swagger/Model/New-Order.ps1
wing328: Fixed via 3902199

samples/client/petstore/powershell_test/src/IO.Swagger/Model/New-Pet.ps1
wing328: Fixed via 3902199

samples/client/petstore/powershell_test/src/IO.Swagger/Model/New-ApiResponse.ps1
wing328: Fixed via fae84f8

modules/swagger-codegen/src/main/resources/powershell/Build.ps1.mustache
wing328: For this, we can reuse the same logic from C# server stub generator, which also requires a unique GUID for project/solution file. I'll do it in my next update.

samples/client/petstore/powershell_test/src/IO.Swagger/API/PetApi.ps1
wing328: I tried to fix it and now the position starts at 1. Is that ok or it must be started at 0?

wing328:
> Perhaps, we could use something like Invoke-AddPet.

What about "Invoke-PetApi-addPet" or "Invoke-PetApiAddPet" as I've seen users putting the same operationId (e.g. getName) in more than 1 endpoints?

(FYI. "operationId" must be unique according to OpenAPI spec 2.0 but there are specs out there not conforming to this constraint)

wing328:
> Can we obtain this info from the generator? operationId and classname tags in mustache look promising.

Sorry what info do you need in order to achieve more than 1 function per PS file?

wing328:
In this [PowerShell file](https://github.com/MeshkDevs/InvokeTwitterAPIs/blob/master/InvokeTwitterAPIs.psm1), I found more than 1 function:

- https://github.com/MeshkDevs/InvokeTwitterAPIs/blob/master/InvokeTwitterAPIs.psm1#L123
- https://github.com/MeshkDevs/InvokeTwitterAPIs/blob/master/InvokeTwitterAPIs.psm1#L196
and more.

Not sure if it's the right way to do it or there are any hidden issues with this approach.

beatcracker:
> `Invoke-PetApiAddPet`

I was thinking along those lines and while this doesn't please my eyes, it looks as close to convention as we can afford for now. I think we should stick with it, at least for foreseeable future.

> Sorry what info do you need in order to achieve more than 1 function per PS file?

I hadn't looked closely at the generator code, but as far as I understand, it reads definitions from the swagger spec and feeds them to the mustache which produces actual code using templates. I had an idea that we could include mustache template in `Build.ps1.mustache` that will render the list of the functions (like `api.mustache`) . Something like this (probably not valid, but just to show an idea):

* `Build.ps1.mustache`
```posh
$Manifest = @{
    FunctionsToExport  = @(
        {{#operations}}
        {{#operation}}
            '{{{classname}}}-{{{operationId}}}'
        {{/operation}}
        {{/operations}}
    )
}
```

> Not sure if it's the right way to do it or there are any hidden issues with this approach.

There is nothing wrong with this approach, this is a perfectly valid PowerShell module, albeit a slightly ungainly. 

Let me try to explain things from the historical perspective:

---

Originally PowerShell hadn't a package repository. This meant that people would package/install modules and functions as they saw fit. Later, the lack of the central repository, became severe limiting factor, since languages like perl/python/.NET make package installation easy and PowerShell fell behind. This and the advancement of the CI/CD pipelines eventually lead to the creation of the [PowerShell Gallery](https://www.powershellgallery.com/), which is based on the NuGet and provides discovery, installation and versioning for modules/scripts.

So without the PowerShell Gallery, you could just put all your functions in the `psm1` file, optionally supply `psd1` manifest and call it a day. One would install module by copying module folder to the PowerShell system module directory or just importing it manually using `Import-Module` cmdlet. Often, modules included install scripts to automate that kind of tasks (like [Install.ps1](https://github.com/MeshkDevs/InvokeTwitterAPIs/blob/master/Install.ps1)).

But now, we have PowerShell Gallery, which strives to deliver high-quality code and has reasonable [publishing guidelines](https://msdn.microsoft.com/en-us/powershell/gallery/psgallery/psgallery-publishingguidelines).

I feel, that as a long-term goal, swagger-codegen should be able to generate PowerShell modules that are fit to to be published there. Ideally, I'd like to be able to setup CI/CD pipeline that would automatically generate a module from swagger spec, run tests and publish it to the PS Gallery

And here we come to the " one-function-per-file" question. As I've said, module manifest should expose `FunctionsToExport` to make those functions discoverable by search in the PS Gallery. The easiest way to do this is to have one function per file, with the file named after the function. This way one could just enumerate directory contents during module build and update module manifest accordingly.

But it's not the "one-and-only" right way. Since we generate the module and have our own constraints on how we can do it, we can produce any acceptable file/folder structure. We just have to make sure that it fits the definition of the [well-formed module](https://msdn.microsoft.com/en-us/library/dd878350.aspx) and adheres to the PS Gallery [publishing guidelines](https://msdn.microsoft.com/en-us/powershell/gallery/psgallery/psgallery-publishingguidelines).

---

#### TL;DR

It's nice to have a good module which adheres to (un)offical standards and having `FunctionsToExport` is a strong requirement. It doesn't mater how exactly we generate it, it just have to be there.

wing328:
> I was thinking along those lines and while this doesn't please my eyes, it looks as close to the convention as we can afford for now. I think we should stick with it, at least for foreseeable future.

Done via 0f0a586

wing328:
> I feel, that as a long-term goal, swagger-codegen should be able to generate PowerShell modules that are fit to to be published there. Ideally, I'd like to be able to setup CI/CD pipeline that would automatically generate a module from swagger spec, run tests and publish it to the PS Gallery

Yup, I think that's the ultimate goal.

> We just have to make sure that it fits the definition of the well-formed module and adheres to the PS Gallery publishing guidelines.

Totally agreed the auto-generated PowerShell module needs to conform to the guideline.

beatcracker:
Here is the new branch with more fixes: [powershell_generator_fix_1 ](https://github.com/beatcracker/swagger-codegen/tree/powershell_generator_fix_1)

samples/client/petstore/powershell_test/src/IO.Swagger/API/PetApi.ps1
```diff
@@ -0,0 +1,164 @@
...
+    [CmdletBinding()]
+    Param (
+        [Parameter(Position = 1, ValueFromPipeline = $true, ValueFromPipelineByPropertyName = $true, Mandatory = $true)]
+        [Int64]
+        ${petId},
+        [Parameter(Position = 2, ValueFromPipeline = $true, ValueFromPipelineByPropertyName = $true, Mandatory = $false)]
+        [String]
+        ${additionalMetadata},
+        [Parameter(Position = 3, ValueFromPipeline = $true, ValueFromPipelineByPropertyName = $true, Mandatory = $false)]
+        [IO.Swagger.Model.String]
```
beatcracker: Should be just plain `[String]`.

samples/client/petstore/powershell_test/src/IO.Swagger/API/PetApi.ps1
```diff
@@ -0,0 +1,164 @@
+function Invoke-PetApiAddPet {
+    [CmdletBinding()]
+    Param (
+        [Parameter(Position = 1, ValueFromPipeline = $true, ValueFromPipelineByPropertyName = $true, Mandatory = $true)]
```
beatcracker: Position should start from `0`. Unfortunately, I couldn't figure out how to do this with `jmustache`.

wing328: No worry. I'll take care of it

samples/client/petstore/powershell_test/src/IO.Swagger/API/PetApi.ps1
wing328: You're right. Will fix it.

samples/client/petstore/powershell_test/src/IO.Swagger/API/PetApi.ps1
wing328: This's fixed. Will push it out later this weekend.

samples/client/petstore/powershell_test/src/IO.Swagger/API/PetApi.ps1
wing328: Couldn't find a fix at the moment. Will give it another try over the weekend.

samples/client/petstore/powershell_test/src/IO.Swagger/API/PetApi.ps1
wing328: Pushed the fix.

samples/client/petstore/powershell_test/src/IO.Swagger/API/PetApi.ps1
wing328: Found a fix and pushed the commit https://github.com/swagger-api/swagger-codegen/pull/5789/commits/dc31206a7afb0146187163f92f5ea66f21150e17

wing328:
Merged https://github.com/beatcracker/swagger-codegen/tree/powershell_generator_fix_1 into my branch.

modules/swagger-codegen/src/main/resources/powershell/Build.ps1.mustache
wing328: Didn't have a chance to work on this yet.

Will do it later today or tomorrow.

beatcracker:
Great, I'll try to check updated generator towards the evening.

modules/swagger-codegen/src/main/resources/powershell/Build.ps1.mustache
wing328: @beatcracker UPDATE: added GUID support via 50b8795

beatcracker:
> Great, I'll try to check updated generator towards the evening.

Well, that didn't work out, too busy 😞. I'll try to have a look at the new code during the weekend.

wing328:
> Well, that didn't work out, too busy 😞. I'll try to have a look at the new code during the weekend.

No worry. Please review the change when you're free.

beatcracker:
Managed to build and test latest version. There is still one issue in the `New-Pet.ps1`: double `[` in the parameter type. I couldn't find relevant code in the generator. After fixing this issue manually, I'm able to import module and make API calls using generated cmdlets.

```posh
[Parameter(Position = 0, ValueFromPipelineByPropertyName = $true, Mandatory = $true)]
[[String]]
${photoUrls},
```

I haven't tested all commands, since this is time consuming. I could provide template for basic [Pester](https://github.com/pester/Pester) tests so they could be generated alongside with the code.

Unfortunately, it looks like Petstore API allows anyone to change stored data, so it would be hard to test against (e,g, `Invoke-PetApiGetPetById` requires fixed `Id` and there is no way PetStore API can provide it, since anyone can add/delete pets).

Usually this is solved with mocking `Invoke-WebRrequest`  in Pester, but since this module is using binaries this approach is not available.

wing328:
> double [ in the parameter type. I couldn't find relevant code in the generator

so `[[String]]` needs to be replaced as `[String]`, right? I'll take care of that.

>  I could provide template for basic Pester tests so they could be generated alongside with the code.

Good idea. We can generate test files for each Api class (e.g. PetApi.ps1) for easier testing. I will generate some very basic files to start with.

> Unfortunately, it looks like Petstore API allows anyone to change stored data, so it would be hard to test against

In our CIs (e.g. AppVeyor for testing C# API clients), we run a local Petstore server so bad data in the public server won't affect our test.

beatcracker:
> [[String]] needs to be replaced as [String], right?

Yes.

> Good idea. We can generate test files for each Api class (e.g. PetApi.ps1) for easier testing. 

Here is the basic test example for Pester:

```posh
Describe 'IO.Swagger PetStore API' {
    Context 'PetApi' {
        It 'Invoke-PetApiGetPetById' {
            $ret = Invoke-PetApiGetPetById -petId 0
            $ret | Should BeOfType IO.Swagger.Model.Pet
            $ret.id | Should Be 0
        }
    }

    Context 'StoreApi' {
        It 'Invoke-StoreApiGetOrderById' {
            $ret = Invoke-StoreApiGetOrderById -orderId 0
            $ret | Should BeOfType IO.Swagger.Model.Order
            $ret.Id | Should Be  0
        }
    }
}
```

wing328:
Using the following as an example
``` 
        [Parameter(Position = 0, ValueFromPipelineByPropertyName = $true, Mandatory = $true)]
        [[String]]
        ${photoUrls},
```
The parameter is an array of string. If we simply put [String], how can Powershell distinguish between array of string and just string?

beatcracker:
> The parameter is an array of string. If we simply put [String], how can Powershell distinguish between array of string and just string?

For array of strings, it should look like this: `[String[]]`

wing328:
> For array of strings, it should look like this: [String[]]

Thanks. I'll fix it.

beatcracker:
Almost there, now we need to remove one set of square brackets at we'll get correct array notation: `[[String[]]]` --> `[String[]]`.

beatcracker:
Yippee ki-yay, it builds and runs!

wing328:
@beatcracker thanks for your tremendous effort 👏 👏 👏 . Finally we've a working PowerShell generator. There are some enhancements I can think of and I'll create a separate task to track those.

beatcracker:
You're welcome, glad we finally did it.

> There are some enhancements I can think of and I'll create a separate task to track those.

I also plan to have a fresh look at the code/module structure to see if it can be improved.

wing328:
Tweet to promote the new generator: https://twitter.com/wing328/status/877184938344239104

wing328:
Also added @beatcracker as the template creator of PowerShell in the readme: https://github.com/swagger-api/swagger-codegen/blob/master/README.md#template-creator

Issue #5788, opened 2017-06-06
CPPRest generated models are not using inheritance


harishchoragudi:
##### Description

CPPRest's generated models are extending from ModelBase instead of the actual Class specified in REST contract provided as input to swagger-codegen-cli. This classs ModelBase is directly hardcoded in model-header.mustache template. This needs to be fixed. Also there is a method call to ModelBase::fromJson() which takes a map's value but instead a map's entry is being passed. This needs to be corrected in api-source.mustache template.

##### Steps to reproduce
1. Create a contract where the models using inheritance. 
   For example if Class A inherits from class B - but the generated models both class A and B extend from ModelBase.
2. Also in api-source.mustache, this line of code (itemObj->fromJson(item)) does not work if a map is being processed as there is no fromJson() method in ModelBase class that takes a map entry. This is causing a compilation error. So it needs to be fixed to itemObj->fromJson(item.second);

wing328:
Closed via #5791

PR #5791, opened 2017-06-06
Fix for issue#5788 -CPPRest generated models are not using inheritance as specified in contract

harishchoragudi:
### PR checklist

### Description of the PR
This PR will fix the issue#5788
 https://github.com/swagger-api/swagger-codegen/issues/5788

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/utils/ModelUtils.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/utils/ModelUtils.java
@@ -9,6 +9,8 @@ public class ModelUtils {
     /**
      * Searches for the model by name in the map of models and returns it
      *
+     * @param name Name of the model
+     * @param models Map of models
      * @return model
      */
     public static CodegenModel getModelByName(final String name, final Map<String, Object> models) {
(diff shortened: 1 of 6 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/utils/ModelUtils.java
```diff
@@ -0,0 +1,29 @@
+package io.swagger.codegen.utils;
+
+import io.swagger.codegen.CodegenModel;
+
+import java.util.List;
+import java.util.Map;
+
+public class ModelUtils {
+    public static CodegenModel getModelByName(final String name, final Map<String, Object> models) {
```
wing328: @harishchoragudi It would be nice to add a proper docstring explaining what this function does.

harishchoragudi: I could just add something like this but I thought it's pretty self explanatory  looking at  method signature.
   /**
     *
     * @param name  Name of the model
     * @param models Map of models
     * @return  Searches for the model by name in the map of models and returns it
     */

wing328: Agreed that the function name is pretty self explanatory but still we want a proper docstring for newly-added functions if it's not too much work to the contributor.

harishchoragudi: I now added the javadoc.

wing328:
@harishchoragudi thanks for the fix, which has been merged into master.

Issue #5793, opened 2017-06-07
[HASKELL] Incorrect escaping when generating field names


mmosemmle:
##### Description
The 'Type mismatch on container access' issues found by lgtm.com:
https://lgtm.com/projects/g/swagger-api/swagger-codegen/alerts/
means that property names are not properly escaped when haskell code is generated.

##### Swagger-codegen version
master @ec19e8f

##### Swagger declaration file content or url
https://gist.github.com/mmosemmle/c2165d6eb37e77b670c383b37118666d

##### Command line used for generation
```
java -jar swagger-codegen/modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -l haskell -i swagger.yaml -o ./test
```
##### Steps to reproduce
Download the swagger.yaml file from the above link, then do use the above command, and check
./test/lib/TestGeneratedBug/Type.hs
to see a field called Test??-Data in line 22.

##### Suggest a Fix

The fix is fairly simple and I am happy to create a PR to fix it. Thanks.

wing328:
@mmosemmle thanks for reporting the issue and creating a PR for the fix.

Let me know if you need help with the PR.

mmosemmle:
Thanks. Created a PR for the fix. Btw, it may also be useful to try out the PR integration from lgtm.com, which will discover issues like this during a PR and avoid them being merged. Thanks.

wing328:
Closed via #5794 

Thanks for the suggestion about lgtm.com. Will check it out later.

PR #5794, opened 2017-06-07
Fix issue 5793

mmosemmle:
### PR checklist

### Description of the PR

Fix a bug that leads to generated field names not escaping special characters. (#5793)

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/HaskellServantCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/HaskellServantCodegen.java
@@ -9,12 +9,14 @@ import io.swagger.models.Operation;
 import io.swagger.models.Swagger;
 
 import java.util.*;
+import java.util.regex.Pattern;
 
 public class HaskellServantCodegen extends DefaultCodegen implements CodegenConfig {
 
     // source folder where to write the files
     protected String sourceFolder = "src";
     protected String apiVersion = "0.0.1";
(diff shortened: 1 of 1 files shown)


wing328:
I ran "./bin/haskell-servant-petstore.sh" to update the Haskell Petstore sample but found the following change:
```
--- a/samples/server/petstore/haskell-servant/lib/SwaggerPetstore/Types.hs
+++ b/samples/server/petstore/haskell-servant/lib/SwaggerPetstore/Types.hs
@@ -25,7 +25,7 @@ import Data.Function ((&))
 -- | Describes the result of uploading an image resource
 data ApiResponse = ApiResponse
   { apiResponseCode :: Int -- ^ 
-  , apiResponseType :: Text -- ^ 
+  , apiResponse'Underscoretype :: Text -- ^ 
   , apiResponseMessage :: Text -- ^ 
   } deriving (Show, Eq, Generic)
```

Seems like the fix will break the auto-generated code. Would you please take another look?

mmosemmle:
Sorry about the mistake and thanks for catching this. It looks like this happens because the field name type in ApiResponse is a reserved word, so it gets prepended with an underscore, which then gets ignored when camelize is applied to it. I added a check to remove the leading slash in this case. Please take a look when you have time. Thanks.

wing328:
@mmosemmle 

> Sorry about the mistake and thanks for catching this.

No problem and we appreciate your contribution to make the Haskell generator better.

wing328:
@mmosemmle the PR looks good to me 👍

Issue #5746, opened 2017-05-31
[Swift3] Decoder might be a little bit more liberal


rdalverny:
##### Description

It happens that some server implementations send booleans or integers as strings, yet the client library expects a boolean or an integer (this is somewhat messy, agreed - in my case).

At this point, on such case, a generated Swift3 client will `fatalError("Source \(source) is not convertible to type \(clazz): Maybe swagger file is insufficient")` (in [swift3/Models.mustache#L92](https://github.com/swagger-api/swagger-codegen/blob/9070c7133ceacaf96898d32105625a380cea04e2/modules/swagger-codegen/src/main/resources/swift3/Models.mustache#L92)).

Well, the server implementations could be more up to spec, although that's not always the case. The client could just accept only strings and later do the conversion.

However, having the client library cast the server response in the correct type at once would be nice (the Java client library does this, as it relies on gson). More, this also allows the server implementations to stay as they are, or to get up to spec, without having to update the client code.

##### Swagger-codegen version

I am using the HEAD revision at this point.

##### Steps to reproduce

 * in the definition, declare a JSON attribute response for a service to be an integer or a boolean
 * have a particular server send a string-encoded integer or boolean (`"123"` instead of `123`, or `"true"` instead of `true`)
 * generate the client Swift3 library
 * query the service
 * appreciate the decoder fatally complain about it

##### Related issues

I am not sure, I'm new to this (Swift and Swagger). However:
 * #2735, although closed, seems related; I am not sure the suggested solution (use "object" for arbitrary JSON data) is appropriate when we have simple types that may be trivially converted (client expecting an int, server sending an int encoded as a string);
 * #5543 is, the other way around (converting a string into a boolean) - just for context; I don't think it has any significance here;
 * #4382 is, in the way an mismatch is handled (fatalError)
 * could #3596, #2133 be, for some context on the decoding part?

##### Suggest a Fix

I am not sure this is the best way to fix this, or even if there would not be a better, more generic way in Swift. However the following comes to work on my side (apart from fixing the server implementations of course). Would using a specialized library be better ([hkellaway/Gloss](https://github.com/hkellaway/Gloss), [ikesyo/Himotoki](https://github.com/ikesyo/Himotoki) or other - at this point, I really have no idea)?

```diff
--- a/modules/swagger-codegen/src/main/resources/swift3/Models.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift3/Models.mustache
[...]

PR #5795, opened 2017-06-07
[Swift3] Set more liberal Swift3 decoder, behind lenientTypeCast option

rdalverny:
This adds a Swift3 option (`lenientTypeCast`) that, when set to true, generates a client JSON decoder that will accept and cast mistyped (by the server) values to their correct (expected by the client, per specification) type, as here:
 - String => Bool ("true" instead of true),
 - String => Int ("123" instead of 123),
 - NSNumber => String (123 instead of "123").

The point is to allow the same client code to handle several server implementations that may (sadly) not be up to spec, or still be "evolving".

Not a perfect/complete solution, not sure if it should be activated along other casts (I think each language/implementation handles this differently), so kept behind an option for Swift3 at this point.

See: #5746.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Swift3Codegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Swift3Codegen.java
@@ -39,11 +39,13 @@ public class Swift3Codegen extends DefaultCodegen implements CodegenConfig {
     public static final String POD_DOCUMENTATION_URL = "podDocumentationURL";
     public static final String SWIFT_USE_API_NAMESPACE = "swiftUseApiNamespace";
     public static final String DEFAULT_POD_AUTHORS = "Swagger Codegen";
+    public static final String LENIENT_TYPE_CAST = "lenientTypeCast";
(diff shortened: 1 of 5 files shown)


jaz-ah:
+1 looks fine, unfortunate but glad it's off by default

wing328:
The test result (after merging with master) is good: https://travis-ci.org/swagger-api/swagger-codegen/builds/240469935

@rdalverny thanks for your contribution.

rdalverny:
@jaz-ah @wing328 thanks to both of you!

wing328:
@rdalverny I've sync'd the change into 2.3.0 (a branch with breaking changes) via https://github.com/swagger-api/swagger-codegen/commit/022eb0d379a057fd386e04deb16184f1b07302f8

Please help take a look to see if I resolved the merge conflict correctly: https://github.com/swagger-api/swagger-codegen/commit/022eb0d379a057fd386e04deb16184f1b07302f8#diff-940535e66d162a893df844fd9b6b45ffR143

rdalverny:
@wing328 sorry for the delay. Looks good to me, although I'm not sure, as returned values are now wrapped in a .success/.failure (I can rewrite this if you like, as well as to match the same "if" style). I've got to try it.

Issue #5760, opened 2017-06-02
Apache Authentication and Authorization Conf Gen


stkrwork:
##### Description

With the information in the Swagger AuthMethods component, a Apache configuration could be generated.

It might also be useful to extend the AuthMethods Class to include a groups Array, to create the group handling in Apache for HTTP Basic/Digest Authentication/LDAP.

Through this one can generate the server  stub and reroute the calls to the server stub via apache. Apache then takes care of the authentication and authorization.

wing328:
Yes, I think that will be useful. There are discussions about generating a DB schema from OpenAPI/Swagger 2.0 spec so there are unlimited possibilities with Swagger Codegen to generate "things" other than API client, server stub, documentation.

stkrwork:
I did make an implementation for it, but it misuses the scope parameters for groups for apache authorization instead as swagger codegen does not support groups for authorization for http basic requests.

wing328:
> as swagger codegen does not support groups for authorization for http basic requests.

What about using vendor extensions to document that in the spec?

stkrwork:
So the user would already include a vendor extension section in the input spec containing the groups?

Something like this:
`"/pet": {
      "post": {
        "tags": [
          "pet"
        ],
        "summary": "Add a new pet to the store",
        "description": "",
        "operationId": "addPet",
        "consumes": [
          "application/json",
          "application/xml"
        ],
        "produces": [
          "application/xml",
          "application/json"
        ],
        "parameters": [
          {
            "in": "body",
            "name": "body",
            "description": "Pet object that needs to be added to the store",
            "required": true,
            "schema": {
              "$ref": "#/definitions/Pet"
            }
          }
        ],
        "responses": {
          "405": {
            "description": "Invalid input"
          }
        },
        "security": [
          {
            "petstore_auth": [
              "write:pets",
              "read:pets"
            ]
          }
        ]
      },
      "put": {
        "tags": [
          "pet"
        ],
        "summary": "Update an existing pet",
        "description": "",
        "operationId": "updatePet",
        "consumes": [
          "application/json",
          "application/xml"
        ],
        "produces": [
          "application/xml",
          "application/json"
        ],
        "parameters": [
          {
            "in": "body",
            "name": "body",
            "description": "Pet object that needs to be added to the store",
            "required": true,
            "schema": {
              "$ref": "#/definitions/Pet"
            }
          }
        ],
        "responses": {
          "400": {
[...]

PR #5800, opened 2017-06-08
ApacheConfGen for HTTP Basic Authentication Proposed implementation

stkrwork:
### PR checklist

### Description of the PR

This is a work in progress implementation for an Apache Configuration Generation for protecting paths in an api. It only generates the location component of the apache configuration, not the whole configuration


stkrwork:
To see that everything works correctly i temporarily chose isOAuth instead of isBasic, to see if the idea works. In the final version, this will be changed to isBasic in the template.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Apache2Confgen.java
```diff
@@ -0,0 +1,106 @@
...
+import io.swagger.codegen.DefaultCodegen;
+import io.swagger.codegen.SupportingFile;
+
+public class Apache2Confgen extends DefaultCodegen implements CodegenConfig {
+  public static final String USER_INFO_PATH = "userInfoPath";
+  protected String userInfoPath = "/var/www/html/";
+
+  @Override
+  public CodegenType getTag() {
+    return CodegenType.OTHER;
```
wing328: What about introducing a new type "CONFIG" instead of using OTHER?

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Apache2Confgen.java
```diff
@@ -0,0 +1,106 @@
...
+
+import io.swagger.codegen.CliOption;
+import io.swagger.codegen.CodegenConfig;
+import io.swagger.codegen.CodegenConstants;
+import io.swagger.codegen.CodegenOperation;
+import io.swagger.codegen.CodegenType;
+import io.swagger.codegen.DefaultCodegen;
+import io.swagger.codegen.SupportingFile;
+
+public class Apache2Confgen extends DefaultCodegen implements CodegenConfig {
```
wing328: What about naming the class as "Apache2ConfigCodegen" to make it consistent with other generators (clients, servers, documentations)?

wing328:
@stkrwork thanks for the PR. I wonder if you can add the samples under "samples/config/petstore/apache2" so that we can get an idea of what the output looks like.

stkrwork:
The samples are currently generated without the additional properties set.

wing328:
@stkrwork thanks for creating the first config generator for Swagger Codegen 👏 👏 👏  

Discussion for future enhancements can be found in https://github.com/swagger-api/swagger-codegen/issues/5936

Linked to https://github.com/swagger-api/swagger-codegen/issues/5760

PR #5809, opened 2017-06-09
[python] Use dashes for Python PyPI package name since it is a convention

frol:
### PR checklist

### Description of the PR

PyPI packages index conventionally uses dashes instead of underlines in the package names. Yet, Python cannot import a package folder with dashes, so the `packageName` should not have dashes, and since it is only PyPI convention, I have implemented a quick fix. Alternatively, `projectName` configuration may be introduced (like it is done for JS, Swift, and Clojure generators), but I have no experience with Java :(

### DIFF
--- a/modules/swagger-codegen/src/main/resources/python/setup.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/setup.mustache
@@ -5,7 +5,7 @@
 import sys
 from setuptools import setup, find_packages
 
-NAME = "{{packageName}}"
+NAME = "{{packageName}}".replace('_', '-')
 VERSION = "{{packageVersion}}"
 {{#apiInfo}}
 {{#apis}}
--- a/samples/client/petstore/python/setup.py
+++ b/samples/client/petstore/python/setup.py
@@ -14,7 +14,7 @@
 import sys
 from setuptools import setup, find_packages
 
-NAME = "petstore_api"
+NAME = "petstore_api".replace('_', '-')
 VERSION = "1.0.0"
 # To install the library, run the following
 #


wing328:
@frol thanks for the PR. In the usage, we already mention using snake_case for package name: 
```
	packageName
	    python package name (convention: snake_case). (Default: swagger_client)
```
Shall we leave it to the users to decide what naming convention to use rather than enforcing it in setup.py? The users may want to keep naming the package name in non-"snake_case" due to historical reasons.

frol:
@wing328 The package name CANNOT contain dashes by design. However, Python packages index conventionally use dashes in names. For example:

[Flask-SQLAlchemy](https://pypi.python.org/pypi/Flask-SQLAlchemy) package on PyPI is named (as most of the packages there) with a dash between Flask and SQLAlchemy. However, the Python package name is [`flask_sqlalchemy`](https://github.com/mitsuhiko/flask-sqlalchemy/tree/master/flask_sqlalchemy). It is a confusing naming after all; it would have been more clear if they stick to "project name" instead of "PyPI package name" vs "Python package name". The point is that the project on Github and on PyPI is called "flask-sqlalchemy" (with a dash), but once you install that you access Python package with `import flask_sqlalchemy` (with an underscore).

wing328:
@frol thanks for the example.

Shall we create an additional mustache tag (e.g. packageNameInSetupPy) to store the package name with underscore replaced with dash in the python generator instead?

I can do it later if my suggested approach is ok.

frol:
@wing328 I would introduce `projectName` config option (instead of `packageNameInSetupPy`) just to be in line with the other generators (JS, Swift, etc). However, it doesn't matter that much. Other than that, the plan sounds good to me!

wing328:
@frol I'll submit a PR by tomorrow and would need your help to review.

frol:
@wing328 Sure! Feel free to ping me once PR is ready.

wing328:
@frol I submitted https://github.com/swagger-api/swagger-codegen/pull/5836. Please review when you've time. Thanks.

wing328:
Closed via https://github.com/swagger-api/swagger-codegen/pull/5836

PR #5814, opened 2017-06-09
[javascript] Use arrow function with usePromises in ES6 ApiClient.js for superagent callback to preserve `this` context

kemurphy:
### PR checklist

### Description of the PR

Without this change, consuming the API results in `cannot read property 'deserialize' of undefined` from `ApiClient`.  Using a plain anonymous function instead of an arrow function drops the `this` context for `ApiClient` in the superagent callback.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Javascript-es6/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Javascript-es6/ApiClient.mustache
@@ -447,7 +447,7 @@ export default class ApiClient {
         }
 
         {{#usePromises}}return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


kemurphy:
Any way to rerun these checks? Looks like it failed with:
```
[ERR!] GitHub API: GitHub rate limit reached.
    To increase the limit use GitHub authentication.
    See: https://github.com/DefinitelyTyped/tsd#tsdrc
```

wing328:
Let me look into it (I'll update the Petstore samples for JS as well)

cc @dinukadesilva

kemurphy:
I ran the petstore scripts but it looks like none of them use `usePromises`, so nothing really changed.

PR #5819, opened 2017-06-11
[JS] Fix JavaScript string return type

cliffano:
Follow up to #5759 , forgot to also fix javascript-es6.


wing328:
@cliffano thanks for the fix for JS ES6 API client.

PR #5832, opened 2017-06-12
[Java][RestTemplate] fix compilation issue with API client (resttemplate)

simingweng:
### PR checklist

### Description of the PR

adapt the "resttemplate" mustaches for generating client which use [threeten](http://www.threeten.org) together with [jackson-datatype-threetenbp](https://github.com/joschi/jackson-datatype-threetenbp) as default dateLibrary


wing328:
@simingweng thanks for the fix. I tested it locally and the result is good except that `sbt compile` failed.

I wonder if you can update the template for `build.sbt` with the fix as well

wing328:
Closes #5640

simingweng:
@wing328 it seems there has never been a sbt build file created for the "resttemplate" target. I don't have any experience with sbt build system. Probably it would be more efficient to create another issue for missing sbt build support, so someone better than I could transcript maven/gradle to sbt.

wing328:
@simingweng agreed. We'll tackle it separately.

PR merged into master. Thanks for your contribution.

PR #5836, opened 2017-06-13
[Python] add project name to setup.py in python client

wing328:
### PR checklist

### Description of the PR

Add project name (e.g. petstore-api) to setup.py so as to conform to Python convention. By default the project name is automatically determined based on `packageName` by replacing "_" with "-"

Note that I didn't add another CLI option as I want to avoid too many option in customising the output. (developers can still set it via --additional-properties)

(PythonClientCodegen.java was set to Unix file format)


frol:
I suggest a few changes:

1. Semantically, `projectName` should be the "most important" configuration, which would be used to generate `packageName` (meaning that it is the other way around than it is implemented now), BUT given the fact that `packageName` was introduced earlier, it makes sense to have this backward `packageName` to `projectName` generation. Can both `projectName` -> `packageName` and `packageName` -> `projectName` autogenerations be implemented? Does this make sense to you?
2. Use `projectName` in the README template

wing328:
1. Make sense but I don't have time now to implement the logic you suggested. I'll create a task to see if anyone from the community can help on that.

2. Will update it.

PR #5838, opened 2017-06-13
[C++] Server Stub Code Generator based on Pistache

sebymiano:
### PR checklist

### Description of the PR
This PR adds an implementation of a C++ server generator that uses the [Pistache](https://github.com/oktal/pistache) REST framework. 
It uses the [JSON for Modern C++](https://github.com/nlohmann/json/) library to provide JSON serialization/deserialization of the generated C++ API Models.


modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
```diff
@@ -117,7 +117,7 @@ public boolean isRestfulIndex() {
      * @return true if act as Restful show method, false otherwise
      */
     public boolean isRestfulShow() {
-        return "GET".equals(httpMethod) && isMemberPath();
```
wing328: @sebymiano thanks for the new generator. I noticed that you've removed the condition "isMemberPath". Is that causing issue for the Pistache C++ server generator?

`isRestfulShow` and other methods are used by other server stub generators as well so removing "isMemberPath" may have a wider impact.

sebymiano: @wing328 I'm sorry, it was my fault. I changed those methods when I started implementing the generator and right after I noticed they were not necessary, but I forgot to revert the changes :). Now it should be ok. 

FYI, I added `equalsIgnoreCase` rather than `equals` because the `httpMethod` variable is not always uppercase. Do you agree?

wing328: @sebymiano 

> I'm sorry, it was my fault.

No worry. We appreciate your contribution for another C++ server generator.

> FYI, I added equalsIgnoreCase rather than equals because the httpMethod variable is not always uppercase. Do you agree?

Yup, make sense to me.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PistacheServerCodegen.java
```diff
@@ -0,0 +1,412 @@
...
+            parameter.dataType = "std::shared_ptr<" + parameter.dataType + ">";
+        }
+    }
+
+    /**
+     * Location to write model files. You can use the modelPackage() as defined
+     * when the class is instantiated
+     */
+    public String modelFileFolder() {
+        return outputFolder + "/model";
```
wing328: @sebymiano Even though Pistache does not support Windows at the moment (https://github.com/oktal/pistache/issues/6), I'll still file a change later to replace "/" with File.separator similar to what we do in other generators.

wing328:
@sebymiano Thanks for the PR. I did some tests and there's no impact to other generators (e.g. Rails, PHP Slims).

I'll file a PR to make some minor changes (as mentioned in my comment).

wing328:
Added @sebymiano as Pistache template creator via https://github.com/swagger-api/swagger-codegen/commit/66c64495a9dd0f14021c0d01b3a25466f8acf45c

wing328:
I'll also send out a tweet about this new generator shortly (so as to promote the new generator)

frol:
@sebymiano Do I get it right that I cannot run a single server to handle all the namespaces on a single port using this stub? I understand that I can run a reverse proxy server in front of those independent API namespace servers running on different ports, but I just want to confirm that my understanding is correct.

PR #5843, opened 2017-06-14
Html2 speed Up | Raw html using java parser instead of JS parser

JuH33:
Some code to speed up the render of HTML2 documentation. 
- Parse Html markdown with Java's parser instead of JS. 
- Kill some JS that slow down the page loading ( and causes Flash of Unstyled Text (FOUT) )
- Use mustache variable to build html instead of javascript

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/StaticHtml2Generator.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/StaticHtml2Generator.java
@@ -1,14 +1,7 @@
 package io.swagger.codegen.languages;
 
-import io.swagger.codegen.CliOption;
-import io.swagger.codegen.CodegenConfig;
-import io.swagger.codegen.CodegenConstants;
-import io.swagger.codegen.CodegenOperation;
-import io.swagger.codegen.CodegenParameter;
-import io.swagger.codegen.CodegenResponse;
-import io.swagger.codegen.CodegenType;
-import io.swagger.codegen.DefaultCodegen;
-import io.swagger.codegen.SupportingFile;
+import com.samskivert.mustache.Mustache;
+import io.swagger.codegen.*;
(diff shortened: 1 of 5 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/StaticHtml2Generator.java
```diff
@@ -215,6 +209,31 @@ private void preparHtmlForGlobalDescription(Swagger swagger) {
...
+    /**
+     * Format to HTML the enums contained in every operations
+     *
+     * @param parameterList The whole parameters contained in one operation
+     * @return String | Html formated enum
+     */
+    public List<CodegenParameter> postProcessParameterEnum(List<CodegenParameter> parameterList) {
+        for(CodegenParameter parameter : parameterList) {
+            if (parameter.isEnum) {
+                parameter.enumFormatted = "";
```
wing328: @JuH33 thanks for the enhancement. Instead of adding another property "enumFormatted", what about using vendor extensions instead?

e.g. https://github.com/swagger-api/swagger-codegen/blob/9f81aa4bfced6eddbce0131853d27592653ee3e2/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/StaticHtml2Generator.java#L187

JuH33: Hey, yeah np. I can switch to vendor extensions. am gonna do it today.
ps: thanks for the tips, looks better than what i did !

JuH33: Done @wing328 ;)

wing328:
@JuH33 PR merged into master. Thanks for your contribution.

(FYI. I've updated the html2 petstore sample via b631938)

PR #5864, opened 2017-06-17
Support a true "java8" option

Randgalt:
### PR checklist

### Description of the PR

The "dateLibrary" option for java, sadly, sets a mustache value "java8". This change updates this so that "java8" in the mustache libraries means what it should mean - use all java8 classes. In this case, there's no need for the third party Base64 library as java8's JDK has this built in. In my view, the "dateLibrary" value "java8" should be deaprecated but that should be a separate PR.


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java
```diff
@@ -359,10 +367,17 @@ public void processOpts() {
...
         if(additionalProperties.containsKey(DATE_LIBRARY)) {
-            setDateLibrary(additionalProperties.get("dateLibrary").toString());
+            setDateLibrary(additionalProperties.get(DATE_LIBRARY).toString());
             additionalProperties.put(dateLibrary, "true");
         }
 
+        if(additionalProperties.containsKey(JAVA8_MODE)) {
+            setJava8Mode(Boolean.parseBoolean(additionalProperties.get(JAVA8_MODE).toString()));
+            if ( java8Mode ) {
+                additionalProperties.put("java8", "true");
```
ePaul: Hmm, maybe also put the false value in the other case? (Or just `additionalProperties.put("java8", java8Mode.toString());`)

By the way, why are you using `"java8"` here and not `JAVA8_MODE`?

ePaul:
You indicated in the PR comment that you updated the samples, but I don't see any changed samples. Are there no changes?

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java
Randgalt: Well - it was used below. I'll change it.

Randgalt:
> You indicated in the PR comment that you updated the samples 

I ran the script - maybe I did something wrong? I'll check.

cbornet:
Can you check if it's not already fixed on branch 2.3.0 ? A jsr310 flag was introduced to distinguish dates and jdk.

Randgalt:
@cbornet I checked branch 2.3.0 and I see `threetenbp` but that's still only applies to the dateLibrary. This PR is focused on using Java8 classes everywhere. In this instance, there's no need for the 3rd party Base64 library as that's builtin to Java 8 now. 

This PR also "takes ownership" of the `java8` mustache variable. IMO that should not have been used to denote which dateLibrary you want to use.

cbornet:
See https://github.com/swagger-api/swagger-codegen/blob/2.3.0/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java#L382 : there is a jsr310 flag on 2.3.0 that is used specifically for JSR310 dates.

Randgalt:
@cbornet I guess I'm not conveying the point of this PR. This is _not_ about the date library. This is for Java8 classes _in general_. Base64 is part of the Java 8 JDK now. There is no longer any need to import a third party library for this purpose.

cbornet:
@Randgalt OK. Then I don't think we need yet another option (we already have too many) there should be only "java8" option for which we can use Base64

Randgalt:
@cbornet this _not_ a new option. This PR takes the "java8" mustache variable that is set by dateLibrary and uses it to mean what it implies: use java8 classes where needed.  The only thing new is a CLI option to set "java8" - this is needed because it no longer makes sense to have "java8" set by dateLibrary (that part should be deprecated).

cbornet:
@Randgalt oh right, I misread your PR :smile: !
Then LGTM

modules/swagger-codegen/src/test/java/io/swagger/codegen/java/JavaClientOptionsTest.java
```diff
@@ -82,6 +82,8 @@ protected void setExpectations() {
             times = 1;
             clientCodegen.setUseBeanValidation(Boolean.valueOf(JavaClientOptionsProvider.PERFORM_BEANVALIDATION));
             times = 1;
+            clientCodegen.setUseBeanValidation(Boolean.valueOf(JavaClientOptionsProvider.PERFORM_BEANVALIDATION));
+            times = 1;
```
ePaul: This duplication looks wrong. I guess you wanted to check for JAVA_8?

Randgalt: Doh - copy/paste error. I'll fix it.

Randgalt:
How do I rerun  the checks?

Randgalt:

Randgalt:
Second time. Not sure what to do:

```
[ERR!] GitHub API: GitHub rate limit reached.
    To increase the limit use GitHub authentication.
    See: https://github.com/DefinitelyTyped/tsd#tsdrc
```

wing328:
@Randgalt I've restarted the Travis job. If it's still failing, I'll test locally.

Randgalt:
ping

wing328:
Test results after sample update are good: https://circleci.com/gh/swagger-api/swagger-codegen/518

wing328:
@Randgalt PR merged into master. Thanks for your contribution.

I wonder if you can make the same enhancement to 2.3.0 branch as I got a lot of merge conflicts syncing master into 2.3.0.

Randgalt:
@wing328 sure

Randgalt:
https://github.com/swagger-api/swagger-codegen/pull/5955

Issue #5709, opened 2017-05-26
[typescript-angular2] Invalid import statement in variables.ts


hallids:
##### Description

The generated file `variables.ts` has this code as the first line:
```typescript
import { InjectionToken<string> } from '@angular/core';
```
As far as I can tell, this is not valid typescript, and will not compile. It produces this error:

> Module parse failed:
> ...
> generated_api\variables.ts Unexpected token (1:31)
> You may need an appropriate loader to handle this file type.
> | import { InjectionToken } from <string>;
> | from;
> | '@angular/core';

Removing `<string>` fixes the problem and the code compiles: 
```typescript
import { InjectionToken } from '@angular/core';
```

##### Swagger-codegen version

2.3.0-SNAPSHOT (41527ead54c881f88324641809dcabfef9bf9e00)

##### Command line used for generation

`java -jar ./swagger-codegen-cli.jar generate -i /tmp/swagger.json -l typescript-angular2 -o output/dir/`

##### Related issues

I believe this is the same as https://github.com/swagger-api/swagger-codegen/issues/5652, although it looks like the markdown removed `<string>` in that issue so it's not immediately clear.

##### Suggest a Fix

Making the changes below fixed the problem for me.

```diff
diff --git a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngular2ClientCodegen.java b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngular2ClientCodegen.java
index 17c4b6df6..89b70d742 100644
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngular2ClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngular2ClientCodegen.java
@@ -36,7 +36,7 @@ public class TypeScriptAngular2ClientCodegen extends AbstractTypeScriptClientCod
     protected String npmName = null;
     protected String npmVersion = "1.0.0";
     protected String npmRepository = null;
-    protected String injectionToken = "InjectionToken<string>";
+    protected String injectionToken = "InjectionToken";

     public TypeScriptAngular2ClientCodegen() {
         super();
diff --git a/modules/swagger-codegen/src/main/resources/typescript-angular2/variables.mustache b/modules/swagger-codegen/src/main/resources/typescript-angular2/variables.mustache
index f5cdcf76e..42b2bd0bf 100644
[...]

Issue #5570, opened 2017-05-05
Discussion - the naming of angular and angular2 should be corrected


taxpon:
##### Description
- Angular2 is no longer Angular2, just [Angular](https://angular.io/) officially
    - Actually, the latest stable version is v4 and it is announced that the major version of Angular will be updated every 6 months
- Angular (in this repo) is officially called [AngularJS](https://angularjs.org/)

##### Suggest a Fix
- The naming of each library in this repository should be corrected as the official usage (as follows) so that new users of swagger-codegen will not confuse about naming

|before(current)|after|
|---------------|-----|
|typescript-angular (*1)|typescript-angularjs|
|typescript-angular2|typescript-angular (*2)|

##### Note
- It is possible to make another confuse for existing users, because *1 and *2 is the same name in the above table
- IMHO, angular version in package.json of typescript-angular2 should be upgraded to v4 with this update
- If it's ok, I'll file a PR

wing328:
What about consolidating typescript-angular and typescript-angular2 into one and use the "library" option to indicate which version (1x, 2x, 4x) of Angular to use in 2.3.0 branch?

kenisteward:
@wing328 the real (and I think only) delineation that needs to be made is if it is AngularJS (With Scopes) or Angular (With RXJS which is as of 2)

What you proposed I think may require more maintenance than requires. (update new library for every version change) when all we really need to know is basically old scoping or new scoping.

I think taking your same idea and instead of version using something like Legacy / Current  or Scope / RxJS would be more fitting as I don't see major changes happening to the way new Angular's HTTP works.

It would definitely be a 2.3.0 breaking changes milestone.

sebastianhaas:
I upvote @taxpon's suggestion. As mentioned in #5866 I feel like supporting both AngularJS and Angular (>2.x) with one single generator is probably making things more complicated.

I'll suggest a something like a `--ngVersion` flag to determine whether the API client should be generated for use with Angular 2, 3, 4 whatever, so that we can comply with the latest best practices (InjectionToken instead of OpaqueToken) while preserving backwards compatibility withing the Angular (2) family.

austin94:
I agree with @taxpon in that the naming should be `typescript-angularjs` for 1.X and `typescript-angular` for 2+.

The `--ngVersion` flag that @sebastianhaas suggested is probably the best approach for now since the Angular releases will be happening with more frequency.

wing328:
[...]

PR #5866, opened 2017-06-18
[Typescript-Angular2]  Removed invalid BASE_PATH InjectionToken type

austin94:
### PR checklist

Bug: #5709 

### Description of the PR

Currently in `variables.ts` the import of `InjectionToken` is failing to compile due to the included <string> type.  My original plan was to hard-code the type in the template, but unfortunately OpaqueTokens do not take a type.  My solution is to not include the type at all for basePath.

I have tested this against the existing Petstore application and the type isn't a requirement for base types.

### Unrelated to my changes
It appears the `typescript-angular2-petsore.sh` files haven't been ran in a while.  Several of those changes are unrelated to my change, but I am including them along with this PR.  Let me know if the unrelated items should be removed!


wing328:
Thanks for the PR.

CC @taxpon

sebastianhaas:
I'm not sure if this is the path we want to follow. If we do this, we're accepting a regression (loosing additional type safety) for all those who are using an up-to-date version of Angular, in order to ensure backward compatability.

@wing328 do you see a way to get rid of the parameterized type just for the import while preserving it for any other usages?

topce:
agree with @sebastianhaas 
@wing328 
It could be done like this  in variables.mustache:
import { {{{injectionTokenImport}}} } from '@angular/core';

export const BASE_PATH = new {{{injectionToken}}}('basePath');
export const COLLECTION_FORMATS = {
'csv': ',',
'tsv': '   ',
'ssv': ' ',
'pipes': '|'
}

in TypeScriptAngular2ClientCodegen.java : 
package io.swagger.codegen.languages;

import java.io.File;
import java.text.SimpleDateFormat;
import java.util.Date;
import java.util.List;
import java.util.Map;

import io.swagger.codegen.CliOption;
import io.swagger.codegen.CodegenModel;
import io.swagger.codegen.CodegenParameter;
import io.swagger.codegen.CodegenOperation;
import io.swagger.codegen.SupportingFile;
import io.swagger.models.ModelImpl;
import io.swagger.models.properties.ArrayProperty;
import io.swagger.models.properties.BooleanProperty;
import io.swagger.models.properties.FileProperty;
import io.swagger.models.properties.MapProperty;
import io.swagger.models.properties.ObjectProperty;
import io.swagger.models.properties.Property;

public class TypeScriptAngular2ClientCodegen extends AbstractTypeScriptClientCodegen {
    private static final SimpleDateFormat SNAPSHOT_SUFFIX_FORMAT = new SimpleDateFormat("yyyyMMddHHmm");

    public static final String NPM_NAME = "npmName";
    public static final String NPM_VERSION = "npmVersion";
    public static final String NPM_REPOSITORY = "npmRepository";
    public static final String SNAPSHOT = "snapshot";
    public static final String USE_OPAQUE_TOKEN = "useOpaqueToken";
    public static final String INJECTION_TOKEN = "injectionToken";
    public static final String INJECTION_TOKEN_IMPORT = "injectionTokenImport";
    public static final String WITH_INTERFACES = "withInterfaces";

    protected String npmName = null;
    protected String npmVersion = "1.0.0";
    protected String npmRepository = null;
    protected String injectionTokenImport = "InjectionToken";
    protected String injectionToken = "InjectionToken<string>";

    public TypeScriptAngular2ClientCodegen() {
        super();
        this.outputFolder = "generated-code/typescript-angular2";

        embeddedTemplateDir = templateDir = "typescript-angular2";
        modelTemplateFiles.put("model.mustache", ".ts");
        apiTemplateFiles.put("api.mustache", ".ts");
        typeMapping.put("Date","Date");
        apiPackage = "api";
        modelPackage = "model";


        this.cliOptions.add(new CliOption(NPM_NAME, "The name under which you want to publish generated npm package"));
        this.cliOptions.add(new CliOption(NPM_VERSION, "The version of your npm package"));
        this.cliOptions.add(new CliOption(NPM_REPOSITORY, "Use this property to set an url your private npmRepo in the package.json"));
        this.cliOptions.add(new CliOption(SNAPSHOT, "When setting this property to true the version will be suffixed with -SNAPSHOT.yyyyMMddHHmm", BooleanProperty.TYPE).defaultValue(Boolean.FALSE.toString()));
        this.cliOptions.add(new CliOption(USE_OPAQUE_TOKEN, "When setting this property to true, OpaqueToken is used instead of InjectionToken", BooleanProperty.TYPE).defaultValue(Boolean.FALSE.toString()));
        this.cliOptions.add(new CliOption(WITH_INTERFACES, "Setting this property to true will generate interfaces next to the default class implementations.", BooleanProperty.TYPE).defaultValue(Boolean.FALSE.toString()));
    }

    @Override
    protected void addAdditionPropertiesToCodeGenModel(CodegenModel codegenModel, ModelImpl swaggerModel) {
        codegenModel.additionalPropertiesType = getSwaggerType(swaggerModel.getAdditionalProperties());
        addImport(codegenModel, codegenModel.additionalPropertiesType);
    }

    @Override
    public String getName() {
        return "typescript-angular2";
    }

    @Override
    public String getHelp() {
        return "Generates a TypeScript Angular2 client library.";
    }

    @Override
    public void processOpts() {
        super.processOpts();
        supportingFiles.add(new SupportingFile("models.mustache", modelPackage().replace('.', File.separatorChar), "models.ts"));
        supportingFiles.add(new SupportingFile("apis.mustache", apiPackage().replace('.', File.separatorChar), "api.ts"));
        supportingFiles.add(new SupportingFile("index.mustache", getIndexDirectory(), "index.ts"));
        supportingFiles.add(new SupportingFile("configuration.mustache", getIndexDirectory(), "configuration.ts"));
        supportingFiles.add(new SupportingFile("variables.mustache", getIndexDirectory(), "variables.ts"));
        supportingFiles.add(new SupportingFile("gitignore", "", ".gitignore"));
        supportingFiles.add(new SupportingFile("git_push.sh.mustache", "", "git_push.sh"));

        if(additionalProperties.containsKey(NPM_NAME)) {
            addNpmPackageGeneration();
        }

        if(additionalProperties.containsKey(USE_OPAQUE_TOKEN) && Boolean.valueOf(additionalProperties.get(USE_OPAQUE_TOKEN).toString())) {
            this.setOpaqueToken();
        }
        additionalProperties.put(INJECTION_TOKEN, this.injectionToken);
        additionalProperties.put(INJECTION_TOKEN_IMPORT, this.injectionTokenImport);

        if(additionalProperties.containsKey(WITH_INTERFACES)) {
            boolean withInterfaces = Boolean.parseBoolean(additionalProperties.get(WITH_INTERFACES).toString());
            if (withInterfaces) {
                apiTemplateFiles.put("apiInterface.mustache", "Interface.ts");
            }
        }
    }

    private void addNpmPackageGeneration() {
        if(additionalProperties.containsKey(NPM_NAME)) {
            this.setNpmName(additionalProperties.get(NPM_NAME).toString());
        }

        if (additionalProperties.containsKey(NPM_VERSION)) {
            this.setNpmVersion(additionalProperties.get(NPM_VERSION).toString());
        }

        if (additionalProperties.containsKey(SNAPSHOT) && Boolean.valueOf(additionalProperties.get(SNAPSHOT).toString())) {
            this.setNpmVersion(npmVersion + "-SNAPSHOT." + SNAPSHOT_SUFFIX_FORMAT.format(new Date()));
        }
        additionalProperties.put(NPM_VERSION, npmVersion);

        if (additionalProperties.containsKey(NPM_REPOSITORY)) {
            this.setNpmRepository(additionalProperties.get(NPM_REPOSITORY).toString());
        }

        //Files for building our lib
        supportingFiles.add(new SupportingFile("README.mustache", getIndexDirectory(), "README.md"));
        supportingFiles.add(new SupportingFile("package.mustache", getIndexDirectory(), "package.json"));
        supportingFiles.add(new SupportingFile("typings.mustache", getIndexDirectory(), "typings.json"));
        supportingFiles.add(new SupportingFile("tsconfig.mustache", getIndexDirectory(), "tsconfig.json"));
    }

    private String getIndexDirectory() {
        String indexPackage = modelPackage.substring(0, Math.max(0, modelPackage.lastIndexOf('.')));
        return indexPackage.replace('.', File.separatorChar);
    }

    @Override
    public String getTypeDeclaration(Property p) {
        Property inner;
        if(p instanceof ArrayProperty) {
            ArrayProperty mp1 = (ArrayProperty)p;
            inner = mp1.getItems();
            return this.getSwaggerType(p) + "<" + this.getTypeDeclaration(inner) + ">";
        } else if(p instanceof MapProperty) {
            MapProperty mp = (MapProperty)p;
            inner = mp.getAdditionalProperties();
            return "{ [key: string]: " + this.getTypeDeclaration(inner) + "; }";
        } else if(p instanceof FileProperty || p instanceof ObjectProperty) {
            return "any";
        } else {
            return super.getTypeDeclaration(p);
        }
    }

    @Override
    public String getSwaggerType(Property p) {
        String swaggerType = super.getSwaggerType(p);
        if(isLanguagePrimitive(swaggerType) || isLanguageGenericType(swaggerType)) {
            return swaggerType;
        }
        return addModelPrefix(swaggerType);
    }

    private String addModelPrefix(String swaggerType) {
        String type = null;
        if (typeMapping.containsKey(swaggerType)) {
            type = typeMapping.get(swaggerType);
        } else {
            type = swaggerType;
        }

        if (!isLanguagePrimitive(type) && !isLanguageGenericType(type)) {
            type = "models." + swaggerType;
        }
        return type;
    }

    private boolean isLanguagePrimitive(String type) {
        return languageSpecificPrimitives.contains(type);
    }

    private boolean isLanguageGenericType(String type) {
        for (String genericType: languageGenericTypes) {
            if (type.startsWith(genericType + "<"))  {
                return true;
            }
        }
        return false;
    }

    @Override
    public void postProcessParameter(CodegenParameter parameter) {
        super.postProcessParameter(parameter);
        parameter.dataType = addModelPrefix(parameter.dataType);
    }

    @Override
    public Map<String, Object> postProcessOperations(Map<String, Object> operations) {
        Map<String, Object> objs = (Map<String, Object>) operations.get("operations");
        List<CodegenOperation> ops = (List<CodegenOperation>) objs.get("operation");
        for (CodegenOperation op : ops) {
            // Convert httpMethod to Angular's RequestMethod enum
            // https://angular.io/docs/ts/latest/api/http/index/RequestMethod-enum.html
            switch (op.httpMethod) {
                case "GET":
                    op.httpMethod = "RequestMethod.Get";
                    break;
                case "POST":
                    op.httpMethod = "RequestMethod.Post";
                    break;
                case "PUT":
                    op.httpMethod = "RequestMethod.Put";
                    break;
                case "DELETE":
                    op.httpMethod = "RequestMethod.Delete";
                    break;
                case "OPTIONS":
                    op.httpMethod = "RequestMethod.Options";
                    break;
                case "HEAD":
                    op.httpMethod = "RequestMethod.Head";
                    break;
                case "PATCH":
                    op.httpMethod = "RequestMethod.Patch";
                    break;
                default:
                    throw new RuntimeException("Unknown HTTP Method " + op.httpMethod + " not allowed");
            }

            // Convert path to TypeScript template string
            op.path = op.path.replaceAll("\\{(.*?)\\}", "\\$\\{$1\\}");
        }

        return operations;
    }

    public String getNpmName() {
        return npmName;
    }

    public void setNpmName(String npmName) {
        this.npmName = npmName;
    }

    public String getNpmVersion() {
        return npmVersion;
    }

    public void setNpmVersion(String npmVersion) {
        this.npmVersion = npmVersion;
    }

    public String getNpmRepository() {
        return npmRepository;
    }

    public void setNpmRepository(String npmRepository) {
        this.npmRepository = npmRepository;
    }

    public void setOpaqueToken() {
        this.injectionTokenImport = "OpaqueToken";
        this.injectionToken = "OpaqueToken";
    }
}

Best Regards

wing328:
@topce please file a PR with the suggested change so that we can more easily review.

topce:
@wing328  Done 
PR:
https://github.com/swagger-api/swagger-codegen/pull/5874

austin94:
@wing328 It looks the new PR is based against `master`, just wanted to confirm that these should be going against the upcoming 2.3.0 release?

Either way, I added the suggested changes to this PR so this should meet @sebastianhaas requirements.

sebastianhaas:
@austin94 thanks, LGTM now!

wing328:
@sebastianhaas thanks for reviewing the change.

@austin94 thanks for the quick turnaround to incorporate the suggested changes by @topce 

I merged the change locally and ran `./bin/typescript-angular2-petstore-all.sh` to update all the Petstore Angular2 samples. Then in the `samples/client/petstore/typescript-angular2/npm` folder, I got the following errors running `npm run build`:
```
variables.ts(1,10): error TS2305: Module '"/private/tmp/swagger-codegen/samples/client/petstore/typescript-angular2/npm/node_modules/@angular/core/index"' has no exported member 'InjectionToken'.
```
I don't think it's caused by this PR but still would appreciate if you guys can take a look.

sebastianhaas:
@austin94 you need to increase the Angular version in `package.mustache` from 2.x to 4.x in order to make sure `InjectionToken` is available. 

@wing328 I wonder how this could have worked before though. It actually can't, not with the `InjectionToken` being imported.

austin94:
Is 2.3.0 targeting Angular 4.x specifically? If so it's trivial for me to bump the version.

Since that's the default target version for the @angular/cli it kind of makes sense.  We still have the `--useOpaqueToken` flag for those that need to degrade the experience.

sebastianhaas:
> Is 2.3.0 targeting Angular 4.x specifically?

I actually don't know about that, but since it's a wip future release branch, I'm pretty confident we should target the latest Angular family (4.x).

> We still have the `--useOpaqueToken` flag for those that need to degrade the experience.

Yes, but those people would have to manually edit their `package.json` as well, which would then get overridden every time they re-generate their API client. But anyhow, I guess that is the best solution for now and for this PR that fixes a rather annoying problem.

wing328:
> Is 2.3.0 targeting Angular 4.x specifically?

I don't think so.

There's a related discussion about support Angular 4.x in https://github.com/swagger-api/swagger-codegen/issues/5570

> I wonder how this could have worked before though. It actually can't, not with the InjectionToken being imported.

Right that's why I said the issue is not related to this change.

There's another TS generator targeting Angular 1.x: `typescript-angular`. One option is to create `typescript-angular 4` for Angular 4.x but personally I would avoid introduce another typescript generator with a different library/framework if typescript-angular2 can cover both 2.x and 4.x with additional option (e.g. library)

sebastianhaas:
> There's another TS generator targeting Angular 1.x: typescript-angular. One option is to create typescript-angular 4 for Angular 4.x but personally I would avoid introduce another typescript generator with a different library/framework if typescript-angular2 can cover both 2.x and 4.x with additional option (e.g. library)

I totally agree. Angular 1.x (aka AngularJS) and Angular > 2.x in the same generator is probably too much to ask, there is not much overlap.

It would be nice to have a `--library` switch to select the target Angular version, that would also render `--useOpaqueToken` obsolete since there is no `InjectionToken` in Angular <4.x and `OpaqueToken` has been deprecated from Angular >=4.x on.

The library flag would determine the Angular version included in the `package.json`, and determine whether to use  `InjectionToken` or `OpaqueToken`. 

I'll see if I can provide a PR in the next day(s), but if anyone is able and willing to do it before me I'll be happy to review :) @austin94

austin94:
@sebastianhaas that sounds like a good plan.  With the creation of #5570, should I just leave this PR as is?

Also, as an outside perspective, I could see the maintaining of a `--library` flag as prohibitive in the future.  To match the _official_ angular story there is now just AngularJS and Angular.  The breaking changes from 2->4 and beyond are going to happen with more frequency in the future with the assumption that the vast majority of people are using the `@angular/cli` and always have a clear upgrade path to the newest version.

Just some random thoughts.

sebastianhaas:
Re this PR: I think it would be nice if @wing328 could merge this as-is, it doesn't break anything that isn't already broken, and fixes a rather serious problem (importing InjectionToken with typedef, which causes a compiler error).

It would be great if you could post your thoughts on supporting multiple Angular versions in #5570 as well. :)

wing328:
Merged into 2.3.0 as suggested

jamieathans:
PR #5882 that was closed in preference to this PR by @wing328 included a change to `typescript-angular2/configuration.mustache` which added nullable instance fields to the `class Configuration` to match `interface ConfigurationParameters`.

This PR doesn't seem to include this change and therefore compilation will break if using  `"strictNullChecks": true`.

Is this an oversight or was it explicitly left out of this PR?

sebastianhaas:
@jamieathans good catch, I don't think so, but it's probably good to have it in a separate PR anyways so it can be discussed and reviewed more easily.

topce:
@jamieathans  yes I fix it in separate commit
Agree with @sebastianhaas  new PR
Regarding naming @wing328 
it would make sense 
To call angular 1 angularjs
And angular 2+ just Angular I think it is google recommendation
Another improvement would be to add command line param
So user can specify previous version of angular 
For example 2 and code is generated to be compiled with this version
Without param compile agains latest and greatest

sebastianhaas:
@topce would you please share your thoughts on naming issues in https://github.com/swagger-api/swagger-codegen/issues/5570?

PR #5871, opened 2017-06-19
[JavaScript] Fixed *WithHttpInfo methods in 'usePromises' mode to be ES5 compatible

frol:
### PR checklist

### Description of the PR

This is a bug-fix PR for the change introduced in #4902. The syntax was not ES5 compatible, thus this change should be applied to the ES5 JS generator.


wing328:
@frol thanks for the fix. I think the CI (Travis) is using ES6+ so I'll create a separate task to use ES5 to catch the issue moving forward using other CI providers (e.g. Shippable)

Issue #5856, opened 2017-06-16
[QT5][CPP] Wrong support of PATCH operation


sabras75:
##### Description

There is a bug for the “PATCH” operation in the QT5CPP templates. In the template file

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/qt5cpp/HttpRequest.cpp.mustache#L286

There is this code:

	else if (input->http_method == "DELETE") {
	  manager->deleteResource(request);
	}
	else {
	  QBuffer buff(&request_content);
	  manager->sendCustomRequest(request, input->http_method.toLatin1(), &buff);
	}

Documentation of QT5 states:

QBuffer::QBuffer(QByteArray *byteArray, QObject *parent = Q_NULLPTR)

Constructs a QBuffer that uses the QByteArray pointed to by byteArray as its internal buffer, and with the given parent. The caller is responsible for ensuring that byteArray remains valid until the QBuffer is destroyed, or until setBuffer() is called to change the buffer. QBuffer doesn't take ownership of the QByteArray.

Since the variable “request_content” is allocated on the stack, this is clearly wrong and a bug. The construction of QBuffer is designed this way so that whenever you write to the buffer, it is also written to the byte array that it is pointing to.

##### Swagger-codegen version

2.2.3-SNAPSHOT

##### Suggest a Fix

Changing QBuffer to use a QByteArray solves the issue for me since there is no real use-case for using a QBuffer.

sabras75:
I'll provide a pull request shortly

PR #5875, opened 2017-06-19
Fix#5856 - Add support for PATCH

sabras75:
### Description of the PR
 Add support for PATCH verb. This was a lifecycle bug. 

Correction use newer QT5 api when available, falls back on old api (with corrected calling code) for backward compatibility.


sabras75:
It seems like the error reported by travis CI is not related to the code, but to the travis/github infrastructure 👍 

> To increase the limit use GitHub authentication.

wing328:
@sabras75 kudos for supporting both new and old version of QT 👍

wing328:
Closes #5856

PR #5879, opened 2017-06-19
[Java Play Framework] Fix to the default values

JFCote:
Fix to the default values. It was not using the right values and didn't take care of the "default" parameter.

### PR checklist


modules/swagger-codegen/src/main/resources/JavaPlayFramework/paramDefaultValue.mustache
```diff
@@ -1 +1 @@
-{{#isBoolean}}false{{/isBoolean}}{{#isInteger}}0{{/isInteger}}{{#isDouble}}0.0{{/isDouble}}{{#isLong}}0L{{/isLong}}{{#isFloat}}0.0{{/isFloat}}{{#isString}}""{{/isString}}
\ No newline at end of file
+{{#defaultValue}}{{#isBoolean}}{{{defaultValue}}}{{/isBoolean}}{{#isInteger}}{{{defaultValue}}}{{/isInteger}}{{#isDouble}}{{{defaultValue}}}{{/isDouble}}{{#isLong}}{{{defaultValue}}}{{/isLong}}{{#isFloat}}{{{defaultValue}}}{{/isFloat}}{{#isString}}"{{{defaultValue}}}"{{/isString}}{{/defaultValue}}{{^defaultValue}}null{{/defaultValue}}
```
ePaul: Do you really need the case distinction here when it is the same each time?

JFCote: For string, it is between `"`.
Maybe I could only do a distinction between "string" and non string.
I'll try to update my code in the next couple of days :)
Thanks for the review

JFCote:
Here is the fixed version.

Issue #5895, opened 2017-06-21
[Java][Spring][Server] When java8 and interfaceOnly are true, generated interface default methods should return HTTP 501 Not Implemented


gabrielbauman:
##### Description

Currently when java8 and interfaceOnly are true and we're generating a java/spring server, swagger-codegen produces java8 default implementations for each API method. These implementations return 200 OK.

    default ResponseEntity<Void> doSomething() {
        // do some magic!
        return new ResponseEntity<Void>(HttpStatus.OK);
    }

It would be useful if the default implementation returned 501 Not Implemented instead. 

    default ResponseEntity<Void> doSomething() {
        // do some magic!
        return new ResponseEntity<Void>(HttpStatus.NOT_IMPLEMENTED);
    }

This would be a nice default behaviour that would make sense to clients. If I forget to implement an interface method on a concrete controller class and it returns OK, it can cause some head scratching.

##### Swagger-codegen version

2.2.2 (maven plugin)

wing328:
Closed via #5897 (merged into master)

PR #5897, opened 2017-06-21
[Java][Spring] Make generated java8 interface default implementations return HTTP 501 Not Implemented

gabrielbauman:
When using JDK8 and generating default interface method implementations, return HTTP 501 Not Implemented rather than 200 OK by default. (#5895)

### PR checklist

### Description of the PR

This simple PR makes generated spring/java 8 interface default methods return HTTP 501 Not Implemented rather than 200 OK. This would be a better default behavior, particularly in the case where swagger-codegen is run on each build cycle using the interfaceOnly option and generated files are never modified. Concrete controller implementations could then implement the generated interface methods, and it would be easier to tell when the concrete controller was missing an implementation from the client side.

I have targeted the PR against 2.3.0 since it's a behavior change, but this also applies cleanly against master.


wing328:
@gabrielbauman thanks for the PR.

I found that it contains 10 commits that are not related to this fix. Can you try creating a new branch/PR by cherry-picking 171151f91edaba65c6a34f9e926a8dc316450e9c ?

wing328:
cc @cbornet @jfiala

wing328:
@gabrielbauman do you need help cherry-picking 171151f in a new branch?

wing328:
Merged into master via 8f9569168192728fb630552cfe6ba8e3a74b9b17

PR #5899, opened 2017-06-22
Use id -g for determining the gid instead of -u

fbe:
Fix wrong group id lookup in run-in-docker.sh

### PR checklist

### Description of the PR

Tiny change, the run-in-docker.sh uses id -u for determining the group id, change that to id -g and everything is fine if the user has a gid != uid.

Tested on my local machine.


wing328:
@fbe thanks for the fix. I also tested locally and the result is good.

PR #5901, opened 2017-06-22
[Python] Increase range of valid status codes

### LINKED ISSUE #5105: [Python] rest.py throws exception for non-error HTTP status responses
##### Description

rest.py throws an exception if the response HTTP status code is not 200 or 206.  There are a lot of other valid HTTP status codes that are not errors.  HTTP 102 Processing, HTTP 201 Created, HTTP 202 Accepted, HTTP 204 No Content....   

https://github.com/swagger-api/swagger-codegen/blob/7de1eeb987b9ae04460f78fe86f32a1bd4f01be7/modules/swagger-codegen/src/main/resources/python/rest.mustache#L194-L195

##### Steps to reproduce

Call any API that returns HTTP status 201, 202, 204, 102....

Swagger generated Python sdk will throw an APIException

##### Suggest a Fix

Shouldn't the code in rest.py not throw for any status code in the range 100..299?

sbko:
swagger-client shouldn't throw ApiException if status code in range 200-299

closes #5105

### DIFF
--- a/modules/swagger-codegen/src/main/resources/python/rest.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/rest.mustache
@@ -191,7 +191,7 @@ class RESTClientObject(object):
             # log response body
             logger.debug("response body: %s", r.data)
 
-        if r.status not in range(200, 206):
+        if not 200 <= r.status <= 299:
             raise ApiException(http_resp=r)
 
         return r
--- a/samples/client/petstore/python/petstore_api/rest.py
+++ b/samples/client/petstore/python/petstore_api/rest.py
@@ -200,7 +200,7 @@ class RESTClientObject(object):
             # log response body
             logger.debug("response body: %s", r.data)
 
-        if r.status not in range(200, 206):
+        if not 200 <= r.status <= 299:
             raise ApiException(http_resp=r)
 
         return r


sbko:
@wing328 do I need to add to commit all changes that were made by ./bin/python-petstore.sh? what about untracked folder `samples/client/petstore/python/.swagger-codegen/`?

sbko:
CI failed with unrelated error: `[ERR!] GitHub API: GitHub rate limit reached.`

wing328:
@sbko I'll run the test locally later. Thanks for your contribution.

modules/swagger-codegen/src/main/resources/python/rest.mustache
```diff
@@ -191,7 +191,7 @@ class RESTClientObject(object):
             # log response body
             logger.debug("response body: %s", r.data)
 
-        if r.status not in range(200, 206):
+        if r.status not in range(200, 300):
```
wing328: @sbko should it be 299 instead of 300?

@frol

sbko:
@wing328 python range is exclusive so range(200, 300) will become [200, 201, ..., 209, 299]. you know what, I will change it to check if greater or less

frol:
@sbko Yeah, `200 <= r.status <= 299` is much easier to reason about than range objects.

frol:
@wing328 LGTM

wing328:
@sbko PR merged into master. Thanks for your contribution

@frol thanks for reviewing the change

Issue #5888, opened 2017-06-20
[JavaScript] master branch now generates ES6 JS code instead of ES5


frol:
##### Description

@dinukadesilva Adding `useES6` configuration option with a default value `true` in #5593 introduced a *breaking change* because the behavior for `generate -l javascript` with no options now generates ES6 JavaScript code instead of ES5.

By the way, the `./bin/javascript-es6*` generators set `useEs6=true` while the option name is `useES6` (`S` is capital).

##### Swagger-codegen version

Master branch

##### Related issues

#5593 

##### Suggest a Fix

Change the default value to `false`.

dinukadesilva:
Thanks for this @frol 
I'll look into this

wing328:
I'll work on this tomorrow if no one has time to pick this up.

wing328:
Filed #5907 to fix it.

PR #5907, opened 2017-06-23
[JavaScript] fix JS shell scripts, add log to show template version

wing328:
### PR checklist

### Description of the PR

- fix JS shell scripts
- add log to show template version
- remove log in default codegen

Closes https://github.com/swagger-api/swagger-codegen/issues/5888


frol:
@wing328 `javascript-promise-es6` output is ES5, but it should be ES6 since the `-es6` suffix suggest that.

wing328:
@frol yup, it should. I'm still troubleshooting that.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
```diff
@@ -268,6 +268,8 @@ public void processOpts() {
         }
         if (additionalProperties.containsKey(USE_ES6)) {
             setUseES6(convertPropertyToBooleanAndWriteBack(USE_ES6));
+        } else {
+            setUseES6(false);
```
frol: Can the `useES6` initialization value be set to `false` to avoid this explicit call? It just seems to be easier to reason about the code if there is some default which changes only when the config option is present.

frol:
P.S. It would be nice to squash the PR to drop the unnecessary double-conversion ES6 -> ES5 -> ES6.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
wing328: The reason why I called it there is to show the line/log which templates to use, e.g.
```
[main] INFO io.swagger.codegen.languages.JavascriptClientCodegen - Using JS ES6 templates
```
So that users do not need to guess which one is being used and much easier for users to catch any issue moving forward in case the wrong version is used.

wing328:
> P.S. It would be nice to squash the PR to drop the unnecessary double-conversion ES6 -> ES5 -> ES6.

Yup, all PRs merged into master/2.3.0 are squashed.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
frol: I meant to change the

```java
protected boolean useES6; // default is ES5
```

to

```java
protected boolean useES6 = false; // default is ES5
```

wing328: Done.

frol: In this case, is there still a need for `setUseES6(false)`?

frol:
Other than the last comment, everything looks good to me!

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
wing328: Yes, to log the line saying it's using JS ES5 templates

frol: OK then.

cliffano:
@wing328 @frol I'm a bit confused. Is `javascript` lang still meant to generate ES5 by default?
I was testing master and it generated ES6.

wing328:
@cliffano For 2.2.3 (stable version we just released), it should default to ES5

For current master (2.3.0 with breaking changes), we change the default to ES6 so as to push the adoption of ES6 JS client.

cliffano:
@wing328 Thanks, I found some references in other issues stating that master would be ES5, I understand now that master at that time was slated for 2.2.3 which is out now. So master has been ES6 post-release.

Another question. `npm test` used to be the way to run unit tests against the generated ES5 code, which no longer works for ES6. What is the exact command that's expected to run unit tests for ES6?

As a side note, I've started my effort in listing the commands that's used by all languages (if applicable for the language) to do the followings:
- resolving dependencies
- running unit tests
- building a package artifact
- installing the package locally
- generating documentation
- publishing the package to a public repository (e.g. pypi, rubygems, maven central, etc)
Here's the config file: https://github.com/cliffano/swaggy-c/blob/master/conf/tasks.yml

wing328:
> Another question. npm test used to be the way to run unit tests against the generated ES5 code, which no longer works for ES6. What is the exact command that's expected to run unit tests for ES6?

I think `npm test` should be the right command but there may be issues with ES6 client at the moment that @CodeNinjai is trying to fix.

wing328:
> As a side note, I've started my effort in listing the commands that's used by all languages (if applicable for the language) to do the followings:

Nice effort 👍 

Please kindly create a new ticket (issue) to track it so that the community can help out and give feedbacks.

PR #5911, opened 2017-06-23
[typescript-angular2] Enumerating response types and codes from api

damienpontifex:
### PR checklist

### Description of the PR

The swagger spec can define multiple responses. So this allows two things:
1. The return type to reflect the multiple possible return types
2. The return value from the map to map to a correct type (json(), text(), undefined etc)

This also has the advantage of enabling further type checking for those enabling typescript strict null checks.

One final thing is utilising typescript string interpolation with `${variable}` instead of string replacement. Simplifies the construction of the URL path.

_n.b. I did update the sample angular2 out of the RC dependencies. Assuming this is fine as 1. it's an RC 2. even with the updates, still targeting minor version of 2.x so should be no breaking changes with the sample_

### DIFF
--- a/modules/swagger-codegen/src/main/resources/typescript-angular2/api.service.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular2/api.service.mustache
@@ -85,24 +85,21 @@ export class {{classname}} {
      {{/summary}}
      {{#allParams}}* @param {{paramName}} {{description}}
      {{/allParams}}*/
-{{^isResponseFile}}
-    public {{nickname}}({{#allParams}}{{paramName}}{{^required}}?{{/required}}: {{{dataType}}}, {{/allParams}}extraHttpRequestParams?: any): Observable<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}{}{{/returnType}}> {
-{{/isResponseFile}}
-{{#isResponseFile}}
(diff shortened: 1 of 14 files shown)


modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache
```diff
@@ -50,13 +50,17 @@ export class {{classname}} {
      * {{notes}}
      {{#allParams}}* @param {{paramName}} {{description}}
      {{/allParams}}*/
-    public {{nickname}}({{#allParams}}{{paramName}}{{^required}}?{{/required}}: {{{dataType}}}, {{/allParams}}extraHttpRequestParams?: any): Observable<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}{}{{/returnType}}> {
+    public {{nickname}}({{#allParams}}{{paramName}}{{^required}}?{{/required}}: {{{dataType}}}, {{/allParams}}extraHttpRequestParams?: any): Observable<undefined | {{#responses}}{{#dataType}}{{{dataType}}} | {{/dataType}}{{/responses}}{}> {
```
wing328: @damienpontifex seems to me the method signature has changed and it should be filed against 2.3.0 instead.

Also there's another PR (merged) that updated the method signature in a similar way: https://github.com/swagger-api/swagger-codegen/pull/5972. Please pull the latest 2.3.0 to take a look.

wing328:
@damienpontifex when you've time, I wonder if you can rebase the change on the latest master to resolve the merge conflicts. Thanks.

cc @TiFu @taxpon @sebastianhaas  @kenisteward

damienpontifex:
I’ve been thinking on whether this is the best way to handle multiple response types. In my usage, it’s more likely to have one success type, but possibly multiple types of errors. Just reconsidering if I can move these error types to throw or cause an error callback rather than complicating the main path of success with multiple types.

Appreciate any feedback, but will think and tinker on it.

PR #1752, opened 2015-12-22
Csharp Enums

justusth:
I implemented enums for C#, I had some problems with the merge but it looks good now.

I also added so default values are sent, we are sending data to C modules and they require all the fields of the api to be set. We have a pretty large system stack with a transaction system that runs C on VMS, C# modules, nodejs servers to route messages etc and hopefully you can agree that this change is reasonable.

PR #5917, opened 2017-06-23
fix npe :Unboxing of '(count < numVars) ? true : null' produce npe

Sodala:
Unboxing the boolean will throw java.lang.NullPointerException if the condition is false

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CSharpClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CSharpClientCodegen.java
@@ -548,7 +548,7 @@ public class CSharpClientCodegen extends AbstractCSharpCodegen {
                 int count = 0, numVars = codegenProperties.size();
                 for(CodegenProperty codegenProperty : codegenProperties) {
                     count += 1;
-                    codegenProperty.hasMore = (count < numVars) ? true : null;
+                    codegenProperty.hasMore = count < numVars;
                 }
                 codegenModel.vars = codegenProperties;
             }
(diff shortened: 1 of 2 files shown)


ePaul:
Do you have an example where the NPE happened? (e.g. a API definition + command line used to generate from it)?

Sodala:
If the null was intentional then we need to use Boolean an not boolean.  I have an example but it is a bit big so I will try to keep the essentials. I come back to you as soon as it is done

ePaul:
**Some history exploration:**

It looks like this line (and the surrounding code with `reconcileInlineEnums`) was added in f4773d3333c263 (2015-11-20, which ultimately seems to belong to #1752 from @justusth, merged in 2016-03-03).
At that time CodegenProperty.hasMore (like many more properties in this class) was a `Boolean` and using `null` was the way of doing it, it seems. #4326 (from 2016-12-07, from @sreeshas, merged 2016-12-09) changed that (and updated the placed it was set from null to false) – but it forgot updating CSharpClientCodegen.
Obviously there also were no tests where this line was covered (and there still are none).

**Suggestion**
It would be nice if we had a unit test which covers this whole reconcileInlineEnums method of CSharpClientCodegen. That one would obviously fail at the point where your pull request fixes it.
Alternatively, or additionally, one of the samples should have a case where this happens.

Sodala:
I added a test that covers the method in question. In fact I copied a test that already existed for java and I adjusted it for csharp.

Note: I had small problems with the case where in csharp is case sensitive when adding models in the map of definitions and not in java. I do not know if it is normal to have a difference between the two tests except  the choice of the Codegen implementation

wing328:
> Note: I had small problems with the case where in csharp is case sensitive when adding models in the map of definitions and not in java. I do not know if it is normal to have a difference between the two tests except the choice of the Codegen implementation

When you've time, please open a new issue with a sample spec to track it.

wing328:
@Sodala thanks for the fix and the test cases.

PR #5920, opened 2017-06-24
[PHP] Add support for Generation Gap pattern

### LINKED ISSUE #5784: [PHP] Add support for Generation Gap pattern
##### Description

Generation Gap pattern is useful for extending codes which generated.

###### Generation Gap pattern

Generation Gap martinfowler.com
https://martinfowler.com/dslCatalog/generationGap.html
> Generation Gap is about keeping the generated and handwritten parts separate by putting them in different classes linked by inheritance.

###### Sample

```php
class HandwrittenPet extends Pet
{
    // just a sample.
    // it has room for improvement.
    public function canNotBuy()
    {
        return $this->status === ‘pending’ || $this->status === ‘sold’;
    }
}

$api = new PetApi();

// I'd like to add `setReturnType()` to specify the ReturnType.
$api->setReturnType(‘\Path\To\HanwrittenPet’)’

$pet = $api->getPetById(1);
var_dump($pet);
// class \Path\To\HanwrittenPet

var_dump($pet->canNotBuy());
// bool
```

##### Swagger-codegen version

2.3.0+

--

How you do you feel about it?

ackintosh:
closes #5784 

### PR checklist

##### Sample

```php
class MyPet extends Pet{}

$api = new \Swagger\Client\Api\PetApi();
$api->setReturnType('\MyPet');

$api->getPetById(1);
// MyPet

$api->findPetsByStatus('sold');
// MyPet[]

```


modules/swagger-codegen/src/main/resources/php/api.mustache
```diff
@@ -71,6 +76,40 @@ use {{invokerPackage}}\ObjectSerializer;
...
+        }
+
+        return $this->specifiedReturnType = $returnType;
+    }
+
+    /**
+     * @params string $returnType
+     * @return string
+     */
+    protected function hydrationType($returnType)
```
ackintosh: I worried about this method name 🤔 
Please tell me if there is a good idea 📝

ackintosh:
@baartosz @dkarlovi cc @wing328 Please review 🙏 ✨

wing328:
@ackintosh thanks for the PR. Quick question if you don't mind. Do I need to set the return type every time creating a new `Api` object or there's a default (e.g. Pet) as defined in the spec?
```
$api->setReturnType('\MyPet');
```
and what if `\Swagger\Client\Api\PetApi()` also returns other models (e.g. User, Order) (which may not be a good design but technically the API owner can do that)?

ackintosh:
By default, `Api` returns the model which defined in the spec.

```php
$api = new \Swagger\Client\Api\PetApi();

// default
$api->getPetById(1);// Pet

$api->setReturnType('\MyPet');
$api->getPetById(1);// MyPet

// Revert to the default
$api->setReturnType(null);
$api->getPetById(1);// Pet
```

It is effected only when inheriting the default return value.
https://github.com/swagger-api/swagger-codegen/pull/5920/commits/c53977f4c005ee83a7b1f74af5aaba39f60eed31#diff-005d8600260d9990694047063ef4020aR101

```php
$api = new \Swagger\Client\Api\PetApi();
$api->getPetById(1);// Pet
$api->getUser(); // User

$api->setReturnType('\MyPet');

$api->getPetById(1);// MyPet
$api->getUser(); // User
```

baartosz:
looks like it would be actually simpler to add custom wrapper around generated model.

```
$pet = new MyCustomPet($api->getPetById(1));
```
MyCustomPet would add some new behaviours while proxying calls all other method calls to model class.

dkarlovi:
I like @baartosz solution here, it's very clean. The part 

> proxying all other method calls to model class

will probably include some magic method setup, but we could ship a trait too?

If having this done by the API, we should probably alias on model level, not API level.

ackintosh:
Thanks! I think @baartosz solution is better. 💡 
My PR complicates the behavior of API 💦

PR #5922, opened 2017-06-24
[csharp] Support inheritance instead of duplicating parent properties in derived classes

manuc66:
### PR checklist

### Description of the PR
Took the work done in PR #4080: "[csharp] Don't duplicate parent properties in derived classes".

Fixed the generated code for:
- validation with the problem of calling explicit implementation on the base class.
- overriding method `ToJson()` instead of hiding
- re-use the parent GetHashCode
- defining the default value of discriminator property in child type to the `typeName`

Fixed some ".bat":
- `csharp-dotnet2-petstore.bat was missing`
- `csharp-property-changed-petstore.bat` did not contains the right arguments vs the `.sh` version

Included an leveraged a helper library in order to de-serialize a child class with `Newtonsoft.Json`, the original library is hosted here: https://github.com/manuc66/JsonSubTypes (this feature is not available in  `Newtonsoft.Json`, see: https://github.com/JamesNK/Newtonsoft.Json/issues/1331


manuc66:
Seems the Travis error is not related to this changeset, could the build be re-triggered ?

samples/client/petstore-security-test/csharp/SwaggerClient/src/IO.Swagger/Client/JsonSubTypes.cs
```diff
@@ -0,0 +1,281 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace JsonSubTypes
+{
+    //  Copyright 2017 Emmanuel Counasse
```
wing328: @manuc66 Can we remove the copyright note and license here? The auto-generated code in any languages should be unlicensed by default.

wing328:
@manuc66 don't worry about the Travis error. I'll restart it and test locally if it's still failing. (C# related tests are covered by AppVeyor)

samples/client/petstore-security-test/csharp/SwaggerClient/src/IO.Swagger/Client/JsonSubTypes.cs
manuc66: I chose to include the small library as source code instead of an independently licensed package dependency that would increase constraints both on the consumer and producer side. I thought it would not be a problem as I saw that copyright notices exist in other generated files. 

However, I think the copyright notice could be removed as long a little notice explain that it came from 'https://github.com/manuc66/JsonSubTypes'.

What do you think about it?

wing328: @manuc66 What about putting down something similar to https://github.com/swagger-api/swagger-codegen/blob/64df75d8b6bcf3c2320e822671a0b1ed7d2efa1c/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L2931 ?

wing328:
cc @jimschubert

samples/client/petstore-security-test/csharp/SwaggerClient/src/IO.Swagger/Client/JsonSubTypes.cs
manuc66: OK, it's done ;-)

manuc66:
Conflicts solved

wing328:
@manuc66 I performed a few more tests and the result is good. Thanks for the contribution.

PR #5934, opened 2017-06-27
Enable snapshot push

fehguy:


.travis.yml
```diff
@@ -99,3 +99,9 @@ script:
 
 env:
   - DOCKER_GENERATOR_IMAGE_NAME=swaggerapi/swagger-generator DOCKER_CODEGEN_CLI_IMAGE_NAME=swaggerapi/swagger-codegen-cli
+
+after_success:
+  - if [ $SONATYPE_USERNAME ] && [ -z $TRAVIS_TAG ] && [ $TRAVIS_PULL_REQUEST == false ]; then
```
wing328: @fehguy instead of using `$TRAVIS_PULL_REQUEST == false`, what about using the following to ensure changes to the master will be deployed to maven repo?

"$TRAVIS_BRANCH" = "master"

I just did a test and found that branches in swagger-codegen (mainly created by you and me) will also trigger the `mvn clean deploy` command

I've filed https://github.com/swagger-api/swagger-codegen/pull/5946 with the fix.

PR #5944, opened 2017-06-28
Docs and examples need model name to be passed as baseType

yonpols:
### PR checklist

### Description of the PR

The change made by PRhttps://github.com/swagger-api/swagger-codegen/pull/5715 introduced an error in the name of models (both files and docs) where names such as `V1TestModel` would be document as `v1TestModel`. 

**Note:** several of the changes seen on the generated `petstore` files are not introduced by this change. Running the script before applying the changes on the PR generate the same changes.


wing328:
@yonpols thanks for the PR, which looks reasonable to me although I don't see any Petstore sample update (python, php, ruby) when testing locally.

Issue #5753, opened 2017-05-31
Unable to generate anything with multiple files yaml spec


JFCote:
##### Description

I'm trying to generate the JavaPlayFramework with a very simple multiple files yaml spec but it doesn't work. I also tried with other server and I receive the same error. I was advised to try to validate the spec using the brand new "validate" function but I still receive the same error.

What is strange is that I am able to look at this spec using swagger-editor in local. There is not error at all. Just put both file in the swagger-editor folder and see for yourself.

There are 2 files (find the content of them below):

- swagger.yaml
- definitions.yaml

Here is the error I get every time

```
[main] INFO io.swagger.parser.Swagger20Parser - reading from swagger-editor/swagger.yaml
Exception in thread "main" java.lang.StringIndexOutOfBoundsException: String index out of range: -7
        at java.lang.String.substring(Unknown Source)
        at io.swagger.models.refs.GenericRef.computeSimpleRef(GenericRef.java:86)
        at io.swagger.models.refs.GenericRef.<init>(GenericRef.java:29)
        at io.swagger.models.properties.RefProperty.set$ref(RefProperty.java:56)
        at io.swagger.models.properties.RefProperty.<init>(RefProperty.java:18)
        at io.swagger.util.PropertyDeserializer.propertyFromNode(PropertyDeserializer.java:214)
        at io.swagger.util.PropertyDeserializer.propertyFromNode(PropertyDeserializer.java:286)
        at io.swagger.util.PropertyDeserializer.deserialize(PropertyDeserializer.java:137)
        at io.swagger.util.PropertyDeserializer.deserialize(PropertyDeserializer.java:39)
        at com.fasterxml.jackson.databind.ObjectMapper._convert(ObjectMapper.java:3584)
        at com.fasterxml.jackson.databind.ObjectMapper.convertValue(ObjectMapper.java:3508)
        at io.swagger.parser.util.SwaggerDeserializer.property(SwaggerDeserializer.java:974)
        at io.swagger.parser.util.SwaggerDeserializer.definition(SwaggerDeserializer.java:789)
        at io.swagger.parser.ResolverCache.loadRef(ResolverCache.java:135)
        at io.swagger.parser.processors.ExternalRefProcessor.processRefToExternalDefinition(ExternalRefProcessor.java:34)
        at io.swagger.parser.processors.ExternalRefProcessor.processRefProperty(ExternalRefProcessor.java:119)
[...]

PR #5951, opened 2017-06-29
[CPP] Fixing some more ordering of checks

mmosemmle:
### PR checklist

### Description of the PR

This issue is similar to 8237e49, in which some type checking has in the wrong order. It was picked up by
https://lgtm.com/projects/g/swagger-api/swagger-codegen/alerts/
as 'Contradictory Type checks' (at the very top). In this case, these are not really causing bugs as the generated code compiles anyway (won't be the case for java). Note that it requires changes to the petstore samples as it now generates the correct default values for long (with an 'L' in the end). 

Also note that the problem in #5753 would have been picked up if Pull request integration: 
https://lgtm.com/projects/g/swagger-api/swagger-codegen/ci/
is enabled.


wing328:
@mmosemmle the fix looks good to me.

cc @stkrwork @Danielku15 @sebymiano

Issue #5949, opened 2017-06-29
Various variable name escape issues


m-y-mo:
##### Description

Using the code search engine in, lgtm.com various toVarName or toParamName methods does not seem to have properly escaped special characters:

https://lgtm.com/query/2023430003/project:12070014/lang:java/

The above shows methods toVarName and toParamName that do not call sanitizeName. Not all of these are problematic, as special characters may have been escaped in another way, but some of them needs fixing. In particular, in the case of c++, this can lead to uncompilable code.
##### Swagger-codegen version
master branch: cbd7301

##### Swagger declaration file content or url
e.g. For cpp.
##### Steps to reproduce

```
java -jar swagger-codegen/modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -l cpprest -i cpp.yaml -o ./test
```
##### Suggest a Fix

I will create a PR for the c++ case, but perhaps others can take a look at the other languages to see if it is ok. Once a method is found to be correct, I can update the query to exclude it (or others can take a look at the docs https://lgtm.com/docs/ql/primer/ql-primer). Once everything is either fixed or excluded, we can close the issue.
Note: The link posted to the query result is cached for the commit cbd7301 and will require updating from time to time. To do so, copy the content of the query console on the page, enter swagger-api/swagger-codegen in the search box, then click 'Query this project' on the right, paste back the content to the console and click run once it is compiled.

PR #5952, opened 2017-06-29
[CPP] Fix cpp escape issue part of #5949

mmosemmle:
### PR checklist

### Description of the PR

This change escapes special characters in the property names in the three c++ generators, while preserving the petstore samples. Part of #5949

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCppCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCppCodegen.java
@@ -7,36 +7,36 @@ import io.swagger.models.properties.Property;
 
 abstract public class AbstractCppCodegen extends DefaultCodegen implements CodegenConfig {
 
-	@Override
-	public String toVarName(String name) {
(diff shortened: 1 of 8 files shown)


wing328:
@mmosemmle thanks for the PR. Please use 4-space instead of tab in Java files. Otherwise, the CI (travis) tests will fail.

stkrwork:
With these changes, does everything compile and run (especially for cpprest, as a full client gets generated)?

mmosemmle:
Hi,

Sorry about the lack of context in the PR. The specific case that this is
trying to fix is when variable names are not properly escaped, e.g.
input-name under properties, which will result in method names like
getInput-name() etc. in the corresponding class in model (Paper.h,
Paper.cpp in this case) This then gives uncompilable code. I have checked
that after the changes, the names of these methods are correct (i.e.
becomes getInputName() etc.) If you don't mind taking a look and see
whether the code do indeed compile and run properly when used as a cpprest
api, that would be great. Thank you very much for your help.

Best Regards,

Man Yue Mo

stkrwork:
looks fine to me.

wing328:
@mmosemmle thanks for refactor common operations into an abstract class "AbstractCppCodegen"

Issue #3870, opened 2016-09-26
Default Java client does not support Content-Type "application/xml"


cacti77:
I have a swagger.yaml with a `post` method that produces `application/json` and this for consumes:

```yaml
    consumes:
        - application/xml;charset=UTF-8
```
I generated the Java client from the swagger.yaml using swagger-codegen-cli-2.2.1.jar and the `--lang java` option. I did not specify a `--library` option. (I doubt it's relevant but, just in case, I also specified these options in an associated json config file: apiPackage, groupId, artifactId and artifactVersion.)

However when I issue a POST request with an XML string as input try to call the `post` megenerated code I get this error in the client:
```
com.mycompany.myapi.ApiException: Content type "application/xml;charset=UTF-8" is not supported
        at com.mycompany.myapi.ApiClient.serialize(ApiClient.java:889)
        at com.mycompany.myapi.ApiClient.buildCall(ApiClient.java:1105)
        at com.mycompany.myapi.client.DefaultApi.modelsPostCall(DefaultApi.java:523)
        at com.mycompany.myapi.client.DefaultApi.modelsPostWithHttpInfo(DefaultApi.java:544)
```
The generated code looks like this:
```java
    /**
     * Serialize the given Java object into request body according to the object's
     * class and the request Content-Type.
     *
     * @param obj The Java object
     * @param contentType The request Content-Type
     * @return The serialized request body
     * @throws ApiException If fail to serialize the given object
     */
    public RequestBody serialize(Object obj, String contentType) throws ApiException {
        if (obj instanceof byte[]) {
            // Binary (byte array) body parameter support.
            return RequestBody.create(MediaType.parse(contentType), (byte[]) obj);
        } else if (obj instanceof File) {
            // File body parameter support.
            return RequestBody.create(MediaType.parse(contentType), (File) obj);
        } else if (isJsonMime(contentType)) {
            String content;
            if (obj != null) {
                content = json.serialize(obj);
            } else {
                content = null;
            }
            return RequestBody.create(MediaType.parse(contentType), content);
        } else {
            throw new ApiException("Content type \"" + contentType + "\" is not supported");
        }
    }
```
In my case `obj` is a `String`. It looks like you don't support `application\xml`.  Is this true? If so, how soon can we expect a fix please? Alternatively can you recommend a workaround I can use? I don't mind hacking the generated code to make it work until a proper fix is released.

##### Related issues

[...]

PR #5956, opened 2017-06-29
withXml with Jackson for Java clients

mobreza:
# Description of the PR

I needed `application/xml` support in a Java client. Thought the best way was to add `withXml` flag and then generate the Java clients accordingly. All Java clients need work, but the **resttemplate** library does its job as is.

`withXml` is respected as one of the `configOptions` for Java clients in your maven plugin:

```xml
	<configOptions>
		<withXml>true</withXml>
		...
```

I've updated the `CodegenModel` and `CodegenProperty` to respect Swagger's **xml** props.


mobreza:
This is in relation to #3870.

PR #5994, opened 2017-07-05
Fix Erlang auth method code generation

wagerlabs:
Broken code is generated when no authentication methods are supplied
in the API schema. Fix this by providing code to fall back to.


wing328:
@wagerlabs thanks for the fix, which looks good to me.

Issue #6002, opened 2017-07-06
Create a swift4 code generator and templates


ehyche:
##### Description

Swift 4 (announced at WWDC 2017 in June 2017) introduces some features which make serialization and deserialization in Swift much simpler:

* The Codable protocol, which is actually two protocols (Decodable and Encodable)
* Because of protocol extensions, for most cases, all the model classes have to do is simply declare their conformance to the Codable protocol.

I have written an Xcode playground which explores lots of details about the Codable protocol here:

https://github.com/ehyche/xcode-playgrounds

in the LearningAboutCodable.playground file.

So I have created a "swift4" language for swagger-codegen which greatly simplifies the mustache templates, and thus the generated models. Now, most models simply have the properties and that's it. For example, here is the Pet class:

```
open class Pet: Codable {

    public enum Status: String, Codable { 
        case available = "available"
        case pending = "pending"
        case sold = "sold"
    }
    public var id: Int64?
    public var category: Category?
    public var name: String?
    public var photoUrls: [String]?
    public var tags: [Tag]?
    /** pet status in the store */
    public var status: Status?

    public init() {}

}
```

##### Swagger-codegen version

I am proposing adding this to master. I have created the swift4 language on a branch, and will be creating a PR for it.

##### Command line used for generation

```
./bin/swift4-petstore-all.sh
```

##### Steps to reproduce

New language.

ehyche:
I have this new "swift4" language ready to go on a branch off of master. But I guess I need to get the proper permissions before I can push this branch up.

ehyche:
Note that you need an Xcode 9 beta in order to use the LearningAboutCodable.playground and the swift4 language.

wing328:
@ehyche thanks for offering help to add the Swift4 generator 👏 👏 👏 

Let us know if you need help with the PR.

cc @jaz-ah @Edubits @jaz-ah

ehyche:
@wing328 : ah, I just re-read the instructions for contributing. I need to first fork the repo.

PR #6010, opened 2017-07-07
Add a swift4 client generator

ehyche:
### PR checklist

### Description of the PR

This PR implements the issue referenced here: https://github.com/swagger-api/swagger-codegen/issues/6002

This adds a new "swift4" generator. Swift 4 added the "Codable" protocol, which is just a combination of the "Encodable" and "Decodable" protocols. These protocols greatly simplify serialization to JSON and deserialization from JSON in Swift. For example, the Pet model object now just looks like:

```
open class Pet: Codable {

    public enum Status: String, Codable { 
        case available = "available"
        case pending = "pending"
        case sold = "sold"
    }
    public var id: Int64?
    public var category: Category?
    public var name: String?
    public var photoUrls: [String]?
    public var tags: [Tag]?
    /** pet status in the store */
    public var status: Status?

    public init() {}

}

open class Category: Codable {

    public var id: Int64?
    public var name: String?

    public init() {}

}

open class Tag: Codable {

    public var id: Int64?
    public var name: String?

    public init() {}

}
```

As compared to the "swift" and "swift3" generators, there is now no need for all of the generated Decoders in Models.mustache, and there is now no need for the JSONEncodable protocol encodeToJSON() method that previously each model object needed to implement. So this greatly simplified the model objects and reduced the size of the generated code.

NOTE: As of this writing, Swift 4 is only present in the Xcode 9 betas. The swift4 generator samples will not compile in Xcode 8, as it does not have support for Codable.

With this PR, I have:

- Created a new Swift4Codegen generator which is based on the existing Swift3Codegen generator
- Created new templates for swift4 which were based on existing swift3 templates but simplified significantly.
- Included swift4 samples which are based on the swift3 samples with some slight changes
- Made sure that all of the current swift4 unit tests pass

Future work:

- Add more unit tests to test edge cases like:
  - Missing required parameters
  - Invalid date and data in the JSON
  - Invalid JSON


jaz-ah:
@ehyche this is super cool - i'll have time a little later today to take a look - thx!

ehyche:
Thanks, @jaz-ah ! If you decide to pull down the branch and try it out running the unit tests yourself in Xcode, keep in mind you'll need to do that with the Xcode 9 beta.

jaz-ah:
understood - i'm using Xcode9 Beta2 right now - is that ok?

modules/swagger-codegen/src/main/resources/swift4/Podspec.mustache
```diff
@@ -0,0 +1,21 @@
...
+  s.license = {{#podLicense}}{{& podLicense}}{{/podLicense}}{{^podLicense}}'Proprietary'{{/podLicense}}{{#podHomepage}}
+  s.homepage = '{{podHomepage}}'{{/podHomepage}}{{#podSummary}}
+  s.summary = '{{podSummary}}'{{/podSummary}}{{#podDescription}}
+  s.description = '{{podDescription}}'{{/podDescription}}{{#podScreenshots}}
+  s.screenshots = {{& podScreenshots}}{{/podScreenshots}}{{#podDocumentationURL}}
+  s.documentation_url = '{{podDocumentationURL}}'{{/podDocumentationURL}}
+  s.source_files = '{{projectName}}/Classes/Swaggers/**/*.swift'{{#usePromiseKit}}
+  s.dependency 'PromiseKit', '~> 4.2.2'{{/usePromiseKit}}{{#useRxSwift}}
+  s.dependency 'RxSwift', '~> 3.4.1'{{/useRxSwift}}
+  s.dependency 'Alamofire', '~> 4.0'
```
jaz-ah: I would make this 4.5 since that is when Alamofire fixed Xcode9 related issues.

ehyche:
@jaz-ah : Yes, Xcode 9 beta 2 works - that is what I used to develop this.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Swift4Codegen.java
```diff
@@ -0,0 +1,640 @@
...
+                        "AnyObject",
+                        "Any")
+        );
+        reservedWords = new HashSet<>(
+                Arrays.asList(
+                    // name used by swift client
+                    "ErrorResponse", "Response",
+
+                    // swift keywords
+                    "Int", "Int32", "Int64", "Int64", "Float", "Double", "Bool", "Void", "String", "Character", "AnyObject", "Any", "Error", "URL",
```
jaz-ah: let's add Codable as a reserved word here

jaz-ah:
@ehyche just a couple minor comments if you want to add those I think we're good to merge!

ehyche:
@jaz-ah : thanks for the review! I'm updating and re-generating the samples now. Will push up changes shortly.

ehyche:
@jaz-ah : I have made updates per your review comments and pushed changes up.

jaz-ah:
great - looks like we're good to go @wing328 ... @ehyche we'll have to keep an eye on bug fixes in swift3 to make sure they make their way into swift4 for a while, it's the only disadvantage to the extra generator.

wing328:

@ehyche thanks for the contribution 👍 

I'll try to setup CI to cover the Swift4 Petstore clients later and let you guys know if I need any help.

cc @jgavris @Edubits

wing328:
FYI. Added @ehyche as the template creator of `swift4` in the project README and sent out the following tweet to help promote the new generator:

https://twitter.com/wing328/status/883748380718977024

(please help retweet)

jgavris:
This is great work, thanks. Let's make sure to keep cherry-picking the patches to the Swift 3 generator up to Swift 4 as we make them.

amelnychuck:
Hi all,

First off let me just say great work so far @ehyche — glad to see someone jumping on this after the WWDC announcement.

A couple of quick comments:
- It looks like there are still quite a few references to the JSONEncodable previous way of handling things that I imagine can be removed or replaced now. (Please correct me if I'm wrong.)
- After running one of my APIs through the swift4 generator, I'm receiving a lot of errors similar to the following for most all of my API methods:
screen shot 2017-07-11 at 1 42 11 am
screen shot 2017-07-11 at 1 41 29 am
screen shot 2017-07-11 at 1 43 21 am

A lot of these seem to be related to the fact that while the Codable protocol is referenced, it's not conformed to properly, and both Encodable and Decodable must be conformed to correctly. This is mostly a problem when you have custom objects in your Model rather than basic types like strings and bools.

Anyways, great work again and I look forward to seeing the generator further improved/fixed before the GM hits!

ehyche:
@amelnychuck : Thanks! Yes, there is only one place that JSONEncodable is still used, and it is on my ToDo list to remove that last reference. Thanks for the reminder.

Regarding the errors:

1. There is no problem referencing non-basic types in your models, like this:

```
open class PriorityList: Codable {
...
public var selectedCommand: RelayCommand?
}
```
as long as RelayCommand also conforms to Codable. Is RelayCommand also a generated model object? If so, it should also conform to Codable.

2. In the request builder, the fact that I'm seeing RequestBuilder<Any> means that something unexpected happened in the generation. Since getBuilder<T:Codable> always assumes that the type parameter conforms to Codable. So I would never expect T to be Any. This probably indicates a corner case I haven't yet considered.

Is your schema publicly accessible? If so, I'd be glad to take a look at it, and find and fix the issues in the current swift4 generator.

amelnychuck:
@ehyche
Regarding point 1. Yes exactly, RelayCommand does not conform to Codable, but it is generated.

RelayCommand.swift (in the Models folder of the swift 4 output) contains the following:
```
//
// RelayCommand.swift
//
// Generated by swagger-codegen
// https://github.com/swagger-api/swagger-codegen
//

import Foundation


public typealias RelayCommand = Any
```

Looks like the generator only makes it a type alias of Any. (Possibly this is an issue with my swagger specification?) I'm guessing this is also related to point 2.

Unfortunately my schema isn't publicly accessible (I can see if I can create a sample that reproduces the issue though), but I'll keep looking into it and if I can give you any more information to help diagnose the issue with the generation, let me know what I can do to help.

ehyche:
@amelnychuck : Looks like RelayCommand was generated by this part of model.mustache:

```
import Foundation

{{#description}}

/** {{description}} */{{/description}}
{{#isArrayModel}}
public typealias {{classname}} = [{{arrayModelType}}]
{{/isArrayModel}}
```
So I would expect {{arrayModelType}} to be something other than "Any" - something like another generated model, or a primitive Swift type.

amelnychuck:
Still looking into the issue, but I agree.

On further examination, it looks like removing all references to RelayCommand in the swift3 version of the SDK solved that issue and I am able to build a test project containing the swift3 sdk and Alamofire successfully. Looks like that's something I'll need to solve as to why the generator is making that type Any...

That being said, when I remove all the references to RelayCommand in my Swift4SDK and sample project (at which point swift3 worked correctly), errors related to that seem to go away, but I still get the list of errors for every instance of `let requestBuilder: RequestBuilder<Any>.Type = SwaggerClientAPI.requestBuilderFactory.getBuilder()`
in the project.

screen shot 2017-07-11 at 10 26 29 am

Streel88:
I've got a question, I need to populate the CodingKeys enum in swift4 classes, to map response keys to variables. I can see it in the [mustache](https://github.com/swagger-api/swagger-codegen/blob/2be2ee080be58e19591d0e8aa6bd8e1d80938fcc/modules/swagger-codegen/src/main/resources/swift4/model.mustache) template (line 75), but I can't find documentation on how to define my yaml to generate it.

wing328:
https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml contains various edge cases (enum models, enum properties, etc)

You may want to use it to generate the Swift 4 client to see how different mustache tags are populated.

Streel88:
Found my issue, i need to use the 2.3.0 Generator not the 2.2.3 generator

Issue #6011, opened 2017-07-07
[SWIFT 3] Invalid redeclaration of 'encodeToJSON()'


julien-h-webcenter:
Hi,

I have a problem with a swagger 2.3.0 to generate swift 3 code. i have an error because the protocol encodeToJSON is redefined two times.
 In Extesions.swift : 
extension Data: JSONEncodable {
    func encodeToJSON() -> Any {
        return self.base64EncodedString(options: Data.Base64EncodingOptions())
    }
}

In Data.swift
open func encodeToJSON() -> Any {
        var nillableDictionary = [String:Any?]()
        nillableDictionary["attribut"] = self.attribut
        nillableDictionary["valeur"] = self.valeur

        let dictionary: [String:Any] = APIHelper.rejectNil(nillableDictionary) ?? [:]
        return dictionary
    }

I don't understand why ?

Thanks.

Swagger-codegen version : 2.3.0

Command line used for generation:
java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -i http://myurl/api.yaml -l swift3 -o /Users/Documents/Swagger

jaz-ah:
@julien-h-webcenter what error are you getting - what compiler are you using?

julien-h-webcenter:
Hi,

thanks for your answer.
Error : 
<img width="532" alt="capture d ecran 2017-07-10 a 09 56 55" src="https://user-images.githubusercontent.com/17987431/28008119-73bdda4a-6556-11e7-8c49-4ec732bec917.png">

i'm using xcode 8.3.3 for compiler

jaz-ah:
ah I see what's happening @julien-h-webcenter  - you have an object in your swagger named "Data" so a Data.swift is created, and when Extensions.swift says:
extension Data: JSONEncodable {
the compiler thinks it's your Data instead of the Foundation library's Data.   so a couple temporary solutions for now:
1) rename the model object inside your Data.swift class to _Data (I know it's generated but that would be a workaround)
2) rename Data inside your Swagger to be something else if possible

the best thing of course would be to add "Data" to the list of reserved Swift keywords in Swift3Codegen.java so this gets fixed automatically... would you have time to contribute that change or would you like me to add?

julien-h-webcenter:
Thanks.

Yes i can to try to add the Keywords data on the Branch 2.3.0.

It's my first PR ;)

wing328:
[...]

PR #6021, opened 2017-07-11
fix add "Data" to reserved Keywords (#6011)

julien-h-webcenter:
### Description of the PR

Add "Data" to Reserved Keywords


wing328:
Applying the same enhancement to Swift 2.x, 4.x generators via 4a8d9c3

Issue #4857, opened 2017-02-26
Deserialization failed due to change of property name


wing328:
##### Description

The issue is originally reported via https://github.com/swagger-api/swagger-codegen/issues/4805#issuecomment-280615622

For typescript-angular2:

```
"Request": {
                "device_type": {
                    "type": "string"
                },
            }
        },
```

generates:

```
export interface Request { 
    deviceType?: string;
}
```

The deserialization does not work as there's no mapping between `deviceType` and `device_type` (original attribute name in JSON)

##### Swagger-codegen version

latest master and 2.3.0

cc @Ondra09, @ataraxus, @Tifu

wing328:
@Ondra09, @ataraxus may I know if you've tried the `typescript-node` generator (not typescript-angular2) in 2.3.0 branch to see if the deserialization works? I'd a look at the code and I think it should work.

ataraxus:
I tried till now only the typescript-angular2 generator. If it helps, I could also test the node one.

ard61:
@wing328 I've encountered this same issue (deserialization failed because no property name mapping in typescript-angular2). 

So, I implemented property name mappings for typescript-angular2 in my fork of [swagger-codegen:2.3.0](https://github.com/swagger-api/swagger-codegen/compare/2.3.0...ard61:fix_issue4857), reusing @TiFu 's solution from PR #4264 . I am ready to submit a pull request with this fix, if you want. 

I have also come across PR #4267, which discussed the long-term goal of implementing property name mappings in a common way across all Typescript code generators. Is this still being implemented (which would make my fix unnecessary)?

Thanks!

wing328:
@ard61 please submit a PR as I don't think anyone has started working on a solution for Typescript-angular2 regarding this problem.

PR #4264, opened 2016-11-26
Typescript Property Renaming Fix

TiFu:
First: First time contributor, so feel free to yell at me :-).

### PR checklist

- [x] Read the [contribution guildelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)
- [x] Filed the PR against the correct branch: master for non-breaking changes and `2.3.0` branch for breaking (non-backward compatible) changes.

Second Point: Is it enough if i recreated the samples for typescript-node (e. g. I don't have to recreate typescript-angular samples)? I only changed one file in swagger-codegen/src/main/resources/typescript-node. 

### Description of the PR

Reference: #2766
This also fixes #2776

The Typescript-Node-Client did not parse the JSON returned from requests, so that properties, which were renamed (e. g. from mounted_at in the response to mountedAt in the model), still appeared as the non-renamed version (e. g. mounted_at) in the model object.

This PR fixes this issue by adding property maps and parsing the JSON recursively.

NOTE:
I'm not sure if this change breaks backwards compatibility, so I filed the PR against 2.3.0.

I hope that I did not miss any cases, I need to handle. It would be great if someone with more experience with swagger-codegen, than I have, could take a look at the code.

PR #6024, opened 2017-07-11
Fix issue4857

ard61:
### PR checklist

### Description of the PR

Fixes issue #4857 Deserialization failed due to change of property name for the typescript-angular2 client, re-using the solution in PR #4264:
* adds a property name map to each model, turning them from an interface into a class
* adds global objects tracking created classes and enums
* adds a class, ObjectSerializer, responsible for translating property names (and serializing/deserializing dates), and makes the generated api use this class at serialization/deserialization. 

Breaks backward compatibility I think, because we change the models from interfaces to classes, which will cause issues for users reusing the models in their code and assigning object literals (because of the added constructor / properties).
If that is undesirable, I'm happy to refactor the property name maps into a separate entity, leaving the models as interfaces.


wing328:

wing328:
cc @Vrolijkx @taxpon @sebastianhaas @kenisteward @TiFu

ard61:
Thanks for pointing that out! Somehow I hadn't added the e-mail address I use for commits to my GitHub account...

samples/client/petstore/typescript-angular2/default/api/pet.service.ts
```diff
@@ -117,7 +118,7 @@ export class PetService {
                 if (response.status === 204) {
                     return undefined;
                 } else {
-                    return response.json() || {};
+                    return ObjectSerializer.deserialize(response.json(), "Array&lt;Pet&gt;") || {};
```
grEvenX: Shouldn't this be `"Array<Pet>"` instead of `"Array&lt;Pet&gt;"`?

sebastianhaas:
I will take at look at this on weekend.

samples/client/petstore/typescript-angular2/default/api/pet.service.ts
TiFu: This seems to be related to {{returnType}} in line 99 in api.service.mustache.This is also an issue for the ts-node code generator. 
Is the returnType encoded by the program generating the input for mustache?

wing328: PLease use `{{{returnType}}}` instead to avoid HTML encoding performed by mustache.

TiFu: I'll submit a PR to fix the issue in ts-node.

modules/swagger-codegen/src/main/resources/typescript-angular2/api.service.mustache
```diff
@@ -254,7 +255,7 @@ export class {{classname}} {
             method: {{httpMethod}},
             headers: headers,
 {{#bodyParam}}
-            body: {{paramName}} == null ? '' : JSON.stringify({{paramName}}), // https://github.com/angular/angular/issues/10612
+            body: {{paramName}} == null ? '' : JSON.stringify(ObjectSerializer.serialize({{paramName}}, "{{dataType}}")), // https://github.com/angular/angular/issues/10612
```
TiFu: This causes the same issue `{{returnType}}` caused. => `{{{dataType}}}` to prevent html encoding

samples/client/petstore/typescript-angular2/default/api/pet.service.ts
ard61: Thanks! Fixed now.

modules/swagger-codegen/src/main/resources/typescript-angular2/types.mustache
```diff
@@ -0,0 +1,142 @@
...
+            return new Date(data);
+        } else {
+            if (enumsMap[type]) {// is Enum
+                return data;
+            }
+
+            if (!typeMap[type]) { // dont know the type
+                return data;
+            }
+            let instance = new typeMap[type]();
```
ard61: We're returning an instance of the modal class. 
Before this fix, we were returning a plain object.
They are type-compatible, but some testing frameworks e.g. Jasmine, do more extended checks for equality:
```
class Foo {
  prop: string
}
describe('Jasmine type checking', () => {
  it('fails', () => {
    foo = new Foo();
    foo.prop = 'Hello';
    expect(foo).toBeEqual({'Hello'});  // Error: 'Expected foo to be a type of Object, but is Foo(prop: "Hello")'
});
```
Which means that some users (me included) would be having to rewrite unit-tests.

ard61:
I'm also having second thoughts about adding the property name maps directly in the model interfaces, because (if I understand this correctly) they are exposed for the users to reuse in their own code. 

Since the property name maps are there just to enable the ObjectSerializer to function correctly, it seems to me that it would make sense to separate them from the models proper.

Any thoughts?

ard61:
Here are the changes I would do. I'll revert the commits if you think we should leave the property name maps as they currently are, in the model class.

TiFu:
IMO the whole serializer should be hidden from the user. It is an implementation detail of the API module and the user should not rely on this functionality. This enables us to change it without breaking the code of users. The user should only interact with the model objects and the api objects.

kenisteward:
@tifu I second this

iamsamwood:
Is this PR still being developed?

PR #6025, opened 2017-07-11
[Java Play Framework] Fix collection utils

JFCote:
### PR checklist

### Description of the PR

Major bug found while using the generator for query parameters that were collection. The util classe was not doing the good thing at all (I copied this from another generator but did not understand that the classe was doing the opposite of what I wanted).

I also applied the fix to form and header params.

Tested in production in my code and it works well. 

@wing328 : I don't understand why the swagger.json file in the changes is modified. I didn't change anything. In fact, I never add any code that generate the examples so I don't know why they are there in the first place! If you understand the reason, please let me know. :)


wing328:
>  I don't understand why the swagger.json file in the changes is modified. I didn't change anything. In fact, I never add any code that generate the examples so I don't know why they are there in the first place! If you understand the reason, please let me know. :)

I've no idea off the top of my head. Will take a look later after merge.

Issue #6029, opened 2017-07-12
[Python]Bug in api_client.py __call_api


zorrofox:
##### Description

In api_client __call_api function [this line](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/python/api_client.mustache#L93) use a new init Configuration:
```python
config = Configuration()
```
and I think we should use the parameter configuration which has be assigned to the attribute  like this:
```python
config = self.configuration
```

wing328:
Closed via #6033

PR #6033, opened 2017-07-12
[Python] fixed bug in api_client.py `__call_api` method

zorrofox:
### PR checklist

### Description of the PR
fix the issue #6029


wing328:
@zorrofox thanks for the PR. The fix looks good to me but I'll defer to the Python experts to have a look.

cc @frol @taxpon

wing328:
cc @baartosz as well who performs the refactoring work.

frol:
It seems to me that the target PR branch should be 2.3 instead of master.

frol:
I think you can simply drop the `config` assignment and simply use `self.configuration.` directly. BTW, there is only a single place where it is used: https://github.com/swagger-api/swagger-codegen/blob/e1e5ac4d379234ac8d4525deb8fa7be983cc1e8c/modules/swagger-codegen/src/main/resources/python/api_client.mustache#L111

wing328:
> It seems to me that the target PR branch should be 2.3 instead of master.

We released 2.2.3 (previous master) over the weekend and master is now 2.3.0.

samples/client/petstore/python/.swagger-codegen/VERSION
```diff
@@ -1 +1 @@
-2.2.3-SNAPSHOT
\ No newline at end of file
+2.3.0-SNAPSHOT
```
frol: > We released 2.2.3 (previous master) over the weekend and master is now 2.3.0.

If this is the case, why does this diff and many other unrelated changes appear?

Anyway, the fix by itself is correct, so if the other matters are fine, it is good to merge this in.

wing328: @frol thanks for reviewing. Very good questions. I'll update the samples after the merge (some changes are due to changes in Swagger Parser)

PR #6040, opened 2017-07-12
[Java] Escape descriptions in javadocs

lwander:
The [java8 doclint](http://openjdk.java.net/jeps/172) rejects unescaped
HTML chars such as `<`, making some generated clients unbuildable with
java8. Seems a few property descriptions were using the `{{{` instead
of `{{` preventing those HTML chars from being escaped properly.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/modelEnum.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/modelEnum.mustache
@@ -11,7 +11,7 @@ import com.google.gson.stream.JsonWriter;
 {{/gson}}
 
 /**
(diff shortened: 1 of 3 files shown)


wing328:
I found that `modules/swagger-codegen/src/main/resources/Java/modelInnerEnum.mustache` needs to be fixed as well.

I've filed https://github.com/swagger-api/swagger-codegen/pull/6061 to fix it as well as C# client, server (e.g. Nancyfx) and Java server (e.g. JAX-RS)

lwander:
Awesome, thanks @wing328

wing328:
@lwander np. https://github.com/swagger-api/swagger-codegen/pull/6061 has been merged into master. Please pull the latest to get those fixes.

lwander:
Great, I appreciate it

PR #6042, opened 2017-07-13
[Typescript-Fetch] new typescript-fetch template base on typescript-angular2

isman-usoh:
### PR checklist

- [x] Ran the shell/batch script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/typescript-fetch-petstore.sh` and `./bin/security/typescript-fetch-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates)

### Description of the PR

new typescript-fetch template base on typescript-angular2


TiFu:
Does this change take care of mapping the property names? See #4857 for a description of the problem.

MohamedHamouGisaia:
I use a REST api where the json response is configured to be `snake_case`. And I use swagger-codegen 2.3.1 to generate a typescript-fetch client  where the interfaces are `camelCase` (I set 
`modelPropertyNaming` to `camelCase`)

I want to keep this configuration but I don't know how to map the `snake_case` response with the `camelCase` attributes at deserialization time. Methods in the generated api return `response.json()`

Am I missing something in my `--input-spec` file  or not using the right version? 

Thank you very much

PR #6043, opened 2017-07-13
[Typescript-node] Fixed html escape bug in returnType and dataType

TiFu:
### PR checklist

### Description of the PR

Fixes the issue discovered in #6024 for typescript-node. This PR removes the HTML encoding for `returnType` and `dataType` when serializing and deserializing the data types.

I am not sure if removing the encoding might have security implications. The security tests tries to inject code via e.g. the api key, version or the host, but does not do the same for data types. Could someone please check if this is a possible issue?


wing328:
@TiFu thanks for the fix, which looks good to me.

Issue #3483, opened 2016-07-29
[Java] Definitions of Primitive Types are Generated as Objects with no accessors


tjboudreaux:
##### Description

In our definitions file, we have some definitions that are done for specifying constraints on primitives.  Below is an example of an `integer` definition that we use for a Phone Number.

When doing code generation for a Java client library, this definition gets generated as a model, but there is no way to get or set it's value.
##### Swagger-codegen version

2.1.6
##### Swagger declaration file content or url

```
PhoneNumber:
  type: integer
  minLength: 10
  maxLength: 10
  minimum: 0
```

This is the generated code for this definition: https://gist.github.com/tjboudreaux/4263e3329e2239557155e0f87f6a8b0c
##### Command line used for generation

swagger-codegen generate -i api/swagger/swagger.json -l java -o builds/android -c
##### Steps to reproduce
1. Add a definition to your swagger file that is a primitive type.
2. Generate a java client library.
##### Related issues

No related issues.
##### Suggest a Fix

I believe the issue is that if a definition is a primitive, it should either: 
1. generate from a model that has a property called value to store it's value 
2. or it needs to be treated as a primitive with no model class.

wing328:
@tjboudreaux thanks for reporting the issue. Is `PhoneNumber` defined as a property in a model?

wing328:
Btw, stable version 2.2.0 has been released. Please give it a try as it contains a lot of enhancements.

tjboudreaux:
@wing328, yes it's defined as a property on multiple models. Here's one example:

```
OrderRestaurant:
  type: object
  required:
    - id
    - name
    - address
    - phone_number
  properties:
    id:
      type: string
    name:
      type: string
    address:
      $ref: '#/definitions/Address'
    phone_number:
      $ref: '#/definitions/PhoneNumber'
```

We also tried the stable version and that did not solve this specific issue.

wing328:
Swagger parser does not support that at the moment: https://github.com/swagger-api/swagger-parser/issues/243

As a workaround, please define the primitive type inline instead of using $ref

abhijith-prabhakar:
Is any checkin done to fix this?  Looking to see if I can use master SNAPSHOT for this

abhijith-prabhakar:
A note here if and when fix is provided.  If you inline primitive type and build the contract, jackson serializes it differently than having an Object property.
Swagger Codegen currently generates a Java model class for primitive json(although no property is present).

[...]

PR #6050, opened 2017-07-13
Issue: #3483 - fixed objects defined as extension of primitives - brute force

joevalerio:
### PR checklist

### Description of the PR

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java
@@ -24,6 +24,7 @@ import io.swagger.codegen.CodegenParameter;
 import io.swagger.codegen.CodegenProperty;
 import io.swagger.codegen.DefaultCodegen;
 import io.swagger.models.Model;
+import io.swagger.models.ModelImpl;
 import io.swagger.models.Operation;
 import io.swagger.models.Path;
 import io.swagger.models.Swagger;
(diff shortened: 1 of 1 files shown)


joevalerio:
brute forced the change only in java after codegenModel is created, I was not sure if the base model should be updated, as that affects all other languages.  But atleast this gives you an idea into what needs to change.

joevalerio:
Why would changes to the abstractJavaCodeGen affect C-Sharp Code?

wing328:
@joevalerio please ignore that or rebase on the latest master, which has the issue fixed.

wing328:

wing328:
@joevalerio can you please also elaborate a bit more on what issue you're trying to fix? Can you share a spec so that we can more easily reproduce the issue?

joevalerio:
Sorry, thought I linked this to issue [#3483](https://github.com/swagger-api/swagger-codegen/issues/3483). In a swagger I am processing a definition defined like this:

  LanguageCode:
    description: 'ISO 639-1 language code'
    type: string
    minLength: 2
    maxLength: 2
    example: 'fi'

For java this should create fields of type String with validation to only allows 2 characters.  I am balking on the validation with this PR, but the defect I reference refers to this creating an object with no accessor.  That part was fixed, however all the references to that object still exist.  This is an issue for definitions that extends any primitive type.  So I just change the codegen model to correctly generate the right code for all java codegen targets.  I was not sure how this effects other languages, so i did not want to do this work in the base swagger model, or the base translation to the codegen model.

FWIW, don't care about credit, just want it to work.

joevalerio:
This was fixed in 2.2.3...

joevalerio:
ok, been using 2.2.3, and now can get what I want.  If you use spring, the order of the definitions seem to matter for the original issue.  I am using java with library: feign to get what I want.

PR #6076, opened 2017-07-16
Bugfix: Generated typescript-angular code fails to compile when models.ts file is empty

reda-alaoui:
### Description of the PR
Generated typescript angular code fails to compile when `models.ts` file is empty (no model definition).
This is because an empty js or ts file is not considered as a valid ES6 module.
Importing it fails the compilation.

With this fix, empty `models.ts` is no more imported by api files.


wing328:
@reda-alaoui my local test result is good. Thanks for the contribution.

cc @Vrolijkx @taxpon @sebastianhaas @kenisteward @TiFu

wing328:
Note that `typescript-angular` has been renamed to `typescript-angularjs` in the latest master (2.3.0)

PR #6079, opened 2017-07-16
Added not null check in code generator.

Dschinghis-Kahn:
### Description of the PR
When generating client libs for the EVE Online ESI API a null pointer exception occured preventing the libs from being build.

### Reproduce the error
git clone https://github.com/swagger-api/swagger-codegen
cd swagger-codegen
mvn clean package
  -i https://esi.tech.ccp.is/latest/swagger.json \
  -l java \
  -o samples/client/eve-online-esi-api/java

### Fix
I traced the root cause of the null pointer exception and prevented it by a not-null-check when adding the object in question.

**Caution: I have absolutely no idea what the code were I implemented the fix does!!!**
Please check the fix for any side effects. I only tested the fix with the petstore and eve specs and both can build their client libs.


wing328:
@Dschinghis-Kahn I think doing a null check is reasonable. Thanks for contributing the fix.

Issue #5905, opened 2017-06-22
[New Generator] Rust API client generator


wing328:
##### Description

We want to generate Rust API client so that Rust programmers can consume REST APIs more easily with Rust.

cc @frol 

##### Swagger-codegen version

Latest master

##### Suggest a Fix

If anyone wants to contribute the generator or help out, please reply to let us know. 

(You don't need to know Java in order to contribute the generator)

farcaller:
> You don't need to know Java in order to contribute the generator

Why am I deep into java code now, then? 😄 

On a serious note: I'm working on rust client generator based on hyper.rs.

wing328:
The PowerShell API client generator was created with the help of an experienced PowerShell developer who does not know Java.

We can start the same process if you're interested creating a Rust client generator in Swagger Codegen.

farcaller:
I'd be up for any help in re. java code (I can reasonably read java). Can do the rust part.

wing328:
@farcaller thanks for offering help on this.

We can reuse the same [methodology](https://github.com/swagger-api/swagger-codegen/issues/4320#issuecomment-296198548) we used for PowerShell API client by first creating a Rust wrapper/SDK for [PetStore](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.yaml) by focusing on a few endpoints.
- [GET](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.yaml#L156-L183)
- [POST (form)](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.yaml#L184-L218)
- [POST (JSON body)](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.yaml#L31-L56)

The Rust code can be put in the folder `samples/client/petstore/rust`.

Then we can reverse-engineer the Rust code to create the mustache templates and the generator in Java.

farcaller:
SGTM. I'll work on the example and will submit a PR

farcaller:
There you go: https://github.com/swagger-api/swagger-codegen/pull/6092

It almost works with the sample node server (it actually crashes the server, but I'm very sure it's the server's fault).

BenjaminGill-Metaswitch:
[...]

PR #6092, opened 2017-07-17
[New Generator] Rust API client generator

farcaller:
### PR checklist

### Description of the PR

Tracking PR for Rust client implementation.

The first commit adds a subset of apis requested in #5905.

### DIFF
--- a/samples/client/petstore/rust/Cargo.toml
+++ b/samples/client/petstore/rust/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "petstore-client"
+version = "0.1.0"
+authors = ["Vladimir Pouzanov <farcaller@gmail.com>"]
+
+[dependencies]
+serde = "*"
+serde_derive = "*"
+serde_yaml = "*"
+serde_json = "*"
+base64 = "*"
+futures = "*"
+hyper = "*"
+url = "*"
+
+[dev-dependencies]
+tokio-core = "*"
--- a/samples/client/petstore/rust/examples/client.rs
+++ b/samples/client/petstore/rust/examples/client.rs
@@ -1,0 +1,44 @@
+extern crate petstore_client;
+extern crate hyper;
+extern crate tokio_core;
+extern crate futures;
+
+use hyper::Client;
+use hyper::client::HttpConnector;
+use tokio_core::reactor::Core;
+use futures::{Future};
+
+fn main() {
+    let mut core = Core::new().expect("failed to init core");
+    let handle = core.handle();
+
+    let http_connector = HttpConnector::new(4, &handle);
+
+    let client = Client::configure().connector(http_connector).build(&handle);
+    let api = "http://localhost:8080/v2";
+
+    let mut new_pet = petstore_client::models::Pet::new("barker".to_owned(), vec![]);
+    new_pet.set_id(1337);
+
+    let work = petstore_client::apis::add_pet(api, &client, &new_pet)
+    .and_then(|_| {
+        petstore_client::apis::update_pet_with_form(
+            api,
+            &client,
+            1337,
+            "barko",
+            "escaped")
+    })
+    .and_then(|_| {
+        petstore_client::apis::get_pet_by_id(
+            api,
+            &client,
+            1337)
+    })
+    .and_then(|pet| {
+        println!("pet: {:?}", pet);
+        futures::future::ok(())
+    });
+
+    core.run(work).expect("failed to run core");
+}
--- a/samples/client/petstore/rust/examples/client.rs
+++ b/samples/client/petstore/rust/examples/client.rs
@@ -15,12 +15,16 @@ fn main() {
     let http_connector = HttpConnector::new(4, &handle);
 
     let client = Client::configure().connector(http_connector).build(&handle);
-    let api = "http://localhost:8080/v2";
+    let api = "http://petstore.swagger.io:80/v2";
 
-    let mut new_pet = petstore_client::models::Pet::new("barker".to_owned(), vec![]);
-    new_pet.set_id(1337);
+    let new_pet = petstore_client::models::Pet::new("barker".to_owned(), vec![]).with_id(1337);
 
-    let work = petstore_client::apis::add_pet(api, &client, &new_pet)
+    let apicli = petstore_client::apis::client::APIClient::new(
+        petstore_client::apis::configuration::Configuration::new(
+            Client::configure().connector(HttpConnector::new(4, &handle)).build(&handle)));
+
+    let work = apicli.pet_api().add_pet(&new_pet)
+    // petstore_client::apis::add_pet(api, &client, &new_pet)
     .and_then(|_| {
         petstore_client::apis::update_pet_with_form(
             api,
--- a/samples/client/petstore/rust/src/apis/add_pet_api.rs
+++ b/samples/client/petstore/rust/src/apis/add_pet_api.rs
@@ -1,0 +1,27 @@
+use hyper;
+use serde_json;
+use futures::{Future, Stream};
+use futures;
+
+use super::models;
+use super::Error;
+
+pub fn add_pet<C: hyper::client::Connect>(
+    prefix: &str,
+    cli: &hyper::client::Client<C>,
+    pet: &models::Pet,
+    ) -> Box<Future<Item = (), Error = Error>> {
+    let mut req = hyper::Request::new(
+        hyper::Method::Post,
+        format!("{}/pet", prefix).parse().unwrap());
+    let serialized = serde_json::to_string(pet).unwrap();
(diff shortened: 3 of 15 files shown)


farcaller:
I'll add a few comments on the specifics inline

samples/client/petstore/rust/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "petstore-client"
```
farcaller: these settings would be exposed as flags.

samples/client/petstore/rust/examples/client.rs
```diff
@@ -0,0 +1,44 @@
+extern crate petstore_client;
```
farcaller: a sample file, not to be part of generated api

samples/client/petstore/rust/src/apis/add_pet_api.rs
```diff
@@ -0,0 +1,27 @@
+use hyper;
+use serde_json;
+use futures::{Future, Stream};
+use futures;
+
+use super::models;
+use super::Error;
+
+pub fn add_pet<C: hyper::client::Connect>(
```
farcaller: rust naming guidelines dictate `snake_case` in function names

samples/client/petstore/rust/src/apis/add_pet_api.rs
```diff
@@ -0,0 +1,27 @@
...
+    prefix: &str,
```
farcaller: I'm not sure about what's the common pattern for api use, really. This sample client will require two things: a http (or https) prefix as a string and the hyper.rs Client (the web client) plus any arguments.

samples/client/petstore/rust/src/apis/add_pet_api.rs
```diff
@@ -0,0 +1,27 @@
...
+    cli: &hyper::client::Client<C>,
+    pet: &models::Pet,
+    ) -> Box<Future<Item = (), Error = Error>> {
+    let mut req = hyper::Request::new(
+        hyper::Method::Post,
+        format!("{}/pet", prefix).parse().unwrap());
+    let serialized = serde_json::to_string(pet).unwrap();
```
farcaller: no error handling for brevity of the example.

samples/client/petstore/rust/src/apis/add_pet_api.rs
```diff
@@ -0,0 +1,27 @@
...
+    req.headers_mut().set(hyper::header::ContentType::json());
+    req.headers_mut().set(hyper::header::ContentLength(serialized.len() as u64));
+    req.set_body(serialized);
+
+    Box::new(
+        cli.request(req).and_then(|res| { res.body().concat2() })
+        .map_err(|e| Error::from(e))
+        .and_then(|_| futures::future::ok(()))
```
farcaller: the api doesn't dictate the return value so nothing is returned even though the sample server returns a pet instance.

npm server choked up on `{"id":1337,"category":null,"name":"barker","photoUrls":[],"tags":[],"status":null}` here, http://petstore.swagger.io worked fine.

samples/client/petstore/rust/src/apis/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+
+impl From<serde_json::Error> for Error {
+    fn from(e: serde_json::Error) -> Self {
+        return Error::Serde(e)
+    }
+}
+
+use super::models;
+
+mod get_pet_by_id_api;
```
farcaller: there's a single namespace, so modules (and files) would have to be named differently from the actual functions.

samples/client/petstore/rust/src/models/pet.rs
```diff
@@ -0,0 +1,29 @@
+/// a Pet
+///
+/// A pet for sale in the pet store
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Pet {
+    id: Option<i64>,
+    category: Option<super::Category>,
+    name: String,
+    #[serde(rename = "photoUrls")] photo_urls: Vec<String>,
```
farcaller: again, snake case here, and renamed fields need an annotation.

wing328:
@farcaller thanks for the Rust Petstore sample, I'll start the reverse engineering process tomorrow and let you know if I've any question.

farcaller:
I poked more around the go generated code and brought the example to the similar level of api. The changes, basically, are that there's an APIClient now, that incapsulates access to the API, and a dedicated pet_api.rs that's supposedly implements all the Pet apis. This extension fully builds on the previous example, though, so it would be fine if the generator produced the functional API as in original commit, and I can tackle the object API later myself when I'll have a better understanding of swagger internals.

wing328:
Quick question: would it be possible to group `add_pet_api.rs`, `get_pet_by_id_api.rs`, `update_pet_with_form_api.rs` into a single rs file (e.g. pet_api_func.rs)?

farcaller:
Absolutely. I'll update the PR

wing328:
@farcaller I've submitted the initial version via https://github.com/swagger-api/swagger-codegen/pull/6105

Please comment on the code in that PR and we'll fix the remaining issues one by one.

samples/client/petstore/rust/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[dependencies]
+serde = "*"
+serde_derive = "*"
+serde_yaml = "*"
+serde_json = "*"
+base64 = "*"
+futures = "*"
+hyper = "*"
+url = "*"
```
lucab: Would you mind annotating which dependency version is the codegen targeting?

samples/client/petstore/rust/src/apis/get_pet_by_id_api.rs
```diff
@@ -0,0 +1,23 @@
...
+use futures::{Future, Stream};
+
+use super::models;
+use super::Error;
+
+pub fn get_pet_by_id<C: hyper::client::Connect>(
+    prefix: &str,
+    cli: &hyper::client::Client<C>,
+    pet_id: i64,
+    ) -> Box<Future<Item = models::Pet, Error = Error>> {
```
lucab: (Here and it other generated methods)
Would you consider returning a more complex type? The Item here is losing any additional information about the response (header, status, etc) which I've found useful in some situations (e.g. token authenticated endpoints redirecting to token provider). For reference, reqwest does this with a richer `Response` [struct + trait](https://github.com/seanmonstar/reqwest/blob/8296c5e28e00a09ce45e0e33f8b10a8d15105a90/src/response.rs#L27).

samples/client/petstore/rust/Cargo.toml
farcaller: yep, as soon as I figure those myself :-) so far the generated code doesn't compile

samples/client/petstore/rust/src/apis/get_pet_by_id_api.rs
farcaller: ack.

wing328: For other generators (e.g. PHP, Ruby), there's a corresponding "WithHttpInfo" method (e.g. getPetByIdWithHttpInfo) to return a list with the response (if any), HTTP status code and HTTP headers.

Not sure if Rust supports returning a list. If not, we will need to create an ApiResponse object with response, status and headers as properties.

For Rust, we can follow similar approach by creating a corresponding `WithHttpInfo` method or we can default the method (e.g. getPetById) to return the response together with status code and headers.

farcaller: I think we can return response object along with the parsed body. Let's focus on making the current implementation compileable and then we can improve.

wing328: It's also worth sharing other API clients usually come with a corresponding "Async" method (e.g. [addPetAsync](https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/csharp/SwaggerClient/src/IO.Swagger/Api/PetApi.cs#L572)) for async function call. Not sure how that should be done in Rust but worth thinking ahead with this future requirement.

farcaller: All this code is async-only. I don't think that it makes sense to have a matching sync version.

wing328: Ah ok. Good to know. One less thing to consider then.

wing328:
Merging this (Rust Petstore sample) into master first similar to what we've done for PowerShell and other generators that follow similar approach so as to give credits back to the contributor.

For further discussion on the Rust generator, please continue in https://github.com/swagger-api/swagger-codegen/pull/6105, which is work-in-progress.

PR #6097, opened 2017-07-18
[Java Play-Framework] Remove unneeded deps

### LINKED ISSUE #6095: [Java-PlayFramework] Unnecessary dependencies are added to the project
##### Description

Generated project has playWs, playCache and playJdbc as dependencies, but they don't seem to be used anywhere.

##### Swagger-codegen version

2.2.3

edrevo:
### PR checklist

### Description of the PR

Fixes #6095

### DIFF
--- a/modules/swagger-codegen/src/main/resources/JavaPlayFramework/build.mustache
+++ b/modules/swagger-codegen/src/main/resources/JavaPlayFramework/build.mustache
@@ -6,5 +6,9 @@ lazy val root = (project in file(".")).enablePlugins(PlayJava)
 
 scalaVersion := "2.11.7"
 
-libraryDependencies += {{#useSwaggerUI}}"org.webjars" % "swagger-ui" % "2.2.10-1"{{/useSwaggerUI}}
-libraryDependencies += {{#useBeanValidation}}"javax.validation" % "validation-api" % "1.1.0.Final"{{/useBeanValidation}}
+{{#useSwaggerUI}}
+libraryDependencies += "org.webjars" % "swagger-ui" % "2.2.10-1"
+{{/useSwaggerUI}}
+{{#useBeanValidation}}
+libraryDependencies += "javax.validation" % "validation-api" % "1.1.0.Final"
+{{/useBeanValidation}}
--- a/modules/swagger-codegen/src/main/resources/JavaPlayFramework/build.mustache
+++ b/modules/swagger-codegen/src/main/resources/JavaPlayFramework/build.mustache
@@ -6,11 +6,5 @@ lazy val root = (project in file(".")).enablePlugins(PlayJava)
 
 scalaVersion := "2.11.7"
 
-libraryDependencies ++= Seq(
-javaJdbc,
-cache,
-javaWs{{#useSwaggerUI}},
-"org.webjars" % "swagger-ui" % "2.2.10-1"{{/useSwaggerUI}}{{#useBeanValidation}},
-"javax.validation" % "validation-api" % "1.1.0.Final"
-{{/useBeanValidation}}
-)
+libraryDependencies += {{#useSwaggerUI}}"org.webjars" % "swagger-ui" % "2.2.10-1"{{/useSwaggerUI}}
+libraryDependencies += {{#useBeanValidation}}"javax.validation" % "validation-api" % "1.1.0.Final"{{/useBeanValidation}}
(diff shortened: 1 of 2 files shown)


wing328:
@edrevo thanks for the PR, which looks good to me.

cc @JFCote

modules/swagger-codegen/src/main/resources/JavaPlayFramework/build.mustache
```diff
@@ -6,11 +6,5 @@ lazy val root = (project in file(".")).enablePlugins(PlayJava)
...
 
-libraryDependencies ++= Seq(
-javaJdbc,
-cache,
-javaWs{{#useSwaggerUI}},
-"org.webjars" % "swagger-ui" % "2.2.10-1"{{/useSwaggerUI}}{{#useBeanValidation}},
-"javax.validation" % "validation-api" % "1.1.0.Final"
-{{/useBeanValidation}}
-)
+libraryDependencies += {{#useSwaggerUI}}"org.webjars" % "swagger-ui" % "2.2.10-1"{{/useSwaggerUI}}
```
JFCote: This is not good. If you don't use swaggerUI or beanValidation, it will show `libraryDependencies += `
And it will not work. 
We should stick with the `Seq` notation which is good practice and give the ability to concatenate multiple dependencies in an array.

Finally, the other dependencies are there by default because they are part of the "default sample app" created by play framework which this generator is based on. I think that even if the generated code is not using them by default, most if not all users will use them. If they don't, they can always delete them.

samples/server/petstore/java-play-framework/build.sbt
```diff
@@ -6,10 +6,5 @@ lazy val root = (project in file(".")).enablePlugins(PlayJava)
...
 scalaVersion := "2.11.7"
 
-libraryDependencies ++= Seq(
-javaJdbc,
-cache,
-javaWs,
-"org.webjars" % "swagger-ui" % "2.2.10-1",
-"javax.validation" % "validation-api" % "1.1.0.Final"
-)
+libraryDependencies += "org.webjars" % "swagger-ui" % "2.2.10-1"
```
JFCote: Like I said in my previous comment, this will only work because the default value for `useSwaggerUI` and `useBeanValidation` is true. It will crash if the user change the default value.

modules/swagger-codegen/src/main/resources/JavaPlayFramework/build.mustache
edrevo: Gah, you are right. i'll fix the code.

With regard to SBT best practices and `+=` vs `++=`, I've never seen any recommendation on one vs the other. It is not a problem to change it, though. One SBT best practice this project isn't following is using a Dependencies.scala file to declare dependencies ([source](http://jsuereth.com/scala/2013/06/11/effective-sbt.html), [example](https://github.com/playframework/playframework/blob/master/framework/project/Dependencies.scala)). Just mentioning that in case you want me to change that too.

edrevo:
Wrt removing the deps, I do think generating lean projects is a good idea. Most users will not remove libs that were added by default and there is no obvious way of detecting whether they can be safely removed, other than removing one by one and see if the project still compiles. If they end up needing the libs they can always add them (which will be easy to detect, since the compiler will complain). Just my 2c.

@wing328, should I fix the PR or will it be closed?

JFCote:
@edrevo : After some reflection, I think you are right. Change your PR to this and I will approve it and ask @wing328 to merge! B**y the way, take notice that I have updated the webjars version of swagger-ui  in the example below to the latest. I've test it on my side and it works.**

```
{{#useSwaggerUI}}libraryDependencies += "org.webjars" % "swagger-ui" % "3.0.19"{{/useSwaggerUI}}
{{#useBeanValidation}}libraryDependencies += "javax.validation" % "validation-api" % "1.1.0.Final"{{/useBeanValidation}}
```

wing328:
> @wing328, should I fix the PR or will it be closed?

@edrevo please use the suggested fix by @JFCote

wing328:
Btw, you can do the following instead to avoid blank lines:
```
{{#useSwaggerUI}}
libraryDependencies += "org.webjars" % "swagger-ui" % "3.0.19"
{{/useSwaggerUI}}
```

edrevo:
Sorry for the mega-delay. I have updated the code with your comments. Thanks everyone for the feedback!

wing328:
Looks like the latest update has addressed the concerns from @JFCote

Issue #5886, opened 2017-06-20
[PowerShell] Enhancements to PowerShell generator


wing328:
##### Description

Here are some enhancements for the PowerShell generator:

- [ ] Option to specify `ClientPath` (which is hard coded to "$ScriptDir\..\..\csharp\SwaggerClient" at the moment"
- [ ] Integration tests using AppVeyor (similar to other API clients, we want to perform testings with the Petstore server (local) to ensure nothing breaks after changes)
- [ ] Review authentication support: HTTP basic, API key, OAuth2
- [ ] Add auto-generated documentation for API endpoints and models
- [ ] Update auto-genreated test files (APIs, models)

##### Swagger-codegen version

Latest master

##### Suggest a Fix

If anyone wants to work on these enhancements or have other suggestions on making the PowerShell generator better, please reply to let us know.

wing328:
Added C# client Path option via https://github.com/swagger-api/swagger-codegen/pull/5887

cc @beatcracker

markdepalma:
@wing328 I am looking into using the PowerShell generator. The part I'm not getting is how to set parameters for things like the authorization header and basePath at runtime here. I see how you would do this in the plain csharp and other versions, but I don't see any of the ApiClient exposed in the generated PS scripts. Outside of modifying the csharp code to hardcode these variables, how should you be doing this?

Issue #5400, opened 2017-04-15
[Python] Codegen for definition with no properties leads to missing models


mbohlool:
Similar to #4804 

Models with no property are valid according to [swagger 2.0](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject) spec:

```
...
Unlike previous versions of Swagger, Schema definitions can be used to describe primitive and arrays as well.
...
```
but code generator does not support them. My suggestion is to create classes that extends primitive types in this case. e.g., for something like this:

```
"MyString": {
"Description": "My string description",
"type": "string",
}
```

generator should generate something similar to:

```
...
class MyString(string):
   """My string description"""
   pass
...
```

Also if there is no type, we should assume it is an free range schema and class should extend "object".

wing328:
@mbohlool thanks for reporting the issue. When I clicked on https://github.com/kubernetes-incubator/client-python/blob/master/kubernetes/docs/V1Patch.md, I got a 404.

Is there a spec we can more easily reproduce the issue?

cc @scottrw93 @baartosz @frol @taxpon @scop

mbohlool:
@wing328 That is exactly the issue. We have a v1.Patch that is not resulted in generating a model.

https://github.com/kubernetes-incubator/client-python/blob/e69d4fe204b98f7d7ee3ada3996b4f5fbceae5fe/scripts/swagger.json

I've linked to a specific commit because I am implementing a workaround for now and this file will be changed soon. But in the commit I linked about, the file has the model but generated client does not.

mbohlool:
You can also look at `v1.Time` and `resource.Quantity` models that has the same problem in the above spec.

taxpon:
@wing328 Do you know how other language generators are working regarding this issue?

wing328:
@taxpon not that I'm aware of. It looks to me it's an issue with Swagger Parser skipping models with no properties.

@mbohlool I found the following:
```
parameters: [
{
  name: "body",
  in: "body",
  required: true,
  schema: {
    $ref: "#/definitions/v1.Patch"
  }
}
],
```
Can you provide an example of what the JSON body looks like?

A workaround is to define a dummy property (e.g. id)

mbohlool:
@wing328 That is a free json object. The server interpret it based on the context. e.g. if it is a Patch on `Pod`, then it would look for `Pod`'s fields to patch. This could be better designed but still a valid swagger.

wing328:
[...]

PR #6105, opened 2017-07-18
[Rust] add new Rust client generator

wing328:
### PR checklist

### Description of the PR

Add Rust client generator based on the Petstore samples provided by @farcaller (https://github.com/swagger-api/swagger-codegen/pull/6092)


farcaller:
In the generated code for `ApiResponse` model, `Type` field is renamed to `Type_` (should be underscorecased too)

farcaller:
>_(\ just removed all the comments accidentally.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RustClientCodegen.java
```diff
@@ -0,0 +1,515 @@
...
+        typeMapping.put("float", "f32");
+        typeMapping.put("double", "f64");
+        typeMapping.put("boolean", "bool");
+        typeMapping.put("string", "String");
+        typeMapping.put("UUID", "String");
+        typeMapping.put("date", "string");
+        typeMapping.put("DateTime", "String");
+        typeMapping.put("password", "String");
+        typeMapping.put("file", "File");
+        // map binary to string as a workaround
```
farcaller: Must be mapped to `Vec<u8>`, String can only hold a vaild utf-16 byte sequence.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RustClientCodegen.java
```diff
@@ -0,0 +1,515 @@
...
+        // the correct solution is to use []byte
+        typeMapping.put("binary", "String");
+        typeMapping.put("ByteArray", "String");
+        typeMapping.put("object", "Object");
+
+        importMapping = new HashMap<String, String>();
```
farcaller: Not needed in rust, we can import everything by absolute paths, also saves the reserved namespace from pollution.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RustClientCodegen.java
```diff
@@ -0,0 +1,515 @@
...
+        additionalProperties.put("modelDocPath", modelDocPath);
+
+        modelPackage = packageName;
+        apiPackage = packageName;
+
+        supportingFiles.add(new SupportingFile("README.mustache", "", "README.md"));
+        supportingFiles.add(new SupportingFile("git_push.sh.mustache", "", "git_push.sh"));
+        supportingFiles.add(new SupportingFile("gitignore.mustache", "", ".gitignore"));
+        supportingFiles.add(new SupportingFile("configuration.mustache", apiFolder, "configuration.rs"));
+        supportingFiles.add(new SupportingFile("api_client.mustache", apiFolder, "api_client.rs"));
```
farcaller: The equivalent is generated in `client.rs`

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RustClientCodegen.java
```diff
@@ -0,0 +1,515 @@
...
+        supportingFiles.add(new SupportingFile(".travis.yml", "", ".travis.yml"));
+
+        supportingFiles.add(new SupportingFile("client.mustache", apiFolder, "client.rs"));
+        supportingFiles.add(new SupportingFile("api_mod.mustache", apiFolder, "api.rs"));
```
farcaller: Should be `mod.rs`

modules/swagger-codegen/src/main/resources/rust/.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: go
```
farcaller: I think `language: rust` is all you need for rust testing here.

modules/swagger-codegen/src/main/resources/rust/Cargo.mustache
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "{{{pacakgeName}}}"
```
farcaller: typo, should be `packageName`

modules/swagger-codegen/src/main/resources/rust/api_mod.mustache
```diff
@@ -0,0 +1,44 @@
...
+use super::models;
+
+{{#apiInfo}}
+{{#apis}}
+{{#operations}}
+{{#operation}}
+{{#summary}}
+//{{{summary}}}
+{{/summary}}
+mod {{classFilename}}_func;
```
farcaller: This doesn't exist?

modules/swagger-codegen/src/main/resources/rust/api_mod.mustache
```diff
@@ -0,0 +1,44 @@
...
+pub use self::{{classFilename}}::{{{operationId}}};
+{{#-last}}
+
+mod {{classFilename}}_api;
```
farcaller: this generates references to e.g. `pet_api_api`, not `pet_api`

farcaller: Both `mod` entries should be outside of {{operations}} loop or the generated code gets multiple mod declarations.

modules/swagger-codegen/src/main/resources/rust/client.mustache
```diff
@@ -0,0 +1,55 @@
...
+{{/apis}}
+{{/apiInfo}}
+    }
+  }
+
+{{#apiInfo}}
+{{#apis}}
+{{#operations}}
+{{#operation}}
+  pub fn {{classFilename}}(&self) -> &{{classFilename}}::{{classname}}{
```
farcaller: This generates multiple identical entries, is it missing `{{#-last}}` / `{{/-last}}`?

modules/swagger-codegen/src/main/resources/rust/model.mustache
```diff
@@ -0,0 +1,57 @@
+{{>partial_header}}
+{{#models}}
+{{#model}}
+{{#description}}
+// {{{classname}}} : {{{description}}}
```
farcaller: Should be `///` here and everywhere the api docs are expected.

Is `description` guaranteed to be a single line?

wing328:
@farcaller thanks for the review note. I'll work on those shortly.

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -0,0 +1,102 @@
...
+    pub fn new(configuration: Rc<configuration::Configuration<C>>) -> {{{classname}}}Impl<C> {
+        {{{classname}}}Impl {
+            configuration: configuration,
+        }
+    }
+}
+
+{{#operations}}
+{{#operation}}
+pub trait {{classname}} {
```
farcaller: Must be outside of the loop (single trait with multiple `fn`s within.

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -0,0 +1,102 @@
...
+    fn {{{operationId}}}(&self, {{#allParams}}{{paramName}}: {{{dataType}}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) -> Box<Future<Item = (), Error = Error>>;
```
farcaller: `Item = ()` specifies the return type (`()` is the return type of none). It should be resolved to the appropriate model.

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -0,0 +1,102 @@
...
+}
+{{/operation}}
+{{/operations}}
+
+
+{{#operations}}
+{{#operation}}
+impl<C: hyper::client::Connect>{{classname}} for {{classname}}Impl<C> {
```
farcaller: ditto on impl outside of the loop

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -0,0 +1,102 @@
...
+    fn {{{operationId}}}(&self, {{#allParams}}{{paramName}}: {{{datatype}}}{{/allParams}}) -> Box<Future<Item = (), Error = Error>> {
```
farcaller: `{{datatype}} ends up being an empty string here?

farcaller: Ditto on `Item = ...`

farcaller: How swagger deals with different types that map to a single downstream type?

e.g. in a model, a string field is `String`, but in argument, a string is usually a `&str`.

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -0,0 +1,102 @@
...
+        let configuration: &configuration::Configuration<C> = self.configuration.borrow();
+        let mut req = hyper::Request::new(
+            hyper::Method::{{httpMethod}},
+            format!("{}{{path}}", configuration.base_path).parse().unwrap());
```
farcaller: {{path}} resolves to strings like `/user/{username}`, in case it has args, they must be substituted by `{}` and added to macro arguments list.

modules/swagger-codegen/src/main/resources/rust/model_mod.mustache
```diff
@@ -0,0 +1,6 @@
+{{#models}}
+{{#model}}
+mod {{{classname}}};
```
farcaller: Should be {{classFilename}}

modules/swagger-codegen/src/main/resources/rust/model.mustache
```diff
@@ -0,0 +1,57 @@
...
+{{/description}}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct {{classname}} {
+{{#vars}}
+  {{#description}}
+  // {{{description}}}
+  {{/description}}
+  #[serde(rename = "{{baseName}}")] {{name}}: {{^required}}Option<{{/required}}{{{datatype}}}{{^required}}>{{/required}}{{#hasMore}},{{/hasMore}}
```
farcaller: datatype for an array is `Vec<TYPE>`, not `[]TYPE`

farcaller: Also, the TYPE, if it's another model, isn't available in this scope and should be referenced as `super::MODEL_NAME` (from the parent file)

farcaller: I assume there's no sense to have an optional array as it's same as empty array in swagger semantics? Then arrays shouldn't be wrapped in `Option<>`.

Also: see next one

modules/swagger-codegen/src/main/resources/rust/model.mustache
```diff
@@ -0,0 +1,57 @@
...
+}
+
+impl {{classname}} {
+  {{#description}}
+  // {{{description}}}
+  {{/description}}
+  pub fn new({{#requiredVars}}{{name}}: {{{datatype}}}{{^-last}}, {{/-last}}{{/requiredVars}}) -> {{classname}} {
+    {{classname}} {
+      {{#vars}}
+      {{name}}: {{#required}}{{name}}{{/required}}{{^required}}{{#isListContainer}}Vec:new(){{/isListContainer}}{{#isMapContainer}}Hash:new(){{/isMapContainer}}{{^isContainer}}None{{/isContainer}}{{/required}}{{#hasMore}},{{/hasMore}}
```
farcaller: If the array can truly be optional, then init it with `None` (a value of Option)

modules/swagger-codegen/src/main/resources/rust/model.mustache
```diff
@@ -0,0 +1,57 @@
...
+      {{/vars}}
+    }
+  }
+
+  {{#vars}}
+  pub fn set_{{name}}(&mut self, {{name}}: {{{datatype}}}) {
+    self.{{name}} = Some({{name}});
```
farcaller: `Some(...)` only applies to `Option<>` (optional) fields. If the field isn't optional, assign as `self.{{name}} = {{name}}`

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RustClientCodegen.java
wing328: Shall we name it as `api_client.rs` instead to make consistent with other generators?

farcaller: SGTM.

modules/swagger-codegen/src/main/resources/rust/api.mustache
wing328: We can use vendor extensions to annotate that. Do you have a mapping for all different primitive types?

modules/swagger-codegen/src/main/resources/rust/api.mustache
wing328: so it becomes `/user/{}`, right? and how to add it to the macro arguments list?

modules/swagger-codegen/src/main/resources/rust/api.mustache
farcaller: I think String -> &str is the only special case, actually.

wing328:
I've pushed https://github.com/swagger-api/swagger-codegen/pull/6105/commits/af9537e806a36961dc77afc91b48f67b5cad642b to incorporate your suggested fixes but I didn't have a chance to verify the Rust generator's output yet.

Please continue to comment on the auto-generated code so that we can move this forward. Thanks!

wing328:
> In the generated code for ApiResponse model, Type field is renamed to Type_ (should be underscorecased too)

Fixed via https://github.com/swagger-api/swagger-codegen/pull/6105/commits/e22a48f3012b5d988ae8c7ac7f0556c99ecc4962

modules/swagger-codegen/src/main/resources/rust/api.mustache
wing328: Fixed via 9ff1cc4. Please pull the latest to take a look

modules/swagger-codegen/src/main/resources/rust/api.mustache
farcaller: Yes. It would look like `format!("{}{}", configuration.base_path, path)` assuming that `path` was defined in the function argument.

farcaller: I mean.. `format!("{}/user/{}", configuration.base_path, user)` in this particular case.

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -0,0 +1,103 @@
...
+    fn {{{operationId}}}(&self, {{#allParams}}{{paramName}}: {{#isString}}&str{{/isString}}{{^isString}}{{{dataType}}}{{/isString}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) -> Box<Future<Item = {{^returnType}}(){{/returnType}}{{#returnType}}{{{returnType}}}{{/returnType}}, Error = Error>> {
+        let configuration: &configuration::Configuration<C> = self.configuration.borrow();
+        let mut req = hyper::Request::new(
+            hyper::Method::{{httpMethod}},
+            format!("{{path}}", configuration.base_path).parse().unwrap());
+
+        {{#hasQueryParams}}
+        {{#queryParams}}
+        {{#required}}
+        /// TODO query parameter {{baseName}}({{paramName}})
```
farcaller: Here and everywhere where the comment isn't an apidoc comment it must remain `//` or it's a compilation failure (i.e. `///` is used only before methods / functions / fields)

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -0,0 +1,103 @@
...
+
+
+impl<C: hyper::client::Connect>{{classname}} for {{classname}}Impl<C> {
+{{#operations}}
+{{#operation}}
+    fn {{{operationId}}}(&self, {{#allParams}}{{paramName}}: {{#isString}}&str{{/isString}}{{^isString}}{{{dataType}}}{{/isString}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) -> Box<Future<Item = {{^returnType}}(){{/returnType}}{{#returnType}}{{{returnType}}}{{/returnType}}, Error = Error>> {
+        let configuration: &configuration::Configuration<C> = self.configuration.borrow();
+        let mut req = hyper::Request::new(
+            hyper::Method::{{httpMethod}},
+            format!("{{path}}", configuration.base_path).parse().unwrap());
```
farcaller: format string must start with verbatim `{}` for the base_path: `"{}{{path}}"`.

Actually, to make it less work, you can skip rewriting the `path`, e.g. if the path is `/user/{userid}` render it like this:

`format!("{}/user/{userid}", configuration.base_path, userid=userid)`

modules/swagger-codegen/src/main/resources/rust/api_mod.mustache
```diff
@@ -0,0 +1,44 @@
...
+
+{{#apiInfo}}
+{{#apis}}
+mod {{classFilename}};
+{{#operations}}
+{{#operation}}
+{{#summary}}
+///{{{summary}}}
+{{/summary}}
+pub use self::{{classFilename}}::{{{operationId}}};
```
farcaller: Actually, the way code is structured, we have one use per imported mod:

`pub use self::{{classFilename}}::{{classname}}`. We don't import the operations.

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -0,0 +1,103 @@
...
+    fn {{{operationId}}}(&self, {{#allParams}}{{paramName}}: {{{dataType}}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) -> Box<Future<Item = {{^returnType}}(){{/returnType}}{{#returnType}}{{{returnType}}}{{/returnType}}, Error = Error>>;
+{{/operation}}
+{{/operations}}
+}
+
+
+impl<C: hyper::client::Connect>{{classname}} for {{classname}}Impl<C> {
+{{#operations}}
+{{#operation}}
+    fn {{{operationId}}}(&self, {{#allParams}}{{paramName}}: {{#isString}}&str{{/isString}}{{^isString}}{{{dataType}}}{{/isString}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) -> Box<Future<Item = {{^returnType}}(){{/returnType}}{{#returnType}}{{{returnType}}}{{/returnType}}, Error = Error>> {
```
farcaller: if returntype references a model, it should be prefixed with `super::`

farcaller: Notice, for arrays of models, that's `Vec<super::Pet>`, not e.g. `super::Vec<Pet>`.

farcaller:
I'm looking at this method:

```yaml
  /store/inventory:
    get:
      tags:
        - store
      summary: Returns pet inventories by status
      description: Returns a map of status codes to quantities
      operationId: getInventory
      produces:
        - application/json
      parameters: []
      responses:
        '200':
          schema:
            type: object
            additionalProperties:
              type: integer
              format: int32
      security:
        - api_key: []
```

And I'm confused of the return type. How is it expected to look in json?

modules/swagger-codegen/src/main/resources/rust/api.mustache
wing328: OK. I'll need more time to handle this.

wing328:
> And I'm confused of the return type. How is it expected to look in json?

```
{
  "cat": 10,
  "dog", 2,
  "rabbit", 32
}
```
Ref: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#model-with-mapdictionary-properties

modules/swagger-codegen/src/main/resources/rust/api.mustache
wing328: Done via c8ff9a9

wing328:
More fixes via c8ff9a9

For "///", can you point out in the template where it should be used? Is it some sort of method/field documentation?

farcaller:
> Ref: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#model-with-mapdictionary-properties

Ok, so for these the mapped type would be `::std::collections::HashMap<String, T>` where T is the target type.

modules/swagger-codegen/src/main/resources/rust/model.mustache
farcaller: `///` here.

modules/swagger-codegen/src/main/resources/rust/model.mustache
```diff
@@ -0,0 +1,57 @@
...
+{{#model}}
+{{#description}}
+// {{{classname}}} : {{{description}}}
+{{/description}}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct {{classname}} {
+{{#vars}}
+  {{#description}}
+  // {{{description}}}
```
farcaller: `///` here.

modules/swagger-codegen/src/main/resources/rust/model.mustache
```diff
@@ -0,0 +1,57 @@
...
+  {{/description}}
+  #[serde(rename = "{{baseName}}")] {{name}}: {{^required}}Option<{{/required}}{{^isPrimitiveType}}{{^isContainer}}super::{{/isContainer}}{{/isPrimitiveType}}{{{datatype}}}{{^required}}>{{/required}}{{#hasMore}},{{/hasMore}}
+{{/vars}}
+}
+
+impl {{classname}} {
+  {{#description}}
+  // {{{description}}}
```
farcaller: `///` here.

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -0,0 +1,103 @@
...
+
+
+impl<C: hyper::client::Connect>{{classname}} for {{classname}}Impl<C> {
+{{#operations}}
+{{#operation}}
+    fn {{{operationId}}}(&self, {{#allParams}}{{paramName}}: {{#isString}}&str{{/isString}}{{^isString}}{{{dataType}}}{{/isString}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) -> Box<Future<Item = {{^vendorExtensions.x-returnTypeInMethod}}(){{/vendorExtensions.x-returnTypeInMethod}}{{#vendorExtensions.x-returnTypeInMethod}}{{{.}}}{{/vendorExtensions.x-returnTypeInMethod}}, Error = Error>> {
+        let configuration: &configuration::Configuration<C> = self.configuration.borrow();
+        let mut req = hyper::Request::new(
+            hyper::Method::{{httpMethod}},
+            format!("{}{{path}}", configuration.base_path{{#allParams}}, {{paramName}}={{paramName}}{{/allParams}}));
```
farcaller: `body` leaks in here, it shouldn't (it's handled down in line 83)

farcaller: The naming is getting mixed up, e.g.

`format!("{}/pet/{petId}", configuration.base_path, pet_id=pet_id, api_key=api_key))`

here `petId` comes from the original field name, but the assignment is done to `pet_id`.

farcaller: also, on the previous one: there's api_key, and it's not mapped to the URI path. Where's that coming from?

farcaller: Also here. I noticed that there are more arguments that I didn't expect to see, I guess those are query arguments?

They need to be processed like this:

```rust
// if query params are empty ...
let mut req = hyper::Request::new(
    hyper::Method::{{httpMethod}},
    format!("{}{{path}}", configuration.base_path{{#allParams}}, {{paramName}}={{paramName}}{{/allParams}}));
// else
let query = url::form_urlencoded::Serializer::new(String::new())
    {{#allParams}}.append_pair("{{paramName}}", {{paramName}}){{/allParams}} // only for query params
    .finish();
let mut req = hyper::Request::new(
    hyper::Method::{{httpMethod}},
    format!("{}{{path}}?{}", configuration.base_path, query{{#allParams}}, {{paramName}}={{paramName}}{{/allParams}}));

```

farcaller: For clarification, `format!(...)` builds an absolute URL here, starting with the scheme and down to query arguments. We get the scheme and base path from `configuration.base_path`, then the path is hard-coded static into the string, then the query arguments follow.

With `format!("{}somepath/{option}/whatever?{}", A, B, option=C)`, the first `{}` is resolved to A, the second (which is in the ver end of the string) is resolved to B, and `{option}` is resolved to C via named argument.

farcaller: So yes, answering your question, only path parameters and query parameters (if any) will end up in here.

wing328: We need to deal with other type of parameters as well: header, query, form.

Please have a look at Go API client API templates or other templates on how we handle all these parameters using a hash/dictionary:

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/go/api.mustache#L41-L47

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/go/api.mustache#L103-L186

For API key, it's defined in security defintions and it's handled in the {{authMethod}} block: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/go/api.mustache#L188-L203

Is there a way we can put the hyper::Request logic in api_client.rs instead while the API layer (e.g. pet_api.rs, user_api.rs) is only responsible for putting different parameters (e.g. header, form, etc) into a hash/dictionary? Later if there's demand for using another HTTP lib in Rust, we can simply make api_client.rs swappable.

wing328:
Fixed docstring and added support for dictionary/hash via 73ceea0

modules/swagger-codegen/src/main/resources/rust/api.mustache
farcaller: Ok, so, the code would look like:

```rust
        let configuration: &configuration::Configuration<C> = self.configuration.borrow();

        let method = hyper::Method::{{httpMethod}};
        
        // { if empty query args }
        let uri = format!("{}{{path}}", configuration.base_path{{#allParams}}, {{paramName}}={{paramName}}{{/allParams}}));
        // { else }
        let query = url::form_urlencoded::Serializer::new(String::new())
            {{#queryParams}}.append_pair("{{paramName}}", {{paramName}}){{/queryParams}} // only for query params
            .finish();
        let uri = format!("{}{{path}}{}", configuration.base_path, query{{#allParams}}, {{paramName}}={{paramName}}{{/allParams}}));
        // { end }
        
        let mut req = hyper::Request::new(method, uri);
        
        // { if header args }
        let mut headers = req.headers_mut();
        {{#headerParams}}
        headers.set_raw("{{paramName}}", "{{paramValue}}");
        {{/headerParams}}
        // { end }

        // { if body args }
        let serialized = serde_json::to_string(body).unwrap();
        req.headers_mut().set(hyper::header::ContentType::json());
        req.headers_mut().set(hyper::header::ContentLength(serialized.len() as u64));
        req.set_body(serialized);
        // { end }

        // send request
        Box::new(
            configuration.client.request(req).and_then(|res| { res.body().concat2() })
            .map_err(|e| Error::from(e))
            // { if no response expected }
            .and_then(|_| futures::future::ok(()))
            // { else }
            .and_then(|body| {
                let parsed: Result<{{returnType}}, _> = serde_json::from_slice(&body);
                parsed.map_err(|e| Error::from(e))
            }).map_err(|e| Error::from(e))
            // { end}
        )
```

This covers path, query, header params. Trying to figure where do form params land and there's also the "file" thing that requires multipart upload.

As on extracting the logic, I think it's better to have version 1 with code explicit. There are a few options on how to extract the logic, but I'm not sure which one would be more effective in terms of calls/memory allocation and still researching that.

farcaller:
> Fixed docstring and added support for dictionary/hash via 73ceea0

It's getting rendered as e.g.

`super::::std::collections::HashMap<String, i32>`

notice that there shouldn't be `super::` in this case (the i32 is a primitive type), but if it was a pet, it'd be `::std::collections::HashMap<String, super::Pet>`.

wing328:
> it'd be ::std::collections::HashMap<String, super::Pet>

Should be fixed.

> As on extracting the logic, I think it's better to have version 1 with code explicit. There are a few options on how to extract the logic

I'm not saying version 1 is bad. My suggestion is based on experience working on other generators. I think it's just a matter of time we need to refactor common logic in ApiClient class to accommodate more feature requests.

> but I'm not sure which one would be more effective in terms of calls/memory allocation and still researching that.

We can work on the performance as a day 2 requirement. First and foremost, we'll need to release a functional Rust API client to collect feedbacks from the community.

wing328:
TODO: add logic to check optional/required (`{{#required}}`) parameters (query, form, header, body)

wing328:
For template changes/fixes, please feel free to submit a PR directly to `rust_generator` branch to avoid the overhead in back-and-forth communications.

farcaller:
> I'm not saying version 1 is bad. My suggestion is based on experience working on other generators. I think it's just a matter of time we need to refactor common logics into ApiClient class to accommodate more feature requests.

Ack.

> For template changes/fixes, please feel free to submit a PR directly to rust_generator branch to avoid the overhead in back-and-forth communications.

#6114

After this PR, there are only a handful of issues left.

modules/swagger-codegen/src/main/resources/rust/model.mustache
```diff
@@ -0,0 +1,57 @@
...
+/// {{{classname}}} : {{{description}}}
+{{/description}}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct {{classname}} {
+{{#vars}}
+  {{#description}}
+  /// {{{description}}}
+  {{/description}}
+  #[serde(rename = "{{baseName}}")] {{name}}: {{^required}}Option<{{/required}}{{^isPrimitiveType}}{{^isContainer}}super::{{/isContainer}}{{/isPrimitiveType}}{{{datatype}}}{{^required}}>{{/required}}{{#hasMore}},{{/hasMore}}
```
farcaller: in here, if it's an array, it's rendered as e.g. `Option<Vec<Tag>>`, `super::` is missing from the child struct name.

modules/swagger-codegen/src/main/resources/rust/model.mustache
```diff
@@ -0,0 +1,57 @@
...
+  pub fn new({{#requiredVars}}{{name}}: {{{datatype}}}{{^-last}}, {{/-last}}{{/requiredVars}}) -> {{classname}} {
+    {{classname}} {
+      {{#vars}}
+      {{name}}: {{#required}}{{name}}{{/required}}{{^required}}{{#isListContainer}}None{{/isListContainer}}{{#isMapContainer}}Hash:new(){{/isMapContainer}}{{^isContainer}}None{{/isContainer}}{{/required}}{{#hasMore}},{{/hasMore}}
+      {{/vars}}
+    }
+  }
+
+  {{#vars}}
+  pub fn set_{{name}}(&mut self, {{name}}: {{{datatype}}}) {
```
farcaller: Same as above, if it's an array, `super::` needs to be added to the child struct name.

Also, `super::` is generally missing if it's a non-primitive type.

Can we rewrite literally any use of a model into an absolute path? E.g. for model T, it should be referenced as `::models::{{classnameFile}}::{{classname}}`. Then we don't need to deal with super anywhere.

wing328: Let's go with `::models::{{classnameFile}}::{{classname}}` instead.

wing328: Yup, just pushed 7f9ced9. Please take a look when you've time.

farcaller:
Finished dev [unoptimized + debuginfo] target(s) in 1.19 secs

it finally builds! Not sure if it's functional, though.

farcaller:
Failing curiously on a bigger swagger file: https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json at the definition of `"io.k8s.apimachinery.pkg.apis.meta.v1.Patch"`:

```
   "io.k8s.apimachinery.pkg.apis.meta.v1.Patch": {
    "description": "Patch is provided to give a concrete name and type to the Kubernetes PATCH request body."
   },
```

doesn't generate any stub struct.

wing328:
> Failing curiously on a bigger swagger file: https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json

There are some known issues with the Kubernetes swagger spec (not the issue with the spec but how the Swagger Parser handles certain object reference and there are some workarounds mentioned before)

farcaller:
> there are some workarounds mentioned before

Can you point me at any? Now that it compiles, I'd like to try the generated code in some real-world scenarios, and k8s is a thing of enough complexity to surface any usage concerns.

wing328:
@farcaller I need some time to dig those out.

I would suggest release a beta version that works with Petstore server so that others can start contributing to it to make it better. (we usually create an enhancement list ([example](https://github.com/swagger-api/swagger-codegen/issues/5886)) after a beta release of new generator)

wing328:
@farcaller Here is the related issue: https://github.com/swagger-api/swagger-codegen/issues/5400

lucab:
@farcaller FYI https://github.com/camallo/k8s-client-rs. I have an half-baked private branch with a manually coded generator based on openapi-rs, but I'll gladly switch to something more maintainable. Sorry for the the OT.

wing328:
@lucab If you've time, I wonder if you can try the latest Rust Petstore client generated by this PR and let us know if you've any feedback. Thanks!

farcaller:
> I would suggest releasing a beta version that works with Petstore server

I'll write a few examples, but I'm pretty confident in that you can call most of petstore APIs. Is there any swagger-global test case to follow?

wing328:
> I'll write a few examples, but I'm pretty confident in that you can call most of petstore APIs. Is there any swagger-global test case to follow?

@farcaller What about the following tests as a starting point?
Add - AddPet
Update - UpdatePet
Get - GetPetById
Delete - DeletePet

farcaller:
Sorry, I've got a few busy days. Will get to tests later this week.

wing328:
@farcaller no problem. Please take your time.

wing328:
@farcaller what about releasing an alpha/beta version so that Rust developers can test it out?

BenjaminGill-Metaswitch:
Sorry - I'm about to throw a spanner in the works.

We (@Metaswitch) have been working on an implementation of Rust client/server generation in swagger-codegen for nearly a year (for use in a micro-service framework). It works well enough that we're immanently going to be using it in production.

We're currently working on open-sourcing our implementation, which should occur within the next couple of weeks.

I've only had a cursory glance at this implementation - the main difference seems to be that we have a single `Api` trait that the client uses and the server implements.

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -0,0 +1,96 @@
...
+use futures;
+use futures::{Future, Stream};
+
+use super::{Error, configuration};
+
+pub struct {{{classname}}}Impl<C: hyper::client::Connect> {
+    configuration: Rc<configuration::Configuration<C>>,
+}
+
+impl<C: hyper::client::Connect> {{{classname}}}Impl<C> {
```
quodlibetor: FWIW I don't see a lot of `Impl` suffixes in Rust. I believe it would be more idiomatic to make this something like `{{{classname}}}Connection` or rename the trait to `{{{classname}}}Client`.

samples/client/petstore/rust/README.md
```diff
@@ -0,0 +1,97 @@
+# Go API client for swagger
```
quodlibetor: Also seems a bit off.

wing328:
> Sorry - I'm about to throw a spanner in the works.

@BenjaminGill-Metaswitch we look forward to the Rust client and server generators by @Metaswitch 

Haskell also starts with a client and server generator and now there're suggestions to create another Haskell client generator to meet different requirements.

We'll keep this Rust client generator and later see if we can combine both into one via CLI options.

modules/swagger-codegen/src/main/resources/rust/README.mustache
```diff
@@ -0,0 +1,96 @@
+# Go API client for {{packageName}}
```
wing328: Yup, it should be replaced by "Rust API client ..."

modules/swagger-codegen/src/main/resources/rust/api.mustache
wing328: @quodlibetor do you mind submitting a PR with the suggested change (together with the fix to replace "Go" with "Rust")?

samples/client/petstore/rust/README.md
wing328: Good catch. I'll do a grep for "Go" before release to make sure we don't have any Go-related code/doc left over in the Rust generator and templates.

quodlibetor: Sure thing, I ought to be able to do it tomorrow. Is the standard to regenerate the petshop code as part of rename PRs?

wing328: Yes, we usually run the shell script (./bin/) or windows batch file (.\bin\windows\) to regenerate the Petstore samples so that CI (travis, circleci, etc) can verify the change (there's no CI setup for Rust Petstore sample yet)

quodlibetor: Cool then. I'll put up the PR and you can decide if it fits.

quodlibetor:
Created #6247 , a PR against this branch.

wing328:
> Created #6247 , a PR against this branch.

@quodlibetor PR merged into master. Thanks for the contribution.

In your opinion, do you think the Rust generator is ready for alpha/beta release?

farcaller:
I ran a few tests against the petstore and it worked as expected for me.

wing328:
> I ran a few tests against the petstore and it worked as expected for me, apart from the whole thing where I forgot to add getters to api

@farcaller Nice. I'll release it later today by merging it into master and create a separate ticket (issue) to track future enhancements.

Thanks for the great work!

farcaller:
https://github.com/swagger-api/swagger-codegen/pull/6249 fixes getters.

wing328:
Btw, are you guys aware of any official/popular Rust style guide? We'll update https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md#style-guide accordingly.

The ones I found by doing a google search ("rust style guide") do not seem to be popular nor official.

farcaller:
I believe this is where the official style is being developed: https://github.com/rust-lang-nursery/fmt-rfcs, but it's not finalized yet.

wing328:
PR merged into master. 

Updated README to add @farcaller as the Rust template creator.

wing328:
Tweet to promote the new generator: https://twitter.com/wing328/status/894127231034720256

Future enhancements: https://github.com/swagger-api/swagger-codegen/issues/6250

Issue #6111, opened 2017-07-19
[Flask] incomplete reserved words list


chubinou:
##### Description

* some python reserved keywords (self, None) are not escaped properly
* some python3 reserved keywords (True, False) are not escaped

##### Swagger-codegen version

2.2.3

##### Swagger declaration file content or url

```yaml
%YAML 1.0
---
swagger: '2.0'
info:
  version: v1
paths:
  /link:
    get:
      responses:
        "200":
          schema:
            $ref: "#/definitions/True"
        "400":
          schema:
            $ref: "#/definitions/self"
definitions:
  True:
    type: object
    properties:
      True:
        type: string
  self:
    type: object
    properties:
      self:
        type: integer
```

this produce a Link model like

```python
#swagger_server/models/self.py
class Self(Model):
    def __init__(self, self: int=None):
        "..."
        self._self = self
#swagger_server/models/true.py
class True(Model):
    def __init__(self, true: str=None):
        "..."
        
```

##### Suggest a fix/enhancement

setReservedWordsLowerCase is called twice, the second time with an incomplete list

wing328:
Closed via #6112

Issue #6059, opened 2017-07-14
Create language-specified base class


wing328:
##### Description

As the awesome Swagger Codegen community adds more generators (e.g clients, servers) to Swagger Codegen, we want to refactor some generators to create language-specific base class so as to reuse the code that are common to generators for the same langauges. Here are a few examples

[C++](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCppCodegen.java)
[C#](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java) 
[Java](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java)
[Typescript](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractTypeScriptClientCodegen.java)

Here is a list to track progress:

- [ ] Android
- [ ] Bash
- [x] C++
- [x] C#
- [ ] Clojure
- [ ] Dart
- [ ] Elixir
- [ ] Eiffel
- [ ] Flash
- [ ] Go
- [ ] Groovy
- [ ] Javascript
- [x] Java
- [ ] ObjC
- [ ] Perl
- [x] PHP
- [ ] PowerShell
- [ ] Python
- [ ] Ruby
- [ ] Swift
- [x] Scala
- [x] TypeScript

##### Swagger-codegen version

Latest master

##### Suggest a Fix/Enhancement

If anyone wants to work on the base class creation, please reply to let us know.

bvwells:
Hi @wing328, I've added an abstract base class for Go codegen. Let me know if you have any feedback.

PR #6112, opened 2017-07-19
[Flask] self and None are reserved words (#6111)

chubinou:
* setReservedWordsLowerCase was called twice, the second time without self and None

* True, False and nonlocal are python3 reserved words

### PR checklist

### Description of the PR

issue reference #6111

* setReservedWordsLowerCase was called twice, the second time without self and None

* True, False and nonlocal are python3 reserved words

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PythonClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PythonClientCodegen.java
@@ -90,7 +90,7 @@ public class PythonClientCodegen extends DefaultCodegen implements CodegenConfig
         // map uuid to string for the time being
         typeMapping.put("UUID", "str");
 
-        // from https://docs.python.org/release/2.5.4/ref/keywords.html
+        // from https://docs.python.org/3/reference/lexical_analysis.html#keywords
         setReservedWordsLowerCase(
                 Arrays.asList(
                     // local variable name used in API methods (endpoints)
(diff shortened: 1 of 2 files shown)


wing328:

wing328:
@chubinou may I know if you've time to update the Python reserved keyword list as well: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PythonClientCodegen.java#L94 ? (note that there are some local variables that should be kept for Python client only)

Ideally we should create a Python abstract base class to use the same reserved word list: https://github.com/swagger-api/swagger-codegen/issues/6059

chubinou:
@wing328 thanks for the account linking notification, it should be good now.

I'll look at the python client side, it should be pretty quick to do.

wing328:
Python tests passed:
```
Using 3218105851 as seed
___________________________________ summary ____________________________________
  py27: commands succeeded
  py3: commands succeeded
  congratulations :)
```
(the Github rate limit error can be ignore)

wing328:
@chubinou thanks for the PR, which has been merged into master.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PythonClientCodegen.java
```diff
@@ -102,7 +102,7 @@ public PythonClientCodegen() {
                     "and", "del", "from", "not", "while", "as", "elif", "global", "or", "with",
                     "assert", "else", "if", "pass", "yield", "break", "except", "import",
                     "print", "class", "exec", "in", "raise", "continue", "finally", "is",
-                    "return", "def", "for", "lambda", "try", "self", "None"));
+                    "return", "def", "for", "lambda", "try", "self", "nonlocal", "None", "True", "False"));
```
tomer: https://github.com/swagger-api/swagger-codegen/issues/7955#issuecomment-384636747

Can you please add `type` and `id` as well all other built-in functions? These keywords are not prohibited as the reserved words, but it is generally good idea not to use them in code, as they are masking the built-in functions. Since there are already some keywords in the list that are not reserved (including `self`, which is nothing but common naming best practice to reference to the instance), I suggest expanding the list of reserved keywords to include all the [built-in functions](https://docs.python.org/3/library/functions.html).

Issue #6100, opened 2017-07-18
[Java-PlayFramework] Fix dependency injection in the Controllers


tzimisce012:
##### Description

When a controller is generated we have that the constructor has as input the implementation of the interface so no dependency injection can be done.

##### Swagger-codegen version

2.3.0-SNAPSHOT

##### Suggest a fix/enhancement

Instead, the constructor should have the interface of the api.

JFCote:
Hi @tzimisce012 

Thanks for the bug report. Unfortunately, I'm not sure to understand it completely.
Can you please provide an example to illustrate the problem that you have?

Thanks!

tzimisce012:
The generated code is the following:

`public class UsersApiController extends Controller {

    private final UsersApiControllerImp imp;
    private final ObjectMapper mapper;

    @Inject
    private UsersApiController(UsersApiControllerImp imp) {
        this.api = api;
        mapper = new ObjectMapper();
    }`

Instead of using the implementation, the interface should be used, and then the implementation will be used thru dependency injection

JFCote:
@tzimisce012 I will look into it this morning. I'm pretty sure I will need to add a module but it should be easy. Let me check for this. Thanks

wing328:
Closed via #6100

PR #6128, opened 2017-07-20
[JavaPlayFramework] Fix for issue #6100

JFCote:
### PR checklist

### Description of the PR

Fix issue #6100 . Please let me know if something is wrong.


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaPlayFrameworkCodegen.java
```diff
@@ -153,6 +153,9 @@ public void processOpts() {
         supportingFiles.add(new SupportingFile("routes.mustache", "conf", "routes"));
 
         //App/Utils folder
+        if (!this.controllerOnly && this.useInterfaces) {
```
edrevo: Could you add an option to disable emitting the Module or to emit the interfaces only? We are integrating swagger-codegen in our SBT build and we would like to only have the interfaces generated.

samples/server/petstore/java-play-framework-no-swagger-ui/app/Module.java
```diff
@@ -0,0 +1,13 @@
+import com.google.inject.AbstractModule;
+
+import controllers.*;
+
+public class Module extends AbstractModule {
+
+    @Override
+    protected void configure() {
+        bind(PetApiControllerImpInterface.class).to(PetApiControllerImp.class);
```
edrevo: Does this compile? `PetApiControllerImpInterface` doesn't exist, afaict.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaPlayFrameworkCodegen.java
JFCote: Unless I'm mistaken, I don't think it's possible. Since we use injection, if I inject the interface, I must be able to tell Play which implementation implements the interface (in the module). If only the interface exists, Play will not be able to run.

For your problem, you just have to use the normal "swagger-codegen-ignore" file that will ignore file you put in it, like a git ignore.

samples/server/petstore/java-play-framework-no-swagger-ui/app/Module.java
JFCote: In this example, the interface exists. This is the "no-swagger-ui" example, so only the UI is absent.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaPlayFrameworkCodegen.java
edrevo: `swagger-codegen-ignore` unfortunately only works for overwrites. It will not prevent the file being created. For our scenario, we are outputting the generated code to the `target` folder (similar to what Play does with its routes), so we don't want to bind Play to any specific implementation (since that will be done by the non-autogenerated code).

I understand having an `interfaceOnly` option would generate a project that would not run as-is, since the implementations would be missing, but that is exactly what we need. We can delete the unnecessary files from the SBT plugin's side, but that seems hacky and will make updating the dependency on the codegen lib more difficult.

edrevo: Now that I've read the code in more detail, I think what I am asking is what `controllerOnly` should do: emit the controllers, emit the interface, but don't emit an implementation.

wing328: @edrevo I think `swagger-codegen-ignore` will also prevent the file from being created for the first time. Can you give it a try?

JFCote:
@wing328 I will look into this tomorrow to fix the conflict.

JFCote:
@wing328 Wait before merging this, I'm  experimenting some compilation issue. Will update the PR with the good code.

JFCote:
@wing328 Just found that the "example" implementation PR https://github.com/swagger-api/swagger-codegen/pull/5775 has been merged but needed help from the community to be merged.

I will completely remove the lines from this PR because it completely break the PlayFramework generator and I don't understand the goal/meaning of the examples at all.

Will commit a fix this afternoon in the current PR

JFCote:
Ok, I fixed everything. The only thing that is not working right now is the "fake endpoint sample" which was added by @wing328 lately. I will be opening a bug about this because I'm not sure we should support all of the cases.

So @wing328 , if it's ok for you, you can merge this.

wing328:
`samples/server/petstore/java-play-framework/sbt_test_jdk8_only.sh` and `samples/server/petstore/java-play-framework/pom.xml` were removed. I'll add those back after the merge.

PR #6129, opened 2017-07-20
Add additional cli option and properties to swift3 for Objective-C compatibility

ehyche:
### Description of the PR

Currently, in the swift3 language, if you have an optional integer, number, or boolean property in a model, then the generated swift3 model class might look like:

```
class SomeModel {
    var someInt: Int?
    var someFloat: Float?
    var someDouble: Double?
    var someBool: Bool?
}
```

This works fine if you are accessing this model only from Swift code. However, it is very common for iOS codebases to contain both Swift AND Objective-C. If you need to access this model from Objective-C, then those 4 properties are not accessible, since Optional scalars do not translate to Objective-C.

Therefore, in the swift3 language, we want to add some code for Objective-C compatibility:

1. We add a "objcCompatible" boolean command-line option. If objcCompatible=true, then this enables some additional code generation to make these types of properties accessible from Objective-C. If objcCompatible=false, then the generated code is exactly as it currently is. The default is objcCompatible=false.

2. If objCompatible=true, then for these types of Objective-C-inaccessible properties (Optional scalars), then we add a "x-swift-optional-scalar=true" vendor extension in the CodegenProperty.

3. Then, in the model.mustache template, if we see x-swift-optional-scalar=true, then we add an additional computed property which returns an optional NSNumber.

So, for example, when objcCompatible=false (the default case), then the generated code for the "declawed" property of the Cat model looks like:

```
open class Cat: Animal {

    public var declawed: Bool?

    ...
```

But when objcCompatible=true, then it looks like:

```
open class Cat: Animal {

    public var declawed: Bool?
    public var declawedNum: NSNumber? {
        get {
            return declawed.map({ return NSNumber(value: $0) })
        }
    }

   ...
```

### To Do

There is remaining work to do for Objective-C compatibility, particularly around enums. The only Swift enums which are visible from Objective-C are enums where the rawValue is an Int. But hopefully now that we have this cli option, then the remaining work should be template-only changes.

I also need to file a PR to merge this changes to the swift4 language as well.

### PR checklist

### DIFF
--- a/bin/swift3-petstore-objcCompatible.json
+++ b/bin/swift3-petstore-objcCompatible.json
@@ -1,0 +1,7 @@
+{
+  "podSummary": "PetstoreClient",
+  "podHomepage": "https://github.com/swagger-api/swagger-codegen",
+  "podAuthors": "",
+  "projectName": "PetstoreClient",
+  "objcCompatible": true
+}
--- a/bin/swift3-petstore-objcCompatible.sh
+++ b/bin/swift3-petstore-objcCompatible.sh
@@ -1,0 +1,31 @@
+#!/bin/sh
+
+SCRIPT="$0"
+
+while [ -h "$SCRIPT" ] ; do
+  ls=`ls -ld "$SCRIPT"`
+  link=`expr "$ls" : '.*-> \(.*\)$'`
+  if expr "$link" : '/.*' > /dev/null; then
+    SCRIPT="$link"
+  else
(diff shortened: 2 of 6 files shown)


wing328:
@ehyche thanks for the enhancement. I'm sure ObjC programmers will welcome the enhancements.

cc @jaz-ah @jgavris @Edubits @mateuszmackowiak

wing328:
Tests passed via https://travis-ci.org/swagger-api/swagger-codegen/builds/260129714

Issue #4848, opened 2017-02-24
undefined method build_from_hash in Enum data type


ydgiannp:
Hi, Im having an issue.
I am generating a json swagger into a ruby gem, the swagger JSOn contains Enum data type.
The generated class that has an Enum data type does not have build_from_hash method so it always get NoMethodError. the convert_to_type method always call build_from_hash method 

this is the convert_to_type method

![screen shot 2017-02-24 at 9 09 23 pm](https://cloud.githubusercontent.com/assets/13848777/23306329/abf95aa6-fad5-11e6-8120-9b9a2a70052d.png)

this is the AccountCategory class that has Enum data type 

![screen shot 2017-02-24 at 9 09 39 pm](https://cloud.githubusercontent.com/assets/13848777/23306339/b9987124-fad5-11e6-83bc-d2ed07a0dd3c.png)

can

wing328:
@ydgiannp do you mind sharing the spec (e.g. via gist.github.com) so that we can more easily reproduce the issue?

ydgiannp:
here's the spec :) https://gist.github.com/ydgiannp/fd5178bd88b9da10bea12a14ae00ca4a

prsnca:
i have encountered the same issue - my current solution is in the swagger.json - replace the $ref which refers to the enum object by the definition of the $ref itself - regenerate your ruby gem and it should work!

prsnca:
Here is the workaround - 

swagger.json before:

```
...
"enumObject": {
  "type": "string",
  "enum": [
    "VALUE1",
    "VALUE2",
    "VALUE3",
    "VALUE4"
  ],
  "default": "VALUE1"
},
"someObjectUsingEnum": {
  "type": "object",
  "properties": {
    "myEnum": {
      "$ref": "#/definitions/enumObject"
    },
    "myOtherProperty": {
      "type": "string"
    }
  }
}
...
```

swagger.json after:
```
...
"enumObject": {
  "type": "string",
  "enum": [
    "VALUE1",
    "VALUE2",
    "VALUE3",
    "VALUE4"
  ],
  "default": "VALUE1"
},
"someObjectUsingEnum": {
  "type": "object",
  "properties": {
    "myEnum": {
      "type": "string",
      "enum": [
         "VALUE1",
         "VALUE2",
         "VALUE3",
         "VALUE4"
       ],
      "default": "VALUE1"
    },
    "myOtherProperty": {
      "type": "string"
    }
  }
},
...
```

[...]

PR #6131, opened 2017-07-21
Ruby Enums: Fix for issue #4848

NicoEigenmannCW:
### PR checklist

### Description of the PR
Fixes issue with ENUM generation for ruby issue #4848

### DIFF
--- a/modules/swagger-codegen/src/main/resources/ruby/partial_model_enum_class.mustache
+++ b/modules/swagger-codegen/src/main/resources/ruby/partial_model_enum_class.mustache
@@ -1,4 +1,13 @@
   class {{classname}}
     {{#allowableValues}}{{#enumVars}}
     {{{name}}} = {{{value}}}.freeze{{/enumVars}}{{/allowableValues}}
+
+    # Builds the enum from string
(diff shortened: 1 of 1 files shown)


wing328:
Tests passed via https://travis-ci.org/swagger-api/swagger-codegen/builds/256069991

wing328:
@NicoEigenmannCW thanks for the contribution, which has been merged into master.

run-o:
@wing328 @NicoEigenmannCW 
Is this the desired behavior for the build_from_hash function?
It looks like this function currently checks the given value versus the enum's name rather than its actual value.
I might be missing something, but I was under the impression that we'd want to check the value of the enum rather than the constant's name.
If we indeed care about the value, we should replace the `constants.select{|c| c.to_s == value}` line with `constants.select{|c| const_get(c) == value}`

wing328:
@run-o I wonder if you can submit a PR so that we can review the fix more easily. (Later I'll show you where to add a test case to cover it)

run-o:
@wing328 Yep sure, I'll work on that.

Issue #6135, opened 2017-07-21
[kotlin] warning displayed when setting api-package and model-package as not supported, but yet, the parameters are supported


CappCorp:
##### Description

when specifiying the api package or the model package with kotlin, a warning is displayed

`[main] WARN io.swagger.codegen.languages.KotlinClientCodegen - modelPackage with kotlin generator is ignored. Setting this value independently of packageName is not currently supported.
[main] WARN io.swagger.codegen.languages.KotlinClientCodegen - apiPackage with kotlin generator is ignored. Setting this value independently of packageName is not currently supported.
`
The propertly are anyway used by the code generator (as they are used in the super class of the kotlin generator)

##### Swagger-codegen version

2.2.3

##### Command line used for generation

`java -jar ./modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -t modules/swagger-codegen/src/main/resources/kotlin-client -i modules/swagger-codegen/src/test/resources/2_0/petstore.yaml -l kotlin --artifact-id kotlin-petstore-client --api-package my.api --model-package my.model -o samples/client/petstore/kotlin
`
##### Steps to reproduce

run the command line, the warning will be displayed but the code files will still be generated in the proper package

PR #6137, opened 2017-07-21
[kotlin] removed unaccurate warning (#6135)

CappCorp:
### PR checklist

### Description of the PR

#6135 removed warnings for "apiPackage" and "modelPackage" indicating that the properties are not used even though they are
properties are actually taken into account in `super.processOpts();` call

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/KotlinClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/KotlinClientCodegen.java
@@ -219,14 +219,6 @@ public class KotlinClientCodegen extends DefaultCodegen implements CodegenConfig
             LOGGER.warn(CodegenConstants.INVOKER_PACKAGE + " with " + this.getName() + " generator is ignored. Use " + CodegenConstants.PACKAGE_NAME + ".");
         }
 
-        if (additionalProperties.containsKey(CodegenConstants.MODEL_PACKAGE)) {
-            LOGGER.warn(CodegenConstants.MODEL_PACKAGE + " with " + this.getName() + " generator is ignored. Setting this value independently of " + CodegenConstants.PACKAGE_NAME + " is not currently supported.");
-        }
-
-        if (additionalProperties.containsKey(CodegenConstants.API_PACKAGE)) {
-            LOGGER.warn(CodegenConstants.API_PACKAGE + " with " + this.getName() + " generator is ignored. Setting this value independently of " + CodegenConstants.PACKAGE_NAME + " is not currently supported.");
-        }
-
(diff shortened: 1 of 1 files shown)


wing328:
@CappCorp thanks for the PR. (The Travis CI failure can be ignored)

cc @jimschubert

jimschubert:
@CappCorp @wing328 I see to recall there being issues with the templates or logic in DefaultCodegen if you set `PACKAGE_NAME`, `INVOKER_PACKAGE`, `MODEL_PACKAGE`, and `API_PACKAGE` together… the resulting packages are not intuitive and may not be correct in the templates.

I'd have to play around with it a bit, as I don't recall the exact issue. This PR looks correct, though, they are processed in the super method.

jimschubert:
@CappCorp @wing328 I can't seem to reproduce whatever issue lead me to include those warnings. I must have fixed the issue I was having prior to the initial PR for the kotlin api client, and forgot to remove the warnings.

The way I worded the warnings ("Setting X independently of Y is not supported") leads me to believe I was testing `modelPackage` and `apiPackage` without the full package name and implemented the getters for these as `packageName + "/" + modelPackage`, or maybe I was trying to construct a fully qualified package in each of their setters? That's the only way I could see myself warning on their usage in this case.

To verify there's no issue with these three currently, I deleted locally existing Kotlin client code and regenerated with the following command:

```
./bin/kotlin-client-petstore.sh --additional-properties=\
  modelPackage=us.jimschubert.models,\
  apiPackage=us.jimschubert.apis,\
  packageName=io.example.client
```

Everything looks like it generated correctly, and `./gradlew check assemble` succeeded.

tree output of `src/main`:

```
.
`-- kotlin
    |-- io
    |   `-- example
    |       `-- client
    |           `-- infrastructure
    |               |-- ApiAbstractions.kt
    |               |-- ApiClient.kt
    |               |-- ApiInfrastructureResponse.kt
    |               |-- ApplicationDelegates.kt
    |               |-- Errors.kt
    |               |-- RequestConfig.kt
    |               |-- RequestMethod.kt
    |               |-- ResponseExtensions.kt
    |               `-- Serializer.kt
    `-- us
        `-- jimschubert
            |-- apis
            |   |-- PetApi.kt
            |   |-- StoreApi.kt
            |   `-- UserApi.kt
            `-- models
                |-- ApiResponse.kt
                |-- Category.kt
                |-- Order.kt
                |-- Pet.kt
                |-- Tag.kt
                `-- User.kt

9 directories, 18 files
```

CappCorp:
Is it ok to merge the pull request then ? :)

wing328:
I did some tests as well and the output code works even with the warnings.

We'll remove this for the time being.

Issue #6139, opened 2017-07-21
[objc] [Bug] Bool query params sent as either 1 or 0


marcotta:
Description

Boolean properties when sent as query string parameters are being sent as either 0 or 1.

```
?isEnabled=1
?isEnabled=0
```

To match Swagger-UI behaviour it should send true or false instead
```
?isEnabled=true
?isEnabled=false
```

Swagger-codegen version

2.2.2

Swagger declaration file content or url

```
{
    "name": "isEnabled",
    "description": "Enables the flag",
    "in": "query",
    "type": "boolean"
}
```
Command line used for generation

`swagger-codegen generate -i api-spec.json -l objc -c config.json -o APIClient`

wing328:
Closed via #6142

PR #6142, opened 2017-07-21
[objc] Convert boolean query params to true/false (#6139)

marcotta:
Change booleans in query parameters from 0/1 to true/false to
match Swagger UI behaviour

### PR checklist

### Description of the PR
https://github.com/swagger-api/swagger-codegen/issues/6139

Boolean properties when sent as query string parameters are being sent as either 0 or 1.

```
?isEnabled=1
?isEnabled=0
```
To match Swagger-UI behaviour it should send true or false instead

```
?isEnabled=true
?isEnabled=false
```


wing328:
@marcotta the fix looks reasonable to me as boolean should be represented by either "true" or "false"

Issue #6084, opened 2017-07-17
Replace `useJaxbAnnotations` with `withXml`


wing328:
##### Description

Replace `useJaxbAnnotations` with `withXml` to make it consistent with other generators when it comes to XML support.

https://github.com/swagger-api/swagger-codegen/search?utf8=%E2%9C%93&q=useJaxbAnnotations&type=

##### Swagger-codegen version

Latest master

##### Related issues/PRs

https://github.com/swagger-api/swagger-codegen/pull/5962

##### Suggest a fix/enhancement

If anyone wants to work on this enhancement, please reply to let us know.

mobreza:
@wing328 I'll give it a go.

wing328:
@mobreza thank you 👍

PR #6152, opened 2017-07-24
withXml replaces useJaxbAnnotations

mobreza:
### PR checklist

### Description of the PR

Replace useJaxbAnnotations with withXml to make it consistent with other generators when it comes to XML support (#6084).


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java
```diff
@@ -150,7 +150,7 @@ public AbstractJavaCodegen() {
                 .SERIALIZE_BIG_DECIMAL_AS_STRING_DESC));
         cliOptions.add(CliOption.newBoolean(FULL_JAVA_UTIL, "whether to use fully qualified name for classes under java.util. This option only works for Java API client"));
         cliOptions.add(new CliOption("hideGenerationTimestamp", "hides the timestamp when files were generated"));
-        cliOptions.add(CliOption.newBoolean(WITH_XML, "whether to include support for application/xml content type. This option only works for Java API client (resttemplate)"));
+        cliOptions.add(CliOption.newBoolean(WITH_XML, "whether to include support for application/xml content type and include XML annotations in the model"));
```
wing328: @mobreza I still think we should list out which Java generators support this.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaCXFServerCodegen.java
```diff
@@ -64,7 +61,7 @@
     public JavaCXFServerCodegen()
     {
         super();
-
+
```
wing328: Minor suggestion: remove the leading spaces in an empty line.

samples/server/petstore/jaxrs-resteasy/eap-java8/pom.xml
```diff
@@ -34,23 +46,6 @@
...
-                <version>1.6.0</version>
-                <executions>
-                    <execution>
-                        <id>Play Test</id>
-                        <phase>integration-test</phase>
-                        <goals>
-                            <goal>exec</goal>
-                        </goals>
-                        <configuration>
-                            <executable>./mvn_test_jdk8_only.sh</executable>
```
wing328: I think we need to use this (./mvn_test_jdk8_only.sh) to only test in Java8 for this particular project as the test environment (CircleCI) uses both OpenJDK 7 & 8 to run the tests.

mobreza: This was part of regenerating the code in 3466635bcc9514d4749dde45729420b602e0e3ea Was it manually inserted?

wing328: Right. I actually can think of a better way to avoid using shell script to run only with JDK8. I'll do it later after this PR.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java
wing328: @mobreza my understanding is that not all Java API client (e.g. feign) supports the `withXML` option. Please correct me if I'm wrong.

mobreza:
I’m not familiar with all clients, but they tend to use the common Java templates and abstract generator. 

Perhaps we should just say that libs with XML and JSON support allow it (e.g. gson doesn’t).

PR #6159, opened 2017-07-24
More flexible subclassing of ApiClient possible by s/private/protected/g

drd:
### PR checklist

- [X] Filed the PR against the correct branch: master for non-breaking changes and `3.0.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

I found myself in a situation where I needed to change the configured `Feature`s, but
because of all the private variables I was not able to do so. Perhaps this is a bit too
broad of a stroke, but I changed all fields and methods to `protected` instead of `private`.
In this way, future extensibility should be improved.

Also, to solve my particular problem, I added a new empty method called
`performAdditionalClientConfiguration` which will allow subclasses to add specific features,
or do anything else possible with a `ClientConfig`.


wing328:
@drd the enhancement looks reasonable to me. Thanks for your contribution.

Issue #4431, opened 2016-12-19
Swagger Editor C#2.0 client codegen -Authrization denied issue. 


HarprasadShukla:
Hi Team,

We are consuming web API in .net application so generated swagger C# 2.0 client code but missing to pass  Header details Authentication in c# generated code. Without Authentication API responding correctly but when applying UID/PWD in c# generated code it is returning authorization denied. Could you please provide some sample to authenticate/Authorize web API generated code by Swagger C# client code. Just we can pass UID/PWD in config file based on different environment hosted API?
In Swagger Editor during code generation JSON/YAML file not providing API URL and credential. We need to pass in config file as dependents upon environment hosted API.
Note- I do not have issue in code generation. When using generated code in C# Visual Studio and passing credential getting error –Authorization denied.
Do let me know in case further clarification required.

Regards,
Harprasad

jimschubert:
@HarprasadShukla I'm not sure I follow the problem fully.

To clarify, you're generating the `C# 2.0` client. You can hit unauthenticated endpoints without a problem, but authenticated endpoints fail?

How are you attempting to pass the auth when you use the client? Currently, the client supports Basic Auth, API Key, and partially supports OAuth 2.0. Basic Auth should bet set via `Configuration.Username` and `Configuration.Password`.

Some sample code would help me to understand the issue better.

wing328:
@HarprasadShukla closing this as there's no update from you.

Issue #4424, opened 2016-12-19
Swagger CodeGen C# 2.0 client Header Authentication/Authrization issue 


HarprasadShukla:
Hi Team,

We are consuming web API in .net application so generated swagger C# 2.0 client code but missing to pass  Header details Authentication in c# generated code. Without Authentication API responding correctly but when applying UID/PWD in c# generated code it is returning authorization denied. Could you please provide some sample to authenticate/Authorize web API generated code by Swagger C# client code. Just we can pass UID/PWD in config file based on different environment hosted API?
Do let me know in case further clarification required.

Regards,
Harprasad

HarprasadShukla:
It help me lot if redirect ticket to right team in case again routed differently. I am looking quick reply as we stuck in middle of the way.

HarprasadShukla:
Just want to add point- I am successfully able to generate code with JSON/YAML file. Issue is  when within C# code applying UID/PWD in unit test case. API responding Authrisation denied. 

We are not passing any header/Authentication info in YAML/JSON while generating code. Do we need to pass it before generating code? or later just only add in the C# code based on environment configuration.

HarprasadShukla:
Just want to add point- I am successfully able to generate code with JSON/YAML file. Issue is when within C# code applying UID/PWD in unit test case. API responding Authrisation denied.

We are not passing any header/Authentication info in YAML/JSON while generating code. Do we need to pass it before generating code? or later just only add in the C# code based on environment configuration.

Sorry mistakenly clicked closed button.... reopening again...

wing328:
@HarprasadShukla if your REST API supports HTTP basic authentication, then you should document it according to the OpenAPI/Swagger spec: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#basic-authentication-sample.

(I'm not sure if C# 2.0 generator has proper support for HTTP basic auth)

PR #6161, opened 2017-07-24
Csharp basic auth

normrider2001:
### PR checklist

### Description of the PR

Added basic auth to CSharp library as it was missing before.  Issue #4424 and Issue #4431 reported this.  I don't consider this a breaking change as it is just adding functionality that has already been documented.


wing328:
@normrider2001 thanks for the PR. The HTTP basic authentication support is handled in the API layer: https://github.com/normrider2001/swagger-codegen/blob/ae740d1a2fc049acbf9cf9bbf9103fb7f16feba5/modules/swagger-codegen/src/main/resources/csharp/api.mustache#L269-L275

In your swagger spec, do the endpoints have a proper security definition (http basic) defined?

normrider2001:
In my endpoint the security isn't properly defined.  I have to figure out how to get that generated in Swashbuckle since the swagger gets generated from the code.  I will close the PR since this is not necessary.  Thanks @wing328

wing328:
@normrider2001 no problem. Thanks again for contributing an enhancement (PR) to Swagger Codegen. Let us know if you need further assistance.

PR #6166, opened 2017-07-25
[Java][Client] Add Deprecated annotation

hisener:
### PR checklist

### Description of the PR

Add `@Deprecated` annotations to generated Java clients.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/api.mustache
@@ -52,6 +52,9 @@ public class {{classname}} {
    * @return {{returnType}}
    {{/returnType}}
    * @throws ApiException if fails to make API call
+   {{#isDeprecated}}
+   * @Deprecated
+   {{/isDeprecated}}
    */
   public {{#returnType}}{{{returnType}}} {{/returnType}}{{^returnType}}void {{/returnType}}{{operationId}}({{#allParams}}{{{dataType}}} {{paramName}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) throws ApiException {
     Object {{localVariablePrefix}}localVarPostBody = {{#bodyParam}}{{paramName}}{{/bodyParam}}{{^bodyParam}}null{{/bodyParam}};
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/api.mustache
@@ -49,6 +49,9 @@ public class {{classname}} {
    * @return {{returnType}}
    {{/returnType}}
    * @throws ApiException if fails to make API call
+   {{#isDeprecated}}
+   * @Deprecated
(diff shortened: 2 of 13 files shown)


wing328:
@hisener thanks for adding support for `deprecated` 👍

brian-byungsookim:
Is there a reason this PR only added the `@Deprecated` tag in the javadoc and not as an actual method annotation?

hisener:
`@Deprecated` is the actual method annotation. `@deprecated` is related with javadoc. I haven't added the javadoc annotation because `deprecated` field is `boolean` in swagger and does not contain any message.

https://docs.oracle.com/javase/tutorial/java/annotations/predefined.html

brian-byungsookim:
I understand the difference between `@Deprecated` and `@deprecated` and I know that the `deprecated` field in OAS 2.0 is a `boolean`. 

I'm confused as to what you mean by 'I haven't added the javadoc annotation' when the .mustache files changed in this PR all have the `@Deprecated` annotation in the javadoc, and none as actual method annotations. Could you clarify for me? (I'll leave comments on the changed files)

modules/swagger-codegen/src/main/resources/Java/api.mustache
```diff
@@ -52,6 +52,9 @@ public class {{classname}} {
    * @return {{returnType}}
    {{/returnType}}
    * @throws ApiException if fails to make API call
+   {{#isDeprecated}}
+   * @Deprecated
```
brian-byungsookim: @hisener Isn't this a javadoc comment and not an annotation because the line starts with `*`?

modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/api.mustache
```diff
@@ -49,6 +49,9 @@ public class {{classname}} {
    * @return {{returnType}}
    {{/returnType}}
    * @throws ApiException if fails to make API call
+   {{#isDeprecated}}
+   * @Deprecated
```
brian-byungsookim: Same as above

modules/swagger-codegen/src/main/resources/Java/libraries/resteasy/api.mustache
```diff
@@ -45,6 +45,9 @@ public class {{classname}} {
    * @param {{paramName}} {{description}}{{#required}} (required){{/required}}{{^required}} (optional{{#defaultValue}}, default to {{{.}}}{{/defaultValue}}){{/required}}{{/allParams}}{{#returnType}}
    * @return {{{returnType}}}{{/returnType}}
    * @throws ApiException if fails to make API call
+   {{#isDeprecated}}
+   * @Deprecated
```
brian-byungsookim: Same as above

hisener:
Yes, I see now. You are right.

brian-byungsookim:
Ah okay, thanks for clarifying! I've worked on some of the changes and just wanted to confirm, so I can probably just submit a PR soon.

PR #6174, opened 2017-07-25
[go][client] add http request host override

vdice:
### PR checklist

### Description of the PR

Adds ability to override the [http.request](https://golang.org/pkg/net/http/#Request).Host if `Host` is set on the api client's configuration.  This is necessary when using golang's http library to properly set a different host value on the request, as simply adding the appropriate header does not have the desired effect.  (See also https://github.com/golang/go/issues/7682 for backstory)


wing328:
@vdice thanks for the PR.

cc @antihax @guohuang @neilotoole

wing328:
@vdice the change and the corresponding test case look good to me.

PR #6175, opened 2017-07-25
[Java] Interpret consumes '*/*' as 'application/json'

lwander:
### PR checklist

I'm having trouble with the `./bin/java-petstore-all.sh` script since it's introducing a lot of changes unrelated to this change, even when I run against `upstream/master` with no changes staged. It seems like at some point changes to the templates were not checked in alongside their changes to the petstore examples.

### Description of the PR

Servers can describe their desired (consumed) content type for an operation as "*/*", which the current swagger API sends as the content type for any request against that operation. This is troublesome because "*/*" provides no info to the server as to what is being sent. Given that swagger already defaults to "application/json", this seems appropriate.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/ApiClient.mustache
@@ -519,10 +519,10 @@ public class ApiClient {
    *
    * @param contentTypes The Content-Type array to select from
    * @return The Content-Type header to use. If the given array is empty,
-   *   JSON will be used.
+   *   or matches "any", JSON will be used.
    */
   public String selectHeaderContentType(String[] contentTypes) {
-    if (contentTypes.length == 0) {
+    if (contentTypes.length == 0 || contentTypes[0].equals("*/*")) {
       return "application/json";
     }
     for (String contentType : contentTypes) {


wing328:
@lwander thanks for the PR. The change is for "jersey1.x" Java API client but I believe Kubernetes is using the default (okhttp-gson): https://github.com/kubernetes-client/java/blob/master/kubernetes/pom.xml#L198

I'll file a PR shortly to add the enhancement to other Java's ApiClient.mustache (some other generators such as Ruby, PHP also have such enhancements but we'll review again to ensure all generators have the same enhancement in place)

wing328:
FYI. Filed https://github.com/swagger-api/swagger-codegen/pull/6188 for other Java API clients.

lwander:
Thanks @wing328, I appreciate all the help

chlung:
@wing328 Do you know when v2.3.0 will be released.  thanks.

wing328:
@chlung we hope to release it by the end of Aug.

chlung:
@wing328  thanks.  If we would like to try v2.3.0 locally to build k8s-client java api, do you know the procedures/steps to do that?

wing328:
Please try the snapshot version: https://github.com/swagger-api/swagger-codegen#compatibility

chlung:
@wing328 thanks

Issue #6180, opened 2017-07-26
[Java-PlayFramework] Some non imported classes and non used imports


tzimisce012:
##### Description

In the ControllerImplementations of the Api interfaces ObjectMapper neither @Inject are not being imported.

Also, many other packages or classes are being imported when they are not being used.

##### Swagger-codegen version

2.3.0

##### Swagger declaration file content or url

Fix in newApiController.mustache in the resource folder for language JavaPlayframework.

PR #6185, opened 2017-07-26
Issue imports

tzimisce012:
### PR checklist

### Description of the PR

Fix issue #6180


wing328:
@tzimisce012 Please run `./bin/java-play-framework-petstore-server-all.sh` to update Petstore Java Play samples.

cc @JFCote

tzimisce012:
I already run the scripts and it generates the code under the folder samples. What should I do with it? Should I commit it?

wing328:
@tzimisce012 yes please commit the changes so that the CI (CircleCI) can verify the change.

tzimisce012:
i passed the ci/circle test but i haven't passed the tests for travis even though I passed them the first time. Are travis tests related with the examples of the pet store too?

JFCote:
@wing328 Will review the files tomorrow. Just came back from vacation. Thanks!

bin/samples/server/petstore/java-play-framework-controller-only/.swagger-codegen-ignore
```diff
@@ -0,0 +1,23 @@
+# Swagger Codegen Ignore
```
JFCote: This complete commit is wrong. The samples already exist in "samples", not in "bin/samples". You should revert it and generate is from the right place. Thanks!

JFCote:
General comment: I have an upcoming PR (not submitted yet) that fix most problem of bad import identified here. If you agree @tzimisce012 , I would like to submit it and add some fix in it based on this PR. The reason why I would use mine instead of yours is because it contains a LOT of other improvement related to Lint warning and it has been extensively tested for the last 3 months with a very large project with lots of use cases.  I should be able to submit my PR and include your fixes too within the end of the week. Let me know if that works for you.

Thanks.

tzimisce012:
yes, that is fine for me

modules/swagger-codegen/src/main/resources/JavaPlayFramework/newApi.mustache
```diff
@@ -8,7 +8,9 @@ import java.util.ArrayList;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import javax.inject.Inject;
-
+{{#isResponseFile}}
+import java.io.FileInputStream;
```
JFCote: this will not work because you need to be inside an operation `{{#operation}}` to be able to access this information, it's not a "class" information. To prevent FileInputStream to end up there, I will need to make this dynamic in the JavaPlayFramework codegen and it will not be handled in the mustache file.

modules/swagger-codegen/src/main/resources/JavaPlayFramework/newApi.mustache
```diff
@@ -3,11 +3,12 @@ package {{package}};
 {{#imports}}import {{import}};
 {{/imports}}
 
-import play.mvc.Http;
```
JFCote: Here too we can't remove this because if we have a `Http.MultipartFormData.FilePart` somewhere in the file, it will not work.

modules/swagger-codegen/src/main/resources/JavaPlayFramework/newApi.mustache
```diff
@@ -3,11 +3,12 @@ package {{package}};
...
 import java.util.List;
 import java.util.ArrayList;
-import java.util.HashMap;
```
JFCote: Here too we can't remove, if we have a collection which is a hash, it will not compile

modules/swagger-codegen/src/main/resources/JavaPlayFramework/newApi.mustache
```diff
@@ -3,11 +3,12 @@ package {{package}};
...
-import java.io.FileInputStream;
```
JFCote: Same thing here.
For all of these, we could go and make it dynamic by adding some logic in the JavaPlayFrameworkCodegen.java that deals with that but it would take a good amount of work and testing and I currently don't have the time.

JFCote:
@tzimisce012 @wing328 
Based on my review, I think we should discard (close) this PR.
I will soon submit a PR with reduced number of import/warning but it will be far from perfect. I will try to improve on this in the future but everything will need to be done in the JavaPlayFrameworkCodegen.java in the postProcessing.

PR #6194, opened 2017-07-27
Added all models to postProcessOperations

ToreJohansson:
Added allModels in postProcessOperations so operation code could be  based on model structure.
I am currently using this to make a decision if an operation has a common structure.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConfig.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConfig.java
@@ -151,7 +151,7 @@ public interface CodegenConfig {
 
     Map<String, Object> postProcessModels(Map<String, Object> objs);
 
-    Map<String, Object> postProcessOperations(Map<String, Object> objs);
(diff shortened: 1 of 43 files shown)


wing328:
@ToreJohansson thanks for the PR. Please elaborate more with a use case or spec on the change.

wing328:
@ToreJohansson I prefer to keep `postProcessOperations` as it's and add another function instead:
```
    Map<String, Object> postProcessOperations(Map<String, Object> objs, List<Object> allModels);
``` 
after postProcessOperations and postProcessModels.

ToreJohansson:
It is used to postprocess the operations utilizing extended data. Is it better to run an additional function right after the original one?


/T

ToreJohansson:
Swagger codegen has the model divided from the operations. In Model code you have no idea which operation(s) that model belongs to. The same in the operation code. You have no clue in the operation code which models the operation might have. So if you need to classify the operation depending on the content of the model you can not do that without have the model submitted.
Currently I have a concept of common model validator to trigger from the operation with an annotation. So I need to know if the generated model will match at the operation. A simple check but I can not do it without the model...

wing328:
@ToreJohansson thanks for the work, which has been merged into master.

PR #6211, opened 2017-07-30
[Python] Fix missing authentication settings

cliffano:
### PR checklist

- [v ] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [v ] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [v ] Filed the PR against the correct branch: master for non-breaking changes and `3.0.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

This PR attempts to fix the missing authentication settings, i.e. the api.mustache template initialises `authentication_settings` using `authModes`, however, `authModes` is never part of `operation`.
My first question would be whether it is really the intention to initialise authentication settings using authModes?

Another question I have is about the expected behaviour. I noticed a difference between the behaviour in javascript lang vs python lang. In python, authentication wouldn't inherit security definitions without this fix, hence authentication would've failed. In javascript, even though authentication settings is not set properly due to this authModes issue, the code inherits a default object that already has the default from security definitions.
It's possible that this problem affects more langs.

One technical issue I have is that the template wouldn't process anything if the item key is named `authModes`, however, renaming it to something else works, hence I renamed it to `_authModes` just to get going. Any idea why? Is the word reserved by any chance?


wing328:
@cliffano thanks for the PR. It seems to me it's working fine and here is an example:

https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/python/petstore_api/apis/fake_classname_tags_123_api.py#L117

cliffano:
@wing328 That looks different to what I got pre-patch: https://github.com/swagger-api/swagger-codegen/pull/6211/commits/0d2e55cbb49e40dcdf60b901c39e3c1e8d45f4e2#diff-2f383154dfab39d9250d3786a016f8b6L117 where auth_settings is an empty array, and after the patch it lists the auths declared in security definitions.

I'll try latest master.

cliffano:
@wing328 The sample you showed me actually made me think that the Python generator is doing the right thing, perhaps it's the JavaScript generator that's doing the wrong thing. I was under the impression of the reverse, hence I submitted this PR trying to make Python generator behaves more like the JavaScript generator (declaring all authentication methods in `securityDefinitions`).

Looking at the sample generated Petstore, the fundamental difference is this:
- For Python: test method must specify `auth_settings` as per the example line you posted in previous comment https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/python/petstore_api/apis/fake_classname_tags_123_api.py#L117
- For JavaScript: test function does not specify `authNames` https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/javascript/src/api/Fake_classname_tags123Api.js#L85 , but authentication works because the authentication methods described in `securityDefinitions` is declared at API client level and included in all API calls https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/javascript/src/ApiClient.js#L55 .

So for a given spec file that only declares `securityDefinitions` at the top level, but doesn't declare any `security` at operation level, authentication works for JavaScript client, but not for Python client.

My question is, what is the right/expected behaviour?

wing328:
@cliffano Python API client and JS API client (ES5) should be doing the right thing while JS API clinet (ES6) needs some improvements and there's a PR for it: https://github.com/swagger-api/swagger-codegen/pull/6269

PR #6228, opened 2017-08-03
Added examples for body parameter

ToreJohansson:
Added examples for Body parameter - the reason for this it to be able to auto generate test cases with valid request content. The other option is to manually add x-examples in the specification and use those as input.


modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
```diff
@@ -2274,6 +2274,10 @@ public CodegenOperation fromOperation(String path,
                 } else if (param instanceof BodyParameter) {
                     bodyParam = p;
                     bodyParams.add(p.copy());
+                    if(definitions != null)
+                    {
```
wing328: Please update to `if (definitions != null) {` instead to conform to Java style guide.

wing328: Fixed via d6c9522

PR #6230, opened 2017-08-03
[Java] Fix okhttp-gson datetime converter compilation

lwander:
### PR checklist

### Description of the PR

This commit https://github.com/swagger-api/swagger-codegen/commit/f1e237f76fc7ee8c731e77569c5628da5dff85d8 seemed to have introduce unprompted new properties to the `DateTimeTypeAdaptor`. While this technically changes the intended behavior of having two separate date time formatters for parsing and printing, I can't see reverting this breaking anything since the generated code in its current state cannot compile.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/JSON.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/JSON.mustache
@@ -13,6 +13,7 @@ import com.google.gson.stream.JsonWriter;
 import org.joda.time.DateTime;
 import org.joda.time.LocalDate;
 import org.joda.time.format.DateTimeFormatter;
+import org.joda.time.format.DateTimeFormatterBuilder;
 import org.joda.time.format.ISODateTimeFormat;
 {{/joda}}
 {{#threetenbp}}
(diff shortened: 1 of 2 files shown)


cbornet:
This is not the right way to fix. The 2 formatters are needed as per #4473. It conflicted with another PR that enabled configuring the formatter at runtime.

lwander:
Thanks @cbornet that's the context I needed. I'll update with a proper fix tomorrow.

lwander:
PTAL @cbornet

cbornet:
Removing the other PR is not the good way to solve this either !
After thinking of it, I think there's a better way to solve #4473 using a single formatter which would be by default:
```java
new DateTimeFormatterBuilder().append(ISODateTimeFormat.dateTime(), ISODateTimeFormat.dateOptionalTimeParser()).toFormatter();
```
@lwander can you test this ?

lwander:
Ah I misunderstood, I thought you wanted the formatter not to be configurable at runtime. Will give that a shot

lwander:
PTAL @cbornet

cbornet:
LGTM. Can you also update the samples ?

lwander:
I was thinking of updating the samples in a separate PR because it seems they haven't been updated in some time. I'll file a separate PR to update the java petstore samples (minus this change), then once that's merged update the samples to just include this change.

lwander:
https://github.com/swagger-api/swagger-codegen/pull/6256

cbornet:
And if you could also do it for retrofit (https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/Java/libraries/retrofit2/JSON.mustache#L94) that'd be awesome !

lwander:
Sure thing

PR #6188, opened 2017-07-26
[Java] MIME type "*/*" default to JSON

wing328:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: master for non-breaking changes and `3.0.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

Set MIME type "*/*" default to JSON

PR #6232, opened 2017-08-03
[Java] okhttp-gson content-type fix for */*

lwander:
### PR checklist

### Description of the PR

Finishes work started in #6188 and #6175

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache
@@ -608,11 +608,11 @@ public class ApiClient {
      *
      * @param contentTypes The Content-Type array to select from
      * @return The Content-Type header to use. If the given array is empty,
-     *   JSON will be used.
+     *   or matches "any", JSON will be used.
      */
     public String selectHeaderContentType(String[] contentTypes) {
-        if (contentTypes.length == 0) {
-            return "application/json";
+        if (contentTypes.length == 0 || contentTypes[0].equals("*/*")) {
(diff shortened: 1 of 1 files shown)


modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache
```diff
@@ -608,11 +608,11 @@ public class ApiClient {
      *
      * @param contentTypes The Content-Type array to select from
      * @return The Content-Type header to use. If the given array is empty,
-     *   JSON will be used.
+     *   or matches "any", JSON will be used.
      */
     public String selectHeaderContentType(String[] contentTypes) {
-        if (contentTypes.length == 0) {
-            return "application/json";
+        if (contentTypes.length == 0 || contentTypes[0].equals("*/*")) {
```
wing328: @lwander we've actually implemented the check in `isJsonMime`: https://github.com/lwander/swagger-codegen/blob/a73ee037181933ee9040cc90383f2d4611a49c85/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache#L580

lwander: Right, once `isJsonMime` returns true, this function will return the value `*/*`, which is not a useful content-type header.

wing328: You're right. We'll revise other generators to apply similar fix.

wing328:
FYI. Java Petstore sample updated via 646c2bc

Issue #3500, opened 2016-08-02
[typescript-angular2] enum with numeric name


damienpontifex:
##### Description

When generating with the typescript-angular2 language, an enum is generated with an integer name. The typescript compiler error is:

> error TS2452: An enum member cannot have a numeric name.

I am generating this from a C# Web API that on a bad request returns a [ModelStateDictionary](http://aspnetsource.azurewebsites.net/#Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelStateDictionary.cs,c257c6b46a5ce569) which has a `validationState` of type [ModelValidationState](http://aspnetsource.azurewebsites.net/#Microsoft.AspNetCore.Mvc.Abstractions/ModelBinding/ModelValidationState.cs,686ecdf00ba040d5).

The generated Typescript enum is:

``` typescript
export namespace ModelStateEntry {

    export enum ValidationStateEnum { 
        0 = <any> '0',
        1 = <any> '1',
        2 = <any> '2',
        3 = <any> '3',
    }
}
```
##### Swagger-codegen version

2.2.0
##### Swagger declaration file content or url

The swagger definition for this from the JSON content is:

``` json
"ModelStateEntry": {
    "type": "object",
    "properties": {
        "rawValue": {
            "type": "object"
        },
        "attemptedValue": {
            "type": "string"
        },
        "errors": {
            "type": "array",
            "items": {
                "$ref": "#/definitions/ModelError"
            },
            "readOnly": true
        },
        "validationState": {
            "format": "int32",
            "enum": [
                0,
                1,
                2,
                3
            ],
            "type": "integer"
        },
        "isContainerNode": {
            "type": "boolean",
            "readOnly": true
        },
        "children": {
            "type": "array",
            "items": {
                "$ref": "#/definitions/ModelStateEntry"
            },
            "readOnly": true
        }
    }
}
```

##### Steps to reproduce

Run `swagger-codegen generate` against a web api that returns a model result of `ModelStateDictionary` (C# type)

##### Related issues

Using Swashbuckle configuration of `DescribeAllEnumsAsStrings()` solves the issue, but we may be able to do something here also to prevent the case where this occurs.

##### Suggest a Fix

The [typescript-angular2/model.mustache](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/typescript-angular2/model.mustache) contains the declaration for the enum, but I'm not sure of the proper change to make here.

[...]

Issue #5146, opened 2017-03-21
[TypeScript] String enums allow incorrect widening


kourge:
##### Description

`swagger-codegen` [emits TypeScript enums](https://github.com/swagger-api/swagger-codegen/blob/55443da/modules/swagger-codegen/src/main/resources/typescript-node/api.mustache#L41) in the form of:

```ts
export namespace Foo {
  export enum Bar {
    baz = <any> "baz",
    spam = <any> "spam"
  }
}
```

This is a hack to support string-based enums, since TypeScript's built-in support for enums only support numbers. However, this workaround is extremely dangerous, since it allows incorrectly widening the enum value to the `number` type even though it is a string:

```ts
const x: number = Foo.Bar.baz; // this compiles, even though it shouldn't
const y: string = Foo.Bar.baz; // this does not compile, even though it should
```

##### Swagger-codegen version

As late as 55443da

##### Swagger declaration file content or url

Any declaration file containing a string enum

##### Command line used for generation

Any `swagger-codegen` invocation that generates TypeScript

##### Suggest a Fix

According to a TypeScript dev, since TS 2.1, you can [safely write string-based enums without the builtin `enum` keyword](https://github.com/Microsoft/TypeScript/issues/3192#issuecomment-261720275). This approach has minimal overhead and exploits `keyof`, lookup types, and type inference.

dartzki:
typescript 2.4 will support string enums: https://github.com/Microsoft/TypeScript/pull/15486

xmeng1:
Why the typescript enum type generate with `<any>` which make it duplicate 
```
export enum xxxTypeEnum {
        AAA = <any> 'AAA',
        BBBB = <any> 'BB_BB',
        CCC = <any> 'CCC'
    }
```

And when I use `for (let obj in xxxTypeEnum)`, there will be 4 items, and the `BBBB` will be duplicated with  `BBBB` and `BB_BB`

kourge:
With TypeScript 2.4, you should not need the `<any>` casts when writing a string enum. This gives you the correctly emitted object without duplicates.

gpassero:
Any update on this topic? Removing `<any>` from enum values generated by codegen would be very helpful. So far I have manually removed unnecessary `<any>` casts every time we regenerate code using codegen to keep TypeScript type checks working correctly. It would be nice to not have to change this manually.

Kierchon:
@gpassero Did you find a solution for this?

gpassero:
[...]

PR #6233, opened 2017-08-03
represent swagger enums as union of literal types

### LINKED ISSUE #6206: Typescript: Why not use string-literal types to represent enums of type string?
##### Description

The typescript-angular2 generator currently emits enums in the form 

```
    export enum StatusEnum {
        Open = <any> 'open',
        InProgress = <any> 'inProgress',
        Resolved = <any> 'resolved',
        Closed = <any> 'closed'
    }
```

For enums of type string the translation 

```
    type StatusEnum = "open" | "inProgress" | "resolved" | "closed";
```

would be more type-safe (witness #5146), and more flexible to use. For instance, I could do:

```
    const badgeType: {[K: StatusEnum]: string} = {
        open: "danger",
        inProgress: "warning",
        resolved: "info",
        closed: "success",
    }
```

rather than

```
    const badgeType = <any>{}; // no type safety, because I can not enumerate the enum values
    badgeType[StatusEnum.Open] = "danger";
    badgeType[StatusEnum.InProgress] = "warning";
    badgeType[StatusEnum.Resolved] = "info";
    badgeType[StatusEnum.Closed] = "success";
```

as well as directly refer to enum constants in an angular template:

```
<div [ngSwitch]="issue.state">
  <div *ngSwitchCase="'open'"> 
```

rather than an unnecessary indirection in the template

```
<div [ngSwitch]="issue.state">
  <div *ngSwitchCase="StatusEnum.Open"> 
```
over the component
```
import { StatusEnum} from ...

export class MyComponent {
  StatusEnum = StatusEnum;
}
```
to the imported enum.

##### Swagger-codegen version

master

Generate enums of type string as union of string literals, leaving enums of numeric type as they are (they need special treatment anyway to fix #3500).

bedag-moo:
### PR checklist

### Description of the PR

This PR changes the angular-typescript generator to emit a union of literals for swagger enums, e.g.:

    export type StatusEnum = 'open' | 'inProgress' | 'resolved' | 'closed';

or
    
    export type ResponseCode = 200 | 404;

enabling their easy use in angular templates, structural subtyping
among enums (in particular, different instances of the same enum
are now mutually assignable), improving type safety by preventing
incorrect widening, and permitting numeric enum values
(albeit without descriptive names)

Fixes #6206, #5146, #3500

### DIFF
--- a/modules/swagger-codegen/src/main/resources/typescript-angular/modelEnum.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular/modelEnum.mustache
@@ -1,7 +1,1 @@
-export enum {{classname}} {
-{{#allowableValues}}
-{{#enumVars}}
-    {{{name}}} = <any> {{{value}}}{{^-last}},{{/-last}}
-{{/enumVars}}
-{{/allowableValues}}
-}
+export type {{classname}} = {{#allowableValues}}{{#enumVars}}{{{value}}}{{^-last}} | {{/-last}}{{/enumVars}}{{/allowableValues}};
--- a/modules/swagger-codegen/src/main/resources/typescript-angular/modelEnum.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular/modelEnum.mustache
@@ -1,1 +1,2 @@
 export type {{classname}} = {{#allowableValues}}{{#enumVars}}{{{value}}}{{^-last}} | {{/-last}}{{/enumVars}}{{/allowableValues}};
+
--- a/modules/swagger-codegen/src/main/resources/typescript-angular/modelGeneric.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular/modelGeneric.mustache
@@ -16,13 +16,7 @@ export interface {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{
 export namespace {{classname}} {
 {{#vars}}
     {{#isEnum}}
-    export enum {{enumName}} {
-    {{#allowableValues}}
-        {{#enumVars}}
-        {{{name}}} = <any> {{{value}}}{{^-last}},{{/-last}}
-        {{/enumVars}}
-    {{/allowableValues}}
-    }
+    export type {{enumName}} = {{#allowableValues}}{{#enumVars}}{{{value}}}{{^-last}} | {{/-last}}{{/enumVars}}{{/allowableValues}};
     {{/isEnum}}
 {{/vars}}
 }{{/hasEnums}}
--- a/modules/swagger-codegen/src/main/resources/typescript-angular/modelGeneric.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular/modelGeneric.mustache
@@ -19,4 +19,5 @@ export namespace {{classname}} {
     export type {{enumName}} = {{#allowableValues}}{{#enumVars}}{{{value}}}{{^-last}} | {{/-last}}{{/enumVars}}{{/allowableValues}};
     {{/isEnum}}
 {{/vars}}
-}{{/hasEnums}}
+}{{/hasEnums}}i
+
--- a/modules/swagger-codegen/src/main/resources/typescript-angular/modelGeneric.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular/modelGeneric.mustache
@@ -19,5 +19,5 @@ export namespace {{classname}} {
     export type {{enumName}} = {{#allowableValues}}{{#enumVars}}{{{value}}}{{^-last}} | {{/-last}}{{/enumVars}}{{/allowableValues}};
     {{/isEnum}}
 {{/vars}}
-}{{/hasEnums}}i
+}{{/hasEnums}}
 
--- a/samples/client/petstore/typescript-angular-v2/default/model/order.ts
+++ b/samples/client/petstore/typescript-angular-v2/default/model/order.ts
@@ -33,9 +33,5 @@ export interface Order {
 
 }
 export namespace Order {
-    export enum StatusEnum {
-        Placed = <any> 'placed',
-        Approved = <any> 'approved',
-        Delivered = <any> 'delivered'
-    }
+    export type StatusEnum = 'placed' | 'approved' | 'delivered';
 }
--- a/samples/client/petstore/typescript-angular-v2/default/model/pet.ts
+++ b/samples/client/petstore/typescript-angular-v2/default/model/pet.ts
@@ -35,9 +35,5 @@ export interface Pet {
 
 }
 export namespace Pet {
-    export enum StatusEnum {
-        Available = <any> 'available',
-        Pending = <any> 'pending',
-        Sold = <any> 'sold'
-    }
+    export type StatusEnum = 'available' | 'pending' | 'sold';
 }
--- a/samples/client/petstore/typescript-angular-v2/npm/model/order.ts
+++ b/samples/client/petstore/typescript-angular-v2/npm/model/order.ts
@@ -33,9 +33,5 @@ export interface Order {
 
 }
 export namespace Order {
-    export enum StatusEnum {
-        Placed = <any> 'placed',
-        Approved = <any> 'approved',
-        Delivered = <any> 'delivered'
-    }
+    export type StatusEnum = 'placed' | 'approved' | 'delivered';
 }
--- a/samples/client/petstore/typescript-angular-v2/npm/model/pet.ts
+++ b/samples/client/petstore/typescript-angular-v2/npm/model/pet.ts
@@ -35,9 +35,5 @@ export interface Pet {
 
 }
 export namespace Pet {
-    export enum StatusEnum {
-        Available = <any> 'available',
-        Pending = <any> 'pending',
-        Sold = <any> 'sold'
-    }
+    export type StatusEnum = 'available' | 'pending' | 'sold';
 }
--- a/samples/client/petstore/typescript-angular-v2/with-interfaces/model/Order.ts
+++ b/samples/client/petstore/typescript-angular-v2/with-interfaces/model/Order.ts
@@ -33,9 +33,5 @@ export interface Order {
 
 }
 export namespace Order {
-    export enum StatusEnum {
-        Placed = <any> 'placed',
-        Approved = <any> 'approved',
-        Delivered = <any> 'delivered'
-    }
+    export type StatusEnum = 'placed' | 'approved' | 'delivered';
 }
--- a/samples/client/petstore/typescript-angular-v2/with-interfaces/model/Pet.ts
+++ b/samples/client/petstore/typescript-angular-v2/with-interfaces/model/Pet.ts
@@ -35,9 +35,5 @@ export interface Pet {
 
 }
 export namespace Pet {
-    export enum StatusEnum {
-        Available = <any> 'available',
-        Pending = <any> 'pending',
-        Sold = <any> 'sold'
-    }
+    export type StatusEnum = 'available' | 'pending' | 'sold';
 }
--- a/samples/client/petstore/typescript-angular-v4/npm/model/order.ts
+++ b/samples/client/petstore/typescript-angular-v4/npm/model/order.ts
@@ -33,9 +33,5 @@ export interface Order {
 
 }
 export namespace Order {
-    export enum StatusEnum {
-        Placed = <any> 'placed',
-        Approved = <any> 'approved',
-        Delivered = <any> 'delivered'
-    }
+    export type StatusEnum = 'placed' | 'approved' | 'delivered';
 }
--- a/samples/client/petstore/typescript-angular-v4/npm/model/pet.ts
+++ b/samples/client/petstore/typescript-angular-v4/npm/model/pet.ts
@@ -35,9 +35,5 @@ export interface Pet {
 
 }
 export namespace Pet {
-    export enum StatusEnum {
-        Available = <any> 'available',
-        Pending = <any> 'pending',
-        Sold = <any> 'sold'
-    }
+    export type StatusEnum = 'available' | 'pending' | 'sold';
 }


bedag-moo:
The Travis failure seems to be an infrastructure issue unrelated to this PR:

> To increase the limit use GitHub authentication.
> See: https://github.com/DefinitelyTyped/tsd#tsdrc

bobvanderlinden:
This change is very much appreciated. I'd like to copy my JSON example for the schema into typescript and have it checked. At the moment this isn't possible because all enum values need to be changed to their reference even though it's structurally the same and the server would send that exact JSON as well.

In addition, this would also solve problems with numerical (or other) enums.

That said, this PR at the moment contains a number of unrelated changes, which might lessen motivation to merge for the Swagger maintainers.

samples/client/petstore/typescript-angular2/default/api/store.service.ts
```diff
@@ -73,7 +73,7 @@ export class StoreService {
     }
 
     /**
-     * For valid response try integer IDs with value &lt; 1000. Anything above 1000 or nonintegers will generate API errors
```
bobvanderlinden: This change seems unrelated.

samples/client/petstore/typescript-angular2/default/api/store.service.ts
```diff
@@ -104,7 +104,7 @@ export class StoreService {
     }
 
     /**
-     * For valid response try integer IDs with value &lt;&#x3D; 5 or &gt; 10. Other values will generated exceptions
+     * For valid response try integer IDs with value <= 5 or > 10. Other values will generated exceptions
```
bobvanderlinden: This change seems unrelated.

samples/client/petstore/typescript-angular2/npm/api/pet.service.ts
```diff
@@ -74,8 +74,8 @@ export class PetService {
     }
 
     /**
-     * Add a new pet to the store
      * 
+     * @summary Add a new pet to the store
```
bobvanderlinden: This change seems unrelated.

samples/client/petstore/typescript-angular2/npm/api/pet.service.ts
```diff
@@ -90,8 +90,8 @@ export class PetService {
     }
 
     /**
-     * Deletes a pet
      * 
+     * @summary Deletes a pet
```
bobvanderlinden: This change seems unrelated.

samples/client/petstore/typescript-angular2/npm/api/pet.service.ts
```diff
@@ -107,8 +107,8 @@ export class PetService {
     }
 
     /**
-     * Finds Pets by status
      * Multiple status values can be provided with comma separated strings
+     * @summary Finds Pets by status
```
bobvanderlinden: This change seems unrelated.

samples/client/petstore/typescript-angular2/npm/api/pet.service.ts
```diff
@@ -123,8 +123,8 @@ export class PetService {
     }
 
     /**
-     * Finds Pets by tags
      * Multiple tags can be provided with comma separated strings. Use tag1, tag2, tag3 for testing.
+     * @summary Finds Pets by tags
```
bobvanderlinden: This change seems unrelated.

samples/client/petstore/typescript-angular2/npm/api/pet.service.ts
```diff
@@ -139,8 +139,8 @@ export class PetService {
     }
 
     /**
-     * Find pet by ID
      * Returns a single pet
+     * @summary Find pet by ID
```
bobvanderlinden: This change seems unrelated.

samples/client/petstore/typescript-angular2/npm/api/pet.service.ts
```diff
@@ -155,8 +155,8 @@ export class PetService {
     }
 
     /**
-     * Update an existing pet
      * 
+     * @summary Update an existing pet
```
bobvanderlinden: This change seems unrelated.

samples/client/petstore/typescript-angular2/npm/api/pet.service.ts
```diff
@@ -171,8 +171,8 @@ export class PetService {
     }
 
     /**
-     * Updates a pet in the store with form data
      * 
+     * @summary Updates a pet in the store with form data
```
bobvanderlinden: This change seems unrelated.

samples/client/petstore/typescript-angular2/npm/api/pet.service.ts
```diff
@@ -189,8 +189,8 @@ export class PetService {
     }
 
     /**
-     * uploads an image
      * 
+     * @summary uploads an image
```
bobvanderlinden: This change seems unrelated.

samples/client/petstore/typescript-angular2/npm/api/store.service.ts
```diff
@@ -73,8 +73,8 @@ export class StoreService {
     }
 
     /**
-     * Delete purchase order by ID
-     * For valid response try integer IDs with value &lt; 1000. Anything above 1000 or nonintegers will generate API errors
+     * For valid response try integer IDs with value < 1000. Anything above 1000 or nonintegers will generate API errors
+     * @summary Delete purchase order by ID
```
bobvanderlinden: This change seems unrelated.

samples/client/petstore/typescript-angular2/npm/api/store.service.ts
```diff
@@ -89,8 +89,8 @@ export class StoreService {
     }
 
     /**
-     * Returns pet inventories by status
      * Returns a map of status codes to quantities
+     * @summary Returns pet inventories by status
```
bobvanderlinden: This change seems unrelated.

samples/client/petstore/typescript-angular2/npm/api/store.service.ts
```diff
@@ -104,8 +104,8 @@ export class StoreService {
     }
 
     /**
-     * Find purchase order by ID
-     * For valid response try integer IDs with value &lt;&#x3D; 5 or &gt; 10. Other values will generated exceptions
+     * For valid response try integer IDs with value <= 5 or > 10. Other values will generated exceptions
+     * @summary Find purchase order by ID
```
bobvanderlinden: This change seems unrelated.

samples/client/petstore/typescript-angular2/npm/api/store.service.ts
```diff
@@ -120,8 +120,8 @@ export class StoreService {
     }
 
     /**
-     * Place an order for a pet
      * 
+     * @summary Place an order for a pet
```
bobvanderlinden: This change seems unrelated.

samples/client/petstore/typescript-angular4/npm/configuration.ts
```diff
@@ -2,18 +2,18 @@ export interface ConfigurationParameters {
     apiKeys?: {[ key: string ]: string};
     username?: string;
     password?: string;
-    accessToken?: string;
+    accessToken?: string | (() => string);
```
bobvanderlinden: This change seems unrelated.

samples/client/petstore/typescript-angular4/npm/configuration.ts
```diff
@@ -2,18 +2,18 @@ export interface ConfigurationParameters {
...
-    password: string;
-    accessToken: string | (() => string);
-    basePath: string;
-    withCredentials: boolean;
+    apiKeys?: {[ key: string ]: string};
+    username?: string;
+    password?: string;
+    accessToken?: string | (() => string);
+    basePath?: string;
+    withCredentials?: boolean;
```
bobvanderlinden: This change seems unrelated.

samples/client/petstore/typescript-angular4/npm/package.json
```diff
@@ -35,6 +35,6 @@
     "typescript": "^2.1.5"
   },
   "publishConfig": {
-    "registry": "https://skimdb.npmjs.com/registry"
+    "registry":"https://skimdb.npmjs.com/registry"
```
bobvanderlinden: This change seems unrelated.

bobvanderlinden:
Hmm, I just saw that this PR was made against `2.3.0`, but this is actually a breaking change. Existing implementations that rely on the TypeScript codegen will break. I'm guessing you need to make the PR against `3.0.0`.

bedag-moo:
It's a breaking change for an unreleased feature, so I thought the earlier release was fine as well. (TypeScript support in 2.2.x is pretty much nonexistent AFAIK), but of course I'd be happy to rebase if a maintainer asked.

As you can see from the commit history, the "unrelated changes" were made by the shell scripts the PR checklist instructed me to run. If the maintainers don't want that, they can cherry-pick my first commit instead (and update the PR checklist).

bobvanderlinden:
You're absolutely right. Sorry for the confusion. Tested it locally on an existing project and it is working wonderfully. Hopefully this PR will be merged soon.

bedag-moo:
@wing328: Hm ... why did you change the base branch? This PR builds upon other work in the 2.3.0 branch that (as far as I know) has not yet been merged to master. Do you mean to tell me to rebase on master? 

As an aside, is there anything I can do to accelerate the review process?

wing328:
@bedag-moo as explained in another thread, 2.3.0 is now the master after we released stable version v2.2.3

bedag-moo:
I manually merged my changes (a rebase didn't work as the files had been moved), and regenerated the samples.

bobvanderlinden:
@bedag-moo Could you give this another look?

bedag-moo:
@bobvanderlinden: *What* do you want me to look at?

bobvanderlinden:
Sorry, meant to mention @wing328 . That what I get replying from mobile ;)

@wing328 could *you* give this another look?

wing328:
cc cc @TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx for review.

modules/swagger-codegen/src/main/resources/typescript-angular/modelEnum.mustache
```diff
@@ -1,7 +1 @@
-export enum {{classname}} {
-{{#allowableValues}}
-{{#enumVars}}
-    {{{name}}} = <any> {{{value}}}{{^-last}},{{/-last}}
-{{/enumVars}}
-{{/allowableValues}}
-}
+export type {{classname}} = {{#allowableValues}}{{#enumVars}}{{{value}}}{{^-last}} | {{/-last}}{{/enumVars}}{{/allowableValues}};
```
sebastianhaas: Please don't remove the newline in the end. (https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline)

modules/swagger-codegen/src/main/resources/typescript-angular/modelGeneric.mustache
```diff
@@ -16,13 +16,7 @@ export interface {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{
...
-    {{#allowableValues}}
-        {{#enumVars}}
-        {{{name}}} = <any> {{{value}}}{{^-last}},{{/-last}}
-        {{/enumVars}}
-    {{/allowableValues}}
-    }
+    export type {{enumName}} = {{#allowableValues}}{{#enumVars}}{{{value}}}{{^-last}} | {{/-last}}{{/enumVars}}{{/allowableValues}};
     {{/isEnum}}
 {{/vars}}
 }{{/hasEnums}}
```
sebastianhaas: See above.

modules/swagger-codegen/src/main/resources/typescript-angular/modelGeneric.mustache
```diff
@@ -16,13 +16,8 @@ export interface {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{
...
-        {{{name}}} = <any> {{{value}}}{{^-last}},{{/-last}}
-        {{/enumVars}}
-    {{/allowableValues}}
-    }
+    export type {{enumName}} = {{#allowableValues}}{{#enumVars}}{{{value}}}{{^-last}} | {{/-last}}{{/enumVars}}{{/allowableValues}};
     {{/isEnum}}
 {{/vars}}
-}{{/hasEnums}}
\ No newline at end of file
+}{{/hasEnums}}i
```
sebastianhaas: I'm afraid you accidentally inserted an `i` there 🙈

bedag-moo: Oops, indeed. Fixed.

wing328:
Tests passed via https://travis-ci.org/swagger-api/swagger-codegen/builds/284619142

wing328:
@sebastianhaas thanks for reviewing the change.

@bedag-moo thanks for the PR.

topce:
Hello @wing328 @sebastianhaas @bedag-moo 
Not sure that this is good decisions since TS 2.4 support string enums
and this change  brake existing code if somebody use enums before 
links:
[Enums](https://www.typescriptlang.org/docs/handbook/enums.html) 
[string enums](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#string-enums)

topce:
Hi  guys  
@bedag-moo 
@wing328 
@sebastianhaas 

I do not want to start  debate about this issue
just to argument my opinion ;-) 

I think it solves some problems because string enums before in code generated were cast as any
since version 2.4 of TypeScript it is not needed to be case anymore

`export type Status = 'open' | 'inProgress' | 'resolved' | 'closed';
export type ResponseCode = 200 | 404;
enum ResponseCodeEnum {
    Success = 200,
    NotFound = 404
}
enum StatusEnum {
    Open = 'open',
    inProgress ='inProgress',
    Resolved = 'resolved',
    Closed = 'closed'
}
`
So in my opinion there is not benefits of using type instead of enum:
For types with number you use enum with number 
for types with string you use enum with string as base type
We do not break compatibility when angular 5 is released 
if they support TS 2.4 or later we generate string enums and we do not brake code 

So string enums (not cast to any ) in my opinion solve all problems mention in this PR.
Also for example angular-cli support generation of enums but not types
so angular  users could be more familiar with enums
C# and Java users as well ;-)

Also enums are more informative in your example we give number 200
name ResponseCodeEnum.Success so in my opinion code is more readable

But of course different people different taste.

Maybe to add option how to generate enums 
by default as enum and if some option is specified as union string literal types ?

Just suggestion.

Good work!

bedag-moo:
Removing the cast to <any> does not resolve #6206.

topce:
@bedag-moo 
why not ?

bedag-moo:
Because the typescript compiler gives compilation errors when you try to do the stuff I describe there. In particular, check out the following code:

```

enum IssueStatusEnum {
  Open = 'open',
  InProgress = 'inProgress',
  Resolved = 'resolved',
  Closed = 'closed',
}

// used in a different place in the api, but conceptually the same enum
enum TransitionTargetEnum {
  Open = 'open',
  InProgress = 'inProgress',
  Resolved = 'resolved',
  Closed = 'closed',
}

const badgeType: {[K: IssueStatusEnum]: string} = {
  Open: "danger",
  InProgress: "warning",
  Resolved: "info",
  Closed: "success",
}

let t: TransitionTargetEnum;
t = IssueStatusEnum.Open;
```

which gives

```
Error:(41, 20) TS1023:An index signature parameter type must be 'string' or 'number'.
Error:(49, 1) TS2322:Type 'IssueStatusEnum.Open' is not assignable to type 'TransitionTargetEnum'.
```

topce:
@bedag-moo 
Can you try to compile with TypeScript version 2.4 or later?

topce:
Can you try in playground ?
and send me link 
[https://www.typescriptlang.org/play/](https://www.typescriptlang.org/play/)

bedag-moo:
I did compile with Typescript 2.4, and if you copy the above code to the playground you can see the errors yourself.

topce:
@bedag-moo 
code bellow compiles 
`enum IssueStatusEnum {
  Open = 'open',
  InProgress = 'inProgress',
  Resolved = 'resolved',
  Closed = 'closed',
}

// used in a different place in the api, but conceptually the same enum
enum TransitionTargetEnum {
  Open = 'open',
  InProgress = 'inProgress',
  Resolved = 'resolved',
  Closed = 'closed',
}

const badgeType: {[K: IssueStatusEnum]: string} = {
  Open: "danger",
  InProgress: "warning",
  Resolved: "info",
  Closed: "success",
}

let t: TransitionTargetEnum | IssueStatusEnum;
t = IssueStatusEnum.Open;`

topce:
@bedag-moo 
`let t: TransitionTargetEnum | IssueStatusEnum;`

topce:
@bedag-moo @wing328 @sebastianhaas 
On positive side of this pull request is that generated js code 
is more efficient !
So in my opinion this could be not default option in code generation

topce:
@bedag-moo 
const could be replace by function 
that take enum parametar and return string

bedag-moo:
... and how would you implement that function without sacrificing clarity or type safety?

topce:
switch case by enum or union of enums 
not sure why you have two enums with different name !?
maybe there is some more elegant solution should not be too difficult

topce:
@bedag-moo 
for example 
``` TypeScript
`function getBadgeType1(issueStatusEnum: IssueStatusEnum): string {
  switch (issueStatusEnum) {
    case IssueStatusEnum.Open: {
      return "danger";
    }
    case IssueStatusEnum.InProgress: {
      return "warning";
    }
    case IssueStatusEnum.Resolved: {
      return "info";
    }
    case IssueStatusEnum.Closed: {
      return "success";
    }
  }
}`

**Comment by topce** [comment]
@bedag-moo @sebastianhaas @bobvanderlinden 

If you wanted not to break compatibility with previous version
it was possible by still keeping old enums with cast any
and add new type with union of string literals 
and finally  create new type as union of this type 
With this done in model in api use new union type
So if end user pass old Enum it is OK
if user pass string it also pass.
When migrating to angular 5 just remove any in enum

**Comment by topce** [comment]
@bedag-moo @bobvanderlinden @sebastianhaas @wing328 
for example 
``` TypeScript
`type IssueStatusType = 'open' | 'inProgress' | 'resolved' | 'closed'; 
enum IssueStatusEnum {
  Open = 'open',
  InProgress = 'inProgress',
  Resolved = 'resolved',
  Closed = 'closed',
}
type IssueStatus = IssueStatusType | IssueStatusType;
// export all types so user use what she want
// use in rest api calls IssueStatus 
`
Just trying to help ;-)
Good work guys !!!

**Comment by bobvanderlinden** [comment]
@topce Thanks for the clarification. My main issue was that I couldn't use JSON as test fixtures because `string` wasn't accepted as a value for an enum. Like in [this](https://www.typescriptlang.org/play/#src=enum%20ExampleEnum%20%7B%0D%0A%20%20%20%20ExampleValueA%20%3D%20'a'%2C%0D%0A%20%20%20%20ExampleValueB%20%3D%20'b'%0D%0A%7D%0D%0A%0D%0Aconst%20example%3A%20ExampleEnum%20%3D%20'a'%3B%0D%0A) example.

This PR was a good alternative, as swagger defines possible string or int values and doesn't really define an enum. These concepts are unfortunately not the same in TypeScript.

Your example illustrates a good workaround to keep everyone happy. Even though the implementation doesn't look that pretty to me, it does seem like a good solution for swagger-codegen for the moment.

That said, it is dawning on me that this is probably a deficiency of TypeScript itself. TypeScript should be able to assign a constant `'string'` to an `enum MyString { StringValue = 'string' }`, but at the same time do not allow arbitrary values to be assigned. Would all of you agree that this is a desired feature in TypeScript? In that case I'd gladly make an issue.

**Comment by topce** [comment]
@bobvanderlinden Interesting. Thank you for explanation!

**Comment by bobvanderlinden** [comment]
I found that it isn't really a good workaround when the rest of your code uses `ExampleEnum` or `ExampleType`. You probably do not want to use `Example` in all of your code because it removes the ability to (for instance) do exhaustive switch-cases.

Additionally, the behavior of string-enums and string-unions being differently typed is intended (see [this new post](https://github.com/Microsoft/TypeScript/issues/17690#issuecomment-337975541)). It's unfortunate, but we'll have to deal with that.

That being said, @RyanCavanaugh mentioned [here](https://github.com/Microsoft/TypeScript/issues/17690#issuecomment-321319291) that the intent of string-enums is to be able to change its underlying value without changing all of the code that relies on it. In the case of Swagger and this PR always need to change the code when the swagger definition changes, so I'd say string-union is appropriate and string-enum is not.

**Comment by topce** [comment]
I put in my previous comments pros and cons for both solution.
I understand that @bedag-moo  and  @bobvanderlinden 
value more simplicity then compatibility with previous version .
I respect your opinion and arguments .

**Comment by kenisteward** [comment]
@bobvanderlinden isn't the purpose of a enum so you don't mistype or misasign values?

I have never come into a case where I've had an enum and wanted to assign the value myself instead of just using Enum.Value

Especially with intellisense it's easier to do enum.value than to type in 'somevalue' and have it tell you that it's wrong.  

I hate that I missed this whole conversation.  In the case of enums, I think it is much better practice to just do  Enum.Value

Please let me know why you would like to do    myenum:Enum = 'EnumVal' instead.

I would definitely be more inclined to accept this change if you could still do Enum.Value.

@sebastianhaas Do you have any opinion on this?

**Comment by kenisteward** [comment]
I've found a better solution that i'll be adding soon that will allow for both and backwards comparability to 1.8

**Comment by kenisteward** [comment]
I made improvements on this in #6788 Now we have full backward compatibility with those who use enums in the MS prescribed way as well as those who want to set it via the value itself (which you can technically do with regular enums.  But this makes it to where we should be able to use any string type in the future as well.

```
type MyStringEnum = "member1" | "member2";

const MyStringEnum = {
    Member1: "member1" as MyStringEnum,
    Member2: "member2" as MyStringEnum
};

// implicit typing example
let myVariable = MyStringEnum.Member1; // ok
myVariable = "member2";                // ok
myVariable = "some other value";       // error, desired

// explict typing example
let myExplicitlyTypedVariable: MyStringEnum;
myExplicitlyTypedVariable = MyStringEnum.Member1; // ok
myExplicitlyTypedVariable = "member2";            // ok
myExplicitlyTypedVariable = "some other value";   // error, desired
```
this allows us to not have to check for typescript as everyone by now should be at least past 1.8.  This gives the best of both worlds.

Credit: https://stackoverflow.com/a/35257367 David Sherret.

[TS Playground](https://www.typescriptlang.org/play/#src=type%20MyStringEnum%20%3D%20%22member1%22%20%7C%20%22member2%22%3B%0D%0A%0D%0Aconst%20MyStringEnum%20%3D%20%7B%0D%0A%20%20%20%20Member1%3A%20%22member1%22%20as%20MyStringEnum%2C%0D%0A%20%20%20%20Member2%3A%20%22member2%22%20as%20MyStringEnum%0D%0A%7D%3B%0D%0A%0D%0A%2F%2F%20implicit%20typing%20example%0D%0Alet%20myVariable%20%3D%20MyStringEnum.Member1%3B%20%2F%2F%20ok%0D%0AmyVariable%20%3D%20%22member2%22%3B%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20ok%0D%0AmyVariable%20%3D%20%22some%20other%20value%22%3B%20%20%20%20%20%20%20%2F%2F%20error%2C%20desired%0D%0A%0D%0A%2F%2F%20explict%20typing%20example%0D%0Alet%20myExplicitlyTypedVariable%3A%20MyStringEnum%3B%0D%0AmyExplicitlyTypedVariable%20%3D%20MyStringEnum.Member1%3B%20%2F%2F%20ok%0D%0AmyExplicitlyTypedVariable%20%3D%20%22member2%22%3B%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20ok%0D%0AmyExplicitlyTypedVariable%20%3D%20%22some%20other%20value%22%3B%20%20%20%2F%2F%20error%2C%20desired#src=type%20MyStringEnum%20%3D%20%22member1%22%20%7C%20%22member2%22%3B%0D%0A%0D%0Aconst%20MyStringEnum%20%3D%20%7B%0D%0A%20%20%20%20Member1%3A%20%22member1%22%20as%20MyStringEnum%2C%0D%0A%20%20%20%20Member2%3A%20%22member2%22%20as%20MyStringEnum%0D%0A%7D%3B%0D%0A%0D%0A%2F%2F%20implicit%20typing%20example%0D%0Alet%20myVariable%20%3D%20MyStringEnum.Member1%3B%20%2F%2F%20ok%0D%0AmyVariable%20%3D%20%22member2%22%3B%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20ok%0D%0AmyVariable%20%3D%20%22some%20other%20value%22%3B%20%20%20%20%20%20%20%2F%2F%20error%2C%20desired%0D%0A%0D%0A%2F%2F%20explict%20typing%20example%0D%0Alet%20myExplicitlyTypedVariable%3A%20MyStringEnum%3B%0D%0AmyExplicitlyTypedVariable%20%3D%20MyStringEnum.Member1%3B%20%2F%2F%20ok%0D%0AmyExplicitlyTypedVariable%20%3D%20%22member2%22%3B%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20ok%0D%0AmyExplicitlyTypedVariable%20%3D%20%22some%20other%20value%22%3B%20%20%20%2F%2F%20error%2C%20desired%0D%0A%0D%0A%2F%2F%20not%20able%20to%20set%20the%20internal%20types%20variables%20unless%20it%20exists%0D%0AMyStringEnum.Member1.blah%20%3D%20'test'%20%2F%2F%20error%2C%20desired)

**Comment by bobvanderlinden** [comment]
I'm not sure that's actually the desired behavior. From typescripts perspective that enum now has 4 possible values instead of 2. A switch will always need a `default:` even though this will never be used at runtime.

I still think this is a defect of typescript itself, but as mentioned in https://github.com/Microsoft/TypeScript/issues/17690 it is the intended behavior.

As for swagger, it's a choice of 3 evils. I'm not sure which one is best. If backwards compatibility is imported I'd say to revert this PR. The new suggestion will mess up switch statements in existing projects.

I still like hat we can use example json structures in tests and mocks. Even though auto completion does not always work nicely, the typechecker will still inform you about typos.

It's just a preference, opinions of more people are very welcome.

**Comment by bedag-moo** [comment]
Meta: Isn't it rather late to discuss this after merging the PR? I filed an issue about the proposed change back in July (#6206) and waited weeks before implementation, then I implemented it and it spents months in review, got merged, and *now* we start talking?

But since all interested parties are now here, let me reiterate the reasons for this change:

1. In swagger, enums are an anonymous collection of possible values. Typescript union types are an anonymous collection of possible values. Typesscript enums are named types, and assignment compatibility is checked by name rather than the set of possible values. To preserve assignment compatibility for different uses of an enum in the swagger contract, swagger-codegen would have to reuse the typescript enum declaration. This would require a major extension of swagger-codegen, because such duplicate elimination is not currently supported by the code generation engine (as far as I know).
2. union types offer equivalent compile time type safety to enums
3. union types offer code completion in visual studio code (if you trigger code completion within the string literal, it will suggest possible values), but not (yet?) in IntelliJ. 
4. union types can be used as index types, but enum types can not:
    ```
    const colorOf: {[S in Issue.StatusEnum]: string} = {
      open: 'danger',
      closed: 'success',
      resolved: 'info',
      inProgress: 'warn'
    }
    ```
    which can be used like
    ```
    const color = colorOf[issue.status];
    ````  
    such simple lookup tables are therefore only possible for union types.
5. union types can be easily used in templates, i.e.
    ```
    <div [ngSwitch]="issue.state">
      <div *ngSwitchCase="'open'"> 
    ```
    while enum types additionally require the enum to be imported and published in a component field so we can use it in the template.

Any revision of this PR should continue to support these 5 properties (assignability, type safety, code completion, support for index types, ngSwitch). If we can additionally improve code completion, and not break backwards compatibility, I'm all ears :-)

@kenisteward : That's an interesting idea, because it preserves the above properties while emulating the previous api. (Bob, I could not reproduce your statement about 4 possible values; the playground says there is a variable and a type named MyStringEnum, but typescript seems to treat these as separate declarations?)

Thinking of switch statements however, I did find a small disadvantage of Keni's pattern, because typescript seems unable to verify exhaustiveness for non-literal switch cases:

```
function indirect(e: MyStringEnum) {  // "not all code paths return a value" under -noImplicitReturn
    switch (e) { 
        case MyStringEnum.Member1:
            return 1;
        case MyStringEnum.Member2:
            return 2;
    }
}

function direct(e: MyStringEnum) { // does not error under -noImplicitReturn
    switch (e) {
        case "member1":
            return 1;
        case "member2":
            return 2;
    }
}
```

[playground link](https://www.typescriptlang.org/play/index.html#src=type%20MyStringEnum%20%3D%20%22member1%22%20%7C%20%22member2%22%3B%0D%0A%0D%0Aconst%20MyStringEnum%20%3D%20%7B%0D%0A%20%20%20%20Member1%3A%20%22member1%22%20as%20MyStringEnum%2C%0D%0A%20%20%20%20Member2%3A%20%22member2%22%20as%20MyStringEnum%0D%0A%7D%3B%0D%0A%0D%0A%2F%2F%20implicit%20typing%20example%0D%0Alet%20myVariable%20%3D%20MyStringEnum.Member1%3B%20%2F%2F%20ok%0D%0AmyVariable%20%3D%20%22member2%22%3B%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20ok%0D%0AmyVariable%20%3D%20%22some%20other%20value%22%3B%20%20%20%20%20%20%20%2F%2F%20error%2C%20desired%0D%0A%0D%0A%2F%2F%20explict%20typing%20example%0D%0Alet%20myExplicitlyTypedVariable%3A%20MyStringEnum%3B%0D%0AmyExplicitlyTypedVariable%20%3D%20MyStringEnum.Member1%3B%20%2F%2F%20ok%0D%0AmyExplicitlyTypedVariable%20%3D%20%22member2%22%3B%20%20%20%20%20%20%20%20%20%20%20%20%2F%2F%20ok%0D%0AmyExplicitlyTypedVariable%20%3D%20%22some%20other%20value%22%3B%20%20%20%2F%2F%20error%2C%20desired%0D%0A%0D%0Afunction%20test(e%3A%20MyStringEnum)%20%7B%0D%0A%20%20%20%20switch%20(e)%20%7B%0D%0A%20%20%20%20%20%20%20%20case%20MyStringEnum.Member1%3A%0D%0A%20%20%20%20%20%20%20%20%20%20%20%20return%201%3B%0D%0A%20%20%20%20%20%20%20%20case%20%22member2%22%3A%0D%0A%20%20%20%20%20%20%20%20%20%20%20%20return%202%3B%0D%0A%20%20%20%20%7D%0D%0A%7D)

But as switches over actual enum types are not considered exhaustive either (without a `default`), this merely mirrors the previous state.

Whether the improved backwards compatibility and code completion is worth the added complexity in the generator is a matter of taste; I am fine with either solution as long as my precious union types are not rolled back ;-)

**Comment by kenisteward** [comment]
@bedag-moo My apologizes for not fully analyzing this sooner.  I only commented here to keep the conversation in an organized area when I could have just opened a new issue to append changes to this.

> 1. In swagger, enums are an anonymous collection of possible values. Typescript union types are an anonymous collection of possible values.

This may be true, but when creating these clients we need to think of how people will (and want) to consume them.  Generally, when a person thinks of an enum, (I think) they assume that they are going to use:

```
let e: Enum = Enum.Value
```
and not

```
let e: Enum = 'value'
```

because of this it would be important to support actual enums or at least the facade of being one

> 2 union types offer equivalent compile time type safety to enums
truth

> 3. union types offer code completion in visual studio code (if you trigger code completion within the string literal, it will suggest possible values), but not (yet?) in IntelliJ.

OH.....MY.....GOD how have i missed this freaking feature! lol.  This is great! Instantly just enabled this!  However, because i'm not sure if all IDE's implement this support but I know that the ones that support TS do support enum completion that is why I say we should still support let e: Enum = Enum.Value

> 4 union types can be used as index types, but enum types can not
This works perfectly for me with an enum.  Not sure why it isn't working for you.
[TS Playground](https://tinyurl.com/ybqhunco)

> 5 union types can be easily used in templates
So enums are just objects when they are translated.  Thinking of it in that light, you can't expose an object to the template unless it is a property of that component.  That's just Angular specification.

But to that note you can do exactly the same thing for an enum in the template.

```
// somewhere else
export class SomeType {
   someValue: Enum
   constructor() {
    // sets someValue to what you want it to be
   }
}
// in some componnent
@input() templateDataValue: SomeType;

// in some component template
<div [ngSwitch]="templateDataValue.someValue">
  <div *ngSwitchCase="'open'"> 
```

Did you have some runtime error or other issue?

> Thinking of switch statements however,

I did the same thing using the original type union and it also didn't force a default.  I also didn't get any errors when I took out a whole value.  

Note: I'm using 2.5.2

**Comment by kenisteward** [comment]
@bobvanderlinden This is true. If there is an issue for that why not just program it to return one of the values by default?  ie

```
  // Enum has 3 values
  b (v: Enum) {
    switch (v) {
        case Enum.Value3
             return something();
        case Enum.Value2
            return something2();
       default:
            return something3();
    }
  }
```

The code is strongly typed already so you know it should be one of those values (unless someone did something nefarious like cast to any and back.

bedag-moo:
>> 1 In swagger, enums are an anonymous collection of possible values. Typescript union types are an anonymous collection of possible values.
> 
> This may be true, but when creating these clients we need to think of how people will (and want) to consume them. 

Certainly, and I would not violate expectations lightly, but how else can we make different uses of the same enum assignment compatible? I think assignment compatibility is a justified expectation as well, and since it is related to correctness and type safety, it seems more important than a syntactic difference. 

>> 4 union types can be used as index types, but enum types can not
>
> This works perfectly for me with an enum. Not sure why it isn't working for you.

Oops, I had tried with `{[K: MyEnum]: string}` rather than `{[K in MyEnum]: string}`. The latter indeed works flawlessly, though it is slightly surprising that its maps enum values rather than names. But that's good enough for me :-)

>> 5 union types can be easily used in templates
>
> But to that note you can do exactly the same thing for an enum in the template.

Oh, you're right. It seemed so obvious that since `const e: Enum = 'open'` was illegal, so would `e === 'open'`, but apparently the compiler is more lax with comparisions and switch statements, so this works even in ahead-of-time compilation :-)

topce:
@bedag-moo I report problem (breaking compatibility} same day i spotted it 
Better late than never 
As I explained in previous comments I would prefer to use enums instead of strings or numbers
Use cases that you mention were not relevant for project I am working on
I am more on same page as @kenisteward
but I see your and @bobvanderlinden have different opinion and I respect  that

Issue #6234, opened 2017-08-04
[Async-Scala] Package name should be stripped from the class name


ramzimaalej:
##### Description

When generating a client using Async-scala module, the class name of the generated models contains the package like this "models.WebsiteBodyModel" which causes compilation errors. I tried many config parameters to override this behavior but ended up submitting a PR to fix.

##### Swagger-codegen version

Latest

##### Swagger declaration file content or url

[link to json file to reproduce the error](https://gist.github.com/ramzimaalej/88a079d293ca7f3a099beec97e74db66)

##### Command line used for generation

swagger-codegen generate -i pet.json -l async-scala

##### Steps to reproduce

cd into the folder created by the generator and do: sbt compile

##### Suggest a fix/enhancement

This [PR](https://github.com/swagger-api/swagger-codegen/pull/6235) fixes the problem

wing328:
Please try the latest master to generate scala client, which combines both `scala` and `async-scala` clients into one.

PR #6235, opened 2017-08-04
Async-Scala Strip package name from the class name

ramzimaalej:
This goal of this PR is to fix this [issue](https://github.com/swagger-api/swagger-codegen/issues/6234).

In fact, When generating a client using Async-scala module, the class name of the generated models contains the package like this "models.WebsiteBodyModel" which causes compilation errors.

### DIFF
--- a/modules/swagger-codegen/pom.xml
+++ b/modules/swagger-codegen/pom.xml
@@ -273,6 +273,13 @@
             <version>${diffutils-version}</version>
             <scope>test</scope>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/org.mockito/mockito-core -->
+        <dependency>
+            <groupId>org.mockito</groupId>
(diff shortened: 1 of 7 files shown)


wing328:
@ramzimaalej thanks for the PR. Can you also try the `scala` generator in the latest master, which should have combined both `scala` and `async-scala` generators into one and see if you experience similar issues?

ramzimaalej:
@wing328 I tried the latest code from master, which works fine and generates async client. However, the model class name is not clean. For example, if you fully qualified class name is models.Pet the generated class will have a name like this: ModelsPets

I will raise another PR to fix this issue.

ramzimaalej:
@wing328 I raised this PR to push the fix to the Scala generator as you suggested: https://github.com/swagger-api/swagger-codegen/pull/6293

Issue #4794, opened 2017-02-14
[Lua] Add API client generation


daurnimator:
Add lua API client generation.

Probably using [lua-http](https://github.com/daurnimator/lua-http)

wing328:
@daurnimator sounds good to me :)

May I know if you're working on the Lua generator? Do you need any help from us (the community)?

daurnimator:
> May I know if you're working on the Lua generator?

Not yet. Though I'm considering it.

> Do you need any help from us (the community)?

Indeed: I'm not sure where to start. (I've just had a look around at some of the other generators: shouldn't be too hard to translate...)

wing328:
We recently add the Bash generator via https://github.com/swagger-api/swagger-codegen/pull/4541, which may serve as a good starting point to add a new client generator.

zjarkov:
Hi, didn't try it personally wth swagger but [lua-Spore](http://fperrad.github.io/lua-Spore/) is not sufficient?

daurnimator:
> Hi, didn't try it personally wth swagger but lua-Spore is not sufficient?

I didn't realise lua-Spore had a swagger component. It might actually be suitable for me.

-----------------

I'm wondering if adding lua support to swagger-code is the most useful thing to do. I feel like a fully dynamic-runtime approach like [swagger-js](https://github.com/swagger-api/swagger-js) took might be more appropriate.

wing328:
@daurnimator dynamic client approach (e.g. swagger-js, lua-spore) is definitely one way to do it but some developers prefer static clients generated by Swagger Codegen.

kigsmtua:
@daurnimator  any progress on this

wing328:
@daurnimator @zjarkov @kigsmtua similar to PowerShell and Rust, we can go through the same approach by creating a Lua API client for [PetStore](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.yaml) by focusing on a few endpoints.
- [GET](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.yaml#L156-L183)
- [POST (form)](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.yaml#L184-L218)
[...]

PR #6244, opened 2017-08-05
samples/client/petstore/lua: Add lua sample

daurnimator:
#WIP 

As suggested in https://github.com/swagger-api/swagger-codegen/issues/4794#issuecomment-320245838

This implementation uses [lua-http](https://github.com/daurnimator/lua-http)

### DIFF
--- a/samples/client/petstore/lua/README.md
+++ b/samples/client/petstore/lua/README.md
@@ -1,0 +1,1 @@
+https://github.com/swagger-api/swagger-codegen/issues/4794#issuecomment-320245838
--- a/samples/client/petstore/lua/main.lua
+++ b/samples/client/petstore/lua/main.lua
@@ -1,14 +1,12 @@
--- https://github.com/swagger-api/swagger-codegen/issues/4794#issuecomment-320245838
+local petstore_client = require "petstore_client"
+local petstore_client_pet = require "petstore_client.pet"
 
-local http_request = require "http.request"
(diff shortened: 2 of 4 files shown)


daurnimator:
I just added code that fleshes it out to use "classes".
I'm not sure exactly what you wanted here. So perhaps look at commits one by one?

wing328:
@daurnimator it definitely looks good as a starting point 👍 Here are some feedbacks:

1) Can we put the following inside a function (e.g. function add_pet )?
```
local req = http_request.new_from_uri(string.format("%s/pet/%d", base, petId))
req.headers:upsert("accept", "application/json")
local headers, stream = req:go()
```

2) can we also put the code inside a module (e.g. petstore)?

Questions:
Is there a way to create sub-module so that Lua developers can call the function like "petstore.pet_api.add_pet" for example?

Is there a way to create object with properties for the models defined in the Swagger/OpenAPI spec? (e.g. https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.yaml#L654-L692)

daurnimator:
@wing328 much of what you mentioned was done in the second commit.

wing328:
@daurnimator nice. I'll start the reverse engineering process over the weekend.

wing328:
Merged it into master so that credits go to @daurnimator

PR #6249, opened 2017-08-06
Added trivial getter implementation

farcaller:
This can be improved further by turning ref Options into Options with refs and passing trivial types by copy.


wing328:
> This can be improved further by turning ref Options into Options with refs and passing trivial types by copy.

will add this to the enhancement list.

PR #6252, opened 2017-08-06
[Lua] Add new Lua client generator

wing328:
### PR checklist

### Description of the PR

For https://github.com/swagger-api/swagger-codegen/issues/4794

Still WIP (based on https://github.com/swagger-api/swagger-codegen/pull/6244)


modules/swagger-codegen/src/main/resources/lua/configuration.mustache
```diff
@@ -0,0 +1,43 @@
...
+	UserName      string            `json:"userName,omitempty"`
+	Password      string            `json:"password,omitempty"`	
+}
+
+type APIKey struct {
+	Key 	string
+	Prefix	string
+}
+
+type Configuration struct {
```
wing328: @daurnimator here is the configuration class for Go API client. We may need something similar in Lua API client to store the authentication setting, default header, host, basePath, etc

modules/swagger-codegen/src/main/resources/lua/model.mustache
```diff
@@ -0,0 +1,28 @@
...
+local {{classname}}_mt = {
+  __name = "{{classname}}";
+  __index = {{classname}};
+}
+
+local function cast_{{classname}}(t)
+  return setmetatable(t, {{classname}}_mt)
+end
+
+local function new_{{classname}}({{#vars}}{{name}}{{#hasMore}}, {{/hasMore}}{{/vars}})
```
wing328: @daurnimator  in the constructor method signature, I've included all parameters (required, optional)

modules/swagger-codegen/src/main/resources/lua/model.mustache
```diff
@@ -0,0 +1,28 @@
...
+  return cast_pet({
+    {{#vars}}
+    {{name}} = {{name}};
```
wing328: @daurnimator later we'll need to handle the situation when the JSON key name contains special character (e.g. date-time) (day 2 requirement)

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -0,0 +1,113 @@
...
+        if not body then
+            return nil, err, errno2
+        end
+        stream:shutdown()
+        local result, _, err3 = dkjson.decode(body)
+        if result == nil then
+            return nil, err3
+        end
+        return {{{packagName}}}_{{classname}}.cast(result)
+    else
```
wing328: @daurnimator for 4xx and 5xx status code, we usually throw an exception in other generators and the exception object contains 3 fields: http status code, header and body (error message). Not sure if we can do something similar in Lua.

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -0,0 +1,113 @@
...
+        local body, err, errno2 = stream:get_body_as_string()
+        if not body then
+            return nil, err, errno2
+        end
+        stream:shutdown()
+        local result, _, err3 = dkjson.decode(body)
+        if result == nil then
+            return nil, err3
+        end
+        return {{{packagName}}}_{{classname}}.cast(result)
```
wing328: @daurnimator I notice that Lua can return a list of objects so I would suggest we return the following
- result (e.g. Pet, string, array of Users)
- http status code
- http headers

as it's just a matter of time the developers need the status code and header.

wing328:
@daurnimator  please review my comments and the TODO in the code when you've time. Thanks!

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -0,0 +1,113 @@
...
+local function new_{{classname}}(host, basePath, schemes)
+  local schemes_map = {}
+  for _,v in ipairs(schemes) do
+    schemes_map[v] = v
+  end
+  local default_scheme = schemes_map.https or schemes_map.http
+  // TODO: default basePath to {{{basePath}}}
+  return setmetatable({
+    host = host;
+    basePath = basePath;
```
daurnimator: Could use `basePath = basePath or {{{basePath}}}` to close above TODO?

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -0,0 +1,113 @@
...
+    schemes = schemes_map;
+    default_scheme = default_scheme;
+  }, {{{pacakgeName}}}_mt)
```
daurnimator: typo? pacakgeName

modules/swagger-codegen/src/main/resources/lua/api.mustache
daurnimator: lua-http follows the http2 convention where the status code is just another header (called `:status`): so no need for the extra `status` return value.

modules/swagger-codegen/src/main/resources/lua/api.mustache
daurnimator: You *can* throw an error, but it's considered better practice to return `nil, error, errno`. Where `error` is usually a string, but could be some type of other object.

modules/swagger-codegen/src/main/resources/lua/model.mustache
daurnimator: For non-valid lua identifiers, the key syntax is `["f@0"]`

samples/client/petstore/lua/lua-petstore/.gitignore
```diff
@@ -0,0 +1,40 @@
+# Compiled Lua sources
```
daurnimator: Not sure where you got this .gitignore file from.. it's tremendously broad. You probably don't need one at all..

samples/client/petstore/lua/lua-petstore/api_response.lua
```diff
@@ -0,0 +1,33 @@
...
+  __name = "api_response";
+  __index = api_response;
+}
+
+local function cast_api_response(t)
+  return setmetatable(t, api_response_mt)
+end
+
+local function new_api_response(code, type, message)
+  return cast_pet({
```
daurnimator: You left `cast_pet` in here.

samples/client/petstore/lua/lua-petstore/tag.lua
```diff
@@ -0,0 +1,32 @@
...
+  __name = "tag";
+  __index = tag;
+}
+
+local function cast_tag(t)
+  return setmetatable(t, tag_mt)
+end
+
+local function new_tag(id, name)
+  return cast_pet({
```
daurnimator: `cast_pet`

samples/client/petstore/lua/lua-petstore/store_api.lua
```diff
@@ -0,0 +1,222 @@
...
+
+function store_api:place_order(body)
+    local req = http_request.new_from_uri({
+        scheme = self.default_scheme;
+        host = self.host;
+        path = string.format("%s//store/order", self.basePath, petId);
+    })
+
+    // set HTTP verb
+	req.headers:upsert(":method", "Post")
```
daurnimator: `"Post"` should be all caps I think?
Also this is indented oddly

samples/client/petstore/lua/lua-petstore/store_api.lua
```diff
@@ -0,0 +1,222 @@
...
+        stream:shutdown()
+        return nil, "Unexpected response status code"
+    end
+end
+
+function store_api:place_order(body)
+    local req = http_request.new_from_uri({
+        scheme = self.default_scheme;
+        host = self.host;
+        path = string.format("%s//store/order", self.basePath, petId);
```
daurnimator: double //

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -0,0 +1,113 @@
+{{>partial_header}}
+//package {{packageName}}
+
+{{#operations}}
+local http_request = require "http.request"
+local http_util = require "http.util"
+local dkjson = require "dkjson"
+
+// model import
+local {{{packagName}}}_{{classname}} = require "{{{pacakgeName}}}.{{classname}}"
```
daurnimator: two different typos: pacakgeName packagName

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -0,0 +1,113 @@
...
+
+local {{{packageName}}}= {}
+local {{{pacakgeName}}}_mt = {
```
daurnimator: pacakgeName

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -0,0 +1,113 @@
...
+{{#operation}}
+function {{classname}}:{{operationId}}({{#allParams}}{{#required}}{{paramName}}{{#hasMore}}, {{/hasMore}}{{/required}}{{/allParams}})
+    local req = http_request.new_from_uri({
+        scheme = self.default_scheme;
+        host = self.host;
+        path = string.format("%s/{{{path}}}", self.basePath, petId);
+    })
+
+    // set HTTP verb
+	req.headers:upsert(":method", "{{httpMethod}}")
```
daurnimator: indentation

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -0,0 +1,113 @@
...
+
+    // TODO: create a function to select proper accept
+    local var_accept = { {{#consumes}}"{{{mediaType}}}"{{#hasMore}}, {{/hasMore}}{{/consumes}} }
+    req.headers:upsert("accept", {{#consumes}}{{#-first}}"{{{mediaType}}}"{{/-first}}{{/consumes}})
+
+    // TODO: create a function to select proper content-type
```
daurnimator: selects content-type based on what?

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -0,0 +1,113 @@
...
+    {{#isBasic}}
+    // TODO HTTP basic auth
+    {{/isBasic}}
+    {{#isOAuth}}
+    // TODO oauth
+    {{/isOAuth}}
+    {{/authMethods}}
+
+    // make the HTTP call
+    local headers, stream, errno = req:go()
```
daurnimator: FYI, we'll probably want to pass a timeout to `:go()`

modules/swagger-codegen/src/main/resources/lua/api.mustache
wing328: Sure, let's follow the best practice to return nil, error, errno.

modules/swagger-codegen/src/main/resources/lua/configuration.mustache
daurnimator: Don't we sort of have that already with the `api` class?

modules/swagger-codegen/src/main/resources/lua/api.mustache
wing328: OK. so just return result and headers then.

modules/swagger-codegen/src/main/resources/lua/configuration.mustache
wing328: Yes, and I'm not saying it's a must. My suggestion is to make it more consistent with other language generator. We can put it as a day 2 requirement for reconsideration.

samples/client/petstore/lua/lua-petstore/.gitignore
wing328: Based on https://github.com/github/gitignore/blob/master/Lua.gitignore

modules/swagger-codegen/src/main/resources/lua/api.mustache
wing328: Here is an example for Java (jersey2):

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache#L435-L491

modules/swagger-codegen/src/main/resources/lua/api.mustache
wing328: OK. We can add another configurable property `timeout` for that.

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -91,20 +91,37 @@ function {{classname}}:{{operationId}}({{#allParams}}{{#required}}{{paramName}}{
         return nil, stream, errno
     end
     local http_status = headers:get(":status")
-    if http_status == "200" then
+    if tonumber(http_status) >= 200 or tonumber(http_status) <= 299 then
```
daurnimator: No need to convert to number (and that will have false positives, e.g. a status code of `2e2`)
Instead consider if you just want to check the first character? In which case: `if http_status:sub(1,1) == "2"`

wing328: Sure. will update that in the next push.

modules/swagger-codegen/src/main/resources/lua/model.mustache
```diff
@@ -13,9 +13,9 @@ local function cast_{{classname}}(t)
 end
 
 local function new_{{classname}}({{#vars}}{{name}}{{#hasMore}}, {{/hasMore}}{{/vars}})
-  return cast_pet({
+  return cast_{{classname}}({
     {{#vars}}
-    {{name}} = {{name}};
+    ["{{baseName}}"] = {{name}};
```
daurnimator: Will this escape correctly if e.g. baseName include a double quote?

daurnimator: Also, it would be better to prefer the non-quoted form (saving the `[]` syntax only for invalid lua identifiers)

wing328:
I think these are the remaining items before a beta release:
- [ ] add support for query parameters
- [ ] add support for authentication
- [ ] function to select a proper "accept", "content-type"

daurnimator:
> add support for query parameters

Support how?

>  add support for authentication

What types of auth? Does this require anything more than setting the authorization header?

>  function to select a proper "accept", "content-type"

When/how should this happen?

wing328:
> Support how?

How can we add query parameters (URL query string) to HTTP request in Lua?

> What types of auth? Does this require anything more than setting the authorization header?

HTTP basic auth, API key (header or query), OAuth access token

> function to select a proper "accept", "content-type"

We need to create 2 functions in Lua similar to https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/Java/libraries/jersey2/ApiClient.mustache#L435-L491

wing328:
@daurnimator I've tried to add support for query parameters and authentication via b63ab93. I also added 2 fake query parameters (will be removed before merge) to illustrate what the auto-generated Lua code for query parameters looks like.

Please review when you've time.

wing328:
Quick question: shall we put the API and model files under the sub-folder "api" and "model" respectively?

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -4,6 +4,7 @@
 {{#operations}}
 local http_request = require "http.request"
 local http_util = require "http.util"
+local url_encode = require "http.util".encodeURIComponent
```
daurnimator: No need to make this a local: call it as `http_util.encodeURIComponent` when needed.

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -27,28 +28,50 @@ local function new_{{classname}}(host, basePath, schemes)
...
     default_scheme = default_scheme;
+    http_username = '';
+    http_password = '';
+    api_key = {}
+    access_token = '';
   }, {{{packageName}}}_mt)
 end
 
+// for base64 encoding
+local function base64_encode(s)
```
daurnimator: Use `basexx.to_base64`. See https://github.com/daurnimator/lua-http/blob/master/http/request.lua#L85 for an example

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -27,28 +28,50 @@ local function new_{{classname}}(host, basePath, schemes)
     basePath = basePath or "{{{basePath}}}";
     schemes = schemes_map;
     default_scheme = default_scheme;
+    http_username = '';
```
daurnimator: Don't default these to the empty string; rather, default them to `nil`.

samples/client/petstore/lua/lua-petstore/pet_api.lua
```diff
@@ -231,29 +240,25 @@ function pet_api:find_pets_by_tags(tags)
     end
 end
 
-function pet_api:get_pet_by_id(pet_id)
+function pet_api:get_pet_by_id(pet_id, )
```
daurnimator: This has a trailing `,` (which is a syntax error)

samples/client/petstore/lua/lua-petstore/user_api.lua
```diff
@@ -353,22 +345,20 @@ function user_api:update_user(username, body)
...
-
     // TODO: create a function to select proper content-type
-    local var_accept = { "application/xml", "application/json" }
+    // ref: https://github.com/swagger-api/swagger-codegen/pull/6252#issuecomment-321199879
+    //local var_accept = { "application/xml", "application/json" }
     req.headers:upsert("content-type", "application/xml")
 
 
-
+	req:set_body(dkjson.encode(body))
```
daurnimator: indentation here is incorrect

daurnimator:
Oh, one big thing: lua comments are `--`, not `//`

daurnimator:
> Quick question: shall we put the API and model files under the sub-folder "api" and "model" respectively?

Sure, if that's swagger-api convention.

modules/swagger-codegen/src/main/resources/lua/api.mustache
wing328: Sure. I was worried it would cause issues later when doing string concatenation.

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -0,0 +1,153 @@
...
+    // api key in headers '{{keyParamName}}'
+    req.headers.upsert("{{{name}}}", api_key['{{{keyParamName}}}'])
+    {{/isKeyInHeader}}
+    {{#isKeyInQuery}}
+    /// TODO: api key in query '{{keyParamName}}'
+    {{/isKeyInQuery}}
+    {{/isApiKey}}
+    {{#isBasic}}
+    // HTTP basic auth
+    req.headers.upsert("Authorization", base64_encode(self.http_username .. " " .. self.http_password))
```
daurnimator: This needs to be `:upsert` (with a colon).

Also you're missing the `"basic "` prefix.

modules/swagger-codegen/src/main/resources/lua/api.mustache
daurnimator: It will: but that's intentional. if the auth requires basic auth and they haven't provided a username/password, then an error *should* occur.

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -22,53 +22,42 @@ local function new_{{classname}}(host, basePath, schemes)
...
-
 {{#operation}}
-function {{classname}}:{{operationId}}({{#allParams}}{{#required}}{{paramName}}{{#hasMore}}, {{/hasMore}}{{/required}}{{/allParams}})
+function {{classname}}:{{operationId}}({{#allParams}}{{paramName}}{{#hasMore}}, {{/hasMore}}{{/allParams}})
     local req = http_request.new_from_uri({
         scheme = self.default_scheme;
         host = self.host;
         path = string.format("%s{{{vendorExtensions.x-codegen-path}}}{{#queryParams}}{{#-first}}?{{/-first}}{{{baseName}}}=%s{{^-last}}&{{/-last}}{{/queryParams}}",
-            self.basePath{{#pathParams}} ,{{paramName}}{{/pathParams}}{{#queryParams}} ,url_encode({{paramName}}){{/queryParams}});
+            self.basePath{{#pathParams}} ,{{paramName}}{{/pathParams}}{{#queryParams}} ,http_util.encodeURIComponent({{paramName}}){{/queryParams}});
```
daurnimator: minor thing, but the missing whitespace after `,` will bug me

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -22,53 +22,42 @@ local function new_{{classname}}(host, basePath, schemes)
...
     })
 
-    // set HTTP verb
+    -- set HTTP verb
```
daurnimator: I would suggest having the templater skip this if it's GET.

wing328:
> Sure, if that's swagger-api convention.

@daurnimator I've moved the API and model files into the respective sub-folders "api" and "model" ( I assume pet_api.lua can access the Pet model without issues)

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -22,53 +22,42 @@ local function new_{{classname}}(host, basePath, schemes)
...
     req.headers:upsert(":method", "{{httpMethod}}")
 
     {{#hasConsumes}}
-    // TODO: create a function to select proper accept
-    // ref: https://github.com/swagger-api/swagger-codegen/pull/6252#issuecomment-321199879
-    //local var_content_type = { {{#consumes}}"{{{mediaType}}}"{{#hasMore}}, {{/hasMore}}{{/consumes}} }
+    -- TODO: create a function to select proper accept
+    -- ref: https:--github.com/swagger-api/swagger-codegen/pull/6252#issuecomment-321199879
```
daurnimator: Did you just do a find/replace for `--`? It broke urls

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -22,53 +22,42 @@ local function new_{{classname}}(host, basePath, schemes)
...
+    --local var_content_type = { {{#consumes}}"{{{mediaType}}}"{{#hasMore}}, {{/hasMore}}{{/consumes}} }
     req.headers:upsert("accept", {{#consumes}}{{#-first}}"{{{mediaType}}}"{{/-first}}{{/consumes}})
 
     {{/hasConsumes}}
     {{#hasProduces}}
-    // TODO: create a function to select proper content-type
-    // ref: https://github.com/swagger-api/swagger-codegen/pull/6252#issuecomment-321199879
-    //local var_accept = { {{#produces}}"{{{mediaType}}}"{{#hasMore}}, {{/hasMore}}{{/produces}} }
+    -- TODO: create a function to select proper content-type
+    -- ref: https:--github.com/swagger-api/swagger-codegen/pull/6252#issuecomment-321199879
```
daurnimator: another bad find/replace

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -86,30 +75,30 @@ function {{classname}}:{{operationId}}({{#allParams}}{{#required}}{{paramName}}{
...
     {{#isKeyInQuery}}
-    /// TODO: api key in query '{{keyParamName}}'
+    -- TODO: api key in query '{{keyParamName}}'
     {{/isKeyInQuery}}
     {{/isApiKey}}
     {{#isBasic}}
-    // HTTP basic auth
-    req.headers.upsert("Authorization", base64_encode(self.http_username .. " " .. self.http_password))
+    -- HTTP basic auth
+    req.readers:upsert("Authorization", "Basic " .. basexx.to_base64(self.http_username .. " " .. self.http_password))
```
daurnimator: header key should be lower case.

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -86,30 +75,30 @@ function {{classname}}:{{operationId}}({{#allParams}}{{#required}}{{paramName}}{
...
     {{/isBasic}}
     {{#isOAuth}}
-    // oAuth
-    req.headers.upsert("Authorization", "Bearer " .. self.access_token)
+    -- oAuth
+    req.headers:upsert("Authorization", "Bearer " .. self.access_token)
```
daurnimator: header key should be lower case

wing328:
The query parameters still need some work as I couldn't find an easy way to support optional query parameters and API key in the URL query string.

daurnimator:
> @daurnimator I've moved the API and model files into the respective sub-folders "api" and "model" ( I assume pet_api.lua can access the Pet model without issues)

You'll just have to modify the require string e.g. `require "petstore.store_api"` will need to change to `require "petstore.api.store"`

modules/swagger-codegen/src/main/resources/lua/api.mustache
wing328: Based on the examples I found (e.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication). "Authorization" (header key) is usually in upper case. (HTTP header key is case-insensitive if I'm not mistaken)

modules/swagger-codegen/src/main/resources/lua/api.mustache
wing328: Yup. Sorry just fixed.

modules/swagger-codegen/src/main/resources/lua/api.mustache
wing328: yup, agreed it should be fixed

modules/swagger-codegen/src/main/resources/lua/api.mustache
wing328: OK. Will add this to the enhancement list later for tracking.

wing328:
> You'll just have to modify the require string e.g. require "petstore.store_api" will need to change to require "petstore.api.store"

Done. e.g. 
```
local petstore_pet_api = require "petstore.api.pet_api"
```

daurnimator:
> Based on the examples I found (e.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication). "Authorization" (header key) usually has the first letter in upper case. (HTTP header key is case-insensitive if I'm not mistaken)

In http2 all header fields MUST be lower case. http://httpwg.org/specs/rfc7540.html#HttpHeaders

lua-http borrows these semantics for all http versions.

wing328:
@daurnimator sure. Added some minor fixes via https://github.com/swagger-api/swagger-codegen/pull/6252/commits/c66f7e9548ec55a7d3c643dd894073adc0ee7da0 as well.

modules/swagger-codegen/src/main/resources/lua/partial_header.mustache
```diff
@@ -1,13 +1,13 @@
...
+ 
  {{/appDescription}}
- * {{#version}}OpenAPI spec version: {{{version}}}{{/version}}
- * {{#infoEmail}}Contact: {{{infoEmail}}}{{/infoEmail}}
- * Generated by: https://github.com/swagger-api/swagger-codegen.git
- */
+  {{#version}}OpenAPI spec version: {{{version}}}{{/version}}
+  {{#infoEmail}}Contact: {{{infoEmail}}}{{/infoEmail}}
+  Generated by: https://github.com/swagger-api/swagger-codegen.git
+--]]
```
daurnimator: Close with just `]]`

daurnimator: Btw, not sure if any of the above (e.g. appDescription) might contain `]]`? How do you deal with this when templating?

daurnimator:
> Done. e.g.
> 
> local petstore_pet_api = require "petstore.api.pet_api"

The `api.pet_api` seems redundant: can you make it just `api.pet`?

modules/swagger-codegen/src/main/resources/lua/partial_header.mustache
wing328: What about escaping `]]` into `\]\]` or `] ]`?

wing328:
> The api.pet_api seems redundant: can you make it just api.pet?

Sure. We can do that. {name}_api is simply a convention we've for other langauges. If `api.{name}` looks more natural to you from a Lua programmer perspective, I will make the change.

modules/swagger-codegen/src/main/resources/lua/partial_header.mustache
wing328: Done.

wing328:
And pet_api.lua should be renamed to just pet.lua if I'm not mistaken.

modules/swagger-codegen/src/main/resources/lua/partial_header.mustache
daurnimator: Instead, lua lets you add any number of '=' signs between `]]`.
So e.g. `--[===[` will be closed only by `]===]`

daurnimator:
> Sure. We can do that. {name}_api is simply a convention we've for other langauges. If api.{name} looks more natural to you from a Lua programmer perspective, I will make the change.

I think it looks odd in any language... but perhaps better to stick with swapper-api norms here.

> And pet_api.lua should be renamed to just pet.lua if I'm not mistaken.

Correct.

samples/client/petstore/lua/lua-petstore/api/store_api.lua
```diff
@@ -0,0 +1,230 @@
...
+local http_util = require "http.util"
+local dkjson = require "dkjson"
+local basexx = require "basexx"
+
+-- model import
+local petstore_store_api = require "petstore.api.store_api"
+
+local petstore= {}
+local petstore_mt = {
+  __name = "";
```
daurnimator: empty __name?

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -0,0 +1,142 @@
...
+  __index = {{{packageName}}};
+}
+
+local function new_{{classname}}(host, basePath, schemes)
+  local schemes_map = {}
+  for _,v in ipairs(schemes) do
+    schemes_map[v] = v
+  end
+  local default_scheme = schemes_map.https or schemes_map.http
+  -- TODO: default basePath to {{{basePath}}}
```
daurnimator: This TODO is done.

modules/swagger-codegen/src/main/resources/lua/partial_header.mustache
wing328: Updated as follows:
```
    @Override
    public String escapeUnsafeCharacters(String input) {
        return input.replace("]]", "] ]");
    }
```

daurnimator: I'm more concerned that an `appDescription` might contain sample code (I know some of mine do).
With this replacement, the sample code would become invalid.

samples/client/petstore/lua/lua-petstore/api/store_api.lua
wing328: Fixed.

modules/swagger-codegen/src/main/resources/lua/api.mustache
wing328: Removed

modules/swagger-codegen/src/main/resources/lua/partial_header.mustache
wing328: We need to escape that to avoid code injection as discovered by a security researcher before.

daurnimator: Right. so what I was suggesting is adding a number of = signs so that the lua block comment can safely contain all things inside.

wing328: I think the hacker can insert the exact number of = signs to uncomment the block and insert malicious code, right?

daurnimator: Insert how?

I'm suggesting at templating time you look at what's about to be pasted into the block; and use a one = more than you find in all the untrusted variables.

wing328: > I'm suggesting at templating time you look at what's about to be pasted into the block; and use a one = more than you find in all the untrusted variables.

Agreed with you that is technically a better approach and will put this in the enhancement list for tracking.

daurnimator:
> @daurnimator from your perspective, do you think it's ready for a beta release so as to collect more feedback from the Lua community?

Maybe. Though it's not like we don't have an empty TODO list.

Btw, something else to add would be generating a .rockspec file.

wing328:
> Btw, something else to add would be generating a .rockspec file

I think that's for LuaRocks (lua package manager) so sure let's add that. I'll include it in enhancement list.

daurnimator:
> I think that's for LuaRocks (lua package manager) so sure let's add that. I'll include it in enhancement list.

Where is this enhancement list?

wing328:
I'll create it after merging this PR into master shortly.

modules/swagger-codegen/src/main/resources/lua/.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: go
```
daurnimator: This isn't go :)

modules/swagger-codegen/src/main/resources/lua/README.mustache
```diff
@@ -0,0 +1,96 @@
...
+
+## Documentation For Authorization
+{{^authMethods}} Endpoints do not require authorization.
+{{/authMethods}}{{#authMethods}}{{#last}} Authentication schemes defined for the API:{{/last}}{{/authMethods}}
+{{#authMethods}}
+## {{{name}}}
+{{#isApiKey}}- **Type**: API key 
+
+Example
+```
```
daurnimator: README examples need updating

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -0,0 +1,141 @@
...
+local http_util = require "http.util"
+local dkjson = require "dkjson"
+local basexx = require "basexx"
+
+-- model import
+local {{{packageName}}}_{{classname}} = require "{{{packageName}}}.api.{{classname}}"
+
+local {{{packageName}}}= {}
+local {{{packageName}}}_mt = {
+  __name = "{{{classname}}}";
```
daurnimator: Subjective question: indentation. 
Most lua is indented in tabs or 4 spaces; however some people use 3 spaces due to it being the default in emacs; and the rare person use 8 spaces. However, 2 spaces is pretty rare.
May I suggest swapping to tabs?

modules/swagger-codegen/src/main/resources/lua/api_response.mustache
```diff
@@ -0,0 +1,35 @@
+{{>partial_header}}
+package {{packageName}}
+
+import (
```
daurnimator: I'm not sure what this file is doing here?

modules/swagger-codegen/src/test/resources/2_0/petstore.yaml
```diff
@@ -171,6 +171,17 @@ paths:
           required: true
           type: integer
           format: int64
+        - name: size
```
daurnimator: This needs removing before merging

modules/swagger-codegen/src/main/resources/lua/api.mustache
```diff
@@ -0,0 +1,141 @@
...
+    {{#isBasic}}
+    -- HTTP basic auth
+    req.readers:upsert("authorization", "Basic " .. basexx.to_base64(self.http_username .. " " .. self.http_password))
+    {{/isBasic}}
+    {{#isOAuth}}
+    -- oAuth
+    req.headers:upsert("authorization", "Bearer " .. self.access_token)
+    {{/isOAuth}}
+    {{/authMethods}}
+
```
daurnimator: There seems to be lots of whitespace in generated files around this.

modules/swagger-codegen/src/main/resources/lua/.travis.yml
```diff
@@ -1,8 +1,6 @@
-language: go
+language: lua
```
daurnimator: Sadly travis doesn't support lua. Us lua devs use a tool 'hererocks' to run in travis
https://github.com/mpeterv/hererocks/#using-hererocks-to-set-up-automated-testing

modules/swagger-codegen/src/main/resources/lua/api.mustache
wing328: Good suggestion. Is there any Lua style guide that we can follow?

modules/swagger-codegen/src/main/resources/lua/api_response.mustache
wing328: Removed. (used by Go)

modules/swagger-codegen/src/test/resources/2_0/petstore.yaml
wing328: Removed. Thanks for the reminder

modules/swagger-codegen/src/main/resources/lua/api.mustache
wing328: It shouldn't. It's a feature in mustache template library that the following will not result in any line or whitespace
```
    {{#thisIsFalse}}
    something here...
    {{/thisIsFalse}}
```

modules/swagger-codegen/src/main/resources/lua/api.mustache
daurnimator: I can't think of a **good** one :PEven the one I wrote I don't like any more

modules/swagger-codegen/src/main/resources/lua/.travis.yml
wing328: Will add this to the enhancement list later

wing328: > Sadly travis doesn't support lua.

Any other CI provides better support for Lua?

modules/swagger-codegen/src/main/resources/lua/api.mustache
wing328: Let's use tab then.

modules/swagger-codegen/src/main/resources/lua/.travis.yml
daurnimator: None I know off the top of my head. Most lua devs will just use hererocks

modules/swagger-codegen/src/main/resources/lua/README.mustache
wing328: Yup, I've removed the auto-genreated README for now as it needs some work to output correct examples. Will add to the enhancement list for tracking.

samples/client/petstore/lua/api/store_api.lua
```diff
@@ -0,0 +1,229 @@
...
+
+	-- set HTTP verb
+	req.headers:upsert(":method", "DELETE")
+
+	-- TODO: create a function to select proper content-type
+	-- ref: https://github.com/swagger-api/swagger-codegen/pull/6252#issuecomment-321199879
+	--local var_accept = { "application/xml", "application/json" }
+	req.headers:upsert("content-type", "application/xml")
+
+
```
daurnimator: These are the empty lines I was talking about.

samples/client/petstore/lua/api/store_api.lua
```diff
@@ -0,0 +1,229 @@
...
+		if not body then
+			return nil, err, errno2
+		end
+		stream:shutdown()
+		local result, _, err3 = dkjson.decode(body)
+		-- exception when decoding the HTTP body
+		if result == nil then
+			return nil, err3
+		end
+		return petstore_{}.cast(result), headers
```
daurnimator: `petstore_{}` doesn't seem like what we wanted?

wing328: Right. That method returns an array of object. Not sure how we can deserialize the result back into an array of object in Lua

samples/client/petstore/lua/api/store_api.lua
wing328: Removed.

samples/client/petstore/lua/api/store_api.lua
daurnimator: It's an already an array. We might just want to `cast` the elements of the array to the correct type.

```lua
for _, ob in ipairs(result) do
    cast_pet(ob)
end
```

wing328: OK. What if the result is a dictionary/hash/map?  What should the code look like?

daurnimator: Hrm? a dictionary/hash of what?
`dkjson.decode` will just decode the json object and return a lua table (arrays and dictionaries are the same data structure in lua)

wing328: > dkjson.decode will just decode the json object and return a lua table

That should work.

> Hrm? a dictionary/hash of what?

Here is an example (the key is not fixed):
{
  "cat" : 10,
  "dog": 2,
  "rabbit": 7
}

samples/client/petstore/lua/api/pet_api.lua
```diff
@@ -0,0 +1,423 @@
...
+		if not body then
+			return nil, err, errno2
+		end
+		stream:shutdown()
+		local result, _, err3 = dkjson.decode(body)
+		-- exception when decoding the HTTP body
+		if result == nil then
+			return nil, err3
+		end
+		return petstore_pet.cast(result), headers
```
wing328: @daurnimator should this be `cast_pet(result), headers`?

wing328: Updated to use cast_pet via 6d180af

wing328:
Tweet: https://twitter.com/wing328/status/895703684079734784 to promote the new generator.

wing328:
@daurnimator here is the enhancement list: https://github.com/swagger-api/swagger-codegen/issues/6284

wing328:
Add @daurnimator as the template creator of Lua via https://github.com/swagger-api/swagger-codegen/commit/f4de42608f3038a915015475e53894035ee218d5

PR #6261, opened 2017-08-08
[Java] Support custom ok-http authenticators

lwander:
### PR checklist

### Description of the PR

Users who want to supply custom authenticators can subclass `ApiClient` and override `getClientAuthenticators`.

Waiting on https://github.com/swagger-api/swagger-codegen/pull/6256 to update petstore samples.


wing328:
@lwander thanks for the PR.

cc @cbornet @jfiala @bbdouglas

cbornet:
Is subclassing the best way ? Why are the authentications unmodifiable in the first place ?

lwander:
My guess had been that modifying the authentications during runtime could lead to some strange behavior, especially if removing something like an OAuth authenticator while a token was still alive. But I don't know the underlying reason. Maybe we can restrict modification to only adding authenticators?

wing328:
@lwander I got the same question as @cbornet. If you would like this change to proceed, please make the authentications modifiable.

PR #6276, opened 2017-08-09
fix #6264

### LINKED ISSUE #6264: [Rails5] Reserved word causing codegen to fail
##### Description

Reserved word causes Rails5 codegen to fail.  However, Ruby codegen has a workaround that renames the words that's causing an issue.  Hoping this can be ported to Rails5

##### Swagger-codegen version

2.2.3

`swagger-codegen generate -i swagger.yaml -o ./server/generated-server -l rails5`

Not sure if it's as simple as taking 
https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RubyClientCodegen.java#L478-L483
and putting it in 
https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Rails5ServerCodegen.java#L259
and
https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/Rails5ServerCodegen.java#L259

willyko:
### Description of the PR

Fix #6264


wing328:
@willyko thanks for the fix, which looks good to me.

Issue #6288, opened 2017-08-11
 Java retrofit generated OAuth interceptor class can potentially call itself forever


ali-jawad:
##### Description

When HTTP request from Retrofit client fails due to authorization, even though valid access token is in hand, `OAuth` class takes for granted that access token needs to be refreshed and that is the problem. So it goes ahead and refreshes it. When it succeeds at that, it recursively calls its intercept method again. So in the next run, with a valid new token (again), the same happens on and on.

The fix is to give up recovering from 401 or 403 on second time.

##### Swagger-codegen version

v2.2.3

###### 1) Mocking oauth and api server endpoints
Use wiremock or anything you prefer to have two endpoints. One for access token, and another for the api.

For the token endpoint, ensure different access token is returned by the mock server and 200 application/json response is returned with valid JWT token structure in its body.

For the api endpoint, ensure it returns 403 or 401.

That would mimic a failure due to authorization to the target api endpoint.

###### 2) Generate retrofit client
To do this, have a swagger.yaml file with an arbitrary endpoint and run codegen standalone.

###### 3) Make the call
Make sure your mock server is running with the setup described in step 1. Then call your endpoint through Retrofit client that is configured to use OAuth:
```
      ApiClient client = new ApiClient();

      client.addAuthorization("oAuth",
         new OAuth(
            new OAuthClientRequest.TokenRequestBuilder(getOAuthTokenUrl())
               .setGrantType(GrantType.PASSWORD)
               .setClientId('client-id')
               .setClientSecret('client-secret')
               .setScope('scope')
               .setUsername('username')
               .setPassword('password')
         )
      );

      client.setAdapterBuilder(getRetrofitBuilder(endpoint));

      MyApi api = client.createService(serviceClass);
      
      api.callMyMockApiEndpoint();
```

###### 3) Observe
You will see an infinite recursion of requesting the getting 401 (or 403) and refreshing access token. It is not a surprise if you inspect the logic in `OAuth` class.

##### Suggest a fix/enhancement

The fix is to limit the recursive calls to two times. If we first get 401 or 403, then refresh token successfully, then same thing happens, it is sensible to assume that this is not a token issue and therefore just stop recursion in `OAuth.intercept()` method and return the response to the caller with the 401 or 403 status.

PR #6289, opened 2017-08-11
fix a bug whereby OAuth intercept method calls itself forever

ali-jawad:
When the response is 401 or 403, the interceptor assumes that it is because of token so it keeps retrying the request with new token. However if the response status is due to the fact that authorization to the api endpoint failed, and not due to invalid token, then a recursive operation of updating token and re-requesting the failed request happens.

### PR checklist

For second point, I committed the files affected by my change. There were many generated files by the shell scripts which I ignored. Ran maven integration tests and they pass.

### Description of the PR

This PR fixes the bug described in https://github.com/swagger-api/swagger-codegen/issues/6288 (and above too)

The fix is to not retry making the api request when receiving 401 (or 403) for a second time, and just return the 401 (or 403) response.


ali-jawad:
Sorry there is a small mistake so I closed this PR. Please refer to https://github.com/swagger-api/swagger-codegen/pull/6290 instead

PR #6290, opened 2017-08-11
[Java] fix a bug whereby OAuth intercept method calls itself forever

ali-jawad:
When the response is 401 or 403, the interceptor assumes that it is because of token so it keeps retrying the request with new token. However if the response status is due to the fact that authorization to the api endpoint failed, and not due to invalid token, then a recursive operation of updating token and re-requesting the failed request happens.

### PR checklist

For second point, I committed the files affected by my change. There were many generated files by the shell scripts which I ignored. Ran maven integration tests and they pass.

### Description of the PR

This PR fixes the bug described in https://github.com/swagger-api/swagger-codegen/issues/6288 (and above too)

The fix is to not retry making the api request when receiving 401 (or 403) for a second time, and just return the 401 (or 403) response.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/retrofit/auth/OAuth.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/retrofit/auth/OAuth.mustache
@@ -112,7 +112,7 @@ public class OAuth implements Interceptor {
             Response response = chain.proceed(rb.build());
     
             // 401/403 most likely indicates that access token has expired. Unless it happens two times in a row.
-            if ( response != null && (response.code() == HTTP_UNAUTHORIZED | response.code() == HTTP_FORBIDDEN) && updateTokenAndRetryOnAuthorizationFailure ) {
+            if ( response != null && (response.code() == HTTP_UNAUTHORIZED || response.code() == HTTP_FORBIDDEN) && updateTokenAndRetryOnAuthorizationFailure ) {
                 if (updateAccessToken(requestAccessToken)) {
                     return retryingIntercept( chain, false );
                 }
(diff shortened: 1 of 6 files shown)


wing328:
@ali-jawad thanks for the PR.

cc @cbornet

cbornet:
It's been a long time...
Normally when updating the token the second time, you get the same token and so the condition on updateToken() prevents the request from being played indefinitely.
@ali-jawad do you get a new access token each time ?

ali-jawad:
Thanks for looking into my PR.
> @ali-jawad do you get a new access token each time ?

For my case yes. We use JWT tokens which are self validate-able. So the server does not have to remember them (e.g. to resend the same token). Furthermore, feel free to refer to OAuth 2 RFC which does not state the requirement of the token being same when requested for subsequent time shortly or so.
https://tools.ietf.org/html/rfc6749#section-3.2

I hope this helps.

modules/swagger-codegen/src/main/resources/Java/libraries/retrofit/auth/OAuth.mustache
```diff
@@ -107,11 +111,10 @@ public class OAuth implements Interceptor {
             //Execute the request
             Response response = chain.proceed(rb.build());
     
-            // 401 most likely indicates that access token has expired.
-            // Time to refresh and resend the request
-            if ( response != null && (response.code() == HTTP_UNAUTHORIZED | response.code() == HTTP_FORBIDDEN) ) {
+            // 401/403 most likely indicates that access token has expired. Unless it happens two times in a row.
+            if ( response != null && (response.code() == HTTP_UNAUTHORIZED | response.code() == HTTP_FORBIDDEN) && updateTokenAndRetryOnAuthorizationFailure ) {
```
cbornet: Isn't there a `|` missing ?

cbornet:
OK. In that case your PR is indeed needed !

modules/swagger-codegen/src/main/resources/Java/libraries/retrofit/auth/OAuth.mustache
ali-jawad: Absolutely. Good catch. It's fixed now.

wing328:
@cbornet thanks for reviewing the change.

@ali-jawad thanks for the PR, which has been merged into master.

ali-jawad:
Many thanks

PR #6293, opened 2017-08-11
[Scala] Strip package name from the class name

ramzimaalej:
This goal of this PR is to fix this [issue](https://github.com/swagger-api/swagger-codegen/issues/6234).

In fact, When generating a client using scala module, the class name of the generated models contains the package like this "models.WebsiteBodyModel" which causes compilation errors.

This PR replaces this old one: https://github.com/swagger-api/swagger-codegen/pull/6235

### DIFF
--- a/modules/swagger-codegen/pom.xml
+++ b/modules/swagger-codegen/pom.xml
@@ -278,6 +278,12 @@
            <artifactId>commonmark</artifactId>
            <version>0.9.0</version>
         </dependency>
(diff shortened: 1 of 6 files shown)


wing328:
@ramzimaalej thanks for the fix.

cc @foxmk, who's also working on the Scala client.

wing328:
@ramzimaalej thanks for the PR, which has been merged into master.

Issue #2153, opened 2016-02-16
Provide a JAX-RS 2.0 client template with minimal or zero dependencies.


keithchong:
There are currently Jersey 2 and Jersey 1.18 JAX-RS clients with/without JSON Jackson processing, 
These have additional dependencies on Jackson annotations, base 64 encoding (migbase64), joda-time, etc.

There should be a template that should use strictly JAX-RS 2.0 Client APIs, (no Jersey dependencies) and have minimal (or zero) dependencies on 'third' party jars.

wing328:
@keithchong thanks for the feedback. Is that something you've cycle to contribute?

keithchong:
Yes, I will contribute.

wing328:
@keithchong thanks. Here is the guideline for contributing:

https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md

jfiala:
version: 2.1.5
language: Java

template file: swagger-codegen/src/main/resources/Java/auth/HttpBasicAuth.mustache

Related to #2153

The reference to com.migcomponents.migbase64.Base64 in HttpBasicAuth.mustache could easily be replaced by the already present com.sun.jersey.core.util.Base64:

``` java
// old: import com.migcomponents.migbase64.Base64;
import com.sun.jersey.core.util.Base64;
// old: headerParams.put("Authorization", "Basic " + Base64.encodeToString(str.getBytes("UTF-8"), false));
headerParams.put("Authorization", "Basic " + new String(Base64.encode(str.getBytes("UTF-8")), "UTF-8"));
```
Then the dependency to com.brsanthu:migbase64 could be removed.

arthurdm:
To be consistent with (https://github.com/swagger-api/swagger-codegen/pull/3080), this new client generation code should use "-l java-spec".

arthurdm:
I just realized that client-gen uses config options for different libraries instead of new languages.  in this case then, it should be a new library under -l java.   Would be nice if the client-gen and server-gen were consistent, but ah well, something for the future.

wing328:
@arthurdm thanks for the feedback. We definitely want to keep things consistent and kudos to @cbornet who has already started doing that.

If you want to lend a helping hand, here is a list of tasks you can help out:

https://github.com/swagger-api/swagger-codegen/issues?q=is%3Aopen+is%3Aissue+label%3A%22Client%3A+Java%22+label%3A%22Need+community+contribution%22

[...]

PR #6296, opened 2017-08-11
[2152] Do not inject Jackson dependencies - JAX-RS spec client

keithchong:
### PR checklist

### Description of the PR


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
```diff
@@ -224,7 +225,12 @@ public void processOpts() {
             apiTemplateFiles.put("apiImpl.mustache", "Impl.java");
             apiTemplateFiles.put("rxApiImpl.mustache", ".java");
             supportingFiles.remove(new SupportingFile("manifest.mustache", projectFolder, "AndroidManifest.xml"));
-        } else {
+        } else if ("jaxrs-spec".equals(getLibrary())) {
+        	// Don't include Jackson and GSON
+        	additionalProperties.remove("jackson");
```
cbornet: Do you need to remove them since they were not added ?

keithchong: Good point.   Looks like they can be removed, so it will be an empty block.

keithchong: BTW, the last dependency on migbase64 is in Java/auth/HttpBasicAuth.mustache.  It has the import on com.migcomponents.migbase64.Base64.

We could possibly use the built-in Java 8 (API since 1.8) java.util.Base64, that will force users to use Java 8.

So from:
Base64.encodeToString(str.getBytes("UTF-8"), false));
To:
Base64.getEncoder().encodeToString(str.getBytes("UTF-8"))

What are your thoughts on this?  Then we could close off #2153 with 'zero' third party dependencies.

keithchong:
Two additional notes about the file changes:

a) I did not intentional change these files, but there appears to be two files with conflicting names in samples/client/petstore/typescript-angular2/with-interfaces/model.

apiResponse.ts and ApiResponse.ts

I was not able to keep both when using eGit.

b) Although most of the mustache files are 'similiar' to other libraries, I'd like to more critique on: modules/swagger-codegen/src/main/resources/Java/libraries/jaxrs-spec/ApiClient.mustache.
eg. in the serialize method, I made a comment on multipart/form-data.  In other libraries, it is supported.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
wing328: @keithchong that should be already done in the latest master: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/Java/auth/HttpBasicAuth.mustache#L7-L13

wing328:
@keithchong thanks for the PR. I got errors when running `mvn test` with the Java (jaxrs-spec) petstore client: https://github.com/swagger-api/swagger-codegen/tree/keithchong-2152-DoNotInjectJacksonDependencies-JaxRsSpecClient/samples/client/petstore/java/jaxrs-spec

I wonder if you can take a look to see if you can reproduce the error.

wing328:
@keithchong do you need help reproducing the issue?

keithchong:
Hi @wing328, it appears that the sample tests for jaxrs-spec are not there, correct?

wing328:
@keithchong I manually created the jaxrs-spec petstore samples when testing it. I can recreate that branch tomorrow.

wing328:
UPDATE: added back the samples in that branch, https://github.com/swagger-api/swagger-codegen/tree/keithchong-2152-DoNotInjectJacksonDependencies-JaxRsSpecClient/samples/client/petstore/java/jaxrs-spec

wing328:
@keithchong do you need help reproducing the issue?

PR #6299, opened 2017-08-12
[php] Fixed typo in codedocs (Swaagger -> Swagger)

manuelkiessling:
The mustache template has a typo in the PHP codedoc block.
This PR fixes the template and the resulting sample petstore code files.


wing328:
@manuelkiessling thanks for the fix. 

I also did a grep on "Swaagger" and confirmed only the PHP template has the typo. 
```
$ grep -R -i Swaagger modules/swagger-codegen/src/main/resources/*
modules/swagger-codegen/src/main/resources/php/model.mustache: * @author   Swaagger Codegen team
```

PR #6301, opened 2017-08-12
fix(codegen, nodejs-server): correct reponse.end

### LINKED ISSUE #6292: invalid chunked encoding when generating nodejs-server with docker image
##### Description
I've got a problem with the docker image swaggerapi/swagger-codegen-cli. 
I'm generating a nodejs-server with the docker image. This works without generating any errors. Then I try to run the server with npm install && node index.js. The server starts and I'm able to view the docs. If I try to make an api call I get following error msg from chrome: `ERR_INVALID_CHUNKED_ENCODING`.

I can fix it by adding `'Transfer-Encoding': 'chunked'` to response.writeHead in the generated /utils/writer.js.

Seems to me that either chunked encoding is not working properly or the Transfer-Encoding header is set accidentially.

Or am I doing something wrong?

With the swagger online generator everything works fine.

##### Swagger-codegen version
latest docker image from https://hub.docker.com/r/swaggerapi/swagger-codegen-cli (pushed 9 hours ago)

https://gist.github.com/sandro97git/39289ccde6153015936a42fd43603b8f

docker run -it -v {$PATH}:/local swaggerapi/swagger-codegen-cli:latest generate -i /local/ClubManagerServer_
Api.yaml -l nodejs-server -o /local/newout

##### Steps to reproduce

Generate with docker and run with `node index.js`

Remove Transfer-Encoding header

sandro97git:
Don't send code with response.end because it's already set in writeHead.

Fixes #6292

### DIFF
--- a/modules/swagger-codegen/src/main/resources/nodejs/writer.mustache
+++ b/modules/swagger-codegen/src/main/resources/nodejs/writer.mustache
@@ -39,5 +39,5 @@ var writeJson = exports.writeJson = function(response, arg1, arg2) {
     payload = JSON.stringify(payload, null, 2);
   }
   response.writeHead(code, {'Content-Type': 'application/json'});
-  response.end(payload, code);
+  response.end(payload);
 }


wing328:
@sandro97git thanks for the PR, which has been merged into master.

Issue #6281, opened 2017-08-10
[C#] Add modelPropertyNaming config option


MortenHC:
##### Description
I have recently joined a big project that is using swagger-codegen. However some custom classes has been added to account for a bunch of properties that the swagger input file did not originally contain. E.g. `SomeClass` generated by swagger could have been inherited by `SomeClassCustom` created manually.

As a newcommer to the project I have been given the task to do some cleanup, and add all the properties to the swagger input file, generate again, and then remove the custom classes that would no longer be needed.

The problem is that the custom classes added contains variables named something like:
`Brand__c`, and when added to the input file, swagger will transform it into `BrandC`. I know that swagger is generating PascalCase for C#, but in this case it will break the code previously refering to the `Brand__c` property.

As far as I understand, this could be fixed if the C# generator had the `"modelPropertyNaming": "original"` option like TypeScript, but that option is not currently a part of the C# generator as far as I can tell. In a case like this, I think it would be a nice option to use with the C# generator, even though it would be conflicting with the conventions for C# code.

wing328:
@MortenHC thanks for sharing more about your use case. I agree with you that the only solution is to add the `modelPropertyNaming` option so as to name the C# variables in a different way.

Would you have time to add that option?

jimschubert:
NOTE: The merge PR adds to C# client, per @wing328 suggestion the option should be moved to the base class so all C# based generators can benefit.

@MortenHC thanks for that PR, it's a nice addition.

PR #6305, opened 2017-08-13
[C#] Added modelPropertyNaming option to the csharp generator

MortenHC:
#6281 
Almost 100% identical with the implementation of the option in the Scala generator


wing328:
@MortenHC thanks for the PR. What about making the enhancement to AbstractCSharpCodegen.java so that other C# generators (e.g. NancyFx, etc) can also benefit from this PR?

MortenHC:
Feel free to move things around. I unfortunately don't have the time to further implement and test the option :)

wing328:
@MortenHC that's ok. We'll see if other contributors can help on this.

PR #6323, opened 2017-08-16
Kotlin: Fix trailing comma for data classes

red-hood:
If a data class has no optional parameters, a trailing comma (which was
supposed to delimit the list of required and optional parameters) will be
inserted after the last required paramter, leading to a syntax error in
Kotlin.

### PR checklist

- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.  -> no changes in git except for version file

### Description of the PR

The reason for the change is that Kotlin does not allow trailing commas in the argument list of data classes, which are added by the template renderer in case there are only required Variables, which leads to a syntax error. The change adds the flag hasOptional in the CodegenModel which signifies if a model has optional Variables, and uses this flag in the template to only emit a comma if both hasRequired and hasOptional are true.


red-hood:
It seems the failing build on travis is a generic problem. I do not have much experience with it, so maybe someone else could have look?

wing328:
@red-hood please run `./bin/kotlin-client-petstore.sh` to update Petstore samples so that the CI can verify the change.

cc @jimschubert

jimschubert:
Looks good.

red-hood:
Running ./bin/kotlin-client-petstore.sh upates the version in b/samples/client/petstore/kotlin/.swagger-codegen/VERSION. Is that intended?

red-hood:
The errors in circleci seem to be some dependency problem in the java petstore, most likely not related to my changes.

jimschubert:
@red-hood it'll update depending on what branch you're on. That VERSION file is for reference of client consumers, so it doesn't hurt to update the file in the samples code.

wing328:

@red-hood thanks for the fix, which has been merged into master.

PR #6333, opened 2017-08-16
[python] Fix body check in rest.py

tubaman:
Checking for "trueness" here is not sufficient.  An empty list has a truth
value of false so it will not be encoded as json and sent as the request
body as it should.  Instead we need to check explicitly if body is None.


wing328:
@tubaman the fix looks good to me (but I'm not expert in Python)

cc @taxpon @scottrw93 @frol for review.

frol:
LGTM

@tubaman Thanks!

PR #6340, opened 2017-08-19
[Swift3] Fix unwrapRequired parameter which wouldn't compile

jaz-ah:
### PR checklist

### Description of the PR

As of commit c65e663a11c60023df0d4fb9e87761627f557c1c, enabling the unwrapRequired flag create code that wouldn't compile.  I fixed that and had to add two decoder methods to make sure that it was still functional at runtime as well.   Regenerated the samples (looks like regeneration brought in some other changes to sample code that wasn't related to this PR).


jaz-ah:
@jgavris not sure if you've seen this or if you use the unwrapRequired flag.

jgavris:
I'm not using it yet...do we not have tests for this? Would be great to add a test that makes the build fail before fixing it.

jaz-ah:
nope doesn't look like it ever had tests - it's a good pt @jgavris - do you know how to add a suite in for this or @wing328 would you know?...

PR #6342, opened 2017-08-20
[Java] Handle Long in enum correctly

### LINKED ISSUE #6338: [JAVA] Incompatible types when enum values are hardcoded
##### Description
Improper type casting when values are hardcoded inside enum.

##### Swagger-codegen version

2.2.3

```json
    "swagger": "2.0",
       "X509Certificate": {
            "type": "object",
            "properties": {

                "public_key_length": {
                    "readOnly": true,
                    "enum": [
                        2048,
                        3072
                    ],
                    "type": "integer",
                    "description": "size measured in bits of the public/private keys used in a cryptographic algorithm",
                    "format": "int64"
                }   
		}
	}
```

on MAC OS:
swagger-codegen generate -i test.json -l java

Also with different swagger versions with maven

##### Result (Generated code)
```java
  /**
   * size measured in bits of the public/private keys used in a cryptographic algorithm
   */
  @JsonAdapter(PublicKeyLengthEnum.Adapter.class)
  public enum PublicKeyLengthEnum {
    NUMBER_2048(2048),
    
    NUMBER_3072(3072);  // these are the problematic lines

    private Long value;

    PublicKeyLengthEnum(Long value) {
      this.value = value;
    }

    public Long getValue() {
      return value;
    }

    @Override
    public String toString() {
      return String.valueOf(value);
    }

    public static PublicKeyLengthEnum fromValue(String text) {
      for (PublicKeyLengthEnum b : PublicKeyLengthEnum.values()) {
        if (String.valueOf(b.value).equals(text)) {
          return b;
        }
      }
      return null;
    }

    public static class Adapter extends TypeAdapter<PublicKeyLengthEnum> {
      @Override
      public void write(final JsonWriter jsonWriter, final PublicKeyLengthEnum enumeration) throws IOException {
        jsonWriter.value(enumeration.getValue());
      }

      @Override
      public PublicKeyLengthEnum read(final JsonReader jsonReader) throws IOException {
        Long value = jsonReader.nextLong();
        return PublicKeyLengthEnum.fromValue(String.valueOf(value));
      }
    }
  }

  @SerializedName("public_key_length")
  private PublicKeyLengthEnum publicKeyLength = null;

  @SerializedName("not_before")
  private Long notBefore = null;

  @SerializedName("subject")
  private String subject = null;
```
##### Compilation failure
```java
[ERROR] /Users/ysimeonov/Documents/workspace/test/target/generated-sources/src/xxx/xxx/xxx/client/model/X509Certificate.java:[134,17] incompatible types: int cannot be converted to java.lang.Long
[ERROR] /Users/ysimeonov/Documents/workspace/test/target/generated-sources/src/xxx/xxx/xxx/client/model/X509Certificate.java:[136,17] incompatible types: int cannot be converted to java.lang.Long
[ERROR] -> [Help 1]
```
##### Workaround
I manually modified the following:
```java
    NUMBER_2048(new Long(2048)),
    
    NUMBER_3072(new Long(3072)); 
```

dariota:
### PR checklist

- [X] Filed the PR against the correct branch: master for non-breaking changes and `3.0.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

When used as an enum value, long values (those marked with format: int64 in an OAS spec) were not being suffixed with l. This means autoboxing (as Long is used in place of long) fails, preventing the code from compiling.

Sample broken before this change (build with  `-l java -Dmodels` or you'll get an NPE as there are no paths) and fixed after the change.

```
swagger: 2.0
definitions:
  BrokenEnum:
    type: object
    properties:
      long_value:
        enum:
          - 1
          - 2
        type: integer
        format: int64
```
Resolves #6338 

Note: I ran java-petstore-all.sh to update the samples - I'm unsure which one exactly the plain `java` language maps to, nor whether to just test everything since model changes can in theory affect them all.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractJavaCodegen.java
@@ -1016,9 +1016,11 @@ public abstract class AbstractJavaCodegen extends DefaultCodegen implements Code
 
     @Override
     public String toEnumValue(String value, String datatype) {
-        if ("Integer".equals(datatype) || "Long".equals(datatype) ||
-            "Double".equals(datatype)) {
+        if ("Integer".equals(datatype) || "Double".equals(datatype)) {
             return value;
+        } else if ("Long".equals(datatype)) {
+            // add l to number, e.g. 2048 => 2048l
+            return value + "l";
         } else if ("Float".equals(datatype)) {
             // add f to number, e.g. 3.14 => 3.14f
             return value + "f";
--- a/samples/client/petstore/java/retrofit2-play24/src/main/java/io/swagger/client/model/FormatTest.java
+++ b/samples/client/petstore/java/retrofit2-play24/src/main/java/io/swagger/client/model/FormatTest.java
@@ -13,15 +13,18 @@
 
 package io.swagger.client.model;
 
+import java.util.Objects;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
-
-import javax.validation.Valid;
 import java.math.BigDecimal;
 import java.time.LocalDate;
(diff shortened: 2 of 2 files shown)


wing328:
@dariota thanks for the fix, which looks good to me. Will run some tests on coming weekend and merge accordingly.

Issue #2936, opened 2016-05-23
Using fullJavaUtil flag generates invalid classes


sohrab-:
##### Description

When using the following settings:

- language: jaxrs-cxf
- fullJavaUtil: true
- modelPackage: com.example.models

The generated classes include invalid imports such as

```java
import com.example.models.ArrayList;
import com.example.models.HashMap;
```

Try this with the official `petstore_full.yaml` and you will see `Pet.java` showing this symptoms.

##### Swagger-codegen version

```xml
<groupId>io.swagger</groupId>
<artifactId>swagger-codegen-maven-plugin</artifactId>
<version>2.1.6</version>
```

##### Swagger declaration file content or url

Any YAML, you can see the problem with official `petstore_full.yaml` from Swagger Editor site as well.

##### Command line used for generation

Used Maven Plugin:

```xml
<build>
	<plugins>
		<plugin>
			<groupId>io.swagger</groupId>
			<artifactId>swagger-codegen-maven-plugin</artifactId>
			<version>2.1.6</version>
			<executions>
				<execution>
					<goals>
						<goal>generate</goal>
					</goals>
					<configuration>
						<inputSpec>src/main/api/petstore.yml</inputSpec>
						<output>${project.build.directory}/generated-sources</output>
						<language>jaxrs-cxf</language>
						<apiPackage>com.example.apis</apiPackage>
						<modelPackage>com.example.models</modelPackage>
						<configOptions>
							<fullJavaUtil>true</fullJavaUtil>
							<sourceFolder>.</sourceFolder>
						</configOptions>
						<environmentVariables>
							<models />
							<apis />
						</environmentVariables>
					</configuration>
				</execution>
			</executions>
		</plugin>
	</plugins>
</build>
```

##### Steps to reproduce

Use the maven plugin

##### Related issues

N/A

##### Suggest a Fix

_Sorry had not had time to debug the issue yet._

wing328:
@sohrab- thanks for reporting the issue. `fullJavaUtil` is only supported in the Java API client at the moment:

https://github.com/swagger-api/swagger-codegen/search?utf8=%E2%9C%93&q=fullJavaUtil

Would you have cycle to help us implement it in the JAX-RS generator?

wing328:
Better wording for fullJavaUtil via https://github.com/swagger-api/swagger-codegen/pull/3319

PR #6357, opened 2017-08-23
[Java] Valid generated classes with `fullJavaUtil`

mkw:
Partial fix for #2936; will fix the ArrayList/HashMap import issue.

### PR checklist

### Description of the PR

All this does is avoid trying to import "ArrayList" and "HashMap" when using full java util classes.  Doing so should be save because java.util.* will be referenced via full classnames in that case, and if alternate mappings are provided for these, absolute classes can be used.


wing328:
@mkw thanks for the fix, which looks good to me.

PR #6365, opened 2017-08-23
Kotlin: Use packageName option for all classes

### LINKED ISSUE #6364: [Kotlin] packageName parameter only used for infrastructure
##### Description
When generating clients for Kotlin, the packageName parameter is only used for the infrastructure classes, types and api still use io.swagger.client as a package name prefix. I would expect the packageName parameter to be used for all generated classes.

NB Passing the --api-package and --model-package options to the CLI can be used as a workaround to set the package name for the api and model classes.

##### Swagger-codegen version
2.2.3

It can easily reproduced by petstore declaration in the repository:https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore.yaml

In the repository, run the following command:
./bin/kotlin-client-petstore.sh -DpackageName='foo.bar'

1. clone the repository
2. build the jar: mvn package
3. run the above command: ./bin/kotlin-client-petstore.sh -DpackageName='foo.bar'

##### Related issues/PRs
I am not aware of any, this is specific to Kotlin.

If api-package or model-package are not set, use the packageName parameter to determine the package name for api and models, otherwise use the value of each respectively.

red-hood:
Use the optional packageName parameter for model and api classes,
additionally to the infrastructure classes.

Fixes #6364

### PR checklist

- [ ]x Filed the PR against the correct branch: master for non-breaking changes and `3.0.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

This PR fixes #6364 by using the packageName option to build the package name for the api and model classes unless these are not explicitly set.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/KotlinClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/KotlinClientCodegen.java
@@ -200,6 +200,10 @@ public class KotlinClientCodegen extends DefaultCodegen implements CodegenConfig
 
         if (additionalProperties.containsKey(CodegenConstants.PACKAGE_NAME)) {
             this.setPackageName((String) additionalProperties.get(CodegenConstants.PACKAGE_NAME));
+            if (!additionalProperties.containsKey(CodegenConstants.MODEL_PACKAGE))
+                this.setModelPackage(packageName + ".models");
+            if (!additionalProperties.containsKey(CodegenConstants.API_PACKAGE))
+                this.setApiPackage(packageName + ".apis");
         } else {
             additionalProperties.put(CodegenConstants.PACKAGE_NAME, packageName);
         }


wing328:
@red-hood thanks for the PR, which looks good to me as other Java generators have similar logic if I remember correctly.

cc @jimschubert

jimschubert:
I've brought this up elsewhere, but maybe need to discuss in chat... programmatically accessing these generators and calling `setPackageName` after processOpts will lead to API/models being these old values, which is probably not expected.

Also, I think I modeled this after the Scala client generator logic. Would one of you be willing to check and see if that generator is missing this logic? I don't have access to a computer for the next few days or so, and I'll probably forget to check by then.

jimschubert:
btw other than my comment for discussion, this PR looks good.

wing328:
> Also, I think I modeled this after the Scala client generator logic. Would one of you be willing to check and see if that generator is missing this logic? I don't have access to a computer for the next few days or so, and I'll probably forget to check by then.

I'll do the check.

wing328:
@jimschubert thanks for reviewing.

@red-hood PR merged into master. Thanks for your contribution.

red-hood:
I just noticed that with the above solution it is possible to set api/model package name via the '--' options, but not by specifying system properties, because they are not added as cli options and therefore not handled as dynamic properties (and as such not added to additionalProperties list).

This leads to the question if it is reasonable to allow to set model/api package seperately, or if those should always be set by appending to the package name. In the first case, those options should also be added to the cliOptions, in latter they should be completely ignored.

jimschubert:
@red-hood 
It's possible to use `--additional-properties` to set model/api packages, and these are handled by DefaultCodegen#processOpts (see https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L130).

You're correct that system properties are only merged into additionalProperties from system properties if that property is listed as a cliOption. Otherwise, it's handled via [CodegenConfigurator](https://github.com/swagger-api/swagger-codegen/blob/d31d431071e12ba42a7cbe75bf0087da8b6ddfaa/modules/swagger-codegen/src/main/java/io/swagger/codegen/config/CodegenConfigurator.java#L405).  I would argue that merging system properties into additionalProperties in this way when they're marked a generator cli options is unintuitive and that the `--model-package`, `--api-package` and `--additional-properties` should be the only entry points for setting these values.

Issue #5849, opened 2017-06-15
[Swift3] Can't use property named result.


RyogaK:
##### Description
If I add a property named `result` in response, it conflicts with a variable that used in template.
Because that variable is also named `result`.

##### Swagger-codegen version
2.3.0

##### Command line used for generation
java -jar swagger-codegen-cli.jar generate -i api-docs.json -l swift3 -o HogeAPI --additional-properties responseAs=RxSwift,unwrapRequired=true

##### Steps to reproduce
- Declare a property that named `result` in API response.
- mvn package

##### Suggest a Fix
I think, all variables (not just `result`) should has any prefix to avoid conflict with properties.
e.g) 
x `let result = ...`
o `let _result = ...`
I think `_` is not usually used for property name.

wing328:
@RyogaK right, local variables should be named differently. For other langauges, we use the prefix "localVar": https://github.com/swagger-api/swagger-codegen/search?utf8=%E2%9C%93&q=localVarPath&type=

I'll work on this next weekend (24th, 25th) if no one has started working on this.

cc @jaz-ah @jgavris @Edubits

wing328:
I'm busy with other tasks and may not be able to work on this by the end of Aug.

If anyone wants to work on this, please reply to let us know.

jaz-ah:
i'm hitting this issue as well - will fix in my next changes today/tomorrow.

jaz-ah:
this is now fixed in latest master - @wing328 could you close this one out plz?  thx.

PR #6377, opened 2017-08-25
[Swift3] fix for unwrapRequired compilation when we have a "result" as one of the parameters

jaz-ah:
### PR checklist

### Description of the PR

If you had a parameter named "result" in one of the models getting decoded, the generated code would fail because we would have a:
guard let result = ..... 
and then at the bottom we would have a:
let result =

which would result in duplicate definitions of "result" - I took care of this by just _'ing the result in our mustache file so it should never conflict with any definition of result.

Address: https://github.com/swagger-api/swagger-codegen/issues/5849


jaz-ah:
@jgavris @wing328 for review -thx

jaz-ah:
@wing328 let me know if you have a chance to take a look at this one - I'll followup w/ another change to merge this and some other changes to swift4.

wing328:
@jaz-ah yup, make senses to prefix local variable `result` with underscore. (for other generators, we use something like var_local_{{name}})

Issue #6381, opened 2017-08-26
[Java-Playframework] UUID parameter is not well parsed


tzimisce012:
##### Description

If a query parameter or header has the format UUID, this is not well parsed and the code generated don't compile

##### Swagger-codegen version

2.3.0-SNAPSHOT

##### Swagger declaration file content or url

For a swagger snipet like:

```
          "parameters": {
              "Correlator": {
                "type": "string",
                "format": "uuid",
                "description": "Correlation id for the different services"
              }
            },
```

the code generate is this:

```
String valuecorrelator = request().getHeader("Correlator");
        UUID correlator;
        if (valuecorrelator != null) {
            correlator = (String)valuecorrelator;

```

and this will not compile.

##### Command line used for generation

language JavaPlayFramework

##### Suggest a fix/enhancement

Add the option to parse UUID with `UUID.fromString()` method

JFCote:
This will have to be checked by @wing328 and other people who know the project better than me.

The problem I see is this (from the documentation)

> Primitives have an optional modifier property format. Swagger uses several known formats to more finely define the data type being used. However, the format property is an open string-valued property, and can have any value to support documentation needs. Formats such as "email", "uuid", etc., can be used even though they are not defined by this specification. Types that are not accompanied by a format property follow their definition from the JSON Schema (except for file type which is defined above).

Even if the spec is considered "valid" when using formats such as "email" and "uuid", I understand the doc is considering them as "custom" and I'm not sure tools like swagger-codegen should implements all custom field formats in the world. But maybe it's ok and should be implemented. I will review your PR with my comments in case it should be implemented.

wing328:
Closed via #6382

PR #6382, opened 2017-08-26
Fix Uuid parse

tzimisce012:
### PR checklist

### Description of the PR

This will fix issue #6381


tzimisce012:
What do you think @JFCote?

modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenParameter.java
```diff
@@ -14,7 +14,7 @@
 
     public String example; // example value (x-example)
     public String jsonSchema;
-    public boolean isString, isInteger, isLong, isFloat, isDouble, isByteArray, isBinary, isBoolean, isDate, isDateTime;
+    public boolean isString, isInteger, isLong, isFloat, isDouble, isByteArray, isBinary, isBoolean, isDate, isDateTime, isUuid;
```
JFCote: Since this is a change in the very generic CodegenParameter, this will need the stamps from other collaborator since it can have effects on other generators.

modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenProperty.java
```diff
@@ -39,7 +39,7 @@
     public boolean hasMore, required, secondaryParam;
     public boolean hasMoreNonReadOnly; // for model constructor, true if next properyt is not readonly
     public boolean isPrimitiveType, isContainer, isNotContainer;
-    public boolean isString, isInteger, isLong, isFloat, isDouble, isByteArray, isBinary, isFile, isBoolean, isDate, isDateTime;
+    public boolean isString, isInteger, isLong, isFloat, isDouble, isByteArray, isBinary, isFile, isBoolean, isDate, isDateTime, isUuid;
```
JFCote: Same thing here

modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenResponse.java
```diff
@@ -11,7 +11,7 @@
     public List<Map<String, Object>> examples;
     public String dataType, baseType, containerType;
     public boolean hasHeaders;
-    public boolean isString, isInteger, isLong, isFloat, isDouble, isByteArray, isBoolean, isDate, isDateTime;
+    public boolean isString, isInteger, isLong, isFloat, isDouble, isByteArray, isBoolean, isDate, isDateTime, isUuid;
```
JFCote: same thing here

modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
```diff
@@ -1706,7 +1706,7 @@ public CodegenProperty fromProperty(String name, Property p) {
             property.isFile = true;
         }
         if (p instanceof UUIDProperty) {
-            property.isString = true;
+            property.isUuid = true;
```
JFCote: this seems dangerous to me. You should at least keep the isString = true and add your line just after. But, that would be strange because it would be the first case where the "is" field could be true for 2 variables. For example, right now you can't be isString and isDate at the same time.

modules/swagger-codegen/src/main/resources/JavaPlayFramework/conversionBegin.mustache
```diff
@@ -1 +1 @@
-{{#isBoolean}}Boolean.valueOf({{/isBoolean}}{{#isInteger}}Integer.parseInt({{/isInteger}}{{#isDouble}}Double.parseDouble({{/isDouble}}{{#isLong}}Long.parseLong({{/isLong}}{{#isFloat}}Float.parseFloat({{/isFloat}}{{#isDateTime}}OffsetDateTime.parse({{/isDateTime}}
\ No newline at end of file
+{{#isBoolean}}Boolean.valueOf({{/isBoolean}}{{#isInteger}}Integer.parseInt({{/isInteger}}{{#isDouble}}Double.parseDouble({{/isDouble}}{{#isLong}}Long.parseLong({{/isLong}}{{#isFloat}}Float.parseFloat({{/isFloat}}{{#isString}}(String){{/isString}}{{#isUuid}}UUID.fromString({{/isUuid}}{{#isDateTime}}OffsetDateTime.parse({{/isDateTime}}
```
JFCote: Here, you need to remove the "isString", it is not needed since it's always a string when the conversionBegin is called. If you need this for you case when isUuid and isString are true at the same time, please put the (String) in the isUUid section only.

modules/swagger-codegen/src/main/resources/JavaPlayFramework/conversionEnd.mustache
```diff
@@ -1 +1 @@
-{{#isBoolean}}){{/isBoolean}}{{#isInteger}}){{/isInteger}}{{#isDouble}}){{/isDouble}}{{#isLong}}){{/isLong}}{{#isFloat}}){{/isFloat}}{{#isDateTime}}){{/isDateTime}}
\ No newline at end of file
+{{#isBoolean}}){{/isBoolean}}{{#isInteger}}){{/isInteger}}{{#isDouble}}){{/isDouble}}{{#isLong}}){{/isLong}}{{#isFloat}}){{/isFloat}}{{#isString}}{{/isString}}{{#isUuid}}){{/isUuid}}{{#isDateTime}}){{/isDateTime}}
```
JFCote: Same thing here, remove the isString section, not needed.

tzimisce012:
Ok, i'll wait the second opinion. This looks like a deeper change on the code that involves the codegen in general, and not just the javaplayframework codegen.

JFCote:
@tzimisce012 Yeah, exactly. It involves all generators.

modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
wing328: I would also suggest keeping `isString = true` to make the change backward compatible as we don't have the resource to review all the templates due to this change.

JFCote:
@wing328 The latest changes seems good to me. If the fact that we deal with "custom" format like guid is ok with you, then it's a go for the merge.

Thanks @tzimisce012 for the great work.

wing328:
Thanks @tzimisce012 for the enhancement, which has been merged into master.

cc @cbornet @ePaul

PR #6383, opened 2017-08-27
[Kotlin] Fix model enum generation

red-hood:
Run postProcessModelsEnum on model enums, so that enumVars get populated,
and fix the model enum template.

### PR checklist

### Description of the PR
see above

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/KotlinClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/KotlinClientCodegen.java
@@ -11,6 +11,7 @@ import java.io.File;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
 
 public class KotlinClientCodegen extends DefaultCodegen implements CodegenConfig {
     static Logger LOGGER = LoggerFactory.getLogger(KotlinClientCodegen.class);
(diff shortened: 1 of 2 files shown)


red-hood:
PS I was not able to determine from which context isString is resolved. Neither CodegenModel, allowableValues nor enumVars have such an attribute. Could someone point out to from which context this is taken?

wing328:
> PS I was not able to determine from which context isString is resolved. Neither CodegenModel, allowableValues nor enumVars have such an attribute. Could someone point out to from which context this is taken?

Would adding the model debug help when generating the client?
```
-DdebugModels=true
```

red-hood:
Yes, thanks, that helped. So basically isString is not in the hierarchy of any contexts, but the quotation marks are added by toEnumValue. I updated the pull request accordingly.

wing328:
Kotline is using `modules/swagger-codegen/src/test/resources/2_0/petstore.yaml` to generate Petstore client. We can use https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml, which covers a lot more edge cases (e.g. enum models, enum of integers, etc)

Bijnagte:
Why was this reverted?

wing328:
@Bijnagte it's not reverted and I can find the code change covered by this PR in the latest master: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/KotlinClientCodegen.java#L385-L388

Issue #6371, opened 2017-08-24
[python] Request: Generate an "unsafe" constructor to models that allows missing required elements


jdevera:
I'm testing with the master branch and the generated models are very nicely checking for all required parameters.

However, I am using the client to write tests against an API and I need to be able to create invalid requests.

If would be great if there was a way to construct the model objects without the `required` and without the range validation.

Such option could even be configurable in the generator options.

How would you go about creating such thing? Currently I see the validation happens in the property setters, which are all called in the constructor. This makes it a bit complex to have both the validated paths and the unsafe ones. However, while I want to create invalid objects in my tests, I still want to rely on the model validation for the API **responses**.

Let's discuss this, I have some capacity to work on this.

jdevera:
How about having a `_validate=True` keyword argument to every constructor? We can store the flag in the model and then check it in all property setters to perhaps bypass the validation.

frol:
Just for the reference, a solution was proposed here: https://github.com/swagger-api/swagger-codegen/pull/6390#issuecomment-325954400, so there is no need for changes in Swagger-Codegen.

PR #6390, opened 2017-08-28
Optional model validation in python client

jdevera:
### PR checklist

### Description of the PR

This PR introduces a new parameter _validated, which defaults to True (backwards compatible) in the generated python models for the python client. When set to False, all value setters in the model object will bypass validation.

This becomes particularly useful when attempting to send wrong values to the API on purpose. For instance, with the intention of testing API error responses.

The problem is referenced in issue #6371

### DIFF
--- a/modules/swagger-codegen/src/main/resources/python/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/model.mustache
@@ -48,11 +48,12 @@ class {{classname}}(object):
     }
 {{/discriminator}}
 
-    def __init__(self{{#vars}}, {{name}}={{#defaultValue}}{{{defaultValue}}}{{/defaultValue}}{{^defaultValue}}None{{/defaultValue}}{{/vars}}):
+    def __init__(self{{#vars}}, {{name}}={{#defaultValue}}{{{defaultValue}}}{{/defaultValue}}{{^defaultValue}}None{{/defaultValue}}{{/vars}}, _validated=True):
         """
         {{classname}} - a model defined in Swagger
         """
 
+        self._is_model_validated = _validated
 {{#vars}}
         self._{{name}} = None
 {{/vars}}
@@ -93,6 +94,12 @@ class {{classname}}(object):
         :param {{name}}: The {{name}} of this {{classname}}.
         :type: {{datatype}}
         """
+
+        if not self._is_model_validated:
+            # If this model was built without validation, then simply set the
+            # value here and quickly return, skipping all possible validation
+            self._{{name}} = {{name}}
+            return
 {{#required}}
         if {{name}} is None:
             raise ValueError("Invalid value for `{{name}}`, must not be `None`")
--- a/samples/client/petstore/python/docs/StoreApi.md
+++ b/samples/client/petstore/python/docs/StoreApi.md
@@ -94,7 +94,7 @@ This endpoint does not need any parameter.
 
 ### Return type
 
-[**dict(str, int)**](dict.md)
+**dict(str, int)**
 
 ### Authorization
 
--- a/samples/client/petstore/python/petstore_api/models/additional_properties_class.py
+++ b/samples/client/petstore/python/petstore_api/models/additional_properties_class.py
@@ -40,11 +40,12 @@ class AdditionalPropertiesClass(object):
         'map_of_map_property': 'map_of_map_property'
     }
 
-    def __init__(self, map_property=None, map_of_map_property=None):
+    def __init__(self, map_property=None, map_of_map_property=None, _validated=True):
         """
         AdditionalPropertiesClass - a model defined in Swagger
         """
 
+        self._is_model_validated = _validated
         self._map_property = None
         self._map_of_map_property = None
         self.discriminator = None
@@ -73,6 +74,12 @@ class AdditionalPropertiesClass(object):
         :type: dict(str, str)
         """
 
+        if not self._is_model_validated:
+            # If this model was built without validation, then simply set the
+            # value here and quickly return, skipping all possible validation
+            self._map_property = map_property
+            return
+
         self._map_property = map_property
 
     @property
@@ -93,6 +100,12 @@ class AdditionalPropertiesClass(object):
         :param map_of_map_property: The map_of_map_property of this AdditionalPropertiesClass.
         :type: dict(str, dict(str, str))
         """
+
+        if not self._is_model_validated:
+            # If this model was built without validation, then simply set the
+            # value here and quickly return, skipping all possible validation
+            self._map_of_map_property = map_of_map_property
+            return
 
         self._map_of_map_property = map_of_map_property
 
--- a/samples/client/petstore/python/petstore_api/models/animal.py
+++ b/samples/client/petstore/python/petstore_api/models/animal.py
@@ -45,11 +45,12 @@ class Animal(object):
         '': 'Cat'
     }
 
-    def __init__(self, class_name=None, color='red'):
+    def __init__(self, class_name=None, color='red', _validated=True):
         """
         Animal - a model defined in Swagger
         """
 
+        self._is_model_validated = _validated
         self._class_name = None
         self._color = None
         self.discriminator = 'className'
@@ -76,6 +77,12 @@ class Animal(object):
         :param class_name: The class_name of this Animal.
         :type: str
         """
+
+        if not self._is_model_validated:
+            # If this model was built without validation, then simply set the
+            # value here and quickly return, skipping all possible validation
+            self._class_name = class_name
+            return
         if class_name is None:
             raise ValueError("Invalid value for `class_name`, must not be `None`")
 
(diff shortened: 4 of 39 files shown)


wing328:

jdevera:
That is fixed now, thanks for noticing @wing328

frol:
@jdevera To be honest, I don't think this is the right way to solve the problem you have. I would implement a faker helper which would inspect `Model.swagger_type` dict and set valid values to all the fields. Thus, you don't need to patch swagger-codegen. You may find [`factory_boy`](https://factoryboy.readthedocs.io) and [`faker`](http://faker.readthedocs.io/en/master/) modules handy.

jdevera:
@frol I didn't understand your suggestion, sorry. My problem was that the generated models do validation unconditionally based on the API spec (good thing) but I needed to create model objects with *invalid* values or missing values for required properties. Those invalid values are quite specific and I already have them in my tests. In fact, they used to work before the additional validation was added.

frol:
Given the fact that you won't use `_validate=False` in production, I don't see a point in testing your code with invalid values since invalid values will raise validation error automatically if server response is invalid.

jdevera:
I only use the python client for the testing my server API. If the server response is invalid, that raises validation error because the responses, when parsed, will not have the `_validate=False`. That is okay.

The problem comes when I want to test how the server deals with wrong requests. I found that client side validation was preventing me from even sending these to the server.

frol:
It is unexpected that model validation blocks requests since parameter parsing is separate from model deserialization. Please, share swagger.json snippet.

jdevera:
Still have the certainty that we have a misunderstanding. My problem and the case I am exposing is completely unrelated to deserialization.

I simply need to issue a request with values that violate the schema. For instance, I want to leave a required field unset, or set a field to 5 when the max is defined as 4. With the generated models I am unable to do this because the property setters include validation and those are called by the constructor.

This is the change that I saw as less intrusive and it solves my problem entirely. If there is a way in which I can accomplish this without having to hand write the requests but rather still using the model somehow, I'd love to hear it.

And just to clarify on your point, model validation *does* block requests if you try to write an invalid request since the validation will raise before you get to pass the model object to the API.

frol:
I cannot test the code right now, but I don't understand how model validation can influence the *request*. Model validation should only get called AFTER the *response* is received.

jdevera:
Not only, the model objects are used in two scenarios:
* The one you mention: A response is received and then it is parsed into the model objects
* The one I care about here: You manually build your request using the model objects and then pass it to the API methods

frol:
@jdevera Please, provide the code snippet of what you want to achieve, otherwise "build request using model object" sounds broken to me.

jdevera:
All generated APIs (under the `api` module within the generated swagger_client, take parameters not directly as json, but rather as objects from the model. This is an example from my generated code:

```python
    def create_organization(self, create_organization, **kwargs):
        """
        Create organization
        This method makes a synchronous HTTP request by default. To make an
        asynchronous HTTP request, please pass async=True
        >>> thread = api.create_organization(create_organization, async=True)
        >>> result = thread.get()

        :param async bool
        :param CreateOrganization create_organization: (required)
        :return: CreateOrganizationResponse
                 If the method is called asynchronously,
                 returns the request thread.
```

That `CreateOrganization` parameter type mentioned there is a model object, just like the response type `CreateOrganizationResponse` also generated by the codegen.

```python
class CreateOrganization(object):
    """
    NOTE: This class is auto generated by the swagger code generator program.
    Do not edit the class manually.
    """

    """
    Attributes:
      swagger_types (dict): The key is attribute name
                            and the value is attribute type.
      attribute_map (dict): The key is attribute name
                            and the value is json key in definition.
    """
    swagger_types = {
        'name': 'str',
        'secret': 'int',
    }

    attribute_map = {
        'name': 'name',
        'secret': 'secret',
    }

    def __init__(self, name=None, secret=None, _validated=True):
        """CreateOrganization - a model defined in Swagger"""
        self._is_model_validated = _validated
        self._name = None
        self._secret = None

        self.discriminator = None

        self.name = name
        self.secret = secret
```


The way I use this is:

```python

import swagger_client as sc

org_api = sc.OrganizationApi()
response = org_api.create_organization(
        create_organization=sc.CreateOrganization(name="foobar", secret=2)
    )
```

As you see here, I created my request using the model (building a CreateOrganization object). Then the response will be parsed into a CreateOrganizationResponse object, and I expect that to be always valid, as it comes from the server, and if it isn't, I also want the validation to happen.

However, now I want to attempt to create an organization with a secret value of -1 but in the API spec it is defined as required and has a minumum of 0, so the generated property setter looks like this (without my change)

```python
    @secret.setter
    def secret(self, secret):
        """
        Sets the secret of this CreateOrganization.
    
        :param secret: The secret of this CreateOrganization.
        :type: int
        """
        if secret is None:
            raise ValueError("Invalid value for `secret`, must not be `None`")
        if secret is not None and secret < 0:
            raise ValueError("Invalid value for `secret`, must be a value greater than or equal to `0`")

        self._secret = secret
```

As you can see, the constructor above did:

```python
self.secret = secret
```

Which in turn calls this setter and performs this validation, therefore, when I try to do this:

```
org_api.create_organization(
        create_organization=sc.CreateOrganization(name="foobar2", secret=-1)
    )
```

I will get a ValueError because -1 is not a valid value for `secret`. This happens before anything is sent to the server. Hence my mention of the response.

With my change, this property setter starts with this:

```python
        if not self._is_model_validated:
            self._secret = secret
            return
```
Which bypassed all subsequent validation and allows me to then send the invalid request to the API.

I don't see how this is broken. Let me know if I can help further explain this.

jdevera:
You can also see this pattern in the [example of the petstore (add pet)](https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/python/docs/PetApi.md#add_pet):

```python
from __future__ import print_function
import time
import petstore_api
from petstore_api.rest import ApiException
from pprint import pprint

# Configure OAuth2 access token for authorization: petstore_auth
configuration = petstore_api.Configuration()
configuration.access_token = 'YOUR_ACCESS_TOKEN'

# create an instance of the API class
api_instance = petstore_api.PetApi(petstore_api.ApiClient(configuration))
body = petstore_api.Pet() # Pet | Pet object that needs to be added to the store

try: 
    # Add a new pet to the store
    api_instance.add_pet(body)
except ApiException as e:
    print("Exception when calling PetApi->add_pet: %s\n" % e)
```

Where you can see the body being created as an object from the model (`Pet`)

frol:
```python
create_organization = sc.CreateOrganization.__new__(sc.CreateOrganization)
create_organization.__dict__ = {'_name': None, '_secret': -1}

response = org_api.create_organization(create_organization=create_organization)
```

Will this work for you?

jdevera:
It does!

However, I find the optional parameter to be a lot cleaner and readable on the calling code.

If you are reluctant to include this because it is perceived as confusing for most, how about adding it conditionally based on generator configuration?

I'd be happy to add those changes too.

frol:
You can create a helper function to make your code cleaner:

```python
def with_invalid_data(model_cls, **kwargs):
    model = model_cls.__new__(model_cls)
    assert set(kwargs.keys()) <= set(model_cls.swagger_types.keys())
    model.__dict__ = {'_%s' % name: value for name, value in kwargs.items()}
    return model

create_organization = with_invalid_data(sc.CreateOrganization, name=None, secret=-1)
```

Adding more options to the codegen will increase complexity, so given the fact that you can avoid the changes, and it is only valid for testing purposes (which is not the main goal for the codegen), I would avoid this additional complexity.

jdevera:
@frol Thank you so much for taking all the time to review my changes and propose alternatives. Even for _weaponizing_ one such alternative for me :)

I agree with your assessment and I withdraw my pull request.

jdevera:
@frol Actually, when testing with a more complex case, it failed in `sanitize_for_serialization` because some of the expected fields were not there. I then tried to add them like this:

```python
def make_invalid_model(model_class, **kwargs):
    """
    Create a model object bypassing all member validation
    :param model_class: The class of the generated model that needs to be instantiated
    :param kwargs: All the model parameters, as they would be passed to the constructor
    :return: An instance of model_class
    """

    model = model_class.__new__(model_class)
    # Since we are bypassing the constructor and the setters, make sure at least that the
    # parameters passed correspond to those in the model.
    assert set(kwargs.keys()) <= set(model_class.swagger_types.keys())

    missing_attributes = set(model_class.swagger_types.keys()) - set(kwargs.keys())

    model.__dict__ = {'_' + name : value for name, value in kwargs.items()}
    if missing_attributes:
        model.__dict__.update({ '_' + name : None for name in missing_attributes})
    return model
```
However this will fail in cases there the default value is set in the spec (I'm setting it to None).

This is not a show stopper for me at all, but is there a way to programmatically obtain default values?

frol:
@jdevera You can inspect the constructor (`__init__`) kwargs: https://stackoverflow.com/questions/12627118/get-a-function-arguments-default-value

jdevera:
Just in case anybody else needs this, I ended up implementing this helper to build my invalid model objects:

https://gist.github.com/jdevera/5fed8e8ee23ff71dfa0ee5d686d844fb

frol:
@jdevera Well done!

@wing328 This was a feature request, not a bug, so you may want to adjust the label.

PR #6405, opened 2017-08-30
Java target: okhttp-gson-generated classes cleaned up, OAuth supports…

rdolejsi:
… supplying BearerProvider, ApiClient supports RequestDecorator

### PR checklist

### Description of the PR
- okhttp-gson-generated classes cleaned up, human-readable now
- OAuth supports supplying BearerProvider


wing328:

wing328:
cc @cbornet @ePaul  @bbdg @JFCote @sreeshas @jfiala

JFCote:
As fas as I'm concerned, the change seems good. The code is clean. But I don't know okhttp at all so others advice might be needed.

wing328:
cc @cbornet @ePaul  as well

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
```diff
@@ -152,13 +152,15 @@ public void processOpts() {
...
         supportingFiles.add(new SupportingFile("ApiClient.mustache", invokerFolder, "ApiClient.java"));
+        supportingFiles.add(new SupportingFile("ApiClientHolder.mustache", invokerFolder, "ApiClientHolder.java"));
         if(!"resttemplate".equals(getLibrary())) {
             supportingFiles.add(new SupportingFile("StringUtil.mustache", invokerFolder, "StringUtil.java"));
         }
 
         supportingFiles.add(new SupportingFile("auth/HttpBasicAuth.mustache", authFolder, "HttpBasicAuth.java"));
         supportingFiles.add(new SupportingFile("auth/ApiKeyAuth.mustache", authFolder, "ApiKeyAuth.java"));
         supportingFiles.add(new SupportingFile("auth/OAuth.mustache", authFolder, "OAuth.java"));
+        supportingFiles.add(new SupportingFile("auth/OAuthBearerProvider.mustache", authFolder, "OAuthBearerProvider.java"));
```
cbornet: This file shouldn't be added for retrofit library

modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClientHolder.mustache
```diff
@@ -0,0 +1,9 @@
+{{>licenseInfo}}
+
+package {{invokerPackage}};
+
+public interface ApiClientHolder {
```
cbornet: This doesn't seem to be used anywhere. What is it for ?

cbornet:
@rdolejsi could you explain what the ApiClientRequestDecorator brings compared to an Interceptor ?

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
rdolejsi: I can't make this conditional at the moment as OAuthBearerProvider is required to meaningfully implement OAuth and thus OAuth depends on this interface now - I don't seem to have identified a way to sensibly attach myself within OAuth lifecycle to supply new Bearer - one at the first server hit, a renewed one once the previous one expires.

As this is an interface only and very crucial for our needs, can you please elaborate more on why this can't be included when retrofit library is in place? Maybe I'm overlooking some basic concept..

cbornet: Retrofit has its own OAuth class that doesn't depend on OAuthBearerProvider so this would be dead code.

modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClientHolder.mustache
rdolejsi: Just to elaborate a bit more on this - the client Api's generated by swagger do not implement any interface. ApiClientHolder is what it describes - a holder of ApiClient. It acts for us both as a way to reach ApiClient within generated Api as well as a way to hold more generated Api's and organize them meaningfully (better than using plain Object).

If I think from the top of my head, the other possibility would be to introduce something like SwaggerClient or SwaggerApi which would characterize the generated Api and allow manipulations with it. It would expose ApiClient within any such Api - just like ApiClientHolder does now.

What do you think?

rdolejsi:
RequestDecorator is indeed used to intercept outgoing request builder to decorate the request (with security features, extra headers, etc.). As there are no generic interceptors (just one HttpLoggingInterceptor), I chose the name to denote the pattern it's doing - decorating.

On a second thought, we might as well introduce OutgoingRequestInterceptor to suit wider needs - what do you think?

cbornet:
Okhttp has a generic [Interceptor](http://square.github.io/okhttp/3.x/okhttp/okhttp3/Interceptor.html) interface so how about just providing an `addInterceptor(Interceptor interceptor)` method ?

rdolejsi:
Looks like Interceptor is called only after the request is built, thus allows to work with quite immutable objects. The current RequestDecorator (or be it OutgoingRequestInterceptor) allows you to work with Request.Builder, thus gaining access to the request structure before the actual request is created.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
rdolejsi: Oh, ok. That means it does not have even OAuth itself, right? So the same rules should be applicable as for OAuth class, which is in there already. Or am I missing something?

rdolejsi: Double-checked the generation procedures again and got you - multiple OAuth classes are available based on the used library, I'll change the JavaClientCodegen accordingly.

rdolejsi:
New code version for this PR: OAuthBearerProvider related to base OAuth only, ApiClientHolder.mustache moved to correct scope - tests no longer fail.

cbornet:
@rdolejsi can you generate the java client samples ?

wing328:
@rdolejsi I got some errors after updating Java okhttp-gson samples: 
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project swagger-petstore-okhttp-gson: Compilation failure: Compilation failure:
[ERROR] /private/tmp/swagger-codegen/samples/client/petstore/java/okhttp-gson/src/main/java/io/swagger/client/api/UserApi.java:[377,57] ')' expected
[ERROR] /private/tmp/swagger-codegen/samples/client/petstore/java/okhttp-gson/src/main/java/io/swagger/client/api/UserApi.java:[377,70] not a statement
[ERROR] /private/tmp/swagger-codegen/samples/client/petstore/java/okhttp-gson/src/main/java/io/swagger/client/api/UserApi.java:[377,76] ';' expected
[ERROR] /private/tmp/swagger-codegen/samples/client/petstore/java/okhttp-gson/src/main/java/io/swagger/client/api/UserApi.java:[377,97] not a statement
[ERROR] /private/tmp/swagger-codegen/samples/client/petstore/java/okhttp-gson/src/main/java/io/swagger/client/api/UserApi.java:[377,113] ';' expected
[ERROR] /private/tmp/swagger-codegen/samples/client/petstore/java/okhttp-gson/src/main/java/io/swagger/client/api/UserApi.java:[377,128] not a statement
[ERROR] /private/tmp/swagger-codegen/samples/client/petstore/java/okhttp-gson/src/main/java/io/swagger/client/api/UserApi.java:[377,144] ';' expected
```

wing328:
@rdolejsi may I know if you need help fixing the error? Seems like it's missing a ) or ;

PR #6411, opened 2017-08-31
Make the Rust codegen compile.

ahmedcharles:


wing328:
@ahmedcharles thanks for the PR.

cc @frol @farcaller

farcaller:
LGTM.

modules/swagger-codegen/src/main/resources/rust/model.mustache
```diff
@@ -28,19 +29,28 @@ impl {{classname}} {
   }
 
   {{#vars}}
+  #[allow(non_snake_case)]
```
frol: @wing328 Can we avoid this? I know that other Codegen languages transform names according to the guidelines of the target programming language.

samples/client/petstore/rust/examples/client.rs
```diff
@@ -12,32 +12,19 @@ fn main() {
...
-    let api = "http://petstore.swagger.io:80/v2";
-
     let new_pet = petstore_client::models::Pet::new("barker".to_owned(), vec![]).with_id(1337);
 
     let apicli = petstore_client::apis::client::APIClient::new(
         petstore_client::apis::configuration::Configuration::new(
             Client::configure().connector(HttpConnector::new(4, &handle)).build(&handle)));
 
-    let work = apicli.pet_api().add_pet(&new_pet)
+    let work = apicli.pet_api().AddPet(new_pet)
```
frol: It seems to be not the change this PR introduced, but it would be great if the naming conventions are aligned according to the [recommendations](https://aturon.github.io/style/naming.html) instead of putting `#[allow(non_snake_case)]` everywhere.

modules/swagger-codegen/src/main/resources/rust/model.mustache
ahmedcharles: I agree, I'm just new to swagger-codegen and wasn't sure how to get names exactly right.

In this case, I need two versions of the name, one with a leading underscore and one without, because the warning is due to having two underscores in a row. Do you know how to create two mustache variables, one with a leading underscore for reserved names and one without?

samples/client/petstore/rust/examples/client.rs
ahmedcharles: Yeah, the python generator uses `add_pet`, though I didn't look at the Java code. Do you know how to change the Java code to get this to be the different style?

modules/swagger-codegen/src/main/resources/rust/model.mustache
wing328: @frol Is there any lib in Rust that can handle JSON (de)serialization?

@ahmedcharles please share the spec of the model.

samples/client/petstore/rust/examples/client.rs
wing328: @ahmedcharles please have a look at https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RustClientCodegen.java#L323

modules/swagger-codegen/src/main/resources/rust/model.mustache
frol: @wing328 [Serde](https://github.com/serde-rs/serde) is used here for JSON (de)serialization.

@ahmedcharles You can just use `_{{ var_name }}` in the template if you need a leading underscore.

wing328: @frol I had a look at https://serde.rs/examples.html but couldn't an example how to provide a 1-1 mapping for attribute name (similar to what've done for Java, C#, etc). Do you know how we can map attribute name with Serde?

ahmedcharles: @frol The issue with the warning is that `type` is a reserved word, so the `{{name}}` is `_type`, which when combined with `with_{{name}}` becomes `with__type`, which is what creates the warning. What I need is `{{name}}` = `_type` and `{{rawName}}` = `type`, so then I can do `fn {{name}}(...)` and `fn with_{{rawName}}(...)` and never have it generate names with two underscores in a row.

@wing328 Isn't that what `#[serde(rename = "jsonName")] rust_name: String;` does?

wing328: @ahmedcharles I think so. I missed that.

frol:
@ahmedcharles FYI, it seems that the PR was merged by mistake and it was reverted... Could you, please, continue your work on the better variable naming?

frol:
Hmm, well... I have just noticed that the naming conventions are fine! Great job!

@wing328 Why did you revert the PR?

wing328:
@frol it was not reverted: https://github.com/swagger-api/swagger-codegen/commit/a6561e7d21a26abbcbb2d51ba3270ec3c7798731

PR #6398, opened 2017-08-29
- Added javax.validation.Valid annotation to modules/swagger-codegen/…

carnevalegiacomo:
**Java jaxrs-spec - javax.validation and interfaceOnly parameter management**
- Added javax.validation.Valid annotation to modules/swagger-codegen/src/main/resources/JavaJaxRS/spec/bodyParams.mustache
- Added bean validation and api description to modules/swagger-codegen/src/main/resources/JavaJaxRS/spec/headerParams.mustache
- Added api description to modules/swagger-codegen/src/main/resources/JavaJaxRS/spec/queryParams.mustache
- Added interfaceOnly parameter management to modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaJAXRSSpecServerCodegen.java to generate only interfaces and models.
- Added interfaceOnly parameter management to modules/swagger-codegen/src/main/resources/JavaJaxRS/spec/api.mustache
- Added throws java.lang.Exception to modules/swagger-codegen/src/main/resources/JavaJaxRS/spec/api.mustache

### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: master for non-breaking changes and `3.0.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

(details of the change, additional tests that have been done, reference to the issue for tracking, etc)

PR #6412, opened 2017-08-31
Issue 5431 Support jaxrs client api interfaces

jarlesat:
### PR checklist

### Description of the PR

* Added jaxrs client as interfaces as a new language "jaxrs-interface". The model is still created as pojos. It is implemented by extending JavaJAXRSSpecServerCodegen and override api.mustache with apiInterface.mustache.
* Creation of pom.xml is made optional with a cliOption "createPom", defaulting to false.
* Added shell script to create sample-files.


wing328:
@jarlesat thanks for the PR. I wonder if you can also add an auto-generated pom.xml so that developers can build the library more easily.

cc @bbdouglas @JFCote @sreeshas @jfiala @lukoyanov

jarlesat:
I added an option "createPom" that defaults to false, as that suited our use case best (we use swagger-codegen to generate the java source for the interfaces on every build). If this option is true a pom.xml will be generated. Maybe the default should be changed to true?

I also see that the generated pom.xml should be altered slightly, to generate a jar instead of a war :-)

jarlesat:
Close and reopen to trigger new build.

wing328:
@jarlesat thanks for the PR. Can this be merged into `jaxrs-spec` generator via CLI option instead of creating a separate generator?

jarlesat:
@wing328 Sure. Is _interfaceOnly_ a nice cli-option?

wing328:
@jarlesat Yup, spring generator also has this option:
```
	interfaceOnly
	    Whether to generate only API interface stubs without the server files. (Default: false)
```

jarlesat:
@wing328 Should I rebase on a recent master (followed by a forced push), or merge master into issue-5431? Or something else?

wing328:
> Should I rebase on a recent master (followed by a forced push)

That's what I usually do.

jarlesat:
Summary of changes from master: The language jaxrs-spec has to command line arguments (cli):
* generatePom (default true)
* interfaceOnly (default false)

generatePom defaults true to stay compatible with earlier versions.

If interfaceOnly is true, only transfer object classes and interfaces are generated. The RestApplication.java is skipped. The optionally generated pom.xml is altered to reflect that only interfaces are created (package type is set to jar instead of war, plugin is jar-plugin instead of war-plugin for instance).

Hence interfaceOnly behaves slightly different from what was commited in #6398: 
* The methods does not return Response, but the proper type.
* The horrible _throws Exception_ anti-pattern has been removed.

wing328:
@jarlesat I ran some tests and the result is good. Thanks for your contribution.

PR #6422, opened 2017-09-01
[Swift3] Missing decoders for Array elements

jaz-ah:
### PR checklist

### Description of the PR

Commit f7cbb2c0e62fa169de8d0c7a75465eff57cdbe4c broke the build for array decoders and then b582c46e4ba3a2e38470c02438c7e2b5250c64e2 fixed it, but removed support for when a swagger spec calls for direct return of array objects (not just when arrays are contained within other objects).  Adding that functionality back in - @jgavris @wing328 could you take a look thx.


wing328:
cc @ehyche for review as well

ehyche:
Change looks good. Decoders class was removed in swift4 module, so there is no equivalent change needed in the swift4 module.

PR #6423, opened 2017-09-01
[Swift3] Make sure decode doesn't fail on optional arguments

jaz-ah:
### PR checklist

### Description of the PR

The most recent decoder changes made it so that even if an enumeration that is marked optional, and then the server sends down a new enumeration value that swagger can't interpret yet, the decode for that object completely fails and we have no backward compatibility.  Instead, if decoding fails on an optional argument, we should just continue on and not fill in that object since it's optional.

@wing328 @jgavris for review


wing328:
cc @ehyche for review as well

PR #6429, opened 2017-09-03
[haskell-http-client] Add new haskell-http-client client generator

jonschoning:
### PR checklist

### Description of the PR

Adds new generator template `haskell-http-client` (HaskellHttpClientCodegen.java) which generates a client library for Haskell. 

The existing Haskell generator uses `servant` as the framework to generate bindings.  

In contrast, this uses `http-client` which is a more low-level/fundamental http library.

Another unique thing about this generator is that it ensures at compile-time that the  _content-type_ & _accept_ headers and associated encoders / decoders and are within the specified Produces and Consumes for a given Operation.

A sample is provided at samples/client/petstore/haskell-http-client along with 

* unit tests (serialization)
* integration tests
* example app 

Currently Unsupported Swagger Features

* Auth Methods
    -  headers may be added manually by user
* Default Parameter Values
* Enum Parameters
* `./bin/security/{LANG}-petstore.sh`

All petstore operations are currently succeeding with http status 200 ok (integration tests).

Expecting a few WIP updates


wing328:
@jonschoning I couldn't get the CI (travis) working: https://travis-ci.org/swagger-api/swagger-codegen/builds/271606980

I'll try to fix it later as part of the future enhancement to this generator.

wing328:
Tweet to promote the new generator: https://twitter.com/wing328/status/905108552464793600

jonschoning:
re: CI (travis):

the haskell bits of the build in fact succeeded without any errors. 
the normal tests completed on line 6683
```
Finished in 0.2569 seconds
6 examples, 0 failures
```
the integration tests completed on line 6778
```
Finished in 0.6834 seconds
20 examples, 0 failures
```

what is happening, is that setting up the haskell compiler bits is adding some time, which is causing the whole job to exceed the max time allowed in conjunction with all the other tests that being run.

What I've found with travis, is that

- the first time `stack` is run and must download the haskell compiler (ghc) and the base packages, which adds about 13 minutes to the build time (example: https://travis-ci.org/jonschoning/swaggerpetstore-haskell-http-client/builds/269008795 note: this is on a default travis ubuntu linux box; I am not sure if the osx boxes have the same performance)
- if this first build succeeds and is able to store the `~\.stack` directory into the travis cache, then there is no need to download/install ghc (it is retrieved from the cache) and the whole build finishes in less than 2 minutes (https://travis-ci.org/jonschoning/swaggerpetstore-haskell-http-client/builds/269255239)

It would be ideal if we could get the caches working for the swagger-codegen build, but it looks like they aren't being uploaded.  in any case, it would still have to complete an entire build to get the cache updated.  I am not sure if it's possible to only enable the haskell tests , and if the cache will be re-used.  I don't think caches are shared between branches either.

I suggest leaving the haskell tests commented out.  if desired, testing haskell via travis CI would be possible by commenting out the other sample targets, so that only the haskell sample targets are run.

PR #6431, opened 2017-09-04
Issue 5978

CodeNinjai:
### PR checklist

### Description of the PR

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
@@ -492,7 +492,7 @@ public class JavascriptClientCodegen extends DefaultCodegen implements CodegenCo
         if (name.matches("^\\d.*")) {
             String varName = "var" + name; // e.g. 200data => var200Data (after camelize)
             LOGGER.warn(name + " (variable name starts with number) cannot be used as variable name. Renamed to " + varName);
-            return varName;
+            name = varName;
         }
 
         name = sanitizeName(name);  // FIXME parameter should not be assigned. Also declare it as "final"
@@ -509,7 +509,7 @@ public class JavascriptClientCodegen extends DefaultCodegen implements CodegenCo
 
         // camelize (lower first character) the variable name
         // pet_id => petId
-        name = camelize(name, true);
+        //name = camelize(name, true);
 
         // for reserved word or word starting with number, append _
         if (isReservedWord(name) || name.matches("^\\d.*")) {
@@ -639,7 +639,7 @@ public class JavascriptClientCodegen extends DefaultCodegen implements CodegenCo
             name = varName;
         }
 
-        this.sanitizeName(name);
+        name = this.sanitizeName(name);
         if (p instanceof RefProperty) {
             return " = " + type + ".constructFromObject(data['" + name + "']);";
         } else {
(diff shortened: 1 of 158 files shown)


modules/swagger-codegen/src/main/resources/Javascript/es6/partial_model_generic.mustache
```diff
@@ -65,7 +65,7 @@ export default class {{classname}} {{#parent}}{{^parentModel}}{{#vendorExtension
     * @member {{=< >=}}{<&vendorExtensions.x-jsdoc-type>}<={{ }}=> {{baseName}}{{#defaultValue}}
     * @default {{{defaultValue}}}{{/defaultValue}}
     */{{/emitJSDoc}}
-    {{baseName}} = {{#defaultValue}}{{{defaultValue}}}{{/defaultValue}}{{^defaultValue}}undefined{{/defaultValue}};
+    {{name}} = {{#defaultValue}}{{{defaultValue}}}{{/defaultValue}}{{^defaultValue}}undefined{{/defaultValue}};
```
wing328: @CodeNinjai I still think this should be {{name}} instead of {{baseName}}

e.g. {{baseName}} (original name specified in the spec) can be "modified-date" while {{name}} is the variable name following JS style guide (e.g. modified_date)

cc @cliffano @frol

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
```diff
@@ -500,7 +509,7 @@ public String toVarName(String name) {
 
         // camelize (lower first character) the variable name
         // pet_id => petId
-        name = camelize(name, true);
+        //name = camelize(name, true);
```
wing328: @CodeNinjai Commenting out `camelize` will probably impact JS ES5 output as well. Shall we keep it as it's?

cc @frol @cliffano

frol: @CodeNinjai Given the code style guides for JS, camel-case should be used for variable and method names. Please, bring it back.

Also, it seems that the custom `name.matches(...)` should be simply placed into `sanitizeName` once instead of copy-pasting.

modules/swagger-codegen/src/main/resources/Javascript/es6/partial_model_generic.mustache
frol: @wing328 So that what the change does, doesn't it? It changes `{{ baseName }}` to `{{ name }}`.

modules/swagger-codegen/src/main/resources/Javascript/es6/model_test.mustache
```diff
@@ -1,19 +1,12 @@
...
-    // AMD.
-    define(['expect.js', '../../src/index'], factory);
-  } else if (typeof module === 'object' && module.exports) {
-    // CommonJS-like environments that support module.exports, like Node.
-    factory(require('expect.js'), require('../../src/index'));
-  } else {
-    // Browser globals (root is window)
-    factory(root.expect, root.{{moduleName}});
-  }
+
```
frol: How is this going to work? The module will be empty since the `factory` isn't called.

modules/swagger-codegen/src/main/resources/Javascript/es6/model_test.mustache
```diff
@@ -1,19 +1,12 @@
 {{>licenseInfo}}
+import expect, { createSpy, spyOn, isSpy } from 'expect';
```
frol: I don't see any use of these.

modules/swagger-codegen/src/main/resources/Javascript/es6/partial_model_generic.mustache
wing328: @frol ah sorry. You're right. I misread.

Issue #6451, opened 2017-09-07
Wrong class path for DeferredResult


sasavilic:
##### Description

Generated code for Spring is using wrong class path for DeferredResult. Currently, it is used `org.springframework.web.context.request.DeferredResult`, when actual class path is `org.springframework.web.context.request.async.DeferredResult`. See [here](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/context/request/async/DeferredResult.html).

Fix should be simple, I will provide PR soon.

##### Swagger-codegen version

2.2.3. This bug is present since corresponding feature is introduced.

##### Command line used for generation
I am using swagger-codegen-maven-plugin. Here is the configuration
`
<configuration>
                            <inputSpec>${project.basedir}/src/main/spec/api.yaml</inputSpec>
                            <language>spring</language>
                            <apiPackage>...removed...</apiPackage>
                            <modelPackage>removed...</modelPackage>
                            <configOptions>
                                <sourceFolder>java</sourceFolder>
                                <dateLibrary>java8</dateLibrary>
                                <interfaceOnly>true</interfaceOnly>
                                <java8>true</java8>
                                <responseWrapper>DeferredResult</responseWrapper>
                            </configOptions>
                            <library>spring-mvc</library>
                            <output>${project.build.directory}/generated-sources/server</output>
                        </configuration>
`

##### Suggest a fix/enhancement

I will provide PR.

sanjaygithu:
please add <async>true</async> in <configOptions>
							<configOptions>
								<sourceFolder>/</sourceFolder>
								<dateLibrary>java8-localdatetime</dateLibrary>
								<delegatePattern>false</delegatePattern>
								<interfaceOnly>true</interfaceOnly>
								<serializableModel>true</serializableModel>
								<responseWrapper>DeferredResult</responseWrapper>
								<async>true</async>
							</configOptions>
add this dependency:
<dependency>
	    <groupId>org.openapitools</groupId>
	    <artifactId>jackson-databind-nullable</artifactId>
	    <version>0.1.0</version>
	</dependency>

PR #6452, opened 2017-09-07
Fix class path for DeferredResult

sasavilic:
### PR checklist

See #6451 for more info.

### Description of the PR

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SpringCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SpringCodegen.java
@@ -310,7 +310,7 @@ public class SpringCodegen extends AbstractJavaCodegen
                 additionalProperties.put(RESPONSE_WRAPPER, "org.springframework.util.concurrent.ListenableFuture");
                 break;
             case "DeferredResult":
-                additionalProperties.put(RESPONSE_WRAPPER, "org.springframework.web.context.request.DeferredResult");
+                additionalProperties.put(RESPONSE_WRAPPER, "org.springframework.web.context.request.async.DeferredResult");
                 break;
             case "HystrixCommand":
                 additionalProperties.put(RESPONSE_WRAPPER, "com.netflix.hystrix.HystrixCommand");


wing328:
@sasavilic please run spring-related petstore scripts to update petstore samples so that CIs can verify the change:
```
./bin/spring-all-pestore.sh
```
cc @bbdouglas @JFCote @sreeshas @jfiala @lukoyanov @cbornet @ePaul

sasavilic:
@wing328, I did run script and it was successful, but there is no file changes to commit. Am I missing something?

lukoyanov:
It looks like there are no sh scripts that are using `responseWrapper` option set to `DeferredResult` that's why nothing gets changed in samples.

wing328:
@sasavilic I did a test with your PR but got the following errors:
```
[ERROR] return type org.springframework.http.ResponseEntity<java.lang.String> is not compatible with org.springframework.web.context.request.async.DeferredResult<org.springframework.http.ResponseEntity<java.lang.String>>
[ERROR] /private/var/tmp/spring/issue3/src/main/java/io/swagger/api/UserApiController.java:[57,33] getUserByName(java.lang.String,java.lang.String) in io.swagger.api.UserApiController cannot implement getUserByName(java.lang.String,java.lang.String) in io.swagger.api.UserApi
[ERROR] return type org.springframework.http.ResponseEntity<io.swagger.model.User> is not compatible with org.springframework.web.context.request.async.DeferredResult<org.springframework.http.ResponseEntity<io.swagger.model.User>>
```

Here is the config I used for testing:
```
{
    "responseWrapper": "DeferredResult"
}
```
Did you get similar errors when testing locally in your machine?

cbornet:
I haven't tested but the change seems OK

wing328:
Yup, the change also looks good to me but when I tested I got errors.

Btw, doing the same test on master showed different errors, which seem to be fixed by this PR.

sasavilic:
@wing328, yes I think I got same errors. The problem is default interface implementation when java-8 is enabled, since ResponseEntity can't be converted to DeferredResult. I was not sure how to correct default template so that it supports both. 

I want to apologize for not reporting this later in PR. Since merging PR process was to slow, we patched swagger our self (including template for interface implementation (removing default implementation)) and forgot about this PR. 

But I can tell one thing, once the template is fixed, it works like a charm.

wing328:
@sasavilic no problem. We'll merge this one and open a new issue to track the errors I reported.

Thanks again for the PR 👍

Issue #6006, opened 2017-07-06
[typescript-angular2] No `Content-Type` and `Accept` set in headers


keradus:
##### Description

While generating the API client in typescript-angular2, the `Content-Type` and `Accept` headers are **not** set.

##### Swagger-codegen version

v2.2.2 and 2.2.3-20170703.140356-40 snapshot

##### Swagger declaration file content or url

not available

##### Command line used for generation

`java -jar swagger-codegen-cli.jar generate -i swagger.json -l typescript-angular2`

##### Steps to reproduce

generate the client, send any request via it, but it's also very visible with code itself.

as we see in template: https://github.com/swagger-api/swagger-codegen/blob/v2.2.2/modules/swagger-codegen/src/main/resources/typescript-angular2/api.mustache#L115-L127
or in generated code:
```js
    public resourceFOOGETWithHttpInfo(authorization?: string, extraHttpRequestParams?: any): Observable<Response> {
        const path = this.basePath + '/FOO';

        let queryParameters = new URLSearchParams();
        let headers = new Headers(this.defaultHeaders.toJSON()); // https://github.com/angular/angular/issues/6845
        if (authorization !== undefined && authorization !== null) {
            headers.set('Authorization', String(authorization));
        }

        // to determine the Content-Type header
        let consumes: string[] = [
        ];

        // to determine the Accept header
        let produces: string[] = [
            'application/json'
        ];

        let requestOptions: RequestOptionsArgs = new RequestOptions({
            method: RequestMethod.Get,
            headers: headers,
            search: queryParameters,
            withCredentials:this.configuration.withCredentials
        });
        // https://github.com/swagger-api/swagger-codegen/issues/4037
        if (extraHttpRequestParams) {
            requestOptions = (<any>Object).assign(requestOptions, extraHttpRequestParams);
        }

        return this.http.request(path, requestOptions);
    }
```

the `consumes` and `produces` vars are generated, but not used.

##### Related issues

not aware of

##### Suggest a Fix

the `consumes` and `produces` vars shall be injected into headers

wing328:
@keradus can you try the TS Angular2 generator in 2.3.0 branch? 
```
swagger-codegen|wing328-ts_angular2_typings⚡ ⇒ grep -R '#consumes' modules/swagger-codegen/src/main/resources/typescript-angular2/*
modules/swagger-codegen/src/main/resources/typescript-angular2/api.service.mustache:            {{#consumes}}
[...]

PR #6295, opened 2017-08-11
Add optional support for HttpClient

bedag-moo:
Support HttpClient in addition to Http, clean up generated code

Fixes #6080 

### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: master for non-breaking changes and `3.0.0` branch for breaking (non-backward compatible) changes.

PR #6454, opened 2017-09-08
[TypeScript][Angular] Better support for "Accept", "Content-Type"

stoetti:
### Description of the PR

fix for issue https://github.com/swagger-api/swagger-codegen/issues/6006

checks if the produces or consumes list is null or empty and adds the corresponding headers (Accept, Content-Type), with the value being a list seperated by semicolon, according to http-spec.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache
@@ -175,6 +175,11 @@ export class {{classname}} {
             '{{{mediaType}}}'{{#hasMore}},{{/hasMore}}
             {{/consumes}}
         ];
+
+        if (consumes != null) {
+            headers.set('Content-Type', produces.join(';'));
+        }
+
         let canConsumeForm = this.canConsumeForm(consumes);
         let useForm = false;
 {{#formParams}}
@@ -193,6 +198,10 @@ export class {{classname}} {
             '{{{mediaType}}}'{{#hasMore}},{{/hasMore}}
             {{/produces}}
         ];
+
+        if (produces != null) {
+            headers.set('Accept', produces.join(';'));
+        }
 
 {{#authMethods}}
         // authentication ({{name}}) required
--- a/modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache
@@ -176,7 +176,7 @@ export class {{classname}} {
             {{/consumes}}
         ];
 
-        if ((consumes != null) && (consumes.length > 0)) {
+        if (consumes != null && consumes.length > 0) {
             headers.set('Content-Type', consumes.join(';'));
         }
 
@@ -199,7 +199,7 @@ export class {{classname}} {
             {{/produces}}
         ];
 
-        if ((produces != null) && (produces.length > 0)) {
+        if (produces != null && produces.length > 0) {
             headers.set('Accept', produces.join(';'));
         }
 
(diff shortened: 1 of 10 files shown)


wing328:
@stoetti thanks for the PR. 

cc @TiFu  @taxpon @sebastianhaas @kenisteward @Vrolijkx

modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -175,6 +175,11 @@ export class {{classname}} {
             '{{{mediaType}}}'{{#hasMore}},{{/hasMore}}
             {{/consumes}}
         ];
+
+        if ((consumes != null) && (consumes.length > 0)) {
+            headers.set('Content-Type', consumes.join(';'));
```
TiFu: Just a little code style change: could you please remove the braces around `consumes != null` and `consumes.length > 0`? (same below in line 202). Thanks for the PR.

sebastianhaas:
Hey! Thanks for your PR. I'm afraid though (as discussed in #6006) that it won't be possible just to set the respective `Accept` or `Response` headers according to what the spec says.

As @wing328 mentioned in the source thread:
> We need to implement SelectHeaderContentType (example) and SelectHeaderAccept (example) in ApiClient similar to what we've done for other langauges (e.g. Ruby, PHP, C#).

So, in other words, you need to add a little logic between the list of possible content types and the one that actually gets chosen, based on what we actually support. There is no point in allowing for xml when we only support json parsing, but with your code, as far as I can judge, this could easily happen with an API that reports only support for xml.

stoetti:
Hey!
So what is needed is a logic that only adds json-related mediatypes to the list provided in the Accpet or Content-Type header?

In my opinion the examples @wing328 provides in the comment on the issue also allow xml-related headers to be added as long as there are no application/json entry in produces or consumes.

wing328:
> In my opinion the examples @wing328 provides in the comment on the issue also allow xml-related headers to be added as long as there are no application/json entry in produces or consumes.

@stoetti that's correct. Just to be clear most generators do not support XML payload at the moment with 1 or 2 exceptions (e.g. Java Jersey 1.x or 2.x)

I'm happy to merge this fix into master while you're working on the enhancement to add functions in selecting accept and content-type. Does it sound good to you?

stoetti:
No worries, I will take care of the enhancements as soon as I find the time (after family vacation) and if you want you can leave the pull request open until then.

wing328:
Tests passed with updated TS samples via https://travis-ci.org/swagger-api/swagger-codegen/builds/278882925

wing328:
@stoetti thanks for the PR, which has been merged into master.

Have a nice weekend.

gupbeheer:
@wing328 The code of @stoetti seems to have been deleted in a merge after you merged his code to master. It is not in the current master anymore. Did I overlook something?

wing328:
cc @macjohnny

macjohnny:
@gupbeheer yes, the code was removed, as it caused other issues like https://github.com/swagger-api/swagger-codegen/issues/6588 and as pointed out in @bedag-moo's https://github.com/swagger-api/swagger-codegen/pull/6295#issuecomment-333179969 does not offer any technical enhancement.

wing328:
@gupbeheer can you try the latest master to see if the TS Angular client works for you? If it's still an issue, we might need to add back the function in selecting accept, content-type and fix the bug with the function instead of removing it.

SNAPSHOT version of 2.3.0 (master) can be found in the README.

gupbeheer:
@wing328 @macjohnny Thank you for the quick and clear reply. I tried it with the latest master, but it does not work for me. We use a rest service that uses the mime-types (access and content-types) for versioning. I like the way it was solved as was discussed in #6006. I made my own implementation based on that. I will make a PR so you can check it. I still need to check with the pet-store and the file-upload example. I will make the PR once I have done that.

gupbeheer:
@wing328 @macjohnny the PR is done: #6751

PR #6461, opened 2017-09-09
Zend Expressive fix and upgrade to version 2

BradCrumb:
When you generate a server with ZendExpressive you get several dependency issues with composer and a non working server.

I fixed the dependencies and had to upgrade ZendExpressive to version 2 to get a working version again.

Changes I made:

- Changed composer.json dependencies
- Removed ErrorMiddleware because we can't use it any more
- Added Whoops as ErrorHandler
- Generated samples, because they where not completely up-to-date.


wing328:
@BradCrumb thanks for the PR.

cc @jebentier @dkarlovi @mandrean @jfastnacht @ksm2 @arnested

jfastnacht:
Hey @BradCrumb I can't reproduce your samples with a `java -jar swagger-codegen-cli.jar generate -i http://petstore.swagger.io/v2/swagger.json -l ze-ph -o path\to\ze-ph` am I missing something? Output files in DTO and Handler are a lot less.

BradCrumb:
Hey @jfastnacht . I used the `./run-in-docker.sh /gen/bin/ze-ph-petstore-server.sh` that is provided by de repo itself. It generated the code exactly as committed.

jfastnacht:
Hey @BradCrumb , thank you, I got it. My fault, haven't contributed for a lengthy period of time.

I found one error and some coding style issues.

**Errors**

`src/App/Handler/Fake.php:47` (api.mustache:41)

- `mediaType="*/*"` starts a comment

**Coding Style (mainly PHP-CS-Fixer)**

`models.mustache`
- whitespace after class names in models
- too many blank lines at the end of the file
- arguments missing in PHPDoc

`api.mustache`
- too many new lines after the last use-statement
- throws missing in PHPDoc
- arguments missing in PHPDoc

`src/App/Strategy/Date.php:15` (Date.php.mustache:15)

- line exceeds 120 characters

`src/App/Strategy/DateTime.php:3` (DateTime.php.mustache:3)

- missing blank line after namespace

`@return map[string,int]`

- Is that valid PHPDoc? Never seen that before.

It's working for me, but I can't say much to the specifics of Zend Expressive 2, since I'm not used to it.

BradCrumb:
Hey @jfastnacht ,

I can fix the issues for you. Honestly I didn't touched those files so the coding standard issues and error in mediatype are also in the latest stable version of the codegen ;).

jfastnacht:
Ah sorry, haven't checked whether those were your changes. I'm not sure whether it's better to do it in this PR or open another issue+PR. But it would be great if you could fix it, if you're on it already.

BradCrumb:
@jfastnacht Just committed some Code style fixes.

The problem with `@return map[string,int]` is an issue in `modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractPhpCodegen.java:92`. This looks like an overall PHP problem.

Same thing with `mediaType="*/*"` Since */ ends a comment and/or /* starts a comment. I think we better have to escape it from `modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractPhpCodegen.java:92`. Now no escaping will be done if you take a look at `modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java:2089`.

Maybe we have to create 2 new issues for these problems because we have to test this with several PHP codegen languages.

jfastnacht:
I had another look into the generated samples and found some more issues:

**Fixed in ze-ph**

`src/App/Validator/Type.php` & `public/index.php`

* missing new line at the end of the file

**Unknown reason**

`src/App/DTO/Capitalization.php:42` & `src/App/Handler/Fake.php:36`

* whitespace at the end of the line

`src/App/DTO/ClassModel.php:` & `src/App/DTO/Name.php:35`

* variable name starts with underscore

I can add the issues tomorrow, the other two might be a problem outside of ze-ph too. If so, PR should be fine after fixing Type.php and index.php, moving everything else into seperate issues.

BradCrumb:
@jfastnacht 

Just fixed `Type.php` and `index.php`.

I also analysed `src/App/DTO/Name.php` and `src/App/DTO/ClassModel.php` and I don't think these are real issues. 

The underscore prefix for field `123Number` is necessary because `$123_number` is not allowed in PHP.

The `$_class` variable is also valid, I think, because the field definition of the property in the swagger yml file is also `_class`. See `petsore-with-fake-endpoint-models-for-testing.yml:1087`.

wing328:
cc @jebentier @dkarlovi @mandrean @jfastnacht @arnested @Articus to review as well.

Articus:
First of all - great thanks for supporting generator 👍 . I have only two suggestions about PHP stub itself:

- using Whoops is competely fine, but it might be easier to update `ErrorMiddleware.php.mustache`:
```
<?php
namespace {{invokerPackage}};

use Interop\Http\ServerMiddleware\DelegateInterface;
use Interop\Http\ServerMiddleware\MiddlewareInterface;
use Psr\Http\Message\ResponseInterface as Response;
use Psr\Http\Message\ServerRequestInterface as Request;

class ErrorMiddleware implements MiddlewareInterface
{
    public function process(Request $request, DelegateInterface $delegate)
    {
        $result = null;
        try {
            $result = $delegate->process($request);
            if (!($result instanceof Response)) {
                throw new \RuntimeException(sprintf(
                    'Invalid response: expecting %s, got %s',
                    Response::class,
                    is_object($result)? get_class($result) : gettype($result)
                ));
            }
        } catch (\Exception $error) {
            $result = (new \Zend\Diactoros\Response())->withStatus(500, 'Internal server error');
            $result->getBody()->write((string)$error);
            error_log((string)$error);
        }
        return $result;
    }
}

```
This update willl make ErrorMiddleware compatible with Zend Expressive 2 without external dependencies.

- lines 4 and 5 in `path_handler.yml.mustache`:
```
    default_params:
      middleware: ''
```
are no longer needed for `"articus/path-handler": "0.2.*"` and it is possible to delete them.

Both are only suggestions, it is up to @BradCrumb to decide if they are worth implementing.

wing328:
@Articus thanks for the suggestion.

@BradCrumb I've merged the PR into master. Please feel free to submit another PR with the suggested enhancements.

PR #6487, opened 2017-09-13
Added description to Operation and command line support for generic support files 

ToreJohansson:
Two commits:

1. Added description to Operation. The API includes a description so swagger-codegen only needs to store the content.
2. To add generic support files to a code generation an update for the language implementation is needed. Therefore I added code to add generic support files and their templates to the command line.


modules/swagger-codegen/src/main/java/io/swagger/codegen/config/CodegenConfigurator.java
```diff
@@ -400,6 +403,10 @@ public ClientOptInput toClientOptInput() {
         config.importMapping().putAll(importMappings);
         config.languageSpecificPrimitives().addAll(languageSpecificPrimitives);
         config.reservedWordsMappings().putAll(reservedWordMappings);
+        for(String supportFileTemplate:supportFilesMapping.keySet())
```
wing328: Minor suggestion: keep "{" in this line to conform to Java style guide.

modules/swagger-codegen/src/main/java/io/swagger/codegen/Codegen.java
```diff
@@ -48,6 +48,8 @@ public static void main(String[] args) {
         options.addOption("d", "debug-info", false, "prints additional info for debugging");
         options.addOption("a", "auth", true, "adds authorization headers when fetching the swagger definitions remotely. Pass in a URL-encoded string of name:header with a comma separating multiple values");
         options.addOption("c", "config", true, "location of the configuration file");
+        options.addOption("m", "support", true, "location of the configuration file");
```
wing328: @ToreJohansson thanks for the PR. I think the description should be updated to "maps templates with file names template1=file1,template2=file2." (likely you copied it from the "c" option one line above)

modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
```diff
@@ -19,7 +19,7 @@
             isRestfulIndex, isRestfulShow, isRestfulCreate, isRestfulUpdate, isRestfulDestroy,
             isRestful, isDeprecated;
     public String path, operationId, returnType, httpMethod, returnBaseType,
-            returnContainer, summary, unescapedNotes, notes, baseName, defaultResponse, discriminator;
+            returnContainer, summary, unescapedNotes, notes, baseName, defaultResponse, discriminator, description;
```
wing328: @ToreJohansson I think "notes" is the operation's description. Have you tried it?

ToreJohansson: In here it is defined as "description":
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md

I am running this patch. It works well. Good to get the description field as java-doc in the generated code.

modules/swagger-codegen/src/main/java/io/swagger/codegen/Codegen.java
ToreJohansson: correct :-)

modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
wing328: @ToreJohansson I mean did you try {{notes}}? I agree it should be named "description" instead but it was named "notes" based on historical reason.

modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
```diff
@@ -2057,6 +2057,7 @@ public CodegenOperation fromOperation(String path,
         op.path = path;
         op.operationId = toOperationId(operationId);
         op.summary = escapeText(operation.getSummary());
+        op.description = escapeText(operation.getDescription());
```
wing328: @ToreJohansson seems to be duplicated with
```
         op.unescapedNotes = operation.getDescription();
         op.notes = escapeText(operation.getDescription());
```

modules/swagger-codegen-maven-plugin/src/test/java/io/swagger/codegen/plugin/CodeGenMojoTest.java
```diff
@@ -0,0 +1,116 @@
...
+        CodeGenMojo codeGenMojo = (CodeGenMojo) lookupMojo("generate", pom);
+        assertNotNull(codeGenMojo);
+
+        try
+        {
+            codeGenMojo.execute();
+        } catch (Exception e)
+        {
+            e.printStackTrace();
+        }
```
wing328: @ToreJohansson would be good if we can follow the Java style guide so that "{" won't start in a new line.

PR #6273, opened 2017-08-09
[swift3] empty model with only additional properties

wuf810:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: master for non-breaking changes and `3.0.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

When a class is defined as having no variables but does have additionalProperties, the class would be created incorrectly as a broken typeAlias.

In addition if a class is defined with no variables and no additionalProperties a broken typeAlias was generated. This is now creates an empty jsonencodable class.

This fixes issue #5535

cc @dwoodel

PR #6492, opened 2017-09-13
Swift4: Support object schemas with only additionalProperties.

ehyche:
### Description of the PR

This was implemented in swift3 in this PR: https://github.com/swagger-api/swagger-codegen/pull/6273

Previously, we had implemented the Codable protocol by simply claiming conformance, and making sure that each of our internal classes also implemented the Codable protocol. So our model classes looked like:

```
class MyModel: Codable {
   var propInt: Int
   var propString: String
}

class MyOtherModel: Codable {
   var propModel: MyModel
}
```

Previously, our additionalProperties implementation would have meant an object schema with an additionalProperties of Int type would have looked like:

```
class MyModelWithAdditionalProperties: Codable {
   var additionalProperties: [String: Int]
}
```

But the default implementation of Codable would have serialized MyModelWithAdditionalProperties like this:

```
{
  "additionalProperties": {
    "myInt1": 1,
    "myInt2": 2,
    "myInt3": 3
  }
}
```

The default implementation would put the additionalProperties in its own dictionary (which would be incorrect), as opposed to the desired serialization of:

```
{
  "myInt1": 1,
  "myInt2": 2,
  "myInt3": 3
}
```

So therefore, the only way to support this was to do our own implementation of the Codable protocol. The Codable protocol is actually two protocols: Encodable and Decodable.

So therefore, this change generates implementations of Encodable and Decodable for each generated model class. So the new generated classes look like:

```
class MyModel: Codable {
   var propInt: Int
   var propString: String

   // Encodable protocol methods

   public func encode(to encoder: Encoder) throws {

        var container = encoder.container(keyedBy: String.self)

        try container.encode(propInt, forKey: "propInt")
        try container.encode(propString, forKey: "propString")
    }

    // Decodable protocol methods

    public required init(from decoder: Decoder) throws {
        let container = try decoder.container(keyedBy: String.self)

        propInt = try container.decode(Int.self, forKey: "propInt")
        propString = try container.decode(String.self, forKey: "propString")
    }

}

class MyOtherModel: Codable {
   var propModel: MyModel

   // Encodable protocol methods

   public func encode(to encoder: Encoder) throws {

        var container = encoder.container(keyedBy: String.self)

        try container.encode(propModel, forKey: "propModel")
    }

    // Decodable protocol methods

    public required init(from decoder: Decoder) throws {
        let container = try decoder.container(keyedBy: String.self)

        propModel = try container.decode(MyModel.self, forKey: "propModel")
    }

}
```

### PR checklist


ehyche:
@wing328 @jaz-ah : please review.

wing328:
also cc'ing @Edubits @jgavris for review.

wing328:
@ehyche thanks for the enhancement.

@jgavris thanks for reviewing the change.

Kimnyhuus:
@ehyche Can I ask to to elaborate a bit on the new generated model classes?

I've run in to some issues in my project, as we've upgraded from Swagger Codegen 2.2.3 to 2.3.0 in an attempt to get the project working with Swift 4.

As an example, we've previously had this generated model:

```
open class PhoneBookUserDTO: JSONEncodable {

    public var firstName: String?
    public var lastName: String?
    public var phoneNr: String


    public init(firstName: String?=nil, lastName: String?=nil, phoneNr: String) {
        self.firstName = firstName
        self.lastName = lastName
        self.phoneNr = phoneNr
    }
    // MARK: JSONEncodable
    open func encodeToJSON() -> Any {
        var nillableDictionary = [String:Any?]()
        nillableDictionary["firstName"] = self.firstName
        nillableDictionary["lastName"] = self.lastName
        nillableDictionary["phoneNr"] = self.phoneNr

        let dictionary: [String:Any] = APIHelper.rejectNil(nillableDictionary) ?? [:]
        return dictionary
    }
}
```

Which now looks like this:

```
open class PhoneBookUserDTO: Codable {

    public var firstName: String?
    public var lastName: String?
    public var phoneNr: String

    // Encodable protocol methods

    public func encode(to encoder: Encoder) throws {

        var container = encoder.container(keyedBy: String.self)

        try container.encodeIfPresent(firstName, forKey: "firstName")
        try container.encodeIfPresent(lastName, forKey: "lastName")
        try container.encode(phoneNr, forKey: "phoneNr")
    }

    // Decodable protocol methods
    
    public required init(from decoder: Decoder) throws {
        let container = try decoder.container(keyedBy: String.self)

        firstName = try container.decodeIfPresent(String.self, forKey: "firstName")
        lastName = try container.decodeIfPresent(String.self, forKey: "lastName")
        phoneNr = try container.decode(String.self, forKey: "phoneNr")
    }
}
```

This has an effect on the way the object is initiated, which before could be done like this:
`return PhoneBookUserDTO(phoneNr: token.title)` <- Excerpt from one of my classes

But now has to be initiated with an encoder:
`return PhoneBookUserDTO(from: <#T##Decoder#>)`

Am I supposed to create my own init in these model classes now?
Or should I provide a Decoder for initiating model objects like this?
 - If so, what would be the proper way to do it?

Hope my question makes sense.

Thanks! :-)

ehyche:
@Kimnyhuus : you would now decode these by decoding it directly from JSON data:

```
let someJSONData = """
{
  "firstName": "Joe",
  "lastName": "Schmoe",
  "phoneNr": "000-000-0000"
}
""".data(using: .utf8)!

let result = CodableHelper.decode(PhoneBookUserDTO.self, someJSONData)
let user = result.decodableObj
```

So the only initializers currently generated is the one for decoding from JSON data. However, if you need, I could certainly add the initializer which takes all of the properties. This was present in the swift3 codegen module, but I wasn't sure it was needed in the swift4 module, so I didn't add it. But I could add it to the templates if you need it.

Kimnyhuus:
@ehyche
Thanks for the quick response and explanation.

That would really be great, if you could do that!
Because we use those initializers all over the project, and it would make it way easier for me, when new models are generated from the backend :-)

lightbow:
Thanks @ehyche for this contribution! I agree with @Kimnyhuus —a non-JSON initializer would be greatly appreciated since there are plenty of places in my code where I initialize an object directly.

Also, I'm wondering if there is a way to do this PR without generating tons of extra code when (for most objects) the generated encode/initFromDecoder would just be doing exactly what you'd get for free from the Codeable protocol. Apple's [documentation](https://developer.apple.com/documentation/foundation/archives_and_serialization/encoding_and_decoding_custom_types)  gives an example under the section "Encode and Decode Manually" exactly like the "additionalProperties" problem. It would be a nice win to keep the original benefit of tiny generated files, and only have swagger generate the custom stuff only where necessary with addtionalProperties present. Do you think that would be possible?

PR #6493, opened 2017-09-13
[typescript-angular2] Allow lazy oauth token refresh (async access_token function)

alexrashed:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [ ] Filed the PR against the correct branch: master for non-breaking changes and `3.0.0` branch for breaking (non-backward compatible) changes.
- [x] **Filed against 2.3.0 (and includes breaking changes).**

### Description of the PR

Fixes #6486.

- [x] Add authentication scheme name and necessary scopes as parameters for the `accessToken` function in the `Configuration` as seen in #6130 
- [x] `accessToken` function now returns an `Observable<string>` (allows lazy renewal using the `refreshToken`)
- [x] Update some old ng2 stuff to ng4 (some copy-pasting)
- [x] Added some previously not committed client / petstore code

Unfortunately Travis CI build fails due to some GitHub rate limits ([Travis CI build](https://travis-ci.org/alexrashed/swagger-codegen/builds/275052086)). I successfully executed `mvn clean verify` in the petstore client for ts-ng2 and ts-ng4, but this only invokes the compilation and no integration tests (how can I invoke any ts-ng2/4 IT locally?). But I tested it with my own project and there are no breaking changes to the petstore projects (as they are not using the `accessToken` function but return a static string instead).

PR #6130, opened 2017-07-21
[Typescript-Fetch] Support additionalproperties, Enum, Auth and more.

isman-usoh:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: master for non-breaking changes and `3.0.0` branch for breaking (non-backward compatible) changes.

### Description of the PR
Support additionalproperties and more.
- [x] Upgrade to TypeScript 2
- [x] Use type definition from npm instead of typings, typings is deprecation
- [x] Use Enum instead of String Literal Types
- [x] Use typescript es6 lib for target es5
- [x] Support additionalproperties
- [x] full support JSDoc
- [x] Add snapshot and npmRepository option
- [x] Update typescript-fetch run script for linux
- [x] Create typescript-fetch run script for windows

PR #6496, opened 2017-09-13
[typescript-angular2] Allow lazy oauth token refresh (#6486)

### LINKED ISSUE #6486: [typescript-angular2] Allow lazy oauth token refresh (async access_token function)
##### Description
In case of an _OAuth2 implicit flow_ the current implementation of the `Configuration` class allows the `accessToken` to either be a _string_ or a _function returning a string_ (directly). In order to allow a lazy update of the (already expired) `access_token` using the (still valid) `refresh_token` the function should get the name of the authorization scheme as well as the necessary scopes and should return an `Observable<string>`.

```typescript
export class Configuration {
    ...
    accessToken?: string | ((name: string, scopes?: string[]) => string) => Observable<string>);
    ...
}
```

This would allow an asynchronous implementation of the `accessToken` function where the auth server can be requested for a new token.

##### Swagger-codegen version

`2.3.0-SNAPSHOT`

swagger-codegen-maven-plugin in version 2.3.0-SNAPSHOT for typescript-angular with ngVersion=4

##### Steps to reproduce

Try to create a _Configuration_ object which lazily refreshes the `access_token` (when the access_token has expired, but the _refresh_token_ is still valid).

##### Related issues/PRs

I commented on a PR for the typescript-fetch client suggesting this feature a few weeks ago (this time I mean the angular2 client): https://github.com/swagger-api/swagger-codegen/pull/6130#issuecomment-321244665
This PR also contains the new parameters for the `accessToken` function for the Typescript-Flow client (but does not return an `Observable<string>`): [Configuration class changes in PR 6130](https://github.com/swagger-api/swagger-codegen/pull/6130/files#diff-ad414fb91808bd31b38e08bcbb6feac6)

- Change the _Configuration_ class to allow the _access_token_ function to return an `Observable<string>`
- Forward the name of the authorization scheme as well as the scope to the `access_token` function
- Build the requestOptions asynchronously before executing the http request (flatMap the Observables)

I'm working on a PR right now, I just want to make sure that the feature would be accepted. :)

alexrashed:
Unfortunately it was a pain to update the previous PR / feature-branch, so here's a fresh branch / PR for ticket #6486 replacing PR #6493.

### PR checklist

### Description of the PR

Fixes #6486.
Replaces #6493.

- [x] Add authentication scheme name and necessary scopes as parameters for the `accessToken` function in the `Configuration` as seen in #6130 
- [x] `accessToken` function now returns an `Observable<string>` (allows lazy renewal using the `refreshToken`)
- [x] Generated ts-ng2, ts-ng4 and the ts-ng security client (the security test client seems like it hasn't been generated for a while)


alexrashed:
Looks like the automated Travis build container didn't get enough resources (job exceeded the time limit). The builds worked fine from my forked repo / feature branch (same code):
- [Travis CI](https://travis-ci.org/alexrashed/swagger-codegen/builds/275143428)
- [Circle CI](https://circleci.com/gh/alexrashed/swagger-codegen/1)

wing328:
@alexrashed I'll restart the CI jobs later to give it another try.

cc @TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx for review.

wing328:
@alexrashed when you've time, please rebase on the latest master to resolve the merge conflicts.

alexrashed:
Again the automated Travis build failed, but the one coupled to my forked repo succeeded: https://travis-ci.org/alexrashed/swagger-codegen/builds/280310264

wing328:
@alexrashed no worry. I'll restart the Travis job.

cc @TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx for review.

sebastianhaas:
Thanks for the PR, it looks really promising. I'm going to have to check it out and try locally though. Looking forward to using this!

kenisteward:
@sebastianhaas I looked over this and it seems good.  I also haven't tried it locally but I'll try to verify as well with you. from the other PR he made we had a convo about it as I thought it wasn't needed but this changes only brings benefit to the client so I backed down :)

alexrashed:
I rebased the changes once again and also added a pom file to automatically build the Typescript Angular 4.3 (HttpClient) changes with the samples profile.

kenisteward:
This should work with lower level angular right? I didn't think there was
anything specific to 4.3

alexrashed:
Of course. But recent changes on the master differ in handling the header depending on the new option for using the new HttpClient. So I rebased my changes, adopted them to correctly set the header fields depending on the option and also added a pom file for the previous changes for HttpClient (which weren't done by me, but I realized that the sample client is not built yet when executing the sample maven profile).

kenisteward:
Okay thanks and good catch

bedag-moo:
This seems rather complicated: Why must we extend the code generator, when the HttpClient makes it easy to add a token to all requests by means of an HttpInterceptor?

My app does not even use the accessToken feature of the code generator, but simply declare

```
@Injectable()
export class OpenIdConnectHttpInterceptor implements HttpInterceptor {

  constructor(private oauthService: OAuthService) {}

  intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
    const headers = req.headers.set("Authorization", "Bearer " + this.oauthService.getAccessToken());
    return next.handle(req.clone({headers}));
  }
}
```

and register it with

```
  providers: [
    {provide: HTTP_INTERCEPTORS, useClass: OpenIdConnectHttpInterceptor, multi: true},
  ]
```

It is also worth noting that the above solution supports repeated subscription on the Observable. For instance, `permissions = every5Minutes.switchMap(() => permissionsService.getPermissions()).retry(3)` would send each request with a current token, while your implementation bakes the token into the observable, causing each request to be sent with the original token. 

That is, idiomatic functional reactive programming is only possible if access tokens are evaluated upon subscription rather than being baked into Observables. This is trivial to achieve with an HttpInterceptor, and hard to achieve in the code generator. 

I'd therefore prefer Configuration.accessToken to be deprecated rather than extending its implementation once again to cover more (but not yet all) usecases.

alexrashed:
Interceptor is a nice pattern / feature which solves the less complex configurations fairly elegant. But an interceptor can't access some data needed for more advanced OAuth swagger configs, f.e. the different scopes for each method.

alexrashed:
Also the token isn't static if you assign it a function.
This function is evaluated for every request and therefore is equally powerful as the interceptor in terms of the execution flow.

kenisteward:
Adrian I'm with you here but Alex made fine arguments for this feature at
the time so I gave in. I would love for @tufu @sebastianhaas to chime in on
deprecation of the access token as well since 4.3 let's you handle all
this.

bedag-moo:
@alexrashed: The current PR does not support variable scopes per method either, does it? And the interceptor can inspect the entire request, and pick different scopes accordingly. This why I'd make the interceptor part of the application, so developers can easily modify it to suit their needs.

> Also the token isn't static if you assign it a function.

I don't think you understood the point I was making. Yes, every invocation of the generated api service method gets a fresh token, but in both Http and HttpClient, a request is sent whenever the returned Observable is subscribed to. This can easily happen more than once when employing RXJS operators. Consider the following code:

    currentPermissions = permissionService.getPermissions().repeatWhen(() => every5Minutes);

Here, an Observable is created once, but a request is sent every 5 minutes.

alexrashed:
The PR does support variable scopes per method. And a single interceptor mapping requests to methods / scopes really doesn't seem to be very clean to me...

The latter however is true, the observable is only created once.

bedag-moo:
Ah, I didn't notice the arguments added to `Configuration.accessToken`. Ok, that can be useful and justifies keeping this concern in the code generator. However, I really need token passing to work with RXJS operators, so I could not use this PR in the present form.

alexrashed:
Since there hasn't been any progress and this PR is heavily outdated, I'm closing it.

PR #6500, opened 2017-09-13
Add object import lines to Perl client

### LINKED ISSUE #6313: [perl] Referenced Swagger datatype modules are not loaded
#### Description

When I have nested objects in a perl client, it doesn't work correctly. I get an error on deserialization:

    Can't call method "from_hash" on an undefined value at /lib/WWW/SwaggerClient/Object/Thingy.pm line 133

This happens whenever the type is "number" or when I use a referenced object type. This issue is about the latter case.

##### Swagger-codegen version

2.2.3

This happens in the PetStore example, and I use that in this test case.

##### Steps to reproduce

Save [this file](https://gist.github.com/lbordowitz/d989406d67ef882628a897b964423933) as `test.pl` in the [samples directory](https://github.com/swagger-api/swagger-codegen/tree/7de95c52c1e047d9ae3adb9f01f7e27fdb586963/samples/client/petstore/perl).

Run `perl test.pl` with all the necessary modules installed. You will get the following error message:

    Can't call method "from_hash" on an undefined value at lib/WWW/SwaggerClient/Object/Pet.pm

The reason this problem doesn't come up in the normal test is because all the necessary objects are included as part of the test. This means, in order to use a perl client generated by swagger, you would need to require the end-user to include both the object and all the object's property dependencies transitively. That hardly seem reasonable.

In the ["use" block](https://github.com/swagger-api/swagger-codegen/blob/7de95c52c1e047d9ae3adb9f01f7e27fdb586963/modules/swagger-codegen/src/main/resources/perl/object.mustache#L11-L22) there should be a bundle of lines that refers to all the datatypes, like `use {{moduleName}}::Object::{{datatype}}` for each `datatype`. That way, when the [eval](https://github.com/swagger-api/swagger-codegen/blob/7de95c52c1e047d9ae3adb9f01f7e27fdb586963/modules/swagger-codegen/src/main/resources/perl/BaseObject.mustache#L77) is called the `$_instance` variable won't be undefined. Please make sure to handle ARRAY properties correctly, as well.

lbordowitz:
### PR checklist

### Description of the PR

Added line to import objects that are required to be imported. Intended to fix #6313


lbordowitz:
CircleCI appears to have failed on an unrelated error with `jaxrs-cxf-cd` module. What branch do you recommend I make these changes on?

wing328:

lbordowitz:
Thanks for bringing that up! I never knew that before.

Okay, emails have been updated on my account, so it should go through properly.

wing328:
@lbordowitz thanks for the PR, which has been merged into master.

PR #6501, opened 2017-09-13
Add support for reserved-words-mappings to cpprest

### LINKED ISSUE #6498: [CPPREST] --reserved-words-mappings does not work
Version=2.2.3.
Generated code for cpprest target in which the schema has C++ reserved words will fail to compile.
There is a --reserved-words-mappings argument in the CLI, but it does not work for cpprest.
I have a PR for this.

bandrews-spirent:
* Also handles automatic escaping for reserved words, i.e.
  by default, no need to provide a mapping.
Fix #6498

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCppCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCppCodegen.java
@@ -101,11 +101,30 @@ abstract public class AbstractCppCodegen extends DefaultCodegen implements Codeg
             return sanitizeName(name);
         }
 
+        if (isReservedWord(name)) {
+            return escapeReservedWord(name);
+        }
+
         if (name.length() > 1) {
             return sanitizeName(Character.toUpperCase(name.charAt(0)) + name.substring(1));
         }
 
         return sanitizeName(name);
+    }
+
+    /**
+     * Escapes a reserved word as defined in the `reservedWords` array. Handle
+     * escaping those terms here. This logic is only called if a variable
+     * matches the reseved words
+     *
+     * @return the escaped term
+     */
+    @Override
+    public String escapeReservedWord(String name) {
+        if(this.reservedWordsMappings().containsKey(name)) {
+            return this.reservedWordsMappings().get(name);
+        }
+        return sanitizeName("_" + name);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


bandrews-spirent:
It is not clear why the Travis build failed.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCppCodegen.java
```diff
@@ -109,6 +113,14 @@ public String toVarName(String name) {
     }
 
     @Override
+    public String escapeReservedWord(String name) {
```
stkrwork: Can you add the comment from the original function to this one?

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CppRestClientCodegen.java
```diff
@@ -55,7 +55,7 @@
 
     /**
      * Configures the type of generator.
-     * 
+     *
```
stkrwork: Auto format changes?

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CppRestClientCodegen.java
```diff
@@ -66,7 +66,7 @@ public CodegenType getTag() {
     /**
      * Configures a friendly name for the generator. This will be used by the
      * generator to select the library with the -l flag.
-     * 
+     *
```
stkrwork: See above

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CppRestClientCodegen.java
```diff
@@ -76,7 +76,7 @@ public String getName() {
     /**
      * Returns human-friendly help for the generator. Provide the consumer with
      * help tips, parameters here
-     * 
+     *
```
stkrwork: See above

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CppRestClientCodegen.java
bandrews-spirent: Yes, I have my editor set to removing trailing whitespace.

bandrews-spirent:
Added the original comment back in.

bandrews-spirent:
bump

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CppRestClientCodegen.java
stkrwork: Please keep that whitespace in

bandrews-spirent: Why is that?

wing328:
@bandrews-spirent thanks for the PR, which has been merged into master.

Issue #3816, opened 2016-09-16
Erlang client generator


webron:
_From @loretoparisi on September 16, 2016 13:0_

Is `Erlang` client autogen in the roadmap?

_Copied from original issue: swagger-api/swagger-editor#1064_

wing328:
@loretoparisi we do not have the Erlang API client generator at the moment. 

Recently @galaxie has contributed a Erlang server generator via #3758.

Would you have time to contribute the Erlang API client generator?

loretoparisi:
@wing328 officially [asked](https://www.linkedin.com/groups/60972) the Erlang Developers Group a contribution to this.

tsloughter:
I'm looking to handle this. Can someone link me to any client specific documentation for writing a generator? There is so much to dig through I'm having trouble finding any besides https://github.com/swagger-api/swagger-codegen#making-your-own-codegen-modules

wing328:
@tsloughter thanks for offering help on this. A good starting point is https://github.com/swagger-api/swagger-codegen/wiki/How-to-add-a-generator-for-a-new-language-or-framework

Another good way is to review other PRs to add new generators: https://github.com/swagger-api/swagger-codegen/pulls?q=is%3Apr+is%3Aclosed+label%3A%22Enhancement%3A+New+generator%22

Let me know if you need further assistance.

tsloughter:
Thanks @wing328 

I'm going to open a WIP pull request in hopes to get others interested and giving me feedback if that is ok :)

wing328:
Hi all, the Erlang client generator (beta) by @tsloughter has been merged into master.

Please pull the latest master to give it a try.

PR #6502, opened 2017-09-13
WIP: initial commit for Erlang client generator

tsloughter:
### PR checklist

- [X ] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [ X] Filed the PR against the correct branch: master for non-breaking changes and `3.0.0` branch for breaking (non-backward compatible) changes.

### Description of the PR

When it is reading this resolves https://github.com/swagger-api/swagger-codegen/issues/3816


wing328:
@tsloughter thanks for the new generator.

To modify to create another path-related mustache variable, you want want to follow scalatra approach to do it in the postProcessOperations using vendor extensions: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ScalatraServerCodegen.java#L124

tsloughter:
@wing328 thanks, I'll take a look at the scalatra approach today.

wing328:
@tsloughter Please run "bin/erlang-petstore-client.sh" and include the samples in this PR.

Please also create "bin/windows/erlang-petstore-client.bat" for Windows users as well.

tsloughter:
Added. Still a number of things I need to or want to do. I'll try to put a list in here tomorrow of what I think needs to be done.

wing328:
@tsloughter shall we release the Erlang client generator as alpha/beta so as to collect feedbacks from the community?

tsloughter:
Hey, maybe wait a couple days? I forgot to put together that list I said I would... I'd like to get that and a few small things changed.

tsloughter:
And I mean that I'll hopefully get time to finish those things and figure out what else needs to be done within a couple days, but that if I don't then merging it is also fine at that point to get it moving along and in an alpha state :)

tsloughter:
I pushed some updates. It doesn't yet handle produces types correctly, just always passing the body for the default response to a json decoder. This should be easy to clean up by adding a helper module that takes the Content-Type and the body.

I also want to add arguments for passing authentication to api calls that need it. Or might just add a `Context` variable to every api call that can include that...

wing328:
@tsloughter thanks for the update. I'll take a look this weekend and let you know if I've any feedback.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ErlangClientCodegen.java
```diff
@@ -0,0 +1,393 @@
...
+        // remove " to avoid code injection
+        return input.replace("\"", "");
+    }
+
+    @Override
+    public String escapeUnsafeCharacters(String input) {
+        return input.replace("*/", "*_/").replace("/*", "/_*");
+    }
+
+    class ExtendedCodegenOperation extends CodegenOperation {
```
wing328: @tsloughter Do we still need this Extended class after using vendor extensions to stored the customized path?

tsloughter: Oh, yea, probably. Will clean it up. More might have got left like that because of how I copied the module from another implementation.

tsloughter: Or no, not so simple, would have to look closer at how `ExtendedCodegenOperation` is used in `postProcessOperations` because that is still needed, maybe it can just be replaced with `CodegenOperation` but just doing that naively and running it didn't compile, so I'll need to look more later.

wing328: @tsloughter that's ok. Let's keep it for now.

wing328:
@tsloughter I would suggest doing a alpha release and start engaging other Erlang developers to review and enhance it.

wing328:
UPDATE:
- add you as the template creator of Erlang
- minor formatting change via 444a05c

PR #6510, opened 2017-09-17
[PHP] Fix #6474: Bug with 'format: date' when using --model-name-prefix

### LINKED ISSUE #6474: [PHP] Bug with 'format: date' when using --model-name-prefix
##### Description

When an object property in the Swagger definition is defined as a **date** and swagger-codegen is called with ```--model-name-prefix``` it **incorrectly maps** the object to ```<Prefix>Date``` instead of a ```\DateTime``` object.

##### Swagger-codegen version

I am using swagger-codegen-cli-2.2.3.

##### Swagger declaration file content

```yaml
---
swagger: '2.0'

info:
  title: 'Test Codegen Date'
  version: '1.0'
  
paths:
  '/foo':
    get:
      responses:
        200:
          description: 'Success'
          schema:
            type: object
            properties:
              some_date:
                type: string
                format: 'date'
              some_datetime:
                type: string
                format: 'date-time'
```

```java -jar swagger-codegen-cli-2.2.3.jar generate -i test.yaml -l php --model-name-prefix Foo```

##### Steps to reproduce

1. Save the above example YAML as ```test.yaml```.
2. Run swagger-codegen without a prefix: ```swagger-codegen generate -i test.yaml -l php```.
3. Open the generated file ```SwaggerClient-php/lib/Model/InlineResponse200.php```.
4. See on line 57 that ```some_date``` from ```test.yaml``` is correctly mapped to ```\DateTime```.
5. See on line 58 that ```some_datetime``` from ```test.yaml``` is correctly mapped to ```\DateTime```.
6. Run swagger-codegen with a prefix: ```swagger-codegen generate -i test.yaml -l php --model-name-prefix Foo```.
7. Open the (prefixed) generated file ```SwaggerClient-php/lib/Model/FooInlineResponse200.php```.
8. See on line 57 that ```some_date``` is **incorrectly** mapped to ```FooDate``` where it should be ```\DateTime```.
9. See on line 58 that ```some_datetime``` is **correctly** mapped to ```\DateTime``` even when using a prefix.

##### Related issues/PRs

Issue #4430 looks similar but it is regarding Java generated code, I haven't been able to find a similar issue for PHP.

Unfortunately I was not able to find where in the code this is managed, but since it works correctly for ```format: 'date-time'``` it could be that it's an easy fix once the difference in handling prefixes between ```format: 'date-time'``` and ```format: 'date'``` is identified.

ackintosh:
### PR checklist

### Description of the PR

Fixes issue #6474.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PhpClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PhpClientCodegen.java
@@ -117,6 +117,7 @@ public class PhpClientCodegen extends DefaultCodegen implements CodegenConfig {
         typeMapping.put("string", "string");
         typeMapping.put("byte", "int");
         typeMapping.put("boolean", "bool");
+        typeMapping.put("date", "\\DateTime");
         typeMapping.put("Date", "\\DateTime");
         typeMapping.put("DateTime", "\\DateTime");
         typeMapping.put("file", "\\SplFileObject");
--- a/modules/swagger-codegen/src/test/java/io/swagger/codegen/php/PhpModelTest.java
+++ b/modules/swagger-codegen/src/test/java/io/swagger/codegen/php/PhpModelTest.java
@@ -1,12 +1,14 @@
 package io.swagger.codegen.php;
 
 import io.swagger.codegen.CodegenModel;
+import io.swagger.codegen.CodegenOperation;
 import io.swagger.codegen.CodegenProperty;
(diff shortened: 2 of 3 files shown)


ackintosh:
Please review 🙏 ✨  @arnested @jebentier @dkarlovi @mandrean @jfastnacht cc @wing328

dkarlovi:
This is more a builder change, can't really review that. :)

wing328:
@ackintosh the fix looks good to me. Thanks for adding a test case as well.

Issue #6505, opened 2017-09-15
[Typescript][JQuery] Fix to handle multiple files 


JFCote:
Hi,

This question might apply to other Typescript generator but since I'm maintaining the Typescript-JQuery generator, the question will be linked to that one.

**The scenario:**

You have a yaml like this:

```yaml
  /foos:
    post:
      tags:
        - Bar
      summary: Upload foos
      operationId: uploadFoos
      consumes:
        - multipart/form-data
      parameters:
        - in: formData
          name: file1
          required: true
          type: file
        - in: formData
          name: file2
          required: true
          type: file
        - in: formData
          name: file3
          required: true
          type: file
        - in: formData
          name: file4
          required: true
          type: file
        - in: formData
          name: foo-info1
          required: true
          type: string
        - in: formData
          name: foo-info2
          required: true
          type: string
        - in: formData
          name: foo-info3
          required: true
          type: boolean
      responses:
        200:
          description: OK
          schema:
            $ref: "#/definitions/Blabla"
```

The problem is that it will generate a method that looks like this:

`public uploadFoos(file1: any, file2: any, file3: any, file4: any, foo-info1: string, foo-info2: string, foo-info3: boolean): JQueryPromise<{ response: JQueryXHR; body: models.Blabla;  }> {`

And then later in the typescript code, we have this horror (which doesn't work)

```typescript
reqHasFile = true;
formParams = file1;
reqHasFile = true;
formParams = file2;
reqHasFile = true;
formParams = file3;
reqHasFile = true;
formParams = file4;
```

I don't know what to do in this case. My idea so was that as soon as there is a formData of type "file", replace file1, file2, file3 and file4 with a single FormData parameter. The downside is that the code will no longer validate if the files are there (if required) because I will not have access.

Any ideas? How do the other generators deal with this? I looked the typescript-angular and fetch but it seems to rely on framework object that I don't have access in pure jquery.

Any help would be appreciated. I will be able to provide a PR once I know what to do and what suits best the project direction.

##### Swagger-codegen version

Master

I'm tagging people who might be interested:
@wing328 @mhardorf @leonyu @roni-frantchi @bherila @TiFu @taxpon @sebastianhaas @kenisteward

wing328:
Closed via #6541

PR #6515, opened 2017-09-18
[Typescript-jQuery] Tentative fix for #6505

JFCote:
### PR checklist

### Description of the PR

This is a tentative fix for https://github.com/swagger-api/swagger-codegen/issues/6505

The downside of this fix is that we "concatenate" all formData of type "file" into one called "files". So if there is 4 file parameters, and the first is required and the 3 others are not, then we will validate that "files" is required. If there is only one formData of type "file", we don't concatenate or change the name.

But I don't think there is a perfect solution in this case. Let me know what you think.


JFCote:
@wing328 Closing this PR. I have a much better solution to fix this bug and will create a brand new PR for this.

Issue #4370, opened 2016-12-12
Cpprest does not set Content-Type header on POST requests.


sdrichter:
##### Description
Cpprest does not set the Content-Type header on POST requests and so the server rejects all POST requests from this language.

##### Swagger-codegen version
Latest master build.

##### Swagger declaration file content or url
Any POST command with Content-Type as application/json.

##### Command line used for generation
java -jar swagger-codegen-cli.jar generate -l cpprest -i <yaml file> -o src

##### Steps to reproduce
Run swagger-codegen for language cpprest with a YAML file with at least one POST method that takes a content type of application/json.

##### Related issues
None

##### Suggest a Fix
I have fixed this on my fork by adding the following line to the api-source.mustache file:
headerParams[U("Content-Type")] = requestHttpContentType;

I will issue a pull request shortly.

wing328:
Closed via #4370

PR #6518, opened 2017-09-19
[CPPREST] Fixed multipart files upload implementation

frol:
### PR checklist

### Description of the PR

Before this PR, file upload functionality was completely broken. I have fixed the segmentation fault due to the use of null `postBody` instead of `uploadData` (I believe, it was just a typo) and `Content-Type` didn't include the `boundary` marker.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/cpprest/api-source.mustache
+++ b/modules/swagger-codegen/src/main/resources/cpprest/api-source.mustache
@@ -221,9 +221,6 @@ pplx::task<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}void{{/r
         throw ApiException(415, U("{{classname}}->{{operationId}} does not consume any supported media type"));
     }
 
-    //Set the request content type in the header.
-    headerParams[U("Content-Type")] = requestHttpContentType;
-
     {{#authMethods}}
     // authentication ({{name}}) required
     {{#isApiKey}}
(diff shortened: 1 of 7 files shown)


modules/swagger-codegen/src/main/resources/cpprest/api-source.mustache
```diff
@@ -221,9 +221,6 @@ pplx::task<{{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}void{{/r
         throw ApiException(415, U("{{classname}}->{{operationId}} does not consume any supported media type"));
     }
 
-    //Set the request content type in the header.
-    headerParams[U("Content-Type")] = requestHttpContentType;
```
frol: I dropped this line because it doesn't make any good:

1. It prevents the `request.set_body` from setting a custom Content-Type (it simply ignores the later contentType passed to `request.set_body`, and thus, the multipart boundary marker cannot be set).
2. `request.set_body` sets the Content-Type in the ApiClient anyway.

frol: I have just seen issue #4370, so I will test it more thoroughly and report my findings.

frol: All my tests with POST endpoints work fine on my end:

* URL-encoded forms
* Multipart forms
* JSON body forms (NOTE: it is not enough to just set the Content-Type to make the form JSON-based)

Thus, I conclude that this change is fine.

wing328:

@frol thanks for the PR, which has been merged into master.

Issue #6526, opened 2017-09-20
[bash] Bug in handling security for generating clients


kenjones-cisco:
##### Description
Code reference section:
https://github.com/swagger-api/swagger-codegen/blob/7f6bccb4ed1dd23a1aa0a694c24cff262b8f00ee/modules/swagger-codegen/src/main/resources/bash/client.mustache#L435

If there is no security defined, generated code as following:

```bash
        #
        # Append parameters without specific cardinality
        #
        local collection_type="${operation_parameters_collection_type["${operation}:::${qparam}"]}"
 [[ "${collection_type}" == "" ]]; then
```

If security is defined but only does not include `apiKey`, generated code as following:

```bash
        #
        # Append parameters without specific cardinality
        #
        local collection_type="${operation_parameters_collection_type["${operation}:::${qparam}"]}"
 [[ "${collection_type}" == "" ]]; then
```

If security is defined with two entries for `apiKey` with one in header and one in query, generated code as following:

```bash
        #
        # Append parameters without specific cardinality
        #
        local collection_type="${operation_parameters_collection_type["${operation}:::${qparam}"]}"

        if        if [[ ${qparam} == "auth" ]]; then
```

##### Swagger-codegen version

v2.2.3

##### Command line used for generation

swagger-codegen-cli \
    generate -i api.yml -l bash -c swagger_config.json -o ./ --additional-properties hideGenerationTimestamp=true

```json
{
  "processMarkdown": true,
  "curlOptions": "-k -s",
  "scriptName": "capi-cli",
  "generateBashCompletion": true,
  "generateZshCompletion": false,
  "hostEnvironmentVariable": "CAPI_HOST",
  "basicAuthEnvironmentVariable": "",
  "apiKeyAuthEnvironmentVariable": ""
}
```

##### Suggest a fix/enhancement

https://github.com/swagger-api/swagger-codegen/blob/7f6bccb4ed1dd23a1aa0a694c24cff262b8f00ee/modules/swagger-codegen/src/main/resources/bash/client.mustache#L419-L435

Reporting the issue; still looking at it to determine fix.

PR #6528, opened 2017-09-20
[bash] Bugfix: Handle all different types of security

kenjones-cisco:
The only special handling was for security definition type `apiKey`
in `query`. All the other security configurations should result in the
same generated code.
Moves the handling of the special query parameters outside of the
`parameters without specific cardinality` section.
To cover the scenario where `elif` was being used, simply leverage the
builtin `continue` statement to stop processing the specific query
parameter and continue to the next available query parameter, if any.

Manually test with multiple different combinations.

Resolves: #6526

### PR checklist

### DIFF
--- a/samples/client/petstore/bash/Dockerfile
+++ b/samples/client/petstore/bash/Dockerfile
@@ -39,6 +39,7 @@ For convenience, you can export the following environment variables:\n\
 \n\
 $(tput setaf 3)PETSTORE_HOST$(tput sgr0) - server URL, e.g. https://example.com:8080\n\
 $(tput setaf 3)PETSTORE_API_KEY$(tput sgr0) - access token, e.g. "ASDASHJDG63456asdASSD"\n\
+$(tput setaf 3)PETSTORE_API_KEY$(tput sgr0) - access token, e.g. "ASDASHJDG63456asdASSD"\n\
 $(tput setaf 3)PETSTORE_BASIC_AUTH$(tput sgr0) - basic authentication credentials, e.g.: "username:password"\n\
 \n\
 $(tput setaf 7)Basic usage:$(tput sgr0)\n\
--- a/modules/swagger-codegen/src/main/resources/bash/client.mustache
+++ b/modules/swagger-codegen/src/main/resources/bash/client.mustache
@@ -411,15 +411,7 @@ build_request_path() {
                 query_request_part+="&"
             fi
         fi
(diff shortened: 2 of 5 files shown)


kenjones-cisco:
FYI, Seems that there are issues running macOS builds on Travis currently.
https://www.traviscistatus.com/incidents/h6gwvyq9zky0

wing328:
@kenjones-cisco thanks for the PR.

cc @frol @bkryza

frol:
Looks good to me.

@kenjones-cisco Thank you!

dmann:
It would be worth updating the `README.md` to mention that `security_definitions` need to be specified in order for the authentication code to be generated. I honestly didn't realize I was missing them until I got compilation error trying to run the generated bash cli.

samples/client/petstore/bash/Dockerfile
```diff
@@ -39,6 +39,7 @@ For convenience, you can export the following environment variables:\n\
 \n\
 $(tput setaf 3)PETSTORE_HOST$(tput sgr0) - server URL, e.g. https://example.com:8080\n\
 $(tput setaf 3)PETSTORE_API_KEY$(tput sgr0) - access token, e.g. "ASDASHJDG63456asdASSD"\n\
+$(tput setaf 3)PETSTORE_API_KEY$(tput sgr0) - access token, e.g. "ASDASHJDG63456asdASSD"\n\
```
bkryza: @ kenjones-cisco Why this line is added here? - it is an exact duplicate of the above line. Did you mean to add another env var in here or is it just a 'typo'?

kenjones-cisco: I didn't add it manually it happened when the code was generated.
I just reran the ./bin/ script for bash.

bkryza: Yes you're right, it seems the petstore example has changed since last time the bash examples were updated (in particular this change is probably the [cause](https://github.com/swagger-api/swagger-codegen/commit/82835db6701c41b7888c6e13d718c9936358b9f5#diff-45181bd22b428830b22bb957f1fe4772) in the repo, and accidentally with your PR the issue came up :-) I'll have a look at it and submit a separate PR... I

wing328:
@kenjones-cisco thanks for the PR 👍 

@frol @bkryza thanks for reviewing the change.

Issue #5500, opened 2017-04-27
[Swift3] primitive typed parameters in POST body


TeMob:
##### Description

Hi 
I am struggling with primitive parameters used in the request body of POST requests.
The problem is that primitive parametertypes, such as "string", are forced to be JSON encoded, because all body-parameters are being JSON encoded.
Thats part of the problem. Another problem is that all parameters are being casted to a [String:AnyObject] dictionary, although a plain string will never succeed in doing so.

This is the code that swagger-codegen generated:

```
open class func myRequestWithRequestBuilder(xyz: String) -> RequestBuilder<ABC> {
	let path = ...
	let URLString = ...
	let parameters = xyz.encodeToJSON() as?  [String:AnyObject]  //<- this will always fail for primitives
	
	let url = NSURLComponents(string: URLString)
	
	let requestBuilder..
	return requestBuilder.init(method: "POST", URLString: (url?.string ?? URLString), parameters: parameters, isBody: true)
}
```

##### Swagger-codegen version

2.2.2

##### Swagger declaration file content or url

```yaml
post:
   tags:
      - "MyApp"
   description: "description"
   consumes:
      - "text/plain"
   produces:
      - "application/json"
   parameters:
      - in: "body"
         name: "xyz"
         required: true
         schema:
         type: "string"
```

##### Command line used for generation

java -jar swagger-codegen-cli.jar generate -i /path/to/my.yaml -l swift3 -o swaggeroutput/

wing328:
@TeMob can you give an example of the HTTP body? we want to know the request body looks like.

TeMob:
I can't get it to produce a request, so I can't show you  the HTTP-request it sends.

wing328:
@TeMob I'm not referring to the HTTP request produced by the Swift3 API client.

I want to take a look at the correct HTTP request body to understand what payload is expected by the RESTful backend.

jahertor:
I have the same problem. The codegen are trying to parse every body parameter to JSON, even when in swagger editor I put `String` as input and `consumes` is `text/plain`. This is a problem for example when using JWT.

Expected request:

```
@POST /api/user

Headers:
Content-Type: text/plain

Body:
eyJhbGciOiJSU0EtT0FFUCIsImVuYyI6IkExMjhDQkMtSFMyNTYifQ....

jahertor:
[...]

Issue #5639, opened 2017-05-15
[Swift 3]How to send non dictionary body 


bharathreddy-97:
I am facing a issue while POST and PUT requests. By default Swagger expects a body as dictionary type. I have requests where i have to send json array, String etc. For example this is one sample method which got created while generating 

open class func FunctionName(lbid: Int64, lasid: Int64, luid: Int64, lcasid: Int64? = nil, lupid: Int64? = nil, ptype: String? = nil, cform: Bool? = nil, csform: Bool? = nil, body: [SomeModel]? = nil) -> RequestBuilder<SomeResponseObject> {
        let path = "<some path>"
        let URLString = SwaggerClientAPI.basePath + path
        let parameters = body?.encodeToJSON() as? [String:AnyObject] -----> This line has a problem
 
        let convertedParameters = APIHelper.convertBoolToString(parameters)
 
        let requestBuilder: RequestBuilder<SomeResponseObject>.Type = SwaggerClientAPI.requestBuilderFactory.getBuilder()

        return requestBuilder.init(method: "POST", URLString: URLString, parameters: convertedParameters, isBody: false)
    }
When encodeToJson() method is called it is expecting result to be of dictionary type but I had to pass array here as seen in method signature. When i send array, String parameters is becoming nil. How can i send array without disturbing Swagger code(because new methods will be added and I don't want to change everytime I generate Swagger code). Can I somehow pass my request object directly which will avoid this situation?

wing328:
@bharathreddy-97 How do you define the endpoint? Are you using BodyParameter (array of models)?

If you can share the spec, the community can use it to more easily reproduce the issue.

Issue #5480, opened 2017-04-25
[Swift3] Request body cannot be an array, only a dictionary


davidpasztor:
##### Description

The request body cannot be an array of dictionaries, or in fact an array of any object, it can only be a dictionary (which can be a dictionary containing other dictionaries as values).
The issue comes from the Alamofire implementation that the request body is converted from the query parameters and the query parameters have to be a dictionary and cannot be an array.

##### Swagger-codegen version

2.2.2

##### Swagger declaration file content or url

```yaml
/path/{to}/file:
    post:
      tags:
        - tag1
      operationId: something
      summary: summary
      description: 
        description
      parameters:
        - $ref: '#/parameters/param1'
        - $ref: '#/parameters/param2'
        - $ref: '#/parameters/param3'
        - name: postBody
          description: data to be posted in the body
          in: body
          required: true
          schema:
            type: array
            items:
              $ref: '#/definitions/Item1'
      responses:
        '201':
          description: Created
          schema:
            $ref: '#/definitions/Reply'

        '400':
          description: Bad request
          schema:
            $ref: '#/definitions/_Messages'
        default:
          description: Unexpected error
          schema:
            $ref: '#/definitions/_Messages'
      security:
        - basicAuth: []
```

##### Command line used for generation

swagger-codegen generate -i /path/spec.yaml -l swift3

##### Steps to reproduce

Generate a Swift3 client side code from a swagger spec that contains type: array in  its schema for an object that is in:body.

##### Related issues

https://github.com/swagger-api/swagger-codegen/issues/2483
https://github.com/swagger-api/swagger-codegen/pull/4490

##### Suggest a Fix
The Swift file API.swift generated by swagger-codegen has the following method: 
```Swift
open class RequestBuilder<T> {
    var credential: URLCredential?
    var headers: [String:String]
    let parameters: [String:Any]?
    let isBody: Bool
    let method: String
    let URLString: String
    
    /// Optional block to obtain a reference to the request's progress instance when available.
    public var onProgressReady: ((Progress) -> ())?

    required public init(method: String, URLString: String, parameters: [String:Any]?, isBody: Bool, headers: [String:String] = [:]) {
        self.method = method
        self.URLString = URLString
        self.parameters = parameters
        self.isBody = isBody
        self.headers = headers
        
        addHeaders(SwaggerClientAPI.customHeaders)
    }
    
    open func addHeaders(_ aHeaders:[String:String]) {
        for (header, value) in aHeaders {
[...]

PR #6531, opened 2017-09-20
Swift3: non dictionary body type

anozaki:
Adding support for encoding non dictionary types into the body. This should be a backwards compatible change but does change RequestBuilder<T>'s parameters to Any? type to support more then dictionary.

This pull request will effect all POST/PUT request as it changes how body content is handled.

Should fix: #5500 #5639 #5480


wing328:
@anozaki thanks for the PR.

cc @jaz-ah @Edubits @jgavris @ehyche

modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
```diff
@@ -77,7 +77,18 @@ open class AlamofireRequestBuilder<T>: RequestBuilder<T> {
      configuration (e.g. to override the cache policy).
      */
     open func makeRequest(manager: SessionManager, method: HTTPMethod, encoding: ParameterEncoding, headers: [String:String]) -> DataRequest {
-        return manager.request(URLString, method: method, parameters: parameters, encoding: encoding, headers: headers)
+        if isBody {
```
jaz-ah: I might be missing something but do you need this if block?  if you change parameters to an Any? like you have instead of a dict, won't ParameterEncoding.encode() do exactly what you have here?

anozaki: hmmm... now that I'm looking at how the ParameterEncoding and Manager works I think something like this is cleaner. I think the function show also changed to throws so we can handle errors better. Then change the execute to use do/catch and generate a proper error.

```
let originalRequest = try URLRequest(url: URLString, method: method, headers: headers)
if let encoding = encoding as? JSONEncoding {
    let encodedRequest = try encoding.encode(originalRequest, withJSONObject: parameters)
    return manager.request(encodedRequest)
} else {
    let encodedRequest = try encoding.encode(originalRequest, with: parameters as? Parameters)
    return manager.request(encodedRequest)
}
```

The main problem with Alamofire is that parameter is defined as `Parameters` which is a typedef of [String: Any]. If you want to specify anything else they make you do everything your self.

Let me try making these changes and run some tests.

modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
```diff
@@ -127,11 +136,15 @@ open class AlamofireRequestBuilder<T>: RequestBuilder<T> {
...
-            if let onProgressReady = self.onProgressReady {
-                onProgressReady(request.progress)
+            do {
+                let request = try makeRequest(manager: manager, method: xMethod!, encoding: encoding, headers: headers)
+                if let onProgressReady = self.onProgressReady {
+                    onProgressReady(request.progress)
+                }
+                processRequest(request: request, managerId, completion)
+            } catch {
+                completion(nil, ErrorResponse.HttpError(statusCode: 500, data: nil, error: error))
```
anozaki: Is 500 the right status code here?

modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
```diff
@@ -76,8 +76,15 @@ open class AlamofireRequestBuilder<T>: RequestBuilder<T> {
      May be overridden by a subclass if you want to control the request
      configuration (e.g. to override the cache policy).
      */
-    open func makeRequest(manager: SessionManager, method: HTTPMethod, encoding: ParameterEncoding, headers: [String:String]) -> DataRequest {
-        return manager.request(URLString, method: method, parameters: parameters, encoding: encoding, headers: headers)
+    open func makeRequest(manager: SessionManager, method: HTTPMethod, encoding: ParameterEncoding, headers: [String:String]) throws -> DataRequest {
+        let originalRequest = try URLRequest(url: URLString, method: method, headers: headers)
+        if let encoding = encoding as? JSONEncoding {
```
anozaki: I need to do this here since ParameterEncoding protocol only nows how to take `Parameters?`. JSONEncoding has a special encode function that will take `Any?`.

modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
```diff
@@ -76,8 +76,15 @@ open class AlamofireRequestBuilder<T>: RequestBuilder<T> {
      May be overridden by a subclass if you want to control the request
      configuration (e.g. to override the cache policy).
      */
-    open func makeRequest(manager: SessionManager, method: HTTPMethod, encoding: ParameterEncoding, headers: [String:String]) -> DataRequest {
-        return manager.request(URLString, method: method, parameters: parameters, encoding: encoding, headers: headers)
+    open func makeRequest(manager: SessionManager, method: HTTPMethod, encoding: ParameterEncoding, headers: [String:String]) throws -> DataRequest {
```
anozaki: Added `throws` here... which might break peoples implementation if they are extending `AlamofireRequestBuilder`. They would need to add some type of `try` in there code if they are calling `super.makeRequest`. I can back this out, but originally the manager.request was handling these errors for us.

jaz-ah: I think it would be cleaner to keep the original implementation w/o throws and just return an error if the try's below fail just like in manger.request:
            return request(originalRequest, failedWith: error)

modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
jaz-ah: y that looks right

modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
anozaki: Thats what I wanted to do originally but that method is private. I could try figuring out whats going on in that function and try to mimic it?

jaz-ah: could this logic be pushed into SessionManager.swift? so AlamofireImplementations just calls a helper method in that class which then has access to that function.

anozaki: SessionManager is part of Alamo if I remember correctly. I don't think that will work.

modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
jgavris: I'm not sure about returning a response code if this didn't come from the server. It could be confusing / misleading.

modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
jaz-ah: ah you're correct.  maybe just return some type of error instead (not necessary copying Alamo's logic exactly but something similar)

modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
anozaki: I agree and I'm not really sure HttpError actually make since here. Changing how the error system would be a bigger change?

Also I choose 500 since that was the closest to what processRequest would have done. I think this is the response you would have gotten in the old code. 
`ErrorResponse.HttpError(statusCode: response.response?.statusCode ?? 500, data: response.data, error: response.result.error!)`

jaz-ah: @jgavris hmm we do send the 500 at the end of processRequest in AlamofireImplementations as well - maybe we should avoid doing that - is there a better type of ErrorResponse to return?

modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
anozaki: So change the method signature? Right now it expects DataRequest to be returned. I was trying to avoid changing the method signature but if thats ok I can do that.

modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
jgavris: I see that...yes we should improve it. Maybe we can add a `case ClientError` to `ErrorResposne` enum?

wing328: For other API client (e.g. PHP), we set the status code to 0 in the API exception object/class if the issue happens in the client side (e.g. network issue). Maybe we can follow the same convention in the Swift client.

modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
jaz-ah: @jgavris any thoughts as to how to avoid changing the method signature here so we all don't have to try/catch around these blocks?

jaz-ah: my only thought would be to make a new request in the catch block w/ some bogus url to force an error... but seems messy.

modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
```diff
@@ -127,11 +136,15 @@ open class AlamofireRequestBuilder<T>: RequestBuilder<T> {
...
-            if let onProgressReady = self.onProgressReady {
-                onProgressReady(request.progress)
+            do {
+                let request = try makeRequest(manager: manager, method: xMethod!, encoding: encoding, headers: headers)
+                if let onProgressReady = self.onProgressReady {
+                    onProgressReady(request.progress)
+                }
+                processRequest(request: request, managerId, completion)
+            } catch {
+                completion(nil, ErrorResponse.ClientError(error: error))
```
anozaki: Added ClientError. Please take a look.

modules/swagger-codegen/src/main/resources/swift3/AlamofireImplementations.mustache
anozaki: I though about this for a while and I think I finally have a clean solution. I'll update the pull request in a little bit.

anozaki:
@jaz-ah @jgavris

I've changed how the encoding process work in order to preserve the method signature. Ended up also not really needing the ClientError but might be a good idea to create that in the future.

jaz-ah:
+1 @anozaki nice change. + @wing328

jaz-ah:
@anozaki do you want to keep this PR open?

anozaki:
Sorry accidentally clicked on the wrong button :/

jgavris:
+1 Nice

jaz-ah:
@wing328 this one's good to merge.

anozaki:
@ehyche Looked like there isn't a pull request for swift4 yet. I have my system finally setup so I can create a pr ready for swift 4.

PR #6538, opened 2017-09-21
[Scala] Default Option parameters to None

gmarz:
### PR checklist

### Description of the PR

Defaults `Option` parameters to `None` which makes method calls and models less verbose.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/scala/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/scala/api.mustache
@@ -65,7 +65,7 @@ class {{classname}}(
 {{#allParams}}   * @param {{paramName}} {{description}} {{^required}}(optional{{#defaultValue}}, default to {{{.}}}{{/defaultValue}}){{/required}}
 {{/allParams}}   * @return {{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}void{{/returnType}}
    */
-  def {{operationId}}({{#allParams}}{{paramName}}: {{#required}}{{dataType}}{{#defaultValue}} /* = {{{defaultValue}}}*/{{/defaultValue}}{{/required}}{{^required}}Option[{{dataType}}]{{#defaultValue}} /* = {{{defaultValue}}}*/{{/defaultValue}}{{^defaultValue}} = None{{/defaultValue}}{{/required}}{{#hasMore}}, {{/hasMore}}{{/allParams}}){{#returnType}}: Option[{{returnType}}]{{/returnType}} = {
+  def {{operationId}}({{#allParams}}{{paramName}}: {{#required}}{{dataType}}{{#defaultValue}} /* = {{{defaultValue}}}*/{{/defaultValue}}{{/required}}{{^required}}Option[{{dataType}}]{{#defaultValue}} = None /* = {{{defaultValue}}}*/{{/defaultValue}}{{^defaultValue}} = None{{/defaultValue}}{{/required}}{{#hasMore}}, {{/hasMore}}{{/allParams}}){{#returnType}}: Option[{{returnType}}]{{/returnType}} = {
     val await = Try(Await.result({{operationId}}Async({{#allParams}}{{paramName}}{{#hasMore}}, {{/hasMore}}{{/allParams}}), Duration.Inf))
     await match {
       case Success(i) => Some(await.get)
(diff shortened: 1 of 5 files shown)


wing328:
@gmarz thank for the PR.

cc @clasnake @foxmk @ramzimaalej

modules/swagger-codegen/src/main/resources/scala/api.mustache
```diff
@@ -61,7 +61,7 @@ class {{classname}}(val defBasePath: String = "{{{basePath}}}",
 {{#allParams}}   * @param {{paramName}} {{description}} {{^required}}(optional{{#defaultValue}}, default to {{{.}}}{{/defaultValue}}){{/required}}
 {{/allParams}}   * @return {{#returnType}}{{{returnType}}}{{/returnType}}{{^returnType}}void{{/returnType}}
    */
-  def {{operationId}}({{#allParams}}{{paramName}}: {{#required}}{{dataType}}{{#defaultValue}} /* = {{{defaultValue}}}*/{{/defaultValue}}{{/required}}{{^required}}Option[{{dataType}}]{{#defaultValue}} /* = {{{defaultValue}}}*/{{/defaultValue}}{{^defaultValue}} = None{{/defaultValue}}{{/required}}{{#hasMore}}, {{/hasMore}}{{/allParams}}){{#returnType}}: Option[{{returnType}}]{{/returnType}} = {
+  def {{operationId}}({{#allParams}}{{paramName}}: {{#required}}{{dataType}}{{#defaultValue}} /* = {{{defaultValue}}}*/{{/defaultValue}}{{/required}}{{^required}}Option[{{dataType}}]{{#defaultValue}} = None /* = {{{defaultValue}}}*/{{/defaultValue}}{{^defaultValue}} = None{{/defaultValue}}{{/required}}{{#hasMore}}, {{/hasMore}}{{/allParams}}){{#returnType}}: Option[{{returnType}}]{{/returnType}} = {
```
wing328: @gmarz it's already default to None if `defaultValue` is not provided:
```
{{^defaultValue}} = None{{/defaultValue}}
```

I think the following should be changed from
```
/* = {{{defaultValue}}}*/
```
to
```
 = {{{defaultValue}}}
```
instead to use the default value provided in the spec.

(I do not recall why it's commented. Maybe it's not working for all default values)

ramzimaalej: @gmarz can you tell me what are the steps to reproduce the error, and I will test it.

gmarz: @wing328 sorry for the slow reply...

> (I do not recall why it's commented. Maybe it's not working for all default values)

Yea, I'm not sure. Maybe because it needs to be wrapped in `Some`...I think the current template will try and do `Option[Boolean] = false` which isn't valid?

@ramzimaalej there is no error/issue here. This change just defaults `Option` types to `None` for convenience.

ramzimaalej: @gmarz do you mind resolving the conflicts and I will take a look.

gmarz: @ramzimaalej done :)

wing328: @gmarz I think the current implementation should be able to meet your requirement:
```
/* = {{{defaultValue}}}*/{{/defaultValue}}{{^defaultValue}} = None{{/defaultValue}}
```
which is default to `None` if `defaultValue` is not provided.

In your spec, do you have the `defaultValue` set to something?

ramzimaalej:
@wing328 LGTM

modules/swagger-codegen/src/main/resources/scala/api.mustache
ramzimaalej: @gmarz do you mind double checking if current version solves the problem or not. if the problem is not solved, please create a ticket (if it does not exists) and attach the swagger-doc you used to reproduce the issue

wing328: The default value is commented out anyway so it should not be a problem merging this PR into master if the spec contains default value.  I'll create a separate "issue" to track the `defaultValue` enhancement to Scala API client.

jimschubert: @gmarz @wing328 @ramzimaalej the only issue I see with the previous default value is that if someone were to say `defaultValue = null` it would fail as `Some(null)` defines a value of null, which is unexpected. The optional value should be `Option({{defaultValue}})` to account for this edge case, as `Option(null)` becomes `None` and `Option(nonNull)` becomes `Some(nonNull)`.

PR #6540, opened 2017-09-21
[Scala] Properly handle csv collectionFormat

gmarz:
### PR checklist

### Description of the PR

Adds support for `"collectionFormat": "csv"`, which previously use to cause an exception.

So now

```json
{
  "name": "foo",
  "in": "path",
  "type": "array",
  "items": {
    "type": "string"
  },
  "collectionFormat": "csv"
}
```

will result in

`def someOperation(foo: List[String])`

(Oops...another fix snuck in here -- defaulting `None` for case classes...I can open a separate PR for this if desired)


wing328:
@gmarz thank for the PR.

cc @clasnake @foxmk @ramzimaalej

wing328:
> (Oops...another fix snuck in here -- defaulting None for case classes...I can open a separate PR for this if desired)

Please file a separate PR for defaulting None  in case classes as I found other Scala SDK not doing that so I think it's subject to more discussion before we can add that to the Scala generator.

gmarz:
@wing328 sure thing, I'll try and get to this soon.

gmarz:
> Please file a separate PR for defaulting None in case classes 

@wing328 I've reverted the change from this PR.

Perhaps it's appropriate to add it to #6538 since it's related. Let me know, otherwise I'll open another PR.

wing328:
> Perhaps it's appropriate to add it to #6538 since it's related. Let me know, otherwise I'll open another PR.

I would suggest filing another PR for the change.

Btw, what issues did you run into if not defaulting to None? A spec to reproduce the issue would be helpful in reviewing the change.

gmarz:
> Btw, what issues did you run into if not defaulting to None? A spec to reproduce the issue would be helpful in reviewing the change.

It's not really an issue per se, it's more for convenience. I'll open a PR with an example so we can keep the discussion here relevant to the change in this PR.

wing328:
I also left some comments in https://github.com/swagger-api/swagger-codegen/pull/6538. Please take a look when you've time.

modules/swagger-codegen/src/main/resources/scala/apiInvoker.mustache
```diff
@@ -52,6 +52,9 @@ class ApiInvoker(val mapper: ObjectMapper = ScalaJsonUtil.getJsonMapper,
   def escape(value: String): String = {
     URLEncoder.encode(value, "utf-8").replaceAll("\\+", "%20")
   }
+  def escape(values: List[String]): String = {
```
wing328: Later we will need to add another method `def escape(values: List[String], format: String): String = { ... }` to support other collection formats.

cc @clasnake

gmarz:
@wing328 

> Please file a separate PR for defaulting None in case classes as I found other Scala SDK not doing that so I think it's subject to more discussion before we can add that to the Scala generator.

Done. https://github.com/swagger-api/swagger-codegen/pull/6790

PR #6554, opened 2017-09-23
[python] Add default Configuration

mbohlool:
### PR checklist

### Description of the PR
Adding a default configuration for python client. A use-case for this is a config loader can be called at the beginning of the code to load config in the default location and other API calls do not need to load config again or keep a configuration object.


mbohlool:
cc @wing328 @taxpon @frol

frol:
This looks good to me, but the `test_separate_default_config_instances` test has to be fixed.

mbohlool:
@frol I've changed the implementation and this one passes test. please take another look.

frol:
I would organize everything into a single class with `__new__` / `__call__` / meta class implementation (it is a singleton pattern) + class-methods, because it will solve a few problems with the current approach: 

* `isinstance(config, Configuration)` will work
* the naming will be in line with PEP8 recommendations
* everything will be in a single place

mbohlool:
@frol  good point. I've changed it. Please take another look.

frol:
@mbohlool This looks much cleaner to me! Thank you! 

@wing328 I think, it is ready for merge.

PR #6555, opened 2017-09-24
[Python] Add __hash__ to Python API client's model

### LINKED ISSUE #6475: [Python] models implement __eq__ without __hash__
##### Description

The base model does not implement a hash function, but it does implement __eq__. This is illegal according to: https://docs.python.org/3/reference/datamodel.html#object.__hash__

The concrete side effect of this is that I cannot use swagger generated Python models in Set comparisons; models that are equal do not necessarily have the same hash.

##### Swagger-codegen version

2.2.3

1. Generate any swagger model for Python.
2. 
```
> Model() == Model()
True

> set([Model()]) == set([Model()])
False
```

wing328:
### PR checklist

### Description of the PR

To close #6475 

I'll file another PR later to sync the changes in Python model template to the Python Flask generator.

### DIFF
--- a/samples/client/petstore/python/petstore_api/models/additional_properties_class.py
+++ b/samples/client/petstore/python/petstore_api/models/additional_properties_class.py
@@ -148,3 +148,7 @@ class AdditionalPropertiesClass(object):
         Returns true if both objects are not equal
         """
         return not self == other
+
+    def __hash__(self):
+      return hash((self.map_property,
+          self.map_of_map_property))
--- a/modules/swagger-codegen/src/main/resources/python/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/model.mustache
@@ -226,5 +226,9 @@ class {{classname}}(object):
         Returns true if both objects are not equal
         """
         return not self == other
+
+    def __hash__(self):
+      return hash(({{#vars}}self.{{name}}{{#hasMore}},
+          {{/hasMore}}{{/vars}}))
 {{/model}}
 {{/models}}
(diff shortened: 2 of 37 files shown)


wing328:
cc @mbohlool @taxpon @frol

wing328:
cc @cbornet as well

samples/client/petstore/python/petstore_api/models/additional_properties_class.py
```diff
@@ -148,3 +148,7 @@ def __ne__(self, other):
         Returns true if both objects are not equal
         """
         return not self == other
+
+    def __hash__(self):
+      return hash((self.map_property,
+          self.map_of_map_property))
```
frol: Well, you will get `TypeError: unhashable type: 'dict'` here... I am not sure if this hash implementation make any sense since `dict` behaves the same way:

```
a = {1: 1}
b = {1: 1}
assert a == b  # OK
assert {a} == {b}  # AssertionError
```

wing328: My implementation is based on https://docs.python.org/3/reference/datamodel.html
```
def __hash__(self):
    return hash((self.name, self.nick, self.color))
```
https://stackoverflow.com/questions/2909106/whats-a-correct-and-good-way-to-implement-hash suggests something similar.

But feel free to suggest other ways to implement `__hash__`

mbohlool: Here the model is mutable. it would be wrong to implement a `__hash__` for it. [documentation](https://docs.python.org/3.5/reference/datamodel.html#object.__hash__) says:

> If a class defines mutable objects and implements an __eq__() method, it should not implement __hash__(), since the implementation of hashable collections requires that a key’s hash value is immutable (if the object’s hash value changes, it will be in the wrong hash bucket).

wing328: Thanks for the explanation. Closing this :)

PR #6557, opened 2017-09-24
[PHP] Use parameter name as description if not defined

wing328:
### PR checklist

### Description of the PR

To avoid missing description for parameter, we will use the parameter name as default if parameter description is not defined.

cc @ackintosh @jebentier @dkarlovi @mandrean @jfastnacht @arnested


ackintosh:
LGTM! 👍

wing328:
@ackintosh thanks for the quick review.

Later we can use the same approach for other undefined description (e.g. class, etc)

Issue #6379, opened 2017-08-25
[JAVA] Definitions that are only a type generate references to them instead of the type


christarczon:
##### Description

Definitions that are simply a type aren't generated correctly in Maven and CLI generators. Works **as expected** when generated at editor.swagger.io. I'm trying to generate a Spring Boot client (not available in online editor) but it happens for Java too.

##### Swagger-codegen version

2.2.3, 2.3.0-SNAPSHOT

##### Swagger declaration file content or url

```yaml
swagger: '2.0'
info:
  title: Test
  version: '1'
paths:
  '/list':
    get:
      responses:
        '200':
          description: ''
          schema:
            $ref: '#/definitions/MyList'
definitions:
  MyList:
    type: array
    items:
      $ref: '#/definitions/ListItem'
  ListItem:
    type: string
```

##### Command line used for generation

```java -jar ./modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -l java -i test.yaml -o test-gen```

##### Steps to reproduce

Generate the above yaml using CLI or Maven. The generated code imports and refers to a model that isn't generated.

Expected:
```public class MyList extends ArrayList<String> {```

Actual:
```public class MyList extends ArrayList<ListItem> {```

##### Suggest a fix/enhancement

I don't understand why this is different from the online editor. Besides the reference to the ListItem class that isn't generated, the files are identical.

wing328:
@christarczon can you please give it another try with the latest master? I seem to recall a fix merged into master recently.

christarczon:
@wing328 Just pulled and rebuilt, same thing.

bbdouglas:
I believe this is the same issue as #5382, which was (at least partially) addressed. I was able to reproduce this issue with the latest master, so the previous fix obviously didn't catch all the cases. I will take a look and see what more needs to be done to cover this case as well.

bbdouglas:
@christarczon As a short-term workaround, you can reverse the order of the definitions of ListItem and MyList. The issue is that MyList refers to a definition that wasn't defined lexically above it. The spec you wrote is correct, we just don't support that ordering yet for these kinds of "type only" definitions. A more comprehensive fix should be available soon.

bbdouglas:
I verified that with the latest master this issue has been fixed. I believe this issue can be closed.

Here is the definition of the MyList class now, generated with the spec given in the issue description:

    public class MyList extends ArrayList<String> {

PR #6559, opened 2017-09-25
Pre-calculate type aliases before processing models

bbdouglas:
A type alias in this context is where a model is simply another name for a
primitive type, such as `MyString` in the following model definitions:

    MyList:
      type: array
      items:
        $ref: '#/definitions/MyString'
    MyString:
      type: string

It is valid to use a type alias as a property in another object or array model,
even if the object/array is defined before the alias is, as in the example
above. However, the current alias logic only looks "back" in list of previously
defined models, meaning that `MyList` would not know that `MyString` is an
alias. This change fixes the incorrect behavior by pre-calculating the list of
aliases before any models are processed. It also changes the test endpoint to
verify the correct behavior even when an object is defined before an alias it
uses.

### PR checklist

### Description of the PR

This is a fix for Issue #6379. The problem was that the list of potential alias types was being built up as the models were being generated, and so a composite type (like an object or array) would only be aware of aliases defined above it in the spec. This change pre-calculates all the aliases before processing the first model definition. I have tested it by changing the test spec to use a definition pattern similar to the one in the bug report, and the same client code is generated either way.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -119,7 +119,7 @@ public class DefaultCodegen {
     // Then translated back during JSON encoding and decoding
     protected Map<String, String> specialCharReplacements = new HashMap<String, String>();
     // When a model is an alias for a simple type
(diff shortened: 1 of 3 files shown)


wing328:
cc @JFCote @sreeshas @jfiala @lukoyanov @cbornet @ePaul

PR #6571, opened 2017-09-26
[R] Added ApiClient and fixed other issues

brnleehng:
### PR checklist

### Description of the PR

Super excited to see a R swagger code generator! I've added a couple of changes and now have a fully working swagger client. 

1) Added **ApiClient.R**: Following the Python code generator and other languages, a HTTP client handler adds flexibility for users to build on top of the default swagger code if user needs such as special authorization handling. 

2) Changed Api operators and how they are called. Some parameters are required while others parameters are not. However, the current API operators forces you to have all of the parameters.

3) Fixed toJSON / fromJSON: The sprintf does not work if the values are NULL. Switched it to use lists instead of strings for flexbility
-  Object$new()$fromJSON : Can't call fromJSON from a new object

4) Added auto generated documentation in the mustache templates. It's somewhat ambiguous on how to use the models and operators, hopefully this would help. 

5) Fixed the Description file to use appName. However it doesn't completely solve the issue of generating a R file if appName has spaces - Then it will fail. 

6) Added Response and Element classes in R


modules/swagger-codegen/src/main/resources/r/api.mustache
```diff
@@ -1,56 +1,94 @@
...
+#' }
+#' 
+#' @export
 {{classname}} <- R6::R6Class(
   '{{classname}}',
   public = list(
     userAgent = "{{#httpUserAgent}}{{{.}}}{{/httpUserAgent}}{{^httpUserAgent}}Swagger-Codegen/{{{packageVersion}}}/r{{/httpUserAgent}}",
-    basePath = "{{{basePath}}}",
-    initialize = function(basePath){
-      if (!missing(basePath)) {
```
wing328: @brnleehng thanks for the PR. Does it mean developers (users of the SDK) can no longer use a different base path during runtime when constructing the class?

wing328: Ah please ignore. You've moved "basePath" to apiClient.

brnleehng:
@wing328 Hmm... I'm not sure why I'm seeing this error. I don't think any of my changes would affect the travis CI.

wing328:
@brnleehng the Travis CI (MacOS image) times out from time to time so please kindly ignore and we'll restart the job before the merge.

wing328:
Not for this PR but we probably want to add some tests (syntax checker, Petstore integration tests, etc) as we're seeing more changes to enhance the R generator.

PR #6579, opened 2017-09-27
#6545: Fixing several bugs for Java model generation for enums

### LINKED ISSUE #6545: [Java] Models with arrays of inner number types are wrongly, uncompiable generated
##### Description

Models with arrays of inner number types are wrongly generated. Generally a number type with enum is generated as `NUMBER_0(0)`.
If this number type with enum is part of an array it is generated as: `_0("0")` The `NUMBER` is missing and the value would be generated as String.
This doesn't even compile.

##### Swagger-codegen version

2.2.3

```json
        "ModelWithArrayHavingNumbersWithEnum": {
            "type": "object",
            "properties": {
                "arrayWithNumbers": {
                    "type": "array",
                    "items": {
                        "type": "integer",
                        "format": "int32",
                        "enum": [0, 1, 2]
                    }
                }
            }
        }
```

..

##### Steps to reproduce

..

##### Related issues/PRs

#5712, #5619

..

sdoeringNew:
### PR checklist

- [x] Ran the shell script under `./bin/java-petstore-jersey2.sh`, `./bin/java-petstore-okhttp-gson.sh` to update Petstore sample so that CIs can verify the change.
~Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.~
- [x] Filed the PR against the correct branch: `2.3.0` branch for changes related to OpenAPI spec **?.0**. Default: `master`
~Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.~
- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming langauge: @bbdouglas @JFCote @sreeshas @jfiala @lukoyanov 

### Description of the PR

fixes #6545


sdoeringNew:
Something is wrong with the Travis build. It's always timeouting.

modules/swagger-codegen/src/main/resources/Java/modelEnum.mustache
```diff
@@ -60,7 +60,7 @@ public enum {{#datatypeWithEnum}}{{{.}}}{{/datatypeWithEnum}}{{^datatypeWithEnum
 
     @Override
     public {{#datatypeWithEnum}}{{{.}}}{{/datatypeWithEnum}}{{^datatypeWithEnum}}{{{classname}}}{{/datatypeWithEnum}} read(final JsonReader jsonReader) throws IOException {
-      {{{dataType}}} value = jsonReader.{{#isInteger}}nextInt(){{/isInteger}}{{^isInteger}}next{{{dataType}}}(){{/isInteger}};
+      {{{dataType}}} value = {{#isFloat}}new Float(jsonReader.nextDouble()); // floats shall be restricted so this cast is safe{{/isFloat}}{{^isFloat}}jsonReader.{{#isInteger}}nextInt(){{/isInteger}}{{^isInteger}}next{{{dataType}}}(){{/isInteger}};{{/isFloat}}
```
lukoyanov: Can we overflow here while converting double => float?

modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenModel.java
```diff
@@ -41,6 +41,7 @@
 
     public Set<String> imports = new TreeSet<String>();
     public boolean hasVars, emptyVars, hasMoreModels, hasEnums, isEnum, hasRequired, hasOptional, isArrayModel, hasChildren;
+    public boolean isInteger, isFloat;
```
lukoyanov: This flag names a bit misleading for me (same names exists in CodegenProperty), may be isFloatEnum/isIntegerEnum?

sdoeringNew: Yes, this flag also exists in CodegenProperty. And it has the same name as it has the same meaning.

Imaging this definition:
```
definitions:
  IntegerValue:
    type: integer
    format: int32
  Object:
    type: object
    properties:
      IntegerValue:
        type: integer
        format: int32
```

The outer ```IntegerValue``` is an integer, so the flag ```isInteger == true```.
The inner ```IntegerValue``` is an integer, so the flag ```isInteger == true```.

Why should I call the flag differently?

modules/swagger-codegen/src/test/java/io/swagger/codegen/CodegenTest.java
```diff
@@ -482,10 +482,10 @@ public void fromModelObjectTest() throws NoSuchFieldException, IllegalAccessExce
     @DataProvider(name = "fromModelEnumTest")
     public static Object[][] fromModelEnumTest() {
         return new Object[][]{
-                {"IntegerEnum", true, false, "integer", Arrays.asList(1, 2, 3)},
-                {"LongEnum", false, false, "long", Arrays.asList(1, 2, 3) /* this seems wrong, the SwaggerParser should read the enum values as Long not as Integer */},
-                {"FloatEnum", false, true, "float", Arrays.asList("1", "2", "3") /* this seems wrong, the SwaggerParser should read the enum values as Float not as String */},
-                {"DoubleEnum", false, false, "double", Arrays.asList("1", "2", "3") /* this seems wrong, the SwaggerParser should read the enum values as Double not as String */},
+                {"IntegerEnum", true, false, "integer", Arrays.asList("1", "2", "3") /* @see: https://github.com/swagger-api/swagger-core/issues/2449 */},
```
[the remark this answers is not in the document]
sdoeringNew: If this test fails it is probably because the mentioned bug was solved and the swagger-core was updated in the swagger-codegen.

This link shall help to understand that there was a bug which was solved and because of that the test is now failing and can be corrected.

Without that link the contributor can only guess how the tests might be.


But I can remove the link and write a big fat comment explaining the whole situation and that test might fail in the future and what to do if that happens. I thought a link might shorter and is more capable of explaining the test situation.

modules/swagger-codegen/src/main/resources/Java/modelEnum.mustache
sdoeringNew: Yes. ```new Float(Double.MAX_VALUE)``` will result in ```Infinity```.

That is not the best solution.

To make it better the JsonReader functionality has to be approximately reimplemented. Something like this:
```Java
    final double nextDouble = jsonReader.nextDouble();
    float result = (float) nextDouble;
    if (Float.isInfinite(result)) {
      throw new MalformedJsonException("JSON forbids infinities: " + result + " for " + nextDouble);
    }
```
(Compare with: https://github.com/google/gson/blob/b1fb9ca9a1bea5440bc6a5b506ccf67236b08243/gson/src/main/java/com/google/gson/stream/JsonReader.java#L909)

But this seems like overkill.


My thoughts were:
In case a float value is restricted by enumerations this method will result in ```null``` if the enum value can't be matched. If a decimal number is to big for float, leading to ```Infinity```, the value can't be matched and the result will be ```null```, too.

lukoyanov: > But this seems like overkill.

ok, seems reasonable.

modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenModel.java
lukoyanov: @wing328 plz advice.

wing328: Previously we encounter the use case with array of array of enum. To make it easier to check whether the inner type is an integer, we use `isInteger`. If we need to ensure it's not an enum, we can `{{^isEnum}} .. {{/isEnum}}`

I agree it may be a bit misleading and it's tradeoff to make it easier to handle 2D or 3D array of enum

PR #6581, opened 2017-09-27
CsharpDotNet2Client - Use clientPackage in additionalProperties

pierredeman:
### Description of the PR
PR for review : @mandrean
- Use **clientPackage** from **additionalProperties** if provided.
- Give execution rights for **csharp-dotnet2-petstore.sh**


wing328:
@pierredeman thanks for the PR, which looks good to me.

If no further question or feedback on this PR, I'll merge it on coming Sunday.

cc @jimschubert

jimschubert:
One issue with this is that by importing the client package via usings, we will now clash all classes under the client package with API models.  We can avoid this by not including the usings and fully qualify all types from the client package.

modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java
```diff
@@ -1,54 +1,24 @@
...
-import io.swagger.models.Info;
-import io.swagger.models.License;
-import io.swagger.models.Model;
-import io.swagger.models.Operation;
-import io.swagger.models.Path;
-import io.swagger.models.RefModel;
-import io.swagger.models.SecurityRequirement;
-import io.swagger.models.Swagger;
-import io.swagger.models.Tag;
+import io.swagger.models.*;
```
jimschubert: I've seen it requested in other PRs not to use the star import and leave these as explicit imports. I don't remember who said this, though. @wing328 was this something we enforce, and is it documented anywhere? It could have just been the reviewer's preference

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CsharpDotNet2ClientCodegen.java
```diff
@@ -57,8 +47,9 @@ public void processOpts() {
         super.processOpts();
 
         if (additionalProperties.containsKey(CLIENT_PACKAGE)) {
-            this.setClientPackage((String) additionalProperties.get(CLIENT_PACKAGE));
+            setClientPackage((String) additionalProperties.get(CLIENT_PACKAGE));
         } else {
+            setClientPackage(packageName + ".Client");
```
jimschubert: Should this be done in the constructor, or the setClientPackage backed by a field with a default like others? If using this generator manually from code, we'll now have no default unless the user invokes processOpts.

If clientPackage has a sane default in the base type, please ignore. I haven't looked yet.

jimschubert:
@pierredeman I reviewed officially, and didn't see a new `using` like I thought I had with my first comment. Please ignore that comment. I probably misread the diff somewhere and thought the using was added to model.mustache.

modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java
wing328: Yup, not suggesting star import as mentioned in the style guide but I've seen people changing it a couple of times.

I'm usually ok accepting star import, especially for new contributors so as to encourage them contributing more.

jimschubert: I think I saw the comment on the Java generator. I'm cool with the star, as that's what I usually use as well. I think the argument against it is for type conflicts on merge, but I've only seen one such conflict in the past 3-4 years of working with Java. In C#, I saw it more often but it was easily solved with an alias.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CsharpDotNet2ClientCodegen.java
pierredeman: You're right, just pushed new commit with initialization in constructor

Issue #6592, opened 2017-09-29
[NancyFx] Allow for override of PackageContext


craffael:
##### Description

The current NancyFX generator constructs the C# namespace as follows: `<packageName>.<packageContext>`. Currently `<packageContext>` is set to a (sanitized) version of the swagger basepath.

In cases where the base path is non-simple, this can lead to strange namespaces. For example for `packageName=company.security` and `basePath=/security.svc/v1` the resulting namespace would be `company.security.security_svcv1` which is rather ugly.

##### Swagger-codegen version

Current master branch.

##### Command line used for generation

`java -jar .\swagger-codegen-cli.jar generate -i ..\swagger\pos\swagger.yaml -l nancyfx -o . --additional-properties packageName=company.security`

##### Suggest a fix/enhancement

I suggest to add a new nancyfx specific option called packageContext that allows us to overwrite the packageContext. I.e. we could use the command line

    java -jar .\swagger-codegen-cli.jar generate -i ..\swagger\pos\swagger.yaml -l nancyfx -o . --additional-properties packageName=company.security,packageContext=v1

to get the namespace `company.security.v1` which would be much nicer.

wing328:
Closed via #6593

PR #6593, opened 2017-09-29
[NancyFx] provide option to override package context

craffael:
### PR checklist

- [] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.

### Description of the PR
See #6592 

@mandrean @jimschubert


jimschubert:
Any idea why the generator splits these two properties rather than constructing packageName or package as other server generators do?

Asp.net Core:

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/aspnetcore/controller.mustache

Scalatra:
https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/scalatra/api.mustache

jimschubert:
Also, this change may need to be marked as a breaking change, as it will break any custom templates out in the wild.

craffael:
@jimschubert  I have no clue why the namespace is constructed with the help of the basepath. I agree with you that the solution of Asp.net core and scalatra is much nicer.
As far as I can judge, my proposed pull request is not a breaking change because it only adds additional functionality. But I guess you're playing with thought of replacing `<packageName>.<packageContext>` with `<packageName>`? (which would be a breaking change)

jimschubert:
@craffael I see now, I switched over to the full file before commenting, and thought that you had changed it from packageContext.packageContext, which I think is how it is described in the linked issue.

So yeah, not a breaking change... but I'd definitely push for alignment with properties used in templates.

craffael:
Ah sorry for the confusion, I updated the linked issue. It's of course `packageName.packageContext`. I don't fully get your last comment: Would you rather change the namespace to just `packageName` (which would be a BREAKING change)?

jimschubert:
@craffael Sorry, I seem to have forgotten the last sentence in that last comment.

I don't think it's a blocker or necessary change here, and could be opened as a new issue for the future.

However, I think it would be less of a breaking change now to move forward with package or packageName (whichever is more common) since there are already options in the default codegen to handle those. That way, there's not an extra property just for this generator which will likely become quickly deprecated.

It's a hard call, obviously, because we don't have any sort of telemetry to know the impact or what additional properties are actually being used.

jimschubert:
@wing328 @mandrean thoughts on the above?

wing328:
> As far as I can judge, my proposed pull request is not a breaking change because it only adds additional functionality.

I agree this is not a breaking change

> I agree with you that the solution of Asp.net core and scalatra is much nicer.

Can you please open a new "issue" (ticket) to track it to see if anyone from the community have time to work on the enhancement?

wing328:
> However, I think it would be less of a breaking change now to move forward with package or packageName (whichever is more common)

Both are pretty common but personally I prefer "packageName"
```
swagger-codegen|master ⇒ grep -R "{{package}}" modules/swagger-codegen/src/main/resources/* | wc -l
     156
swagger-codegen|master ⇒ grep -R "{{packageName}}" modules/swagger-codegen/src/main/resources/* | wc -l
     283
```

PR #6622, opened 2017-10-04
[haskell-http-client] add support for auth methods

jonschoning:
* add support for auth methods

* use newtypes for required params

* fix duplicate operationId issues

* prevent aliasing of vendorextension references in fromOperation
    
* add --fast to stack ci build


jonschoning:
fyi - pushed an update to this PR just now  
`* fix duplicate operationId issues`

wing328:
> * fix duplicate operationId issues

Does it address the issue that more than 1 endpoint has the same `operationId`?

jonschoning:
technically, the default generator was creating unique `operationId` for other lang generators, but for this lang generator I was getting more than 1 endpoint with the same operationId because of 2 issues:
* the fact that `DefaultCodegen.addOperationToGroup`, which generates unique operationIds, runs *after* `fromOperation`.  Therefore, the `HaskellHttpClientCodegen.fromOperation` sees non-unique operationIds.  This caused an issue because I calculate derived properties (vendorextensions) from the operationId seen in `fromOperation`.  
* incorrect operationId sanitation  

I fixed this by performing logic to generate new operationId's in `HaskellHttpClientCodegen.toOperationId`

wing328:
@jonschoning thanks for the explanation

(CircleCI tests passed via https://circleci.com/gh/swagger-api/swagger-codegen/1935)

PR #6638, opened 2017-10-08
[haskell-http-client] bug fixes; path & newtype generation issues

jonschoning:
* fix path generation/param-substitution issues

* fix newtype de-duplication issues

* refactoring 

* correct version in comments

* prevent duplicate MimeTypes

* sort parameter newtypes


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/HaskellHttpClientCodegen.java
```diff
@@ -398,13 +394,7 @@ public void preprocessSwagger(Swagger swagger) {
...
         additionalProperties.put("swaggerVersion", swagger.getSwagger());
 
-        //copy input swagger to output folder
-        try {
-            String swaggerJson = Json.pretty(swagger);
-            FileUtils.writeStringToFile(new File(outputFolder + File.separator + "swagger.json"), swaggerJson);
-        } catch (IOException e) {
-            throw new RuntimeException(e.getMessage(), e.getCause());
-        }
+        WriteInputSwaggerToFile(swagger);
```
wing328: @jonschoning What about using NodeJS and other server generator approach? https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NodeJSServerCodegen.java#L299

jonschoning: * the method i used to copy the input swagger is the same method as https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaJAXRSSpecServerCodegen.java#L153-L161

* in `NodeJSServerCodegen.java`,  `swagger.mustache` uses the template variable `swagger-yaml` which appears to be populated in https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NodeJSServerCodegen.java#L386-L402 , but I'm not sure exactly what it's doing.

wing328: OK. Thanks for pointing it out. I wasn't aware JAXRS spec generator is using a different approach.

We'll accept both approach.

Issue #6641, opened 2017-10-09
Swift4: Need additional initializer which uses properties


ehyche:
##### Description

I am filing this issue to track the discussion which took place in this PR for swift3: https://github.com/swagger-api/swagger-codegen/pull/6492

The request was to add an additional initializer for swift4-generated models. This initializer would take all of the properties, and thus would allow you to initialize a model object without having to decode it from JSON.

##### Swagger-codegen version

All

##### Swagger declaration file content or url

This is a feature request. See the discussion at the end of the PR listed above for context.

Kimnyhuus:
Hi' @ehyche!
Do you by any chance know, when this will be merged and available in Swagger?
 - I'm guessing that it will be available in Swagger Codegen 2.3.0, when merged?

wing328:
@Kimnyhuus https://github.com/swagger-api/swagger-codegen/pull/6642 has been merged into master.

Can you pull the latest master or use the SNAPSHOT version to give it a try?

Kimnyhuus:
@wing328 Sure! Thanks for the quick response :-)

wing328:
@Kimnyhuus no problem. Let me know if you still need help on this.

Closing this for the time being as the PR has already been merged.

Kimnyhuus:
@wing328: 
We've now generated code from master on v2.3.0, and the initialisers are present now.
There's one thing missing though. On Swift3 the initialisers had the ability to make all optional model variables '=nil' in the parameter, making it possible to initialise an object with only some or none of the parameters. 

Example:

```
Model:

Open class SwaggerModel: Codable

public var foo: String
public var bar: Int?

public init(foo: String, bar: String?=nil) {
        self.score = score
        self.user = user
}


In class usage:

var swaggerModel = SwaggerModel(foo: "Leet")
swaggerModel.bar = 1337
```
This is not the case with the current version of v2.3.0, which does not generate the '=nil'.
Can you add this again? :-)

ehyche:
@Kimnyhuus : yeah we can add this. I should have added it previously. 

Of course, the "= nil" will only get added if:

- the property is optional (and thus is a Swift Optional); AND
- no default is provided in the schema. If a default is already provided in the schema, then that will be used for the default.

ehyche:
@Kimnyhuus : would mind filing a separate issue for this? That way, I can tie the PR to a new Issue and not one that has already been closed. Thanks!

Kimnyhuus:
[...]

PR #6642, opened 2017-10-09
Swift4: Add additional initializer for initializing model object with properties.

ehyche:
### Description of the PR

This change fixes this issue: https://github.com/swagger-api/swagger-codegen/issues/6641

It adds an additional initializer which allows model objects to be initialized using the properties. For example, if we had this model:

```
        "ErrorInfo": {
          "type": "object",
          "properties": {
            "code": {
              "type": "integer",
              "format": "int32"
            },
            "message": {
              "type": "string"
            },
            "details": {
              "type": "array",
              "items": {
                "type": "string"
              }
            }
          },
          "description": "Example Error object"
        },
```
    
With this change, we generate an initializer for this model object like this:
    
```
        public init(code: Int?, message: String?, details: [String]?) {
            self.code = code
            self.message = message
            self.details = details
        }
```

### PR checklist

- [x ] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [x ] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming langauge.


jaz-ah:
@ehyche is it worth adding an additional check if we have additionalProperties so we don't generate empty init() { } functions?

ehyche:
@jaz-ah : good suggestion. Will add that check.

ehyche:
@jaz-ah : updated to add hasVars checks around the initializer per your suggestion.

I had also forgotten to run the scripts, so I re-ran these scripts:

- bin/swift4-petstore-all.sh
- bin/swift4-petstore-objcCompatible.sh
- bin/swift4-test.sh

jaz-ah:
+1 looks good thx @ehyche - @wing328 ready to merge

wing328:

@ehyche thanks for the enhancement

cc @jgavris @Edubits

jgavris:
Nice! I might have mentioned this before, it's too bad we couldn't have made these `struct`'s and gotten convenience initializers auto-generated since the `Codable` protocol requires an initializer.

ehyche:
@jgavris : that's a really good idea to make them struct's. Then we get value semantics and immutability for free. I've been thinking about that, and will look into it.

PR #6659, opened 2017-10-11
JAXRS-SPEC: fix container return type

carnevalegiacomo:
Fixed container return type.
**Es.**
_type: array
  items:
   type: string_

**Without fix:**
  _List<List<String>>_
**Fixed:**
  _List<String>_

 PR checklist

### Description of the PR


wing328:
Please run the following to update the JAXRS-spec petstore samples so that the CIs can verify the result:
```
./bin/jaxrs-spec-petstore-server.sh
./bin/jaxrs-spec-petstore-server-interface.sh
```

carnevalegiacomo:
https://github.com/swagger-api/swagger-codegen/pull/6659/commits/010d14a90641b1f38ca941febc637795a279b76e

Run
./bin/jaxrs-spec-petstore-server.sh
./bin/jaxrs-spec-petstore-server-interface.sh

wing328:
cc @bbdouglas  @JFCote  @sreeshas  @jfiala  @lukoyanov @cbornet

modules/swagger-codegen/src/main/resources/JavaJaxRS/spec/apiInterface.mustache
```diff
@@ -11,4 +11,4 @@
     }{{/hasAuthMethods}}, tags={ {{#vendorExtensions.x-tags}}"{{tag}}"{{#hasMore}}, {{/hasMore}}{{/vendorExtensions.x-tags}} })
     @ApiResponses(value = { {{#responses}}
         @ApiResponse(code = {{{code}}}, message = "{{{message}}}", response = {{{baseType}}}.class{{#returnContainer}}, responseContainer = "{{{returnContainer}}}"{{/returnContainer}}){{#hasMore}},{{/hasMore}}{{/responses}} })
-    {{>returnTypeInterface}} {{nickname}}({{#allParams}}{{>queryParams}}{{>pathParams}}{{>headerParams}}{{>bodyParams}}{{>formParams}}{{#hasMore}},{{/hasMore}}{{/allParams}});
\ No newline at end of file
+    {{>returnTypeInterface}} {{nickname}}({{#allParams}}{{>queryParams}}{{>pathParams}}{{>headerParams}}{{>bodyParams}}{{>formParams}}{{#hasMore}},{{/hasMore}}{{/allParams}}) throws Exception;
```
cbornet: Why the throws Exception ?

wing328: If we dont' have that in other JAXRS generator, it should be removed (not be added in the 1st place)

cc @carnevalegiacomo

carnevalegiacomo: Hi
i added throws Exception because jaxrs provides the class javax.ws.rs.ext.ExceptionMapper to catch and re-throw exceptions.

Example:
public class CustomExceptionMapper implements ExceptionMapper<Exception> {
    @Override
    public Response toResponse(Exception exception) {
        if (exception instanceof org.springframework.security.access.AccessDeniedException) {
             CustomError error = new CustomError();
             ......
             return Response.status(Response.Status.FORBIDDEN).entity(error).build();
        }
        ......
    }

}

cbornet: Having an interface throwing Exception is bad design for me.

cbornet: `org.springframework.security.access.AccessDeniedException` is a RuntimeException so there's no need to throw Exception in the signature.

carnevalegiacomo: org.springframework.security.access.AccessDeniedException is an example

jarlesat: "throws Exception" is not only a bad design pattern, it is a horrible design pattern, you could in fact consider it a design bug. As @cbornet mentions AccessDeniedException is a RuntimeException, in fact all exception classes in spring are RuntimeExceptions just to avoid the need to declare them in a throws clause. An ExceptionMapper works just as well with no need to uglify your code with a horrible, horrible "throws Exception".

wing328: What about removing "throws Exception"? If needed, the developer can customize the template with the `-t` option to include it.

PR #6664, opened 2017-10-12
[C++] Sanitize operation ids.

fvarose:
### PR checklist

### Description of the PR

Each operation described in a swagger spec has an ID which is used by the generator directly to define the corresponding C++ function name.
This PR makes sure `operationId`s are sanitized, so that valid C++ code is generated (no spaces, dots, etc in function names).


wing328:
@fvarose thanks for the PR.

cc @ravinikam @stkrwork

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCppCodegen.java
```diff
@@ -128,6 +128,11 @@ public String escapeReservedWord(String name) {
     }
 
     @Override
+    public String toOperationId(String operationId) {
+        return sanitizeName(super.toOperationId(operationId));
```
stkrwork: What about the case if the parameter is an empty string? If its not overwritten, it checks for that.

fvarose: This case is covered by the call to the default implementation (`super.toOperationId`), which throws if the string is empty.

wing328: @fvarose what about using something similar to [RubyClientCodegen.java#L619-L635](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RubyClientCodegen.java#L619-L635)?

It just a matter of time we need to handle these edge cases.

fvarose:
@wing328 @stkrwork 
I've added the check for reserved keywords. Unlike the Ruby client, the empty string case is performed by the default implementation, I think it's cleaner that way and avoids code duplication.
I also added the modern C++ keywords that were missing (see [reference list](http://en.cppreference.com/w/cpp/keyword)).

PR #6674, opened 2017-10-13
Swagger eiffel:fix

jvelilla:
### PR checklist

### Description of the PR

Updated API_CLIENT.parameter_to_string feature, missing STRING representation.
Updated PetStore sample to use the latest code.


wing328:
@jvelilla thanks for the PR, which has been merged into master.

Quick question on `eiffel` if you don't mind. Is there a way to do a compile test on the Eiffel Petstore sample to ensure it has no syntax error? I did a google search but couldn't find it.

We can then setup the compile test in the CIs

PR #6687, opened 2017-10-15
[JAXRS-CXF] Add constructors to POJOs

shpandrak:
Before this change, POJOs where generated without regular constructors.This would mean you have to use the default constructor and setters or builder-style methods.
This is problematic when trying to make changes to the swagger file by adding required fields.
Before the change, the model consumer code will compile even when the callers does not supply all required fields when constructing a model object.
After the change, such addition will break the compilation and by that move important model validation from runtime to compile time.
For that reason, Our project code style requires use of constructors, and that is the motivation for this contribution.
This change adds a regular constructor to all POJOs, and specifically to the generated models.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenProperty.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenProperty.java
@@ -37,7 +37,8 @@ public class CodegenProperty implements Cloneable {
     public boolean exclusiveMinimum;
     public boolean exclusiveMaximum;
     public boolean hasMore, required, secondaryParam;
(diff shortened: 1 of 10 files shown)


cbornet:
As already said in #4686, if we have a constructor with required params, then the empty one shall be removed. Also the "all params" constructor should be a cli option.
The required params should also be annotated with `NotNull`

wing328:
@shpandrak thanks for the PR. Do you need help adding the CLI option as mentioned by @cbornet ?

shpandrak:
> @shpandrak thanks for the PR. Do you need help adding the CLI option as mentioned by @cbornet ?

I was planning to do that, though would appreciate help if possible :)

Issue #6400, opened 2017-08-29
[SWIFT3] URL parameters are not escaped


hce:
##### Description

swagger-codegen does not properly escape parametrized parts of a URL.

For example:

        var path = "/recipient/{recipient}"
        path = path.replacingOccurrences(of: "{recipient}", with: "\(recipient)", options: .literal, range: nil)

The "recipient" variable should be URLEscaped, but it isn't. Thus, if recipient contains spaces or illegal characters, the request fails. Depending on the source of the parameters, this could also lead to vulnerable code.

##### Swagger-codegen version

git commit id 18c57a65ed5def618be6495e86fd2d0c73f89882

##### Swagger declaration file content or url

n/a

##### Command line used for generation

not relevnt

##### Steps to reproduce

Use a swagger API that has parameters in the URL, for example: /article/{articleName}, then use "Hello World ../" as articleName

##### Related issues/PRs

n/a

##### Suggest a fix/enhancement

Properly url escape the passed parameters:

let param1 = param.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed)

wing328:
@hce I would suggest you to try the latest master if it's still an issue.

SNAPSHOT version: https://github.com/swagger-api/swagger-codegen#compatibility

hce:
Sorry for the late reply; tested and it works fine! Thanks!

hce:
My mistake, the issue *still* does exist.

Just tested again with the 30a176126aa3600b6252f80b84040a1ab351c125 commit (latest)

wing328:
@hce when you've time, can you also test the `swift4` generator to see if it has similar issues?

cc @ehyche @jaz-ah @jgavris @Edubits

hce:
Confirmed for the swift4 generator!

ehyche:
@hce : thanks for checking on this. I should have a PR fixing this for swift4 next week.

ehyche:
This PR was opened to fix this issue in swift3: https://github.com/swagger-api/swagger-codegen/pull/6705

xiaochow:
@ehyche thanks for your work! The same issue still exists for swift4 generator, I have been manually escaping them in the client side code. I wonder if you could make the changes to that when you have time? Thank you!

wing328:
@xiaochow do you mind filing a PR to fix it in Swift4 generator?

xiaochow:
[...]

PR #6705, opened 2017-10-16
Swift3: properly percent-escape path parameters

ehyche:
### PR checklist

- [x ] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [x ] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming langauge.

### Description of the PR

This change fixes the following issue:

https://github.com/swagger-api/swagger-codegen/issues/6400

The problem was that path parameters were not properly percent-escaped before being placed into the URL path. This leads to creation of an invalid URL, which then fails. So therefore, in the API template where path parameters are handled, we propertly percent escape them, using the characters which are allowed in URL paths.

In addition to this template change, then this PR includes the following changes:

1. Resulting changes in all generated code due to the above template change.
2. I added the objcCompatible run to the swift3-petstore-all.sh so that I could re-generated all of the generated code with a single script.
3. I added a unit test in UserAPITests.swift which verifies that paths are properly escaped.
4. In order to make the unit test work, then I needed access to RequestBuilder<T>.URLString to verify that the path was properly escaped. However, RequestBuilder<T>.URLString had "internal" access control so it was inaccessible from the unit test. So therefore, I made four contants in RequestBuilder<T> to be public. This should not harm anything, since they are constants ("let's") and cannot be changed from the outside of the class after initialization.
5. There were also some stray changes which look like they were caused by having not run bin/swift3-petstore-all.sh in a while.

@jaz-ah @jgavris @Edubits : please review


jaz-ah:
+1 LGTM

jgavris:
Nice. This is definitely a breaking change though right? I assume many people are already working around this by escaping some parameters manually.

ehyche:
@jgavris : yes, folks would have had to work around this bug by passing in already-escaped path params. So it shouldn't break anyone, although their pre-escaping is unnecessary now.

jgavris:
Won't double-encoding the parameters be an issue? That is assuming they see this change and update their request builders appropriately.

ehyche:
@jgavris : yep, you' right. I verified that double-escaping could be an issue in a playground:

```
let pathParam = "/Some/Path With Spaces In It"
let escapedPathParam = pathParam.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed) ?? ""
let escapedEscapedPathParam = escapedPathParam.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed) ?? ""
print("pathParam=\(pathParam)\nescapedPathParam=\(escapedPathParam)\nescapedEscapedPathParam=\(escapedEscapedPathParam)")
```

which prints

```
pathParam=/Some/Path With Spaces In It
escapedPathParam=/Some/Path%20With%20Spaces%20In%20It
escapedEscapedPathParam=/Some/Path%2520With%2520Spaces%2520In%2520It
```

I had thought that:

```
.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed)
```

would be idempotent, but apparently '%' is not an allowed path character.

What do you think? Should I percent-decode and then re-enode? Or should we just make people aware of this potentially-breaking change? In practice, I think seeing path parameters with whitespace should be pretty rare.

hce:
I'd just comply with the (swagger) standard, and if that says that spaces are allowed as a parameter, then they should be correctly escaped (principle of least surprise)

hce:
=> If someone worked around a broken library before, they shouldn't be surprised if their code eventually breaks.

jgavris:
I suppose I agree, it's not ideal but it was broken already. Let's push this through?

hce:
I fixed the problem in my projects by manually modifying the *generated files* after each re-generation, thus a fix wouldn't cause any problems.

wing328:
Tests passed via https://travis-ci.org/swagger-api/swagger-codegen/builds/294707015

Issue #6703, opened 2017-10-16
Allow using help command with run-in-docker.sh/docker-entrypoint.sh


tgifford-webfirst:
##### Description

docker-entrypoint.sh does not find the "help" \@Command even though it is a valid command.

##### Swagger-codegen version

2.3.0-SNAPSHOT (b22fb30891a8c6a576ead24358dbcd77af2f75c8)

##### Steps to reproduce

```sh
$ ./run-in-docker.sh help
++ dirname ./run-in-docker.sh
+ cd .
+ maven_cache_repo=/home/ted/.m2/repository
+ mkdir -p /home/ted/.m2/repository
++ id -u
++ id -g
+ docker run --rm -it -w /gen -e GEN_DIR=/gen -e MAVEN_CONFIG=/var/maven/.m2 -u 1000:1000 -v /home/ted/swagger-codegen:/gen -v /home/ted/.m2/repository:/var/maven/.m2/repository --entrypoint /gen/docker-entrypoint.sh maven:3-jdk-7 help
match
/gen/docker-entrypoint.sh: line 26: exec: help: not found
```

##### Suggest a fix/enhancement

Allow using the help command. Quick workaround patch:

```patch
diff --git a/docker-entrypoint.sh b/docker-entrypoint.sh
index 556b6ff..b6e7804 100755
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -11,7 +11,7 @@ codegen="${cli}/target/swagger-codegen-cli.jar"
 cmdsrc="${cli}/src/main/java/io/swagger/codegen/cmd"
 
 pattern="@Command(name = \"$1\""
-if expr "x$1" : 'x[a-z][a-z-]*$' > /dev/null && fgrep -qe "$pattern" "$cmdsrc"/*.java; then
+if expr "x$1" : 'x[a-z][a-z-]*$' > /dev/null && fgrep -qe "$pattern" "$cmdsrc"/*.java || expr "$1" = 'help' > /dev/null ; then
     # If ${GEN_DIR} has been mapped elsewhere from default, and that location has not been built
     if [[ ! -f "${codegen}" ]]; then
         (cd "${GEN_DIR}" && exec mvn -am -pl "modules/swagger-codegen-cli" -Duser.home=$(dirname MAVEN_CONFIG) package)
```

wing328:
@tgifford-webfirst please submit a PR for the suggested enhancement.

wing328:
[...]

PR #6706, opened 2017-10-16
Allow using help command with run-in-docker.sh

tgifford-webfirst:
### Description of the PR

See #6703. docker-entrypoint.sh doesn't find the "help" command with fgrep. This is a simple workaround--there may be a better approach.


wing328:
I think a better way is to add a "Help.java" class in "modules/swagger-codegen-cli/src/main/java/io/swagger/codegen/cmd"
```
swagger-codegen|r_fix_body ⇒ ls modules/swagger-codegen-cli/src/main/java/io/swagger/codegen/cmd
ConfigHelp.java        Generate.java          Langs.java             Meta.java              Validate.java          ValidateException.java Version.java
```

Your PR is acceptable and let's see if someone from the community can help create the "Help.java" later.

PR #6708, opened 2017-10-16
[Java][JAXRS-CXF] Improve API documentation in the CXF Server stub and Client generation

### LINKED ISSUE #6696: [Java][CXF] Include more of the API documentation in the CXF Server stub generation
##### Description

This is a feature request. I want to include more of the API documentation in the CXF Server stub generation so that the Java code is more complete.

##### Swagger-codegen version

2.3.0

..

..

##### Steps to reproduce

..

##### Related issues/PRs

..

Here is the change that I propose: https://github.com/gustavoapaz/swagger-codegen/commit/05eefe2adbde8a9b8cb44c7bdccd5a81ab9daba3

gustavoapaz:
### PR checklist

### Description of the PR

I included more of the API documentation in the CXF Server stub and Client generation. 

fix #6696

@bbdouglas, @JFCote, @sreeshas, @jfiala, @lukoyanov, @cbornet, @wing328

### DIFF
--- a/modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf/pom.mustache
+++ b/modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf/pom.mustache
@@ -5,7 +5,7 @@
   <packaging>jar</packaging>
   <name>{{artifactId}}</name>
   {{#appDescription}}
-  <description>{{{appDescription}}}</description>
+  <description>{{appDescription}}</description>
   {{/appDescription}}
   <version>{{artifactVersion}}</version>
   <build>
--- a/modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf/pom.mustache
+++ b/modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf/pom.mustache
@@ -4,6 +4,9 @@
   <artifactId>{{artifactId}}</artifactId>
   <packaging>jar</packaging>
   <name>{{artifactId}}</name>
+  {{#appDescription}}
+  <description>{{{appDescription}}}</description>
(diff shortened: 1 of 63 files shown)


JFCote:
I'm not sure if this should be by default or by adding a CLI options. @wing328, what do you think?

In the Play Framework, I don't put any documentation in the generated code since the documentation is the spec itself (YAML). It's like duplicating information at 2 places. But I guess there are scenario when this is needed so a CLI options might be best.

gustavoapaz:
@ricardona what do you think about this?

wing328:
> In the Play Framework, I don't put any documentation in the generated code since the documentation is the spec itself (YAML). 

In some other server generators, I think we do include the doc in the auto-generated code and the IDE should be able to display those when the developers are working on the code.

In API clients, we definitely put the doc (summary, notes, description) in the code as the one using the API clients may not have the spec with them.

JFCote:
@wing328 Yes, for the client it makes total sense but for the server, not so much. But if other server do it, I won't object.

modules/swagger-codegen/src/main/resources/JavaJaxRS/cxf/pom.mustache
```diff
@@ -4,6 +4,9 @@
   <artifactId>{{artifactId}}</artifactId>
   <packaging>jar</packaging>
   <name>{{artifactId}}</name>
+  {{#appDescription}}
+  <description>{{{appDescription}}}</description>
```
wing328: @gustavoapaz I think you will need to use {{appDescription}} instead to use the HTML-escaped value in handling special characters like <, > in the description.

Issue #5104, opened 2017-03-17
[TypeScript] Generators (node, fetch, angular, angular2, jquery) refactoring


wing328:
##### Description

We want to refactor/consolidate all Typescript generators (node, fetch, angular, angular2, jquery) into one single `typescript` generator and support node, fetch, angular, angular2, jquery via the `library` option. 

The goal is to reuse templates (e.g. model) if possible and for easier maintenance/support.

##### Swagger-codegen version

2.3.0 (since it's breaking change)

##### Suggest a Fix

- put the templates under `typescript/{name}`
- put shared templates under `typescript/common` and reuse if possible

If anyone has feedback or wants to work on this, please reply to let us know.

bherila:
@wing328 I am worried the template for this will become too complex/unwieldy, I already feel like editing the mustache templates is mind bending :-o

Also, I'm having buyers remorse about my own investment in the jquery library. Since I end up consuming the Typescript from multiple places (React Native, Node, and Client side) I am considering writing *yet another one* that uses just regular XmlHttpRequest. 

How do folks feel about deprecating lots of these in favor of `XmlHttpRequest`? I don't have time right now but I wouldn't mind putting it up on Freelancer.com or something.

mmkal:
@bherila isn't [fetch api](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) a better candidate as a lowest common denominator? You should be able to use something like [this](https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/typescript-fetch/builds/default/api.ts) it for React Native, Node and Client side.

As an aside, I agree that the current TypeScript offerings aren't quite sufficient. Two things I would like from TypeScript are:

* [`express`](https://expressjs.com/en/4x/api.html) types, for a node server. i.e. Interfaces for functions which produce an express middleware, that you can implement. This would make sure that node servers are abiding by their own contracts (and give accurate intellisense etc. while developing).
* React/React PropTypes.

If @wing328's suggestion would make it easier to add those, I would be all for it.

wing328:
@bherila 

> How do folks feel about deprecating lots of these in favor of XMLHttpRequest? 

Based on my experience with other generators (e.g. Java, PHP, C#, etc), there's always a need for another HTTP library since it's unlikely there's one HTTP library that can meet all the requirements.

> I already feel like editing the mustache templates is mind bending :-o

[...]

PR #6739, opened 2017-10-18
Feature/replace isomorphic fetch with portable fetch

### LINKED ISSUE #6738: typescript-fetch client isomorphic dependency does not work on React Native
##### Description
The generated code from the typescript fetch client generator cannot be used in React Native. There are multiple open issues related to this:
* https://github.com/matthew-andrews/isomorphic-fetch/issues/125 
* https://github.com/matthew-andrews/isomorphic-fetch/issues/59
* https://github.com/matthew-andrews/isomorphic-fetch/issues/48
The maintainer does not comment on these issues, nor does he respond to the posted pull requests to add react native support. So I would love to see the dependency to isomorphic-fetch will be replaced by portable-fetch (fork with react native support)

##### Swagger-codegen version
2.3.0

##### Steps to reproduce
Use any swagger file to generate client and use it in a React Native project. It will give the following error: fetch is not a function

replace the dependency from isomorphic-fetch to portable-fetch

jeff-99:
### Description of the PR

Fixes #6738

I replaced the isomorphic-fetch dependency with a well maintained fork (portable-fetch https://github.com/Knotis/portable-fetch) which supports React Native


wing328:
@jeff-99 thanks for the PR but the CIs failed with the following:
```
> @swagger/typescript-fetch-petstore@1.0.0 postinstall /Users/travis/build/swagger-api/swagger-codegen/samples/client/petstore/typescript-fetch/builds/es6-target
> npm run build

> @swagger/typescript-fetch-petstore@1.0.0 build /Users/travis/build/swagger-api/swagger-codegen/samples/client/petstore/typescript-fetch/builds/es6-target
> tsc --outDir dist/

api.ts(16,32): error TS7016: Could not find a declaration file for module 'portable-fetch'. '/Users/travis/build/swagger-api/swagger-codegen/samples/client/petstore/typescript-fetch/builds/es6-target/node_modules/portable-fetch/fetch-npm-node.js' implicitly has an 'any' type.

  Try `npm install @types/portable-fetch` if it exists or add a new declaration (.d.ts) file containing `declare module 'portable-fetch';`
```

jeff-99:
@wing328 , comitted the missing files. Build now seems to fail at the same step as the master branch.

TiFu:
The client security tests for typescript-fetch still contain isomorphic-fetch. Could you please regenerate the security-test? (see https://github.com/jeff-99/swagger-codegen/blob/feature/replace-isomorphic-fetch-with-portable-fetch/bin/security/typescript-fetch-petstore.sh)

jeff-99:
@TiFu the security tests are regenerated

wing328:
@jeff-99 I'll take another tomorrow and merge accordingly if no question from me. Thanks again for the PR.

@TiFu thanks for reviewing the change.

wing328:
@jeff-99 thanks for the PR, which has been merged into master.

samples/client/petstore/typescript-fetch/builds/es6-target/api.ts
```diff
@@ -35,7 +36,7 @@ export const COLLECTION_FORMATS = {
  * @interface FetchAPI
  */
 export interface FetchAPI {
-    (url: string, init?: any): Promise<Response>;
+    (url: string, init?: any): Promise<any>;
```
masaeedu: @wing328 @jeff-99 Is this change intentional? It reverts #6774.

masaeedu: What I don't understand is: if I run the tests today, without modifying anything, I find that this is changed back to `Response`. So something is broken in the way CI is being done for pull requests, because if the CI was working correctly it would reject this change.

I suspect the fact that there is a merge commit in here somehow confused things. In the future, it is probably best to require all pull-requests to be a single chain of commits that is rebased to the tip of master.

masaeedu:
Aside from the comments above, it is probably not a good idea to constantly be modifying the fetch dependency in here. If you need a special `fetch` implementation to suit your needs, you can just inject it when creating your `*API` instance.

If it wasn't a breaking change I would suggest removing the default fetch from this generator entirely, and make `fetch` a required argument.

kenisteward:
I believe the purpose of this change wasn't to just flip to a special
implementation. I think there was a problem with isomorphic and they
weren't resolving it so the pr changed to another that is a fork of
isomorphic but fixes more issues .  In general that seems fine but if there
is a way to abstract the library from the client and let them know they
need to choose a client that would be great.  I still think having a
default client is great though.

masaeedu:
@kenisteward The current implementation already abstracts the library from the client. So when you instantiate your `WhateverAPI`, you can pass in an instance of `fetch`. For my use case I use `node-fetch`, which I import `fetch` from and inject as an argument for that parameter.

I think what would be better than having `isomorphic-fetch` or `portable-fetch` or whatever as a dependency of the generated library would be to simply use global `fetch` as a default value for the fetch API parameter. That way the user can just import whatever polyfill they prefer before they import the library and things will work fine.

There probably isn't a way to represent this in a good way in the types, but you could make it a runtime error if you instantiate `*API` and the fetch API argument ends up being undefined.

samples/client/petstore/typescript-fetch/builds/es6-target/api.ts
masaeedu: What I don't understand is: if I run the tests today, without modifying anything, I find that this is changed back to Response. So something is broken in the way CI is being done for pull requests, because if the CI was working correctly it would reject this change.

I suspect the fact that there is a merge commit in here somehow confused things. In the future, it is probably best to require all pull-requests to be a single chain of commits that is rebased to the tip of master.

TiFu:
Ideally we would hide all request code (e.g. fetch, and whatever other request methods are out there) behind one unified interface - making future integration with the other typescript code generators far easier and maybe provide one/some default implementations of that interface e.g. for node-fetch, isomorphic-fetch
I believe @kenisteward wanted to look into merging all ts generators (#5104).

Issue #6741, opened 2017-10-19
[JAVA e.g. feign] Need a common interface for generated ApiClient classes


charlescapps:
##### Description

Currently, every time you generate client code in Java for a Swagger API, the `ApiClient` class that is generated will be an entirely different fully qualified class name (FQCN) and there is no common interface indicating what you can do with these objects.

In general, what you can do with an `ApiClient` could be different for each `library` e.g. `feign` may support different options than `resteasy` or `jersey2`. 

We want the ability to have some shared code for configuring these `ApiClient` objects. Unless we want to use reflection for every call (which is brittle, annoying, and potentially slow if we are creating a new client for every HTTP request), we need some common interface.

We unfortunately can't just package the common interface in the generated code; it has to have a constant package name. Why not? Suppose microservice A depended on client libs from microservice B and microservice C, then it would get the same FQCN from maven from both generated client libraries (the FQCN of the common interface)! 

Thus, the only real solution is to define a new module somewhere in this project that has these common interfaces. I think it makes the most sense to define a new module for each java library type, so that we don't pollute the generated code with dependencies they don't need.

##### Swagger-codegen version

2.3.0-SNAPSHOT

##### Swagger declaration file content or url

n/a this happens for any swagger.json. The `ApiClient` class in the generated code maven module has no base interface.

##### Command line used for generation

`config.json`:
```json
{
  "library": "feign",
  "artifactId": "${CLIENT_LIB_ARTIFACT_ID}",
  "groupId": "${SERVICE_GROUP_ID}",
  "artifactVersion": "${VERSION}"
}
```

```bash
java -jar swagger-codegen-cli.jar \
  generate -i swagger.json \
  -l java \
  -o client_lib \
  -c config.json \
  --api-package "${CLIENT_LIB_PACKAGE}.api" \
  --model-package "${CLIENT_LIB_PACKAGE}.model" \
  --invoker-package ${CLIENT_LIB_PACKAGE}
```

##### Steps to reproduce

Look at the output code! The `ApiClient` class doesn't extend anything! :-)

##### Related issues/PRs

I can't find anything quite like this.

##### Suggest a fix/enhancement

I'm about to put up a PR, I would suggest we make new modules in this project that are very tiny maven modules with nothing but minimal dependencies and the common interface for a given Java library binding.

daniilmatafonov:
I'm using swagger-codegen-maven-plugin with version 2.4.5 and rest-assured as testing library. I think common interface needed for this library too.

PR #6742, opened 2017-10-19
Common base class for feign client

charlescapps:
### PR checklist

CC technical committee for Java:

### Description of the PR
Issue https://github.com/swagger-api/swagger-codegen/issues/6741

This PR adds a new maven module called `swagger-codegen-common-feign`. The mustache template for the `pom.xml` for the `feign` client then adds this as a dependency. The generated `ApiClient` class then implements the globally defined interface `io.swagger.codegen.common.feign.SwaggerFeignApiClient`. Bear with me while I explain why I believe this to be the best solution to an important problem. It sounds like a lot to add a new maven module, but these will be very light modules.

The Java codegen client code (e.g. for Feign library, but this applies to any of them) currently has an `ApiClient` class that has a different FQCN for every swagger API. We strongly need a way to define shared code for having a common factory / helper to configure these API clients. 

For example, we have about 6 microservices that use Jersey + Swagger which communicate with each other. We would like to unify these services to all use the Feign API client generated by `swagger-codegen`. We need shared code to configure these clients in a uniform way, e.g.
* Call `setBasePath` to set the base URL to the correct microservice and environment (dev, qa, prod, ...)
* Call `getFeignBuilder` to configure the Retry behavior
* Call `setAccessToken` to set an access token obtained via some other means.

This change will enable more powerful usages of the Feign codegen, and in the future we can do something similar for other Java client libraries.


modules/swagger-codegen-common-feign/src/main/java/io/swagger/codegen/common/feign/SwaggerFeignApiClient.java
```diff
@@ -0,0 +1,34 @@
+package io.swagger.codegen.common.feign;
+
+import feign.Feign;
+import feign.RequestInterceptor;
+
+import java.util.Map;
+
+
+public interface SwaggerFeignApiClient {
```
charlescapps: This is the common interface that all `ApiClient` classes generated for feign-based Java clients will implement. It's a reasonable start with minimal dependencies (only feign) that defines everything we need for our use case.

modules/swagger-codegen/src/main/resources/Java/libraries/feign/ApiClient.mustache
```diff
@@ -32,7 +32,7 @@ import {{invokerPackage}}.auth.*;
 import {{invokerPackage}}.auth.OAuth.AccessTokenListener;
 
 {{>generatedAnnotation}}
-public class ApiClient {
+public class ApiClient implements io.swagger.codegen.common.feign.SwaggerFeignApiClient {
```
charlescapps: This will be on the classpath since I also updated the `pom.mustache`

modules/swagger-codegen/src/main/resources/Java/libraries/feign/pom.mustache
```diff
@@ -274,6 +279,8 @@
     <maven.compiler.source>${java.version}</maven.compiler.source>
     <maven.compiler.target>${java.version}</maven.compiler.target>
     <swagger-core-version>1.5.15</swagger-core-version>
+    <!-- TODO: fix this to a release version after 2.3.0 of swagger-codegen is released-->
+    <swagger-codegen-version>[2.2.9, 3.0.0)</swagger-codegen-version>
```
charlescapps: Need a version range here, since `2.3.0` isn't released yet, and there's no obvious way to get the current version of `swagger-codegen` into a mustache variable. If there's any easy way to dynamically replace this with the maven version, we could do that! 

The common interface should change exceedingly rarely, so I doubt this would be a problem. Would be nice if we could force this to be a release version instead possibly being a snapshot version.

modules/swagger-codegen-common-feign/pom.xml
```diff
@@ -0,0 +1,188 @@
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
```
charlescapps: This is the `pom.xml` for the new module that has the common interface that all feign-based Java codegen clients can use. There are probably some plugins here I can delete. I'll improve this and take any suggestions.

charlescapps:
Travis failed because it didn't receive output for 10 minutes. I don't think this has anything to do with my changes, tests pass locally.

wing328:
@charlescapps please ignore the Travis CI test at the moment (I've been working on this issue...)

The Java Petstore samples are tested by CircleCI, which reported the following issue:
```
Ran 26 tests containing 159 assertions.
0 failures, 0 errors.
[ERROR] Failed to execute goal on project swagger-petstore-feign: Could not resolve dependencies for project io.swagger:swagger-petstore-feign:jar:1.0.0: Failed to collect dependencies at io.swagger:swagger-codegen-common-feign:jar:[2.2.9,3.0.0): No versions available for io.swagger:swagger-codegen-common-feign:jar:[2.2.9,3.0.0) within specified range -> [Help 1]
[ERROR] 
[ERROR] 
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/DependencyResolutionException
[ERROR] 
```

Can you please take a look when you've time?

pom.xml
```diff
@@ -879,6 +879,7 @@
         <module>modules/swagger-codegen-cli</module>
         <module>modules/swagger-codegen-maven-plugin</module>
         <module>modules/swagger-generator</module>
+        <module>modules/swagger-codegen-common-feign</module>
```
cbornet: Maybe `modules/languages/swagger-codegen-common-feign` ? So we could have other modules for other JVM languages/frameworks.

cbornet:
Well, the advantage I see is that the code would be easily testable in an IDE as it would become standard JAVA code instead of mustache template. The drawback is indeed that it would make the language code even more scattered between generator, templates and library artifact.

charlescapps:
I think I missed your comments because I was working on my other pull request and busy at work, sorry!!! 

> To be honest, I don't think it make sense to have a new module to share stuff between generators. The fact that the module contains the name of a specific generator is even worst.

That's true, my concern was that if we have a single, common module, then we are forcing unexpected transitive dependencies to be present for every generator. For example, this requires some feign-specific dependencies, which we don't want to be required for a Spring client. 

However, I'm thinking now that a simpler approach would be to fix all of the generated API Clients to have a better design. They can simply allow passing in the ObjectMapper and other values, then this would be unnecessary.

For example, I made the new client for `google-api-client` simply allow passing in all possible configurable fields: https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/java/google-api-client/src/main/java/io/swagger/client/ApiClient.java#L45

I'm going to close this out, we're not personally consuming feign any more since I added the `google-api-client` support. If we end up needing this, I'd bias towards solving by just making them more configurable, which is a really easy change. 👍

Issue #6649, opened 2017-10-09
Angular 4+ TypeScript coge generator with HttpClient is broken if post / put / patch don't have body


pgrm:
##### Description

Looking at the lines https://github.com/swagger-api/swagger-codegen/blob/f3e19e0fdc1ae6416c37bb637521323d88939a9f/modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache#L241-L252 

If no body and form params are defined in put or post (which is possible), the generated code looks like this:

````
return this.httpClient.put<any>(`${this.basePath}/some-url/${encodeURIComponent(String(id))}`, {
            headers: headers,
            withCredentials: this.configuration.withCredentials
        });
````

which causes 

````
{  
  headers: headers,
  withCredentials: this.configuration.withCredentials
}
````

to be treated and sent as the body. This means a body, which isn't expected by the server, is sent and headers, which are expected, are not sent.

##### Swagger-codegen version

Master branch, latest commit I have is https://github.com/swagger-api/swagger-codegen/commit/bf74e33c52cdb125bcd30e7742d116352f8f9ded

##### Related issues/PRs

This is when the bug was introduced - https://github.com/swagger-api/swagger-codegen/pull/6295

##### Suggest a fix/enhancement

I'm not sure what would be the best way to fix it. For the template it would be easiest to just have always some default body param if necessary, which would be simply null, for the Java backend code it would be nicer to solve it simply on the template, but since you need a check if it's a PATCH, PUT or POST operation, I don't see how to do it in mustache. Maybe introducing a flag like `expectsBody`?

macjohnny:
@pgrm can you check whether this is still an issue with https://github.com/swagger-api/swagger-codegen/pull/6574?

pgrm:
@macjohnny I've just now used the latest master to generate my new client, and the issue is still there

macjohnny:
@pgrm I will try to create a fix and open a new PR

pgrm:
@macjohnny FYI, this is how I solved it for now - https://github.com/apaleo/swagger-codegen/commit/57cf9b6388c78bd821263ff9c02d989be4781caf

macjohnny:
@pgrm glad to see you found a fix. could you please open a new PR with your changes and generate the samples? I guess you should also be aware of https://github.com/swagger-api/swagger-codegen/pull/6496, which will hopefully be merged soon.

topce:
Hi @pgrm @macjohnny 
[...]

Issue #6723, opened 2017-10-18
[TypeScript-Angular2] Swagger Codegen Get Methods not Including Parameters


gabrielfuller:
##### Description

I am using Codegen Java file to create the Typescript-Angular 2 API for my C# web api.  The code gen tool creates the correct code for post and put, but get and delete do not include the parameters correctly.

Here is the generated put method:
```
public organizationsUpdatePut(command?: UpdateOrganizationCommand): Observable<UpdateOrganizationCommandResult> {

        let headers = this.defaultHeaders;

        return this.httpClient.put<any>(`${this.basePath}/Organizations/Update`, command, {
            headers: headers,
            withCredentials: this.configuration.withCredentials,
        });
    }
```
Here is the read method:
```
public organizationsReadGet(command?: ReadOrganizationCommand): Observable<ReadOrganizationCommandResult> {

        let headers = this.defaultHeaders;

        return this.httpClient.get<any>(`${this.basePath}/Organizations/Read`, {
            headers: headers,
            withCredentials: this.configuration.withCredentials,
        });
    }
```

As you can see, in the read method, the command parameter is never added to the HttpClient Get method so my call to the API always fails.  Any ideas how to fix it?  Am I doing something wrong?

##### Swagger-codegen version
3.0.0-20171009.075709-6

##### Command line used for generation
java -jar C:\SimOpSolutions\SOSBrowserClientApp\files\swagger-codegen-cli-3.0.0-20171009.075709-6.jar generate -i http://localhost:5000/swagger/v1/swagger.json -l typescript-angular -o C:\SimOpSolutions\SOSBrowserClientApp\sos-app\src\app\sos-api
pause

macjohnny:
@gabrielfuller can you post the relevant parts of your swagger.json?

wing328:
>3.0.0-20171009.075709-6

3.0.0 is still being worked on to support OAS (OpenAPI spec) 3.0

I would suggest you to use the 2.3.0 SNAPSHOT version for the time being.

defmonk0:
While this is still open, I also want to mention that `httpClient.delete` is not always properly configured either. When generating a delete call that takes body data, it adds it as a parameter to `httpClient.delete`, but the signature of that function only accepts url and options. 

[...]

PR #6754, opened 2017-10-19
Bugfix/6649 angular http client empty body

macjohnny:
### PR checklist

### Description of the PR
fixes https://github.com/swagger-api/swagger-codegen/issues/6649, https://github.com/swagger-api/swagger-codegen/issues/6723


macjohnny:
cc @roni-frantchi @wing328 @sebastianhaas @bedag-moo @krizic @taxpon @kenisteward@watiko @pgrm @topce

modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
```diff
@@ -141,6 +141,15 @@ public boolean isRestfulUpdate() {
     }
 
     /**
+     * Check if act as Restful update method
+     *
+     * @return true if act as Restful update method, false otherwise
+     */
```
pgrm: the comment is wrong :wink: this is what I came up with, but wasn't fast enough to create a PR

    /**
     * Check if body param is allowed for the request method
     *
     * @return true request method is PUT, PATCH or POST; false otherwise
     */

macjohnny: ah, now I see it. I updated the PR accordingly

pgrm:
@wing328 you can mark this one also with bug and typescript

macjohnny:
@wing328 @sebastianhaas the checks passed, how about merging?

defmonk0:
Hello! I was having some trouble with delete calls (it was passing 3 parameters, when the delete method can only take 2), so @macjohnny asked me to try this branch to see if it fixed the problem.

While it does fix the parameter mismatch, it doesn't quite fix the problem altogether. [This is the API I've been using](https://twitter.com/julieeeaustinnn/status/920721266864197632), with the contacts->delete method. As you can see, it wants a list of contact_ids in the body of the request. This PR still has an issue where it doesn't pass body data for delete requests at all (which *is* valid swagger, even if its uncommon).

As I understand it, Angular does not currently support body data in the options of [the delete method](https://angular.io/api/common/http/HttpClient#delete). As an alternative, [the request method](https://angular.io/api/common/http/HttpClient#request) does. If we used request instead, we could have full control over what gets passed around.

I don't have much experience with mustache templates, but I also think it's worth considering using the request method for all of the call types. It would be more consistent, so it might simplify the template in the long run. I'm open to input on this, if anyone has any particular thoughts on using one over the other.

sebastianhaas:
> I don't have much experience with mustache templates, but I also think it's worth considering using the request method for all of the call types. It would be more consistent, so it might simplify the template in the long run. I'm open to input on this, if anyone has any particular thoughts on using one over the other.

@defmonk0 either that, or you could try and file a request with the Angular team. Not sure if body data in a delete is inside specs though, so they might as well not agree on doing this. Would be interesting to do some research on that, maybe I'll find some time today.

macjohnny:
@defmonk0 @sebastianhaas the HTTP standard does not explicitly forbid body data with `DELETE` requests. however, it seems that the body data does not have any semantic meaning, which is why it is discouraged. See discussions e.g. here:
https://stackoverflow.com/questions/14323716/restful-alternatives-to-delete-request-body
https://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request

samples/client/petstore/typescript-angular-v4.3/npm/model/order.ts
```diff
@@ -33,9 +33,7 @@ export interface Order {
 
 }
 export namespace Order {
-    export enum StatusEnum {
-        Placed = <any> 'placed',
-        Approved = <any> 'approved',
-        Delivered = <any> 'delivered'
-    }
+    export type StatusEnum = 'placed' | 'approved' | 'delivered';
```
sebastianhaas: Do you know why you got these changes?

samples/client/petstore/typescript-angular-v4.3/npm/model/tag.ts
```diff
@@ -21,3 +21,5 @@ export interface Tag {
     name?: string;
 
 }
+
```
sebastianhaas: And these empty lines?

samples/client/petstore/typescript-angular-v4.3/npm/model/order.ts
macjohnny: I guess because the v4.3 was not re-generated after the changes introduced in https://github.com/swagger-api/swagger-codegen/pull/6233
this means that 4.3 somehow is not checked by CI. @wing328 can you fix that?

samples/client/petstore/typescript-angular-v4.3/npm/model/tag.ts
macjohnny: same as above

sebastianhaas: I see, and did someone delete the newline at the end? And does your output contain one or two blank lines in the end? It appears to be two, but maybe that's just Github

macjohnny: @sebastianhaas i cannot tell where the newlines come from, but they also appear in e.g. https://github.com/swagger-api/swagger-codegen/pull/6769, so it is just about re-generating the files

macjohnny:
@sebastianhaas so does this PR look good to you?

sebastianhaas:
Yes, I guess so, I'm always in favor of enforcing standards, however, for a project like a codegen it might be sensible to maximize compatibility. Maybe using an option flag though.

macjohnny:
@sebastianhaas alright, but I suggest to merge this PR and discuss the `DELETE` body issue in https://github.com/swagger-api/swagger-codegen/issues/6723 , and eventually create another PR.

pgrm:
@sebastianhaas / @macjohnny / defmonk0 FYI - it seems more of an incomplete typescript restriction, than actually intended use case that DELETE would allow to set a body when using the `request` method, meaning that the angular team might remove this possibility in the future.

My reasoning is based on the following method signature:

````
request<R>(req: HttpRequest<any>): Observable<HttpEvent<R>>;
````

and here are the constructors for HttpRequest:

    constructor(method: 'DELETE' | 'GET' | 'HEAD' | 'JSONP' | 'OPTIONS', url: string, init?: {
        headers?: HttpHeaders;
        reportProgress?: boolean;
        params?: HttpParams;
        responseType?: 'arraybuffer' | 'blob' | 'json' | 'text';
        withCredentials?: boolean;
    });
    constructor(method: 'POST' | 'PUT' | 'PATCH', url: string, body: T | null, init?: {
        headers?: HttpHeaders;
        reportProgress?: boolean;
        params?: HttpParams;
        responseType?: 'arraybuffer' | 'blob' | 'json' | 'text';
        withCredentials?: boolean;
    });
    constructor(method: string, url: string, body: T | null, init?: {
        headers?: HttpHeaders;
        reportProgress?: boolean;
        params?: HttpParams;
        responseType?: 'arraybuffer' | 'blob' | 'json' | 'text';
        withCredentials?: boolean;
    });

sebastianhaas:
@pgrm good catch, even more so, let's merge this PR and discuss options re compatibility with non-standard API definitions in another place.

wing328:
If no more feedback/question on this, I'll merge this tomorrow (Tue)

Issue #6734, opened 2017-10-18
[NancyFx] Support Async


craffael:
##### Description

Currently the NancyFx codegen generates synchronous endpoints. I suggest to add a new nancyfx specific `--additional-property async=true` that can be used to generate async endpoints.

##### Swagger-codegen version

2.3.0-SNAPSHOT

##### Suggest a fix/enhancement

if `async=true` I suggest to do the following modifications:

- add `async` keyword to every route and set second parameter in the square brackets to `true` to mark it as an asynchronous route.
- Make interface methods of Services and the abstract methods in in the abstract service(s) return Tasks + add `async` keyword where necessary.
- add the corresponding `using` statements.

See e.g. [this blogpost](http://blog.jonathanchannon.com/2013/08/24/async-route-handling-with-nancy/) for more details.

@mandrean @jimschubert What do you think?

wing328:
@craffael thanks for the suggestion.

It's worth sharing with you that most API clients (e.g C#, Java, Ruby, Python, etc) come with both sync (e.g. [AddPet](https://github.com/swagger-api/swagger-codegen/blob/c66a0aaa07695276fc3fa6a24bb42d2176d0f5cc/samples/client/petstore/csharp/SwaggerClientNetCoreProject/src/IO.Swagger/Api/PetApi.cs#L36)) and async (e.g. [AddPetAsync](https://github.com/swagger-api/swagger-codegen/blob/c66a0aaa07695276fc3fa6a24bb42d2176d0f5cc/samples/client/petstore/csharp/SwaggerClientNetCoreProject/src/IO.Swagger/Api/PetApi.cs#L216)) methods

For the server generator, I thought the controller classes are just the HTTP handler, which means it must complete its operation (e.g. save a record to database, start a background job to generate reports, etc) in order to return a HTTP response (but I've to say I'm not familiar with the NancyFx generator)

craffael:
@wing328 
1) Do these API clients have a common `optional-property` to use async terminology? If so I could reuse that property instead...

[...]

PR #6755, opened 2017-10-19
[NancyFx] Support Async

craffael:
### PR checklist

- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming langauge. @mandrean @jimschubert

### Description of the PR

This PR introduce a new additional property `async=true` (default: `false`) that generates a NancyFX server stub that uses C#'s asynchronous programming model.

Note that I've also added the new file `bin/nancyfx-petstore-server-async.sh` to generate a corresponding sample implementation under `samples/server/petstore/nancyfx-async/`. This sample implementation currently compiles with VisualStudio 2017.

This PR would resolve issue #6734.


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NancyFXServerCodegen.java
```diff
@@ -46,6 +46,7 @@
     private static final String IMMUTABLE_OPTION = "immutable";
     private static final String USE_BASE_PATH = "writeModulePath";
     private static final String PACKAGE_CONTEXT = "packageContext";
+    private static final String ASYNC = "async";
```
wing328: @craffael I would suggest using "supportsAsync" instead as the "supportsAsync" is used in C# client as well.

craffael: In the C# client the meaning of `supportsAsync` is slightly different:
- It cannot be configured by the user through `--additional-properties` but is instead determined from e.g. the version of the .NET framework.
- If `supportsAsync=true` additional Code is generated if it is supported by the underlying .NET framework version. This makes sense because the C# client is a client, i.e. implementing every endpoint twice (once synchronous, once asynchronous) gives the user more choice. They can decide whether they want to call the sync or async method.

The last point is especially important: Following @jimschubert 's argument (https://github.com/swagger-api/swagger-codegen/issues/6734#issuecomment-337782969) I suggest a different approach for nancyfx because it generates not a client, but a server stub: The user should choose whether he wants to generate an async or a sync server stub (because one cannot implement an endpoint async *and* sync). Therefore its not about whether the environment (.NET Framework) *supports* async but whether the user *wants* async. Because of this I suggest to keep the name of this `--additional-property` as it is.

jimschubert: @craffael what it sounds like @wing328 is getting at is that 'async' on its own may not be clear from CLI about what's actually being made async.

I agree supportsAsync would be confusing because this would give different semantics between client and server generators.

Could we do something like 'asyncServer' or 'asyncEndpoints'? This would be more maintainable in the long run. For example, if we were  (hypothetically) to add EF Code First Data layer and some data providers don't support async DB calls, we may want asynchronous endpoints, with synchronous DB calls. In such a case, just 'async' wouldn't work so we would have to introduce a breaking change to rename it.

craffael: @jimschubert Okay I'm fine with `asyncServer` as well as `asyncEndpoints` but I prefer the `asyncServer` slightly. I will make the corresponding change in a moment.

craffael:
I've now renamed the optional-property and the variable names from `async` => `asyncServer` and fixed a small bug. There is however still another bug present where I'm not 100% sure how to solve it:
Currently the additional property `asyncServer` is passed through to the underlying mustache template parser. In the `api.mustache` template we have expressions such as `{{#asyncServer}}await{{/asyncServer}}` which should only be rendered when `asyncServer=true`. However at the moment they are also rendered for `asyncServer=false` because the value of the mustache objectModel contains under the key `asyncService` the *string* `"false"`. A simple fix would be to add another configuration property to the mustache objectModel (e.g. in `NancyFxServerCodegen.java`) that is not a string but of type `Boolean`. Do you have other suggestions?

modules/swagger-codegen/src/main/resources/nancyfx/api.mustache
```diff
@@ -6,6 +6,7 @@ using Sharpility.Base;
 using {{packageName}}.{{packageContext}}.Models;
 using {{packageName}}.{{packageContext}}.Utils;
 using NodaTime;
+{{#asyncServer}}using System.Threading.Tasks;{{/asyncServer}}
```
wing328: You may to avoid a new line by using the following instead:
```
{{#asyncServer}}
using System.Threading.Tasks;
{{/asyncServer}}
```

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NancyFXServerCodegen.java
```diff
@@ -128,6 +133,10 @@ public void processOpts() {
             setPackageGuid((String) additionalProperties.get(OPTIONAL_PROJECT_GUID));
         }
 
+        if(additionalProperties.containsKey(ASYNC_SERVER)) {
```
wing328: Please add a space after "if" to conform to Java style guide.

craffael: done

modules/swagger-codegen/src/main/resources/nancyfx/api.mustache
craffael: done

jimschubert:
@craffael the true/false should work as expected. I'd suggest a clean/rebuild and if the problem continues pass the debug flags to evaluate further. Let me know if you need assistance.

wing328:
@craffael I could repeat the issue you mentioned and have a fix for it. I'll apply the fix after merging it into master.

wing328:
Fixed via eba4aff

jimschubert:
@wing328 what's the difference between that valueOf and parseBoolean? Just curious, as I thought "false" would be false for both. Sorry if it's a silly question, I always use valueOf and have little experience with the other.

wing328:
The issue is with `(String)additionalProperties.get(ASYNC_SERVER)`, which causes exception as it fails to cast boolean (the value in the JSON file) into String.

I used `valueOf` to make it consistent with other generators. (parseBoolean should give us the same result)

Ref: http://www.luyue.org/boolean-parseboolean-vs-boolean-valueof-vs-boolean-getboolean/

jimschubert:
Ah, I completely missed that. Thanks for the clarification.

craffael:
Ah interesting, I also didn't know about this difference. However the issue is unfortunately not yet resolved. I just made a pull from the current master and run the code with `asyncServer=false` and it generated async code.

craffael:
I think the problem lies in the way how the `--additional-properties` are mapped into the data object that is passed to mustache. This is however outside of the NancyFX codegen:  The additional properties are added to the "template data object" in `DefaultGenerator.java:421`

As far as I can see it there are two options to resolve the issue:
1) Change the way how cli options of type `BooleanProperty.TYPE` are parsed, i.e. such that they are added as booleans into the "template data object" and not as strings
2) Override the `asyncServer` property in the NancyFX codegen so that it has type Boolean and not string.

Although Option 1 would be the nicest, it may break many other templates so that I guess only option 2 is an realistic? Or is there even another possible solution to this?

wing328:
@craffael thanks for the details. I'll take another look tomorrow.

For the time being, users should simply omit `asyncServer=false` to use the async feature.

wing328:
@craffael I just did another test but couldn't repeat the issue.

Did you do a "mvn clean" before rebuilding the JAR after pulling the latest from the master?

craffael:
@wing328 Thanks for looking into it again. Interestingly I checked again at my private computer at home and I still have the issue. Maybe I'm too stupid so if you don't want to bother about this anymore that's also fine for me ;)
But if you're curious: I did exactly the following commands to make a completly fresh build  of the repo and generate a server stub with nancyFx:

    git clone https://github.com/swagger-api/swagger-codegen.git
    cd swagger-codegen/
    mvn clean
    mvn package
    cd modules/swagger-codegen-cli/target/
    java -jar swagger-codegen-cli.jar generate -l nancyfx -i http://petstore.swagger.io/v2/swagger.json -o ~/temp/petstore/ --additional-properties asyncServer=false
    vim ~/temp/petstore/src/IO.Swagger/Modules/PetModule.cs

craffael:
I should also mention that I think the fix in https://github.com/swagger-api/swagger-codegen/commit/eba4affbb412f8be8d52ea8da49ddaf0010481be is correct, but it has probably no effect so far. The method `NancyFXServerCodegen#setAsyncServer()` simply sets the private member variable `asyncServer` but this variable is not used anywhere else (unless there is some reflection magic happening). Originally, I wanted to override the `asyncServer` entry in the "template data object", that is passed to mustache, here. But then I was not sure whether this is the right approach and so I have not yet done that...

wing328:
@craffael my bad. I didn't test it correctly.

I committed 8508434 to fix it. Please pull the latest master to give it another try (my local test result is good).

craffael:
Ah perfect, this fixes it. Thanks or the clarification!
I guess you still have to merge the branch `fix_nancy_option` into `master` (or maybe that's already in progress) ;)

craffael:
(maybe set `asyncServer=true` in `nancyfx-petstore-server-async.sh` before the commit)

wing328:
@craffael thanks for the confirmation 👍 

I thought it's already set in master but I'll double check again.

PR #6758, opened 2017-10-19
revert changes to fakes yaml

jonschoning:
* the yaml currently includes invalid definitions that do not exist: `#/definitions/xmlObject`. 
introduced in 1d89ab6e0 

Revert to previous known good yaml state.

* update template creators


wing328:
@jonschoning thanks for reverting the incorrect change. I'll add the fake endpoint to test XML payload later.

PR #6769, opened 2017-10-20
[TypeScript-Angular] Path URI Encoding Update

### LINKED ISSUE #6762: [Typescript-Angular] Service generation incorrectly attempting to use parameter.
##### Description

Upon generating the code for my angular project, I copied it into my app directory. Immediately, the compiler starts spitting out errors about being unable to find variables. This is an example of the exact error (there are a lot of them):

`ERROR in /home/def_monk/code/eve-calendar/src/app/esi/api/wars.service.ts (163,92): Cannot find name 'war_id'.`

It's pretty easy to identify what it's talking about, since the uses of snakecase are very rare in the files. There is a mismatch in the parameter name and what is used in the return statement. The function in which that particular error occurs is here:

```typescript
/**
 * List kills for a war
 * Return a list of kills related to a war  --- Alternate route: &#x60;/v1/wars/{war_id}/killmails/&#x60;  Alternate route: &#x60;/legacy/wars/{war_id}/killmails/&#x60;  Alternate route: &#x60;/dev/wars/{war_id}/killmails/&#x60;  --- This route is cached for up to 3600 seconds
 * @param warId A valid war ID
 * @param datasource The server name you would like data from
 * @param page Which page of results to return
 * @param userAgent Client identifier, takes precedence over headers
 * @param xUserAgent Client identifier, takes precedence over User-Agent
 */
public getWarsWarIdKillmails(warId: number, datasource?: string, page?: number, userAgent?: string, xUserAgent?: string): Observable<Array<GetWarsWarIdKillmails200Ok>> {
    if (warId === null || warId === undefined) {
        throw new Error('Required parameter warId was null or undefined when calling getWarsWarIdKillmails.');
    }

    let queryParameters = new HttpParams({encoder: new CustomHttpUrlEncodingCodec()});
    if (datasource !== undefined) {
        queryParameters = queryParameters.set('datasource', <any>datasource);
    }
    if (page !== undefined) {
        queryParameters = queryParameters.set('page', <any>page);
    }
    if (userAgent !== undefined) {
        queryParameters = queryParameters.set('user_agent', <any>userAgent);
    }

    let headers = this.defaultHeaders;
    if (xUserAgent !== undefined && xUserAgent !== null) {
        headers = headers.set('X-User-Agent', String(xUserAgent));
    }

    return this.httpClient.get<any>(`${this.basePath}/wars/${encodeURIComponent(String(war_id))}/killmails/`,
    {
        params: queryParameters,
        headers: headers,
        withCredentials: this.configuration.withCredentials,
    });
}
```

##### Swagger-codegen version

Master at time of pull:  [f0b7dae](https://github.com/swagger-api/swagger-codegen/tree/f0b7daeec691ea0fb0d48139cc4610b0a49d285b) (which would mean actual version is 2.3.0, if I understand correctly).

https://esi.tech.ccp.is/latest/swagger.json?datasource=tranquility

```
	-i https://esi.tech.ccp.is/latest/swagger.json?datasource=tranquility \
	-l typescript-angular \
	-o ./clients/esi/
```

##### Steps to reproduce

Generate code, and try to move it into the angular application. The typescript will not be able to compile. I'm used [Angular CLI](https://github.com/angular/angular-cli) to do the project generation and the compile/serve.

##### Related issues/PRs

#4264 seems related, since it looks to be when the change to camelcase was made to match the style guide.

Looking around to try and figure out what the problem was, I eventually made my way to [this line](https://github.com/swagger-api/swagger-codegen/blob/f0b7daeec691ea0fb0d48139cc4610b0a49d285b/modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache#L266) of mustache template for the services. That said, this is my first time using both codegen and angular2+, and I know nothing about using mustache templates, so I have no idea if I'm actually correct in guessing it has something to do with that line of code. Lol. But simply using the camelcase version of the parameter, instead of snakecase, seems like it would fix the error.

defmonk0:
### PR checklist

### Description of the PR

This update replaces the code for updating paths for URI encoding. Previously, this code was not converting from snake_case to camelCase, causing the variables embedded in the path to not match the parameter names they were passed to the function as.

The new method involves finding matches of items in curly-braces and iterating through them. Each match is then searched through for underscores. If one is found, it is removed and the following letter is capitalized.

The more complicated replacement is necessary to guarantee we only change the format of the variables inside the template braces. Otherwise, it might replace parts of the url/path itself, rather than just the variable names.

This should fix #6762.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngularClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngularClientCodegen.java
@@ -1,6 +1,7 @@
 package io.swagger.codegen.languages;
 
 import java.io.File;
+import java.lang.StringBuffer;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Date;
@@ -8,6 +9,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.Set;
 
 import io.swagger.codegen.CliOption;
@@ -82,7 +85,7 @@ public class TypeScriptAngularClientCodegen extends AbstractTypeScriptClientCode
         supportingFiles.add(new SupportingFile("apis.mustache", apiPackage().replace('.', File.separatorChar), "api.ts"));
         supportingFiles.add(new SupportingFile("index.mustache", getIndexDirectory(), "index.ts"));
         supportingFiles.add(new SupportingFile("api.module.mustache", getIndexDirectory(), "api.module.ts"));
-        supportingFiles.add(new SupportingFile("rxjs-operators.mustache", getIndexDirectory(), "rxjs-operators.ts"));        
+        supportingFiles.add(new SupportingFile("rxjs-operators.mustache", getIndexDirectory(), "rxjs-operators.ts"));
         supportingFiles.add(new SupportingFile("configuration.mustache", getIndexDirectory(), "configuration.ts"));
         supportingFiles.add(new SupportingFile("variables.mustache", getIndexDirectory(), "variables.ts"));
         supportingFiles.add(new SupportingFile("encoder.mustache", getIndexDirectory(), "encoder.ts"));
@@ -149,7 +152,7 @@ public class TypeScriptAngularClientCodegen extends AbstractTypeScriptClientCode
     public boolean isDataTypeFile(final String dataType) {
         return dataType != null && dataType.equals("Blob");
     }
-    
+
     @Override
     public String getTypeDeclaration(Property p) {
         Property inner;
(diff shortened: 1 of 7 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngularClientCodegen.java
```diff
@@ -247,8 +250,55 @@ public void postProcessParameter(CodegenParameter parameter) {
...
+            Matcher pathMatcher = pathPattern.matcher(op.path);
+
+            // Prep a string buffer where we're going to set up our new version of the string.
+            StringBuffer pathBuffer = new StringBuffer();
+
+            // Iterate through the matches for the regex.
+            // This should be anything that was contained inside curly-braces.
+            while(pathMatcher.find()) {
+                // Get a substring of our match.
+                String templateSection = op.path.substring(pathMatcher.start(), pathMatcher.end());
```
macjohnny: I would choose a simpler implementation like e.g. https://stackoverflow.com/a/34229320

defmonk0: Hmm. Looking at it now, I agree. It's been a long while since I've coded in Java, so after seeing the original code used regex, I just did the same in case it needed to be more complicated later. That said, it's definitely not the most efficient. 

I'll mess with this is a bit, I should be able to do this with much simpler string manipulation, crawling through 1 character at a time. I should only need one iteration through the string.

defmonk0:
Rebased on master, and re-pushed the branch with the extra commit.

I simplified the implementation. Much cleaner, and more efficient too. Since using split would have still required avoiding underscores outside of the template braces, I figured this iteration method was simpler in the long run. No regex or secondary iteration required.

defmonk0:
I rebased on master again, so that the tests would pass (since the previous commit this was based on was not). @macjohnny thoughts?

wing328:
@defmonk0 thanks for the PR with detailed comments in the code, which looks good to me.

PR #6757, opened 2017-10-19
Bugfix/6727 import http client module

macjohnny:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming langauge.

### Description of the PR

fixes https://github.com/swagger-api/swagger-codegen/issues/6727 and replaces PR https://github.com/swagger-api/swagger-codegen/pull/6729

PR #6788, opened 2017-10-23
Fixed spacing in model.mustache

kenisteward:
### PR checklist

### Description of the PR

@sebastianhaas @wing328 

An extra line was being added in models.  See any an example in the PR here #6757 

Also amended typscript-angular2-all.bat and .sh to include angular 4 and angular 4.3 generation.


kenisteward:
Note @wing328 I see we have run-all-petstore.cmd that's run by appveyor but they are multiple *-all.bat files.  This means tests are taking longer than needed due to double generating templates.   Should we just include *-all.bat and update the appveyor script or should we in the appveyor script just exclude once that have -all in them?

modules/swagger-codegen/src/main/resources/typescript-angular/modelGeneric.mustache
```diff
@@ -10,7 +10,6 @@ export interface {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{
      */
     {{/description}}
     {{name}}{{^required}}?{{/required}}: {{#isEnum}}{{{datatypeWithEnum}}}{{/isEnum}}{{^isEnum}}{{{datatype}}}{{/isEnum}};
-
```
sebastianhaas: I like this

samples/client/petstore/typescript-angular-v2/default/model/ApiResponse.ts
```diff
@@ -17,9 +17,8 @@
  */
 export interface ApiResponse {
     code?: number;
-
     type?: string;
-
     message?: string;
-
 }
+
```
sebastianhaas: But do you know where these two lines come from? Why two?

kenisteward: dangit.... I think I accidentally added another one after the simple eol.  I can go back and change that lol

kenisteward: turns out it wasn't me.  It was there incorrectly from the enum generator

wing328:
> Should we just include *-all.bat and update the appveyor script or should we in the appveyor script just exclude once that have -all in them?

@kenisteward maybe better to exclude *-all.bat

wing328:
@kenisteward the PR looks good

cc @TiFu @taxpon @sebastianhaas @Vrolijkx

samples/client/petstore/typescript-angular-v2/default/model/ApiResponse.ts
sebastianhaas: @kenisteward ah I see. I'll create a PR to get rid of them eventually, unless anybody sees any value in having two empty lines at the end.

kenisteward:
So in the model.mustache we include modelgeneric and enumgeneric. Whenever
there are nested mustaches it seems it takes all the end of file lines from
all of them.  Therefore if we have every mustache with end of file lines we
end up with more than we want in the output.  I've fixed the output by
simply adding one end line on the outer mustache and removing ones from the
inserted.

PR #6796, opened 2017-10-24
Bugfix/6391 retrofit wrong order of params

### LINKED ISSUE #6391: [Android] Retrofit CodeGen : @Path must be before @Query
##### Description

With SwaggerCodeGen 2.2.2 (.jar) we have an issue with retrofit,
a query has @Path and @Query params, and when it's generated, some @Query are before the @Path one. But Retrofit crash in this case

sample : 
--------
```
@GET("/users/{name}/images/")
Call<User> getUserImage(@Path("name") String name, @Query("page") int page)
```
will works

but 

```
@GET("/users/{name}/images/")
Call<User> getUserImage(@Query("page") int page, @Path("name") String name)
```
will crash

##### Swagger-codegen version

SwaggerCodeGen 2.2.2 (.jar) 

generate -i MY_FILE -o OUTPUT_DIR -l java --library=retrofit2 -c specs/myConfigfile.json

giuliopulina:
### PR checklist

Fixes #6391: retrofit2 is crashing when @Query params are defined before @Path param in the api. 
With this change, params are correctly reordered (@Path param are always defined before @Query param).

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
@@ -1,10 +1,13 @@
 package io.swagger.codegen.languages;
 
+import static java.util.Collections.sort;
+
 import com.google.common.collect.LinkedListMultimap;
 import io.swagger.codegen.*;
 import io.swagger.codegen.languages.features.BeanValidationFeatures;
 import io.swagger.codegen.languages.features.GzipFeatures;
 import io.swagger.codegen.languages.features.PerformBeanValidationFeatures;
+
 import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
@@ -48,6 +51,7 @@ public class JavaClientCodegen extends AbstractJavaCodegen
     protected boolean performBeanValidation = false;
     protected boolean useGzipFeature = false;
     protected boolean useRuntimeException = false;
+
 
     public JavaClientCodegen() {
         super();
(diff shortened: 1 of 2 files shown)


samples/client/petstore/java/retrofit/src/test/java/io/swagger/client/api/PetApiTest.java
```diff
@@ -76,33 +76,6 @@ public void testFindPetsByStatus() throws Exception {
     }
 
     @Test
-    public void testFindPetsByTags() throws Exception {
```
wing328: @giuliopulina May I know why the test cases for `findPetsByTags` are removed? Is it throwing errors when run locally?

giuliopulina: yes, I confirm that mvn integration-test was throwing error.

In fact, if I try to call the service: http://petstore.swagger.io/v2/pet/findByTags?tags=friendly

I receive this error:

      <apiResponse>
          <message>something bad happened</message>
          <type>unknown</type>
      </apiResponse>

And also, if I look in swagger UI, I see that this endpoint is deprecated:

image

wing328: The CIs (Travis, CircleCI) are running a local copy of the Petstore server as the public petstore server contains invalid pet data from time to time (since anyone can update the record).

Please add those tests back and the CI tests should pass.

giuliopulina: done 👍

wing328:
@giuliopulina I did some tests locally and the result is good. Thanks for the PR.

florent37:
thanks a lot for your work :)

mmallis87:
I have a related question regarding operations path params specifically: does anyone think that we should enforce path params order based on how they appear in on the URL path instead of relying on the order in the `parameters` array?

For instance when the URL path is "/accounts/{accountId}/users/{userId}" and
```
"parameters": [
          {
            "name": "userId",
            "in": "path",
            "description": "The external user number (int) or user ID Guid.",
            "required": true,
            "type": "string"
          },
          {
            "name": "accountId",
            "in": "path",
            "description": "The external account number (int) or account ID Guid.",
            "required": true,
            "type": "string"
          }
]
```
the generated method will have a signature that looks like `public void myMethod(String userId, String accountId);`. I think it makes more sense to have `public void myMethod(String accountId, String userId);`. Curious to see what you guys think!

giuliopulina:
> I have a related question regarding operations path params specifically: does anyone think that we should enforce path params order based on how they appear in on the URL path instead of relying on the order in the `parameters` array?
> 
> For instance when the URL path is "/accounts/{accountId}/users/{userId}" and
> 
> ```
> "parameters": [
>           {
>             "name": "userId",
>             "in": "path",
>             "description": "The external user number (int) or user ID Guid.",
>             "required": true,
>             "type": "string"
>           },
>           {
>             "name": "accountId",
>             "in": "path",
>             "description": "The external account number (int) or account ID Guid.",
>             "required": true,
>             "type": "string"
>           }
> ]
> ```
> 
> the generated method will have a signature that looks like `public void myMethod(String userId, String accountId);`. I think it makes more sense to have `public void myMethod(String accountId, String userId);`. Curious to see what you guys think!

Hi @mmallis87 for what my opinion may be worth, I think it can make sense: however, I think this behavior should be aligned between all the different generators, not just for the Retrofit one, so it should be an higher level design decision.

mmallis87:
@giuliopulina I agree that this type of change should be aligned across all generators since it's not specific to Java or Retrofit.

PR #6812, opened 2017-10-25
[Ruby] Fix "build_from_hash" function for Ruby enums

run-o:
### PR checklist

### Description of the PR

Modify build_from_hash to compare the given value against the enum's value rather than the enum's constant name.
Unless I'm missing something, it seems to me that this should be the desired behavior for build_from_hash.

For instance, consider the following Ruby enum class:

```
class Enum
  ENUM1 = "enum 1"
  ENUM2 = "enum 2"
  ENUM3 = "enum 3"
  def build_from_hash(value)
    ...
  end
end
```

If we run the following:
```
Enum.new.build_from_hash("enum 2")
```
It'll return "enum 2".
Before this fix, it would return an error since we were comparing "enum 2" against the constant names "ENUM1", "ENUM2" and "ENUM3".

Hope this makes sense!

### DIFF
--- a/modules/swagger-codegen/src/main/resources/ruby/partial_model_enum_class.mustache
+++ b/modules/swagger-codegen/src/main/resources/ruby/partial_model_enum_class.mustache
@@ -6,8 +6,8 @@
     # @param [String] The enum value in the form of the string
     # @return [String] The enum value
     def build_from_hash(value)
-      consantValues = {{classname}}.constants.select{|c| c.to_s == value}
-      raise "Invalid ENUM value #{value} for class #{{{classname}}}" if consantValues.empty?
+      constantValues = {{classname}}.constants.select{|c| {{classname}}::const_get(c) == value}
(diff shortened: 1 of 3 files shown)


run-o:
@wing328 Here's my PR for the Ruby enum fix. Let me know if it makes sense. Thanks!

wing328:
@run-o thanks for the PR. Can you add a test case in a new file named "enum_spec.rb" under the [spec folder](https://github.com/swagger-api/swagger-codegen/tree/master/samples/client/petstore/ruby/spec) for the following enum class:
```
 module Petstore
   class EnumClass
     
     ABC = "_abc".freeze
     EFG = "-efg".freeze
     XYZ = "(xyz)".freeze
```

wing328:
cc @NicoEigenmannCW 

Linked to #6131

wing328:
@run-o please add the test case later when you've time. Thanks.

The fix looks good to me so merging it into master.

PR #6820, opened 2017-10-26
[JAVA] Use application/octet-stream as the default content type

### LINKED ISSUE #6819: [JAVA] application/json is not the best choice for default content-type
##### Description

The default java template (and the "library" java variants) apply a default content-type of `application/json` if it is not specified. :

https://github.com/swagger-api/swagger-codegen/blob/4f3c86aa145e040024ee4d539947cf3b69c60d21/modules/swagger-codegen/src/main/resources/Java/ApiClient.mustache#L526

I don't think it is correct to assume this content-type. This may cause parsing errors on the server side if the content is in fact not valid JSON. A safer, more generic "default" content-type would be `application/octet-stream`, as it is less likely to cause problems with parsing.

There is also some indication in the [HTTP spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec7.html#sec7.2.1) that `application/octet-stream` would be a better default:
> If the media type remains unknown, the recipient SHOULD treat it as type "application/octet-stream". 

So the generated swagger client would be explicitly setting the content-type that a correct server would also assume.

##### Swagger-codegen version

2.3.0 @ 6814530d690db7b88e91fc1aab50f10472f707bd

```yaml
swagger: '2.0'
basePath: /
schemes:
  - http
paths:
  '/toggle':
    post:
      summary: Toggle the switch
      description: If switch is off, toggle to on, and vice versa.
      responses:
        200:
          description: Success
```

java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -l java -i example.yaml

##### Steps to reproduce

Use the generated client to make a POST request, and observe that a header is added: "Content-Type: application/json", even though this is not specified in the YAML, and the request body does not contain JSON data.

##### Related issues/PRs

https://github.com/swagger-api/swagger-codegen/issues/6647

Use `application/octet-stream` as the "default" content-type: #6820

bmordue:
### PR checklist

### Description of the PR

If no content-type header is specified, use `application/octet-stream` as a sensible "default".
Fixes #6819 

cc @bbdouglas @JFCote @sreeshas @jfiala @lukoyanov @cbornet

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/ApiClient.mustache
@@ -523,7 +523,7 @@ public class ApiClient {
    */
   public String selectHeaderContentType(String[] contentTypes) {
     if (contentTypes.length == 0 || contentTypes[0].equals("*/*")) {
-      return "application/json";
+      return "application/octet-stream";
     }
     for (String contentType : contentTypes) {
       if (isJsonMime(contentType)) {
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/feign/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/feign/ApiClient.mustache
@@ -216,7 +216,7 @@ public class ApiClient {
    *   JSON will be used.
    */
   public String selectHeaderContentType(String[] contentTypes) {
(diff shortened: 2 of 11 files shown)


bmordue:
Some problems with this: serializing as application/octet-stream at the client side.

PR #6828, opened 2017-10-27
Add TypeScript default value mappers

### LINKED ISSUE #6765: [typescript-angular] template variable `defaultValue` not set correctly
##### Description

When overriding the mustache files for `typescript-angular`, I found out it does not set the template variable `defaultValue` for strings. I wanted to file a PR for this, but I noticed the `toDefaultValue` in `DefaultCodegen.java` is also ignoring `StringProperty`s.

##### Swagger-codegen version

2.2.3, 2.3.0-latest

N/A

N/A

##### Steps to reproduce

* create a swagger file with a schema with an property of type 'string' and a 'default'
* update `modules/swagger-codegen/src/main/resources/typescript-angular/modelGeneric.mustache` to contain `defaultValue`
* run the codegen
* the generated .ts file has parses `defaultValue` of `'null'`

##### Related issues/PRs

N/A

Either patching `modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java` or `modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractTypeScriptClientCodegen.java` by adding

```
if (p instanceof StringProperty) {
    StringProperty sp = (StringProperty) p;
    if (sp.getDefault() != null) {
        return sp.getDefault();
    }
    return "null";
}
```

to `public String toDefaultValue(Property p)`.

I am willing to do so, but I am not sure it can be added to `DefaultCodegen.java` safely.

trenneman:
Fixes #6765.

### PR checklist

### Description of the PR

Added a method to set the default values for TypeScript conversions.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractTypeScriptClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractTypeScriptClientCodegen.java
@@ -18,11 +18,20 @@ import io.swagger.codegen.CodegenProperty;
 import io.swagger.codegen.CodegenType;
 import io.swagger.codegen.DefaultCodegen;
 import io.swagger.models.properties.ArrayProperty;
+import io.swagger.models.properties.BooleanProperty;
+import io.swagger.models.properties.DateProperty;
+import io.swagger.models.properties.DateTimeProperty;
+import io.swagger.models.properties.DoubleProperty;
 import io.swagger.models.properties.FileProperty;
+import io.swagger.models.properties.FloatProperty;
+import io.swagger.models.properties.IntegerProperty;
+import io.swagger.models.properties.LongProperty;
 import io.swagger.models.properties.MapProperty;
 import io.swagger.models.properties.Property;
+import io.swagger.models.properties.StringProperty;
 
 public abstract class AbstractTypeScriptClientCodegen extends DefaultCodegen implements CodegenConfig {
+    private static final String UNDEFINED_VALUE = "undefined";
 
     protected String modelPropertyNaming= "camelCase";
     protected Boolean supportsES6 = true;
@@ -219,6 +228,49 @@ public abstract class AbstractTypeScriptClientCodegen extends DefaultCodegen imp
             return "any";
         }
         return super.getTypeDeclaration(p);
+    }
+
+    @Override
+    public String toDefaultValue(Property p) {
+        if (p instanceof StringProperty) {
+            StringProperty sp = (StringProperty) p;
+            if (sp.getDefault() != null) {
+                return "\"" + sp.getDefault() + "\"";
+            }
+            return UNDEFINED_VALUE;
+        } else if (p instanceof BooleanProperty) {
+            return UNDEFINED_VALUE;
+        } else if (p instanceof DateProperty) {
+            return UNDEFINED_VALUE;
+        } else if (p instanceof DateTimeProperty) {
+            return UNDEFINED_VALUE;
+        } else if (p instanceof DoubleProperty) {
+            DoubleProperty dp = (DoubleProperty) p;
+            if (dp.getDefault() != null) {
+                return dp.getDefault().toString();
+            }
+            return UNDEFINED_VALUE;
+        } else if (p instanceof FloatProperty) {
+            FloatProperty fp = (FloatProperty) p;
+            if (fp.getDefault() != null) {
+                return fp.getDefault().toString();
+            }
+            return UNDEFINED_VALUE;
+        } else if (p instanceof IntegerProperty) {
+            IntegerProperty ip = (IntegerProperty) p;
+            if (ip.getDefault() != null) {
+                return ip.getDefault().toString();
+            }
+            return UNDEFINED_VALUE;
+        } else if (p instanceof LongProperty) {
+            LongProperty lp = (LongProperty) p;
+            if (lp.getDefault() != null) {
+                return lp.getDefault().toString();
+            }
+            return UNDEFINED_VALUE;
+        } else {
+            return UNDEFINED_VALUE;
+        }
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


wing328:
@trenneman thanks for the PR.

cc @TiFu  @taxpon  @sebastianhaas @kenisteward @Vrolijkx

kenisteward:
@trenneman Thanks for the PR.  The changes look harmless but can I ask why it was made?  I don't think we use these anywhere (included in the actual typescript client types).  code wide i only see this function being used in 
AbstractJavaCodegen.java
ApexClientCodegen.java
DartClientCodegen.java

which none of these are related to typescript.

wing328:
The CI failure (caused by Ruby client) has been fixed in the latest master already.

trenneman:
@kenisteward For the default mustache templates nothing has changed indeed. However, when providing your own mustache files, it is not possible to use the default values as they are set incorrectly.

I've rebased to master.

kenisteward:
@trenneman Thanks for the clarification.

trenneman:
@kenisteward No problem. It would have made sense to put it in the description.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractTypeScriptClientCodegen.java
```diff
@@ -222,6 +231,49 @@ public String getTypeDeclaration(Property p) {
     }
 
     @Override
+    public String toDefaultValue(Property p) {
+        if (p instanceof StringProperty) {
```
TiFu: I'm wondering if the code in this method could be cleaner/shorter.

This method treats DoubleProperty, FloatProperty, IntegerProperty (and others) exactly the same. 
I think a better solution might be to do:

```
if (p instanceof Integerproperty 
    || p instanceof FloatProperty
    || p instanceof ....) {
    if (p.getDefault() != null) return p.getDefault().toString()
} else if (p instanceof StringProperty) {
 /* code here */
}
// default if the property doesn't have a proper default (e.g. DateProperty)
return UNDEFINED_VALUE 
```

I'm not set on this, but the current code has a lot of repetitions of the same statements.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractTypeScriptClientCodegen.java
```diff
@@ -222,6 +231,49 @@ public String getTypeDeclaration(Property p) {
...
+            StringProperty sp = (StringProperty) p;
+            if (sp.getDefault() != null) {
+                return "\"" + sp.getDefault() + "\"";
+            }
+            return UNDEFINED_VALUE;
+        } else if (p instanceof BooleanProperty) {
+            return UNDEFINED_VALUE;
```
TiFu: Wouldn't ```false``` be a better default for a boolean property? (or whatever the default in ```BooleanProperty``` is?)

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractTypeScriptClientCodegen.java
trenneman: I would prefer that as well, but `getDefault()` is not an abstract method of `Property`. Actually, this is the way it is handled in the super class `DefaultCodegen` as well.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractTypeScriptClientCodegen.java
trenneman: In TypeScript, the default is `undefined`, not `false`.

trenneman:
@wing328 The build still fails, can it be re-triggered? I don't think the issue is on my end.

wing328:
@trenneman restarted. MacOS image seems to fail a lot more often so likely we'll switch back to Linux image.

trenneman:
@wing328 OK, thanks!

kenjones-cisco:
After pulling in the latest changes to master, and running `mvn package`, I'm now seeing the following errors:

```
complexPropertyTest(io.swagger.codegen.typescript.typescriptangular2.TypeScriptAngular2ModelTest)  Time elapsed: 0.017 sec  <<< FAILURE!
java.lang.AssertionError: expected [null] but found [undefined]
        at io.swagger.codegen.typescript.typescriptangular2.TypeScriptAngular2ModelTest.complexPropertyTest(TypeScriptAngular2ModelTest.java:124)

listPropertyTest(io.swagger.codegen.typescript.typescriptangular2.TypeScriptAngular2ModelTest)  Time elapsed: 0.003 sec  <<< FAILURE!
java.lang.AssertionError: expected [null] but found [undefined]
        at io.swagger.codegen.typescript.typescriptangular2.TypeScriptAngular2ModelTest.listPropertyTest(TypeScriptAngular2ModelTest.java:91)

simpleModelTest(io.swagger.codegen.typescript.typescriptangular2.TypeScriptAngular2ModelTest)  Time elapsed: 0.003 sec  <<< FAILURE!
java.lang.AssertionError: expected [null] but found [undefined]
        at io.swagger.codegen.typescript.typescriptangular2.TypeScriptAngular2ModelTest.simpleModelTest(TypeScriptAngular2ModelTest.java:45)


Results :

Failed tests:
  TypeScriptAngular2ModelTest.complexPropertyTest:124 expected [null] but found [undefined]
  TypeScriptAngular2ModelTest.listPropertyTest:91 expected [null] but found [undefined]
  TypeScriptAngular2ModelTest.simpleModelTest:45 expected [null] but found [undefined]
```

@wing328

kenisteward:
The tests need to be updated to check for undefined. I don't think they
were

kenisteward:
@trennemen could you update the model tests?

@wing328 are the model tests not run for the ci? If not we need to add them so this doesn't happen again.

kenjones-cisco:
Looks like maybe these were old tests that had been removed or renamed. I ran `mvn clean` and then `mvn package` again and it was successful. 

Hadn't run into this scenario previously, my apologizes.

kenisteward:
Np @tenneman flase alarm haha

wing328:
@kenjones-cisco No worry :)

(The model tests for TS generators (node, angular, etc) are updated as part of this PR)

kenisteward:
K thanks! I didn't realize that

PR #6834, opened 2017-10-27
Swagger Codegen Dockerfile Entrypoint

### LINKED ISSUE #3386: [docker] Hardcoded HTTP port in the generator container
As for [now](https://github.com/swagger-api/swagger-codegen/blob/bd7904d42a2a52afee169880f284b934ae45aa7d/modules/swagger-generator/Dockerfile#L11) generator container listens on port `8080`. To change the HTTP port I have to repeat the command from the `Dockerfile` with additional options:
`docker run -d swagger-generator java -jar /generator/jetty-runner.jar --port 80 /generator/swagger-generator.war`
`--port 80` is my addition to make container listen on port `80`
It would be great to an ability to set HTTP port via environment variables and don't depend on internals of the `Dockerfile`.

jebentier:
### PR checklist

### Description of the PR

I've done the work to add an entrypoint to the code-generator dockerfile in an attempt to resolve any complaints of not being able to easily specify a different port for the java process to run on.

fix #3386

### DIFF
--- a/modules/swagger-generator/entrypoint.sh
+++ b/modules/swagger-generator/entrypoint.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+if [-z $PORT]; then
+  PORT=8080
+fi
+
+java -jar /generator/jetty-runner.jar --port $PORT /generator/swagger-generator.war
--- a/modules/swagger-generator/Dockerfile
+++ b/modules/swagger-generator/Dockerfile
@@ -2,6 +2,7 @@ FROM openjdk:8-jre-alpine
 
 WORKDIR /generator
 
+COPY entrypoint.sh entrypoint.sh
 COPY target/lib/jetty-runner* /generator/jetty-runner.jar
 COPY target/*.war /generator/swagger-generator.war
 
@@ -9,5 +10,5 @@ ENV GENERATOR_HOST=https://generator.swaggerhub.com/api/swagger.json
 
 EXPOSE 8080
 
-CMD ["java", "-jar", "/generator/jetty-runner.jar", "/generator/swagger-generator.war"]
+ENTRYPOINT entrypoint.sh
 


modules/swagger-generator/entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+if [-z $PORT]; then
+  PORT=8080
+fi
+
+java -jar /generator/jetty-runner.jar --port $PORT /generator/swagger-generator.war
```
jimschubert: There are 3 points here…

1. There's no reason to bind this running application inside the container to some other port within the container.  Rebinding to a different host port is more easily done when you execute docker:
    ```
    docker run  -p 8181:8080 swaggerapi/swagger-generator
    ```
    This binds the port 8080 of swagger-generator container to your host port 8181, so you can access the API at http://localhost:8181.
     Each container runs as if it's a separate "machine" on a virtual network. Even if you were running with `--net=host`, I would think it would be an antipattern to assume the user would pass `-e "PORT=8181"` rather than the facilities provided by docker cli. In either case, I would suggest we stick with the standard `-p` and assume users of Docker are aware how to do this.
2. Because of how Docker works, you'll usually want to `exec java …` when creating an entrypoint script like this.  If you don't, your running service usually won't respond to signals sent to the container, and you may break some functionality like `docker wait`. You can verify this by starting the container with your entrypoint script, then exec'ing into that container and running `ps aux`, your main process should always be id=1. This could probably change in future versions  of Docker, and would almost certainly differ across java frameworks as well.
3. There's no way for you to know ahead of time what ports to expose, and only 8080 is exposed in the Dockerfile. This means you either have to revert to the `-p 8181:8080` mentioned in the first point, or `-P` in which case it wouldn't matter what port you've passed in anyway.

jimschubert (changes requested):

modules/swagger-generator/entrypoint.sh
jimschubert: Note… I marked my review as 'request changes', but I think it's best to leave the Dockerfile as-is.  The changes I think need to be made would be in documentation explaining how to change the port if needed (i.e. the `-p 8181:8080` example from above).

jimschubert:
Thanks for this PR. I'm not sure if reviews trigger comment notifications, so I wanted to leave a comment and ask to review the feedback. Have a great night!

PR #6837, opened 2017-10-27
Docker image tags

### LINKED ISSUE #6728: [docker] swagger-codegen-cli docker image - use version tags
##### Description

Currently docker image: https://hub.docker.com/r/swaggerapi/swagger-codegen-cli/ is published only with the "latest" tag. 

##### Suggest a enhancement

Docker image should be versioned with tag value of swagger codegen version, which it contains. This allows predictable usage of it.

jebentier:

### Description of the PR

Added a docker build tag to the circle.yml file so that when the images are pushed, a version tag is pushed as well at latest.

fix #6728

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -10,6 +10,7 @@ machine:
   environment:
     DOCKER_GENERATOR_IMAGE_NAME: swaggerapi/swagger-generator
     DOCKER_CODEGEN_CLI_IMAGE_NAME: swaggerapi/swagger-codegen-cli
+    DOCKER_BUILD_TAG: 2.2.3
 
 dependencies:
   cache_directories:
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -10,6 +10,7 @@ machine:
   environment:
     DOCKER_GENERATOR_IMAGE_NAME: swaggerapi/swagger-generator
     DOCKER_CODEGEN_CLI_IMAGE_NAME: swaggerapi/swagger-codegen-cli
+    DOCKER_BUILD_TAG: 2.2.3
```
wing328: @jebentier thanks for the enhancement. Should this be 2.3.0-SNAPSHOT instead? (the latest master is 2.3.0)

kenjones-cisco: Typically the image tags should match the git tags. To make it dynamic using a git command like such:

```
tag=$(git tag -l --points-at HEAD) ; if [[ -z "$tag" ]]; then echo "latest" ; else echo "$tag" ; fi
```

If there is no tag on the current commit, you get back "latest" otherwise you get the tag which would have been `v2.2.3` if that commit was the one being built.

jimschubert: This could be done more easily in Docker Hub as an automated build. Probably a silly question, but @wing328 do you know if there's a reason it's set up in CI rather than in Docker Hub?

ffledgling:
Bump, I'd like to see this merged. What needs to be modified to get this PR merged (or have tags on the docker images a different way)? Looks like at the least the tag needs to updated to 2.3.0 (but that seems incorrect-ish?) or alternatively the tag needs to be auto-extracted from git. @wing328 ?

circle.yml
wing328: @jimschubert sorry I don't have an answer as it's setup by Tony.

cc @fehguy

wing328: @kenjones-cisco thanks for the explanation. If it's within Docker user's expectation that 2.2.3 can refer to the Docker image built by the latest master (2.3.0), then it's totally fine with me.

wing328:
@ffledgling not that I don't want to merge but I'm busy with other tasks at the moment.

I'll take a deeper look into tags on docker image this coming weekend and merge this PR if I don't have any more question.

wing328:
CircleCI reported the following errors:
```
Successfully built 123b4eebb286
The push refers to a repository [docker.io/swaggerapi/swagger-codegen-cli] (len: 1)
error getting tags for swaggerapi/swagger-codegen-cli: Tag does not exist for 2.2.3
```
Ref: https://circleci.com/gh/swagger-api/swagger-codegen/2610

Anyone has any clue about the error?

We will need to roll back the change if we can resolve this.

jimschubert:
master should be `latest`, while tags should be triggered based on git ref. As I mentioned before, this would be way easier in Docker Hub. To do via CI, you'd need to build from a release branch unless Travis allows building tags.

kenjones-cisco:
The error is because the docker tag is using `CIRCLE_TAG`, but the docker push is using the new environment variable `DOCKER_BUILD_TAG`. Not sure where or if `CIRCLE_TAG` is ever set.

To do this via CI, you will need a script that determines if the current commit being built has a tag, if no tag then it is `latest` which is what most commits to master will be. When a release is tagged, then the commit should have a tag associated it to, and then a tagged version of the image is pushed.

wing328:
OK. I'll have another look tomorrow.

wing328:
I've rolled back this change.

I think what we need to do is to tag the docker image in the next release (in Docker Hub).

ffledgling:
So is the final resolution on this that new releases will have manually added version number in the docker releases? If that's the case, it looks like it did not happen for v2.3.0.

wing328:
@ffledgling yup, and you're right there's no tagging for v2.3.0. I do not have right to do it in dockerhub.

@webron is very busy these days and we'll do it for v2.3.1 instead, which will be released very soon.

Issue #6825, opened 2017-10-26
[Java] Feature request: google-api-client client library


charlescapps:
##### Description

The `google-api-client` [HTTP client](https://github.com/google/google-api-java-client) is a really standard HTTP client used in GCP, and it was used by a Google technology that was basically a clone of swagger (Cloud Endpoints). It's a very common HTTP framework used in GCP, and has good interop with the authentication abstractions provided by Google.

e.g. you can combine HTTPRequestInitializers (`com.google.api.client.http.HttpRequestInitializer`), and Google provides some nice classes that extend `HttpRequestInitializer` such as
* `com.google.api.client.auth.oauth2.Credential` - for OAuth authentication, which handles refreshing tokens and so on
* `com.google.api.client.googleapis.auth.oauth2.GoogleCredential` - which handles service account auth and also refreshes the token automagically.

TL; DR - we really want codegen for the `google-api-client` HTTP library!!

##### Swagger-codegen version

2.2.3. There's no support for this http library, it's not a regression.

##### Swagger declaration file content or url

n/a not a bug.

##### Command line used for generation

After we implement this, I would suggest the `library` is called `google-api-client`.

##### Steps to reproduce

n/a it's a new feature request

##### Related issues/PRs

I don't see anything related. Searching for "google-api-client" turns up 0 results in the issues.

##### Suggest a fix/enhancement

I'm planning to fix this by adding a new library type to the Java client codegen! My PR will be ready soon!

PR #6838, opened 2017-10-27
[Java] Templates to support google-api-client library

charlescapps:
### PR checklist

@bbdouglas (2017/07) @JFCote (2017/08) @sreeshas (2017/08) @jfiala (2017/08) @lukoyanov (2017/09) @cbornet (2017/09)

## Description of the PR
See issue https://github.com/swagger-api/swagger-codegen/issues/6825
This PR adds support for generating client code for the [Google APIs Client for Java](https://github.com/google/google-api-java-client). This HTTP client is not just for Google's APIs, as you can read in the description of their page: 

> The Google APIs Client Library for Java is a flexible, efficient, and powerful Java client library for accessing any HTTP-based API on the web, not just Google APIs.

### Motivation
The `google-api-client` is really standard in GCP for making HTTP requests. It works great with authentication in GCP (or outside GCP), by automatically handling refresh tokens with the google `Credential` class. Furthermore, this provides a clear upgrade path for anyone on Google Cloud Endpoints who wants to migrate to a better REST framework, since Cloud Endpoints uses the `google-api-client`. 

### Changes in this PR
* Add a new library to the Java client codegen called `google-api-client`
* Templates under `modules/swagger-codegen/src/main/resources/Java/libraries/google-api-client`
* A minor tweak to `DefaultCodegen` to have a list of required/body params, so the client can generate overloaded methods that use a `Map<String, Object>` for query params.
* Generated the samples for `google-api-client`, and verified they compile, etc.

### Features of the generated Google ApiClient
* Nice builder-style pattern e.g. `apiClient.myFooApi().doSomeRequest(params...)`
* Generates overloaded API methods to specify query params as a `Map<String, Object>`, and methods ending in `forHttpResponse` that return the raw `HttpResponse` rather than deserializing as a POJO.
* Configurable `basePath`, `HttpTransport`, `HttpRequestInitializer` (used for auth), and `ObjectMapper`.
* Uses a standard Jackson `ObjectMapper` Instead of using the problematic Google-custom-Jackson (e.g.`com.google.api.client.json.jackson2.JacksonFactory`)
* Does not generate auth-related classes under the `auth` package, because I'm not sure what that would gain us; it's really straightforward to use an `HttpRequestInitializer` to provide and access token, e.g. you can use `com.google.api.client.auth.oauth2.Credential` and it can also do token refreshes automatically (or just use an access token if that's all you have).

cc: @drd


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
```diff
@@ -172,10 +173,13 @@ public void processOpts() {
             supportingFiles.add(new SupportingFile("StringUtil.mustache", invokerFolder, "StringUtil.java"));
         }
 
-        supportingFiles.add(new SupportingFile("auth/HttpBasicAuth.mustache", authFolder, "HttpBasicAuth.java"));
-        supportingFiles.add(new SupportingFile("auth/ApiKeyAuth.mustache", authFolder, "ApiKeyAuth.java"));
-        supportingFiles.add(new SupportingFile("auth/OAuth.mustache", authFolder, "OAuth.java"));
-        supportingFiles.add(new SupportingFile("auth/OAuthFlow.mustache", authFolder, "OAuthFlow.java"));
+        // google-api-client doesn't use the Swagger auth, because it uses Google Credential directly (HttpRequestInitializer)
+        if (!"google-api-client".equals(getLibrary())) {
```
JFCote: I think this should go in the GoogleApiClientCodegen.java since it's only useful for you generator.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
```diff
@@ -196,7 +200,7 @@ public void processOpts() {
             apiDocTemplateFiles.remove("api_doc.mustache");
         }
 
-        if (!("feign".equals(getLibrary()) || "resttemplate".equals(getLibrary()) || usesAnyRetrofitLibrary())) {
+        if (!("feign".equals(getLibrary()) || "resttemplate".equals(getLibrary()) || usesAnyRetrofitLibrary() || "google-api-client".equals(getLibrary()))) {
```
JFCote: oh I see other library has done it too... @wing328, is this normal? Since we are using objects, we should'nt have any "if" in the JavaClientCodegen... It should all be in specific file of each generator...
Or there is a good reason for this ??

modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
```diff
@@ -28,6 +28,7 @@
     public List<CodegenParameter> queryParams = new ArrayList<CodegenParameter>();
     public List<CodegenParameter> headerParams = new ArrayList<CodegenParameter>();
     public List<CodegenParameter> formParams = new ArrayList<CodegenParameter>();
+    public List<CodegenParameter> requiredParams = new ArrayList<>();
```
JFCote: I think you need to put `CodegenParameter` in the ArrayList to be able to compile swagger-codegen in Java < 1.7

modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
```diff
@@ -12,7 +12,7 @@
 
 public class CodegenOperation {
     public final List<CodegenProperty> responseHeaders = new ArrayList<CodegenProperty>();
-    public boolean hasAuthMethods, hasConsumes, hasProduces, hasParams, hasOptionalParams,
+    public boolean hasAuthMethods, hasConsumes, hasProduces, hasParams, hasOptionalParams, hasRequiredParams,
```
JFCote: I'm surprised this is needed. Most generator use the params array and loop with "isRequired" to build almost everything. Is there any specific reason to add this to the "generic code"?

charlescapps: Yeah, you can check `{{isRequired}}` but that's not sufficient, because we use `{{hasNext}}` to see if there's a subsequent element, and that isn't aware of `hasRequired`, so it was generating code with 2 commas in a row.

modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
charlescapps: Will do right now! Yes, makes sense to support 1.6.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
charlescapps: Okay, I'll do that right now, sounds like a plan.

charlescapps: FYI this is following the existing pattern. There actually aren't any subclasses of `JavaClientCodegen`; it uses if-statements on the `library` throughout to determine how the configuration of each library. I'd have to figure out how to support a new subclass, it looks like it uses reflection to instantiate the `JavaClientCodegen` with the classes listed in `/Users/charlescapps/Premise/projects/swagger-codegen-charlescapps/modules/swagger-codegen/src/main/resources/META-INF/services/io.swagger.codegen.CodegenConfig`

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
charlescapps: Yeah, it appears there are no subclasses of `JavaClientCodegen`, see above comment, @wing328 probably has a more complete explanation.

JFCote: Ok. It's weird because for example, for the server generator "java play framework" which I maintain, there is: 
`JavaPlayFrameworkCodegen` -> `AbstractJavaCodegen` -> `DefaultCodegen`. 
This prevent having "if" everywhere.

Same thing for Typescript jQuery which I maintain too:
`TypeScriptJqueryClientCodegen` -> `AbstractTypeScriptClientCodegen` -> `DefaultCodegen`

I wonder why the JavaClientCodegen are all mixed in one file. 
For the moment, don't change anything. We will see with @wing328 answers :)

modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenOperation.java
JFCote: Ok makes sense!

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java
charlescapps: That makes sense. Yeah, a refactor into subclasses probably belongs in a separate pull request, but I'm game to do whatever you suggest.

JFCote: This is ok for now and belong in another pull request like you said. :)

wing328: When we added support for other HTTP libraries, we didn't foresee we can support so many libraries (9 HTTP libraries support so far thanks to the awesome community). I agree a separate PR to refactor into subclasses would be ideal.

charlescapps: That makes sense. Is this waiting on any additional reviews? Let me know if any more changes are needed here. Thanks! This will really help out our company in our migration to use Swagger clients for all of our service-to-service communication. (And possibly we could use this for our Android app, too).

wing328: @charlescapps I'll take another look later today and merge accordingly.

wing328:
@charlescapps thanks for the PR, which looks good to me and my test results are good.

cc @lwander

wing328:
UPDATE: Added the new Java API client to CircleCI via https://github.com/swagger-api/swagger-codegen/commit/cf5eba87bd42af752197490037d1506764dfbfcc

charlescapps:
Thanks for being so responsive. I'm excited to use this new client, now we won't need to build from a fork.

wing328:
@charlescapps FYI. I've filed https://github.com/swagger-api/swagger-codegen/pull/6885 to skip the check for body parameter, which can be optional

Ref: https://github.com/swagger-api/swagger-codegen/issues/6878

wing328:
@charlescapps looks like #6885 introduces an issue. In the latest master, I ran `./bin/java-petstore-google-api-client.sh` to update the Java (google-api-client) Petstore sample and notice some changes. Here is the diff for one of them:
```
         return apiClient.getHttpRequestFactory().buildRequest(HttpMethods.POST, genericUrl, content).execute();
     }
 
-    public HttpResponse fakeOuterBooleanSerializeForHttpResponse(Boolean body, Map<String, Object> params) throws IOException {
+    public HttpResponse fakeOuterBooleanSerializeForHttpResponse(Map<String, Object> params) throws IOException {
         Object postBody = body;
         
         UriBuilder uriBuilder = UriBuilder.fromUri(apiClient.getBasePath() + "/fake/outer/boolean");```
```
May I know if you've time to take a look?

PR #6840, opened 2017-10-28
[csharp] fix: ApiReponse Header values

faryu:
The Header values within the ApiResponse class where all "System.String[]". (The string, not a real string array)
With this fix it has the correct header values.

Also added an option to the petstore example in order to use the correct csproj format for the netstandard project generation.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/csharp/ApiResponse.mustache
+++ b/modules/swagger-codegen/src/main/resources/csharp/ApiResponse.mustache
@@ -19,7 +19,7 @@ namespace {{packageName}}.Client
         /// Gets or sets the HTTP headers
         /// </summary>
         /// <value>HTTP headers</value>
-        public IDictionary<string, string> Headers { get; private set; }
+        public IDictionary<string, string[]> Headers { get; private set; }
 
         /// <summary>
         /// Gets or sets the data (parsed HTTP body)
@@ -33,7 +33,7 @@ namespace {{packageName}}.Client
         /// <param name="statusCode">HTTP status code.</param>
         /// <param name="headers">HTTP headers.</param>
         /// <param name="data">Data (parsed HTTP body)</param>
-        public ApiResponse(int statusCode, IDictionary<string, string> headers, T data)
+        public ApiResponse(int statusCode, IDictionary<string, string[]> headers, T data)
         {
             this.StatusCode= statusCode;
             this.Headers = headers;
--- a/modules/swagger-codegen/src/main/resources/csharp/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/csharp/api.mustache
@@ -297,12 +297,12 @@ namespace {{packageName}}.{{apiPackage}}
 
             {{#returnType}}
             return new ApiResponse<{{{returnType}}}>(localVarStatusCode,
-                localVarResponse.Headers.ToDictionary(x => x.{{^netStandard}}Name{{/netStandard}}{{#netStandard}}Key{{/netStandard}}, x => x.Value.ToString()),
+                localVarResponse.Headers.ToDictionary(x => x.{{^netStandard}}Name{{/netStandard}}{{#netStandard}}Key{{/netStandard}}, x => x.Value.ToArray()),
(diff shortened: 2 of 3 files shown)


modules/swagger-codegen/src/main/resources/csharp/ApiResponse.mustache
```diff
@@ -33,7 +33,7 @@ namespace {{packageName}}.Client
         /// <param name="statusCode">HTTP status code.</param>
         /// <param name="headers">HTTP headers.</param>
         /// <param name="data">Data (parsed HTTP body)</param>
-        public ApiResponse(int statusCode, IDictionary<string, string> headers, T data)
+        public ApiResponse(int statusCode, IDictionary<string, string[]> headers, T data)
```
jimschubert: NOTE: Because this generator isn't currently intended to be regenerated over existing code, I wouldn't consider this a breaking change.

jimschubert: n/m. I was looking at this wrong.  This *is* a breaking change.

This should probably be an overloaded constructor which converts the values to a comma separated value per [RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2)

modules/swagger-codegen/src/main/resources/csharp/api.mustache
```diff
@@ -297,12 +297,12 @@ namespace {{packageName}}.{{apiPackage}}
 
             {{#returnType}}
             return new ApiResponse<{{{returnType}}}>(localVarStatusCode,
-                localVarResponse.Headers.ToDictionary(x => x.{{^netStandard}}Name{{/netStandard}}{{#netStandard}}Key{{/netStandard}}, x => x.Value.ToString()),
+                localVarResponse.Headers.ToDictionary(x => x.{{^netStandard}}Name{{/netStandard}}{{#netStandard}}Key{{/netStandard}}, x => x.Value.ToArray()),
```
jimschubert: iirc, this `ToArray()` won't work in .NET 3.5/4.x. Can you regenerate samples and test both the `^netStandard` and `#netStandard` branches here?

modules/swagger-codegen/src/main/resources/csharp/ApiResponse.mustache
faryu: Actually, from the POV of the developer of the consuming application I would prefer this to be a breaking change.
Considering that the previous value field didn't contains any real values at all I don't think there will be many broken builds, because one will notice the wrong values immediately and cannot use it as it is before this fix.
If someone get's a broken build, than he or she will see that there can be multiple values and can handle these as they need.
Combining multiple values back into one delimited string only to have the consumer splitt it up again does not seem to be efficient. If the consumer needs the values as a single combined string, than I would leave that up to the consumer to do so.

modules/swagger-codegen/src/main/resources/csharp/api.mustache
faryu: Thank you for your quick response!

I will take a look at the earlier .NET versions.

faryu:
Now it doesn't break the build of any netframework project.

I digged a bit deeper to locate the cause of the bug and found, that the different RestSharp versions where the cause. RestSharp.portable has a list of values, whereas RestSharp has a combined string.

Now I moved the RestSharp version dependent code into the ApiResponse class.

netstandard / netcore libraries have `IDictionary<string, string[]> Headers`
whereas
netframework has `IDictionary<string, string> Headers.`

modules/swagger-codegen/src/main/resources/csharp/ApiResponse.mustache
faryu: I just fixed it. It's no longer a breaking change for generated netframework clients.

It is a "breaking" (didn't work previously) change for netstandard clients, but they never received the correct values, so I think this is acceptable.

jimschubert:
@faryu Thanks, now that you say the issue that is correctly what I remembered. I couldn't remember the details, just that there is a wonky difference between the two. Ideally, I'd like to invest some time into removing RestSharp, but I haven't had much OSS contribution time for a few months.

The point about the constructor was regarding existing code which may be using the ApiClient directly. I'm ok with your change, since it will affect less consumers. And if it has a negative impact on more than anticipated, we can add the additional constructor. The reason I don't consider it a huge deal is because consumers could just as well modify the csharp templates and add the constructor.

@wing328 thoughts on that?

PR #6842, opened 2017-10-29
[PHP][Symfony] Update Controller.mustache

cydrickn:
### PR checklist

### Description of the  ###

### DIFF
--- a/modules/swagger-codegen/src/main/resources/php-symfony/Controller.mustache
+++ b/modules/swagger-codegen/src/main/resources/php-symfony/Controller.mustache
@@ -136,7 +136,7 @@ class Controller
             return null;
         }
         
-        if ($this->container->get('kernel')->getEnvironment() === 'prod') {
+        if (!$this->container->get('kernel')->isDebug()) {
             return [
                 'message'  => $exception->getMessage(),
             ];
--- a/modules/swagger-codegen/src/main/resources/php-symfony/Controller.mustache
+++ b/modules/swagger-codegen/src/main/resources/php-symfony/Controller.mustache
@@ -135,6 +135,12 @@ class Controller
         if (null === $exception) {
             return null;
         }
+        
+        if ($this->container->get('kernel')->getEnvironment() === 'prod') {
+            return [
+                'message'  => $exception->getMessage(),
+            ];
+        }
 
         return [
             'message'  => $exception->getMessage(),


wing328:
@cydrickn can you elaborate a bit more on the change? Is that a bug fix or enhancement?

cc @naelrashdeen @ksm2 @jebentier @dkarlovi @mandrean @jfastnacht @ackintosh

cydrickn:
Hi @wing328,

The changes is for php symfony to not response to user the class name and not to response the stack trace of errors for prod environment.
The changes will only response the message of the error if the environment is in production.

Thanks

wing328:
@cydrickn thanks for the explanation. If I might, I would suggest adding an "else" clause:
```php
        if ($this->container->get('kernel')->getEnvironment() === 'prod') {
            return [
                'message'  => $exception->getMessage(),
            ];
        } else {
            return [
                'message'  => $exception->getMessage(),
               'type'     => get_class($exception),
               'previous' => $this->exceptionToArray($exception->getPrevious()),
           ];
        }
```
and also including a 1-liner explaining the difference.

modules/swagger-codegen/src/main/resources/php-symfony/Controller.mustache
```diff
@@ -135,6 +135,12 @@ class Controller
         if (null === $exception) {
             return null;
         }
+        
+        if ($this->container->get('kernel')->getEnvironment() === 'prod') {
```
dkarlovi: Do not test against this, check for debug mode.

wing328: Personally, I also prefer the debug flag.

dkarlovi: It makes sense seeing this is debug information. Production environment can also be run in debug mode.

cydrickn: Okay i will change this in using debug

cydrickn: Done changes in 
60a2c7aebf188434ff4713cb4bdce593a1870b8b

wing328:
@cydrickn thanks for the PR, which has been merged into master.

I've updated the Petstore sample for PHP Symfony and removed spaces in the empty line 138 via 8b251555acc7a652e7e6133bbeccd392442466c8

PR #6846, opened 2017-10-29
[Rust] Implement 'object' type conversion

euank:
This contains 2 fixes for the rust codegen.

First, the TODO for "Object" was replaced with the `serde_json::Value` type.
This at least makes it compile; I haven't ensured runtime functionality is correct, but it's certainly no worse than before.

Second, I've updated the provided `Cargo.toml` to include rough version selectors. They correspond to what it's picking now, but will future-proof it a bit more against external crate changes.

cc @frol @farcaller


frol:
LGTM

farcaller:
LGTM

Issue #6690, opened 2017-10-15
C# Enums defined as optional are not nullable [MyEnum (?) myenum]


moreyeel:
##### Description

I am generating C# files from a YAML file. I have an optional enum parameter which I decorate in the template NullValueHandling = NullValueHandling.Ignore meaning if null do not serialize. Like so:

[JsonProperty("days", NullValueHandling = NullValueHandling.Ignore)]
public WeekDays Days { get; set; }

The trouble is that unlike other C# nullable types (int?, bool?) the enum is not generated as nullable. 
This means that the default value (sun) is always sent in my requests which is not the desired behavior. I cannot discern when the user chose not to configure a value at all and when the user configured sun in the aforementioned enum.

##### Swagger-codegen version

swagger-codegen version 2.2.3

##### Swagger declaration file content or url

```yaml

swagger: '2.0'
info:
  version: '1'
  title: My title
host: 'localhost:10010'
basePath: /
schemes:
  - http
  - https
consumes:
  - application/json
produces:
  - application/json
definitions:
  WeekDays:
    type: string
    enum:
      - sun
      - mon
      - tue
      - wed
      - thu
      - fri
      - sat
  My_Class_With_Optional_Enum:
    properties:
      quarantine:
        type: boolean
      grayware:
        type: boolean
      days:
        $ref: '#/definitions/WeekDays'
      
```

##### Command line used for generation

java -jar -Dproject -Dmodels -DmodelTests=false -DmodelDocs=false -DdebugModels=true -DdebugOperations=true swagger-codegen-cli-2.2.3.jar generate -i my.yaml -l csharp -o outputDir -t templateDir --type-mappings Guid?=string

##### Steps to reproduce

Generate the C# files using swagger-codegen-cli-2.2.3.jar
Open the file MyClassWithOptionalEnum.cs
See public WeekDays Days { get; set; }
Expected" public WeekDays ? Days { get; set; }

##### Suggest a fix/enhancement

C# treat enums as a nullable type

moreyeel:
Any ideas about this one?

jimschubert:
@moreyeel I've opened #6851. I was wondering if you could build this branch and evaluate?

moreyeel:
Hi Jim, 

Thank you very much for this, while the solution is not what I expected, it certainly does the job. 
[...]

PR #6851, opened 2017-10-30
[csharp] Treat enum models consistently

jimschubert:
C# works differently from most languages in that enums are not
considered objects. This means default(EnumType) will choose a default
of the first enum option. This isn't desirable because it breaks the
required = false functionality of swagger specs, which defines a
property which isn't required to exist in the message body.

Rather than force consumers to use enum values such as UNSPECIFIED, UNKNOWN,
NOT_SET, etc... we can treat enums as primitives. This means any
non-required enum will become Nullable<EnumType> regardless of whether
it is defined as an inline enum or a referenced enum model.

### PR checklist

### Description of the PR

This allows us to reference enums defined as RefModel using enum logic, rather than object logic. That is, `default(EnumType?)` rather than `default(EnumType)`.

I've made some test-specific modifications to DefaultGenerator, such that integration tests can override the switches we pull from System properties and exclude the swagger codegen metadata (for test maintainability).

See #6690 et al.

/cc @mandrean  @wing328


jimschubert:
I'll rebuild samples after others have reviewed the changes, to limit noise.

jimschubert:
See linked issue, reporter has verified this branch against about 400 models with no additional concerns.

PR #6866, opened 2017-11-02
[haskell-http-client] refactor & promote accept mimetype to request; simplify templates

jonschoning:
* add AuthMethodException

* rename cli prop

* simplify templates


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/HaskellHttpClientCodegen.java
```diff
@@ -681,6 +685,9 @@ private void processReturnType(CodegenOperation op) {
                 op.vendorExtensions.put(X_HAS_UNKNOWN_RETURN, true);
             } else {
                 returnType = "NoContent";
+                if(!op.vendorExtensions.containsKey(X_INLINE_ACCEPT)) {
```
wing328: Minor suggestion: adding a space after `if`

jonschoning: thanks, will use auto code-formatter to fix

Issue #6818, opened 2017-10-26
[Scala][Finch] Further enhancements to the Finch generator


wing328:
##### Description

Here is a list of suggested enhancements to Scala Finch generator by @jfraudeau 

- [ ] Add support for Cookie parameters (OAS 3.0)
- [ ] Query parameters and Header that are not interpreted as strings
- [ ] All arities aren't always handled
- [ ] Improve code quality

##### Swagger-codegen version

Latest master

##### Related issues/PRs

https://github.com/swagger-api/swagger-codegen/pull/6809

##### Suggest a fix/enhancement

If anyone wants to work on these enhancements or have other suggestions, please reply to let us know.

erikwj:
I'd like to work on it, was working on a similar solution to fix the params and headers, also added the security parameters from OAS

wing328:
@erikwj thank you 🙇 

Let us know if you need help with the PR.

matteomazza91:
Since
 * the commit f0b07b2 added the support for security definitions and
 * the commit 1b90a05 exploit it
I want notify that I create the PR #7537. It extends the security definitions with cookie auth (OpenApi 3.0.0).
I hope it can be useful if you want to extend the FinchServerCodegen

PR #6877, opened 2017-11-03
#6818: Update Finch version

erikwj:
@clasnake  @jimschubert
Added scalafix for import cleanup, some rewrites, added support for non-string headers, added support for security definitions (apiKeys only for now)

Adding tests for Finch will be in a next pull request


erikwj:
found some issues, closing this one

Issue #6784, opened 2017-10-23
[CSharp] Issue Generating client when nonPublicApi = false


lunat:
Hi,

Csharp client generator creates all classes and methods as "internal" even if "nonPublicApi":"false".

I am using swagger-codgen version 2.3.0

Thanks,

lunat

jimschubert:
@lunat how are you declaring `nonPublicApi`? This is an additional-property option.

lunat:
I also tried with https://generator.swagger.io.
Setting both
```
"options":{  
      "nonPublicApi":false
}
```
or 
```
"options":{  
      "nonPublicApi":true
}
```
I get internal classes.
Note that if I exclude "nonPublicApi" setting from options, all classes are public.

jimschubert:
Verified with

```
{
  "options": {
    "packageVersion": "1.0.1",
    "nonPublicApi": false
  },
  "swaggerUrl": "https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/examples/v2.0/json/petstore-expanded.json"
}
```

And classes are always internal.

However, I don't think this is C# specific. I've been working on another branch, and the true/false variables which previously processed correctly in Mustache are no longer being treated as true/false. This seems to be causing "false" to be treated as truthy.

I'd have to dig into this a bit to understand what changed, and when it changed. You can see that the visibility modifier in C# hasn't changed since I added it in January: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/csharp/visibility.mustache, so the expectation would be that this code works the same in 2.2.3 as it did in 2.2.2.

jimschubert:
@lunat I opened #6899 to address this issue. However, regenerating all C# code produced a few issues that I'll need to follow up on:

* .NET 4.0 code doesn't compile with the `build.sh` script, complaining that RestSharp needs nuget 2.12 instead of the downloaded 2.8 (even though I have 2.12 installed globally).
* SwaggerClientWithPropertyChanged sample doesn't seem to compile or copy files into artifacts folders on build.
* SwaggerClientNetStandard sample doesn't build with msbuild locally (may be a config issue on my machine), and there are no build scripts included in the generated output.

I'll wait and see how CI handles things, and follow up on the above points later in the week.

PR #6899, opened 2017-11-07
[csharp] clean boolean additional properties 6784

jimschubert:
### PR checklist

### Description of the PR

/cc @mandrean @wing328 

This is a general cleanup of how we store booleans on `additionalProperties`. It seems as though either `"false"` used to be treated in template as `false` or we were more consistent in replacing false strings with false booleans on the `additionalProperties` object in the C# generators. I personally implemented and tested the functionality in the linked issue, so I'm not sure why something like `{{^property}}` is generated regardless of the boolean value now (unless I was crazy when I implemented it 💃 ). I tried to determine why some functionality that has worked before (e.g. `nonPublicApi`) doesn't work in 2.2.3 but I was unable to track down any issue. I assume a git bisect would take quite a long time, and handling booleans explicitly seemed both easier and a best practice.

I've also taken this opportunity to expose both `supportsAsync` and `validatable` as options on the C# generator, since these are valid user-defined settings.

I've also added some warnings to clarify when the generator removes an explicit setting due to lack of support in the target framework or template. This removal of settings may not be ideal, but my changes in this PR were done to remain consistent with other settings which remove values.

see #6784


jimschubert:
I've resolved compilation of .NET 4.0 sample via Jetbrains Rider. The `build.sh` script complains about ambiguity:

```
error CS0104: `ReadOnlyDictionary' is an ambiguous reference between `System.Collections.ObjectModel.ReadOnlyDictionary<TKey,TValue>' and `IO.Swagger.Client.ReadOnlyDictionary<T,K>'
```
I'll need to look more closely and see why this error occurs, considering ReadOnlyDictionary wasn't introduced until .net 4.5.

jimschubert:
So the issue was that the targetFramework (used in project files) was being used for the `mcs -sdk` option's value. These aren't the same values, or the same format. For example, net40 -> 4.

To resolve this issue, I've explicitly added a translation value to `additionalProperties` so we can use the correct mono SDK for the selected target framework.  The issue from my previous comment about `ReadOnlyDictionary` ambiguities between the implementation I added and that of `System.Collections.ObjectModel.ReadOnlyDictionary` had to do with mono falling back to the default SDK.

Once I resolved this issue, I found that specifying `-sdk:4` correctly also caused mono to specify the C# language version correctly and broke compilation because Rider auto-created delegates for the `ReadOnlyDictionary` using `nameof`.  Oddly, I had different behaviors with mono 4.8 and mono 5.4.1.6. Rather than fix this one-off error, I've explicitly mapped `langversion` based on the `sdk` value passed.

After making these changes and cleaning up some of the bash scripts under `bin`, each sample that can be built with a `build.sh` script now build successfully using latest mono.

I know a lot of the commits in the PR are unrelated to the sanitization of booleans, but I figured the build cleanup was directly related.

bin/csharp-petstore.sh
```diff
@@ -26,6 +26,6 @@ fi
 
 # if you've executed sbt assembly previously it will use that instead.
 export JAVA_OPTS="${JAVA_OPTS} -XX:MaxPermSize=256M -Xmx1024M -DloggerPath=conf/log4j.properties"
-ags="generate $@ -i modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml -l csharp -o samples/client/petstore/csharp/SwaggerClientNet40 --additional-properties packageGuid={321C8C3F-0156-40C1-AE42-D59761FB9B6C} -c ./bin/csharp-petstore-net-40.json"
+ags="generate $@ -i modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml -l csharp -o samples/client/petstore/csharp/SwaggerClient --additional-properties packageGuid={321C8C3F-0156-40C1-AE42-D59761FB9B6C}"
```
jimschubert: @wing328 Your commit when you implemented .NET 4.0 generator added `csharp-petstore-net-40.sh` as well as modifying this file. It looked to me like you may not have meant to make `csharp-petstore.sh` compile .NET 4.0?

Feel free to revert this change and the changes in `csharp-petstore-all.sh`.

I would actually be cool with making this file compile .NET 4.5 as the minimal example (since 4.5 is an in-place update to 4.0 anyway), and create a separate script for .NET 3.5.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CSharpClientCodegen.java
```diff
@@ -17,21 +18,33 @@
     private static final String NET45 = "v4.5";
     private static final String NET40 = "v4.0";
     private static final String NET35 = "v3.5";
+    // TODO: v5.0 is PCL, not netstandard version 1.3, and not a specific .NET Framework. This needs to be updated,
+    // especially because it will conflict with .NET Framework 5.0 when released, and PCL 5 refers to Framework 4.0.
```
jimschubert: PCL 5.0 actually refers to .NET Framework 4.6 and higher, .NET Core 1.0 and Windows Universal 10.0. I commented this correctly later in the file. As a TODO, I don't think it's a huge issue.

I'd like to straighten this out before 2.3.0 release.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CSharpClientCodegen.java
```diff
@@ -156,133 +173,160 @@ public CSharpClientCodegen() {
...
+            setValidatable(Boolean.FALSE);
             setSupportsAsync(Boolean.FALSE);
-            if (additionalProperties.containsKey("supportsAsync")) {
-                additionalProperties.remove("supportsAsync");
-            }
-            additionalProperties.put("validatable", false);
-            additionalProperties.put("net35", true);
         } else if (NETSTANDARD.equals(this.targetFramework)) {
+            // TODO: NETSTANDARD here is misrepresenting a PCL v5.0 which supports .NET Framework 4.6+, .NET Core 1.0, and Windows Universal 10.0
+            additionalProperties.put(MCS_NET_VERSION_KEY, "4.6-api");
```
jimschubert: NOTE: I added this, although this `NETSTANDARD` isn't writing `build.sh` at the moment.

bin/csharp-petstore.sh
wing328: I was my mistake. Thanks for fixing it.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CSharpClientCodegen.java
wing328: Yup, 2.3.0 is a release with breaking changes so please feel free to fix these wordings before the release.

PR #6905, opened 2017-11-07
[CppRest] Add `Object` and fix modelbase includes

### LINKED ISSUE #6512: [CPPREST] Arbitrary object type results in a compilation error (missing Object.h)
##### Description

Using Codegen on the swagger specification file containing properties of the arbitrary object type results in the following compilation error:

```
.../model/Inline_response_200.h:25:10: fatal error: Object.h: No such file or directory
 #include "Object.h"
          ^~~~~~~~~~
```

##### Swagger-codegen version

I am using the master branch, but I believe that the issue is there from the initial introduction of CPPREST generator.

```yaml
swagger: "2.0"
info:
  title: "title"
  version: "1.0.0"
paths:
  /pets/:
    get:
      parameters:
      - name: "offset"
        in: "query"
        required: false
        type: "integer"
      responses:
        200:
          description: "successful operation"
          schema:
            type: "object"
            properties:
              extra_info:
                type: "object"
```

##### Steps to reproduce

Just run the generator and try building the module:

```
$ mkdir build
$ cd build
$ cmake -D CPPREST_ROOT=/usr ../
$ make -j 8
```

It seems that there should be `Object.h` implementation (it seems that the standard hash map can be used here).

fvarose:
### PR checklist

### Description of the PR

This PR addresses two issues:

1. ~The `Object.h` header file does not exist, we replace it with `ModelBase.h` which provides the expected functionality (fix #6512). This was seemingly an early copy and paste mistake that wasn't caught by the PetStore example.~
Edit (14/11/2017): this PR provides a simple implementation of `Object`, wrapped around `web::json::value` from casablanca. This fixes #6512.
2. Fixed some relative include paths, to prevent clashes during linking when using two separate APIs that generate identical file names.

I've updated the sample code accordingly and managed to compile it without any problems.
This change has also been tested successfully on a codebase where two separate APIs coexist (which wasn't possible at all before).


fvarose:
cc @stkrwork @ravinikam

fvarose:
Update: I've had a second look and this actually doesn't solve the issue mentioned in #6512..
I will need to revise this by creating an actual `Object` class that derives from `ModelBase` and implements the virtual conversion functions.

fvarose:
This PR is now ready to be reviewed again. It fixes #6512.
@wing328 @stkrwork @ravinikam

wing328:
@fvarose thanks for the enhancement. 

murzic4:
hi all
Im not sure but it looks like types UUID and Number are missing as well

wing328:
@murzic4 do you mind filing a PR to add those?

PR #6928, opened 2017-11-10
[Gatling] Generate gatling scenarios from swagger issue 6927

ryandavis84:
### PR checklist

### Description of the PR
I wanted an easier way to write gatling tests for micro-services that utilize swagger.  

Things that may be out of the norm:
1. I created files from the preProcessSwagger method as i did not want a new file just for every api but one for every type of parameter type inside of an operation.  There are other ways of doing this but i felt like this simplified the approach.  

2. In some strange instance in the model generation it will import a java.util.List even though list is removed from the import mappings.  I cant track down why this has happened so i am manually removing this from the models as they appear.  Maybe this is a bug somewhere in the swagger-codegen

I have tested with various swagger json's, and everything is working as expected (with the exception of the list issue mentioned above)


ryandavis84:
@clasnake @jimschubert I know this isnt really scala as much as it is gatling expression but it does use more scala than anything else.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/GatlingCodegen.java
```diff
@@ -0,0 +1,344 @@
...
+import io.swagger.models.parameters.*;
+import io.swagger.models.properties.*;
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.StringUtils;
+
+import java.io.IOException;
+import java.util.*;
+import java.io.File;
+
+public class GatlingCodegen extends DefaultCodegen implements CodegenConfig {
```
cbornet: Shouldn't it extend `AbstractScalaCodegen` ?

ryandavis84:
Any direction why shippable failed would be helpful, I ran this locally and everything worked.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/GatlingCodegen.java
ryandavis84: I started this project with the generated helper files.  I can change it to that.

ryandavis84: Thank you @cbornet

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/GatlingCodegen.java
```diff
@@ -11,7 +11,7 @@
 import java.util.*;
 import java.io.File;
 
-public class GatlingCodegen extends DefaultCodegen implements CodegenConfig {
+public class GatlingCodegen extends AbstractScalaCodegen implements CodegenConfig {
```
cbornet: The idea behind AbstractScalaCodegen is that a lot of code is factorized in the class. So you should be able to remove a lot of code from GatlingCodegen.

ryandavis84: I will look it over later today and remove whatever I can.  Im assuming this will help with my java.util.List getting imported as well?

cbornet: You should be able to use common importMapping, typeMapping, reserved words and pre/post-processing methods.

ryandavis84: Yeah, i used the importMapping and removed List as you can see but sometimes it ends up getting imported into my model template.

ryandavis84: It did fix the list issue.  Thank you.

ryandavis84: The only other issue that I have is that scala has a reservedWord as type, but i know many apis will have the word type in them. So for the JSON string i build out, it contians the _type as a field.  

Im assuming this is where I could use jackson but even if i do that, when i call {{name}} from the {{vars}} its going to show up as _type anyways.  It would be really nice if the reservedWords gave an additional field to use as variables instead of changing the name itself.

jimschubert: You can escape reserved words directly.

See how I did this in the Kotlin client recently:
https://github.com/swagger-api/swagger-codegen/pull/6874/files

I'd be fine with Scala escaping similarly.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/GatlingCodegen.java
```diff
@@ -160,14 +160,14 @@ public GatlingCodegen() {
...
-     * those terms here.  This logic is only called if a variable matches the reseved words
+     * Gatling does not need the models to have escaped words as it builds models dynamically instead of through
+     * an instance of the object.
      *
      * @return the escaped term
      */
     @Override
     public String escapeReservedWord(String name) {
-        return "_" + name;  // add an underscore to the name
+        return name;  // add an underscore to the name
```
cbornet: Maybe instead of overriding here, remove from AbstractScalaCodegen and override in other concrete Scala codegen classes

ryandavis84: The models for gatling are more for knowledge than for use.  I may remove them all together and just keep the toStringBody as that is really all that is needed.  Since all the objects are constructed from fields that are essentially coming from a CSV and translated through the session expression.  I left them there as it made sense at the time.

wing328:

ryandavis84:
I did what that guide told me to do and now i have all these commits.  Should i decline and just remake the PR from a different branch?

wing328:
@ryandavis84 I would suggest you submit a new PR instead.

PR #6751, opened 2017-10-19
Implemented fix for #6006. Mime-type support

gupbeheer:
Technical committee: @TiFu @taxpon  @sebastianhaas  @kenisteward  @Vrolijkx 

### PR checklist

- [X] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [X ] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [X] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [X] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming langauge.

### Description of the PR
Fix for #6006 and #6454. Since the fix done in #6454 has been removed.
I tested using the Angular-v2 and v4 scripts.
This fix is needed to support rest services that use mime-types for versioning.
The configuration has been expanded with methods that determine the correct Accepts and Content-Type (for non-form posts) mime-type and can be overridden if needed when creating a configuration. (If multiple versions are supported by the rest service, the correct filtering can be done there)

There is one thing I am not completely satisfied with and that is that all generated operations have a 'consumes' block. The reason behind that is that there should be no duplicate code in the 'api.service.mustache'. But I could not find a Mustache section that is equal to '#bodyParam' and '#hasFormParams' together.

PR #6945, opened 2017-11-13
Typescript angular linting fixes

gupbeheer:
Technical committee: @TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx

### PR checklist

### Description of the PR

When I added the generated Typescript Angular code to my project, my 'tslint' did not like the generated code. I updated the code to make my 'tslint' happy. This PR is to share my changes with the community.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache
@@ -1,5 +1,5 @@
 {{>licenseInfo}}
-/* tslint:disable:no-unused-variable member-ordering */
+/* tslint:disable:no-unused-variable member-ordering import-spacing*/
 
 import { Inject, Injectable, Optional }                      from '@angular/core';
 {{#useHttpClient}}
@@ -65,7 +65,7 @@ export class {{classname}} {
      */
     private canConsumeForm(consumes: string[]): boolean {
         const form = 'multipart/form-data';
-        for (let consume of consumes) {
+        for (const consume of consumes) {
             if (form === consume) {
                 return true;
             }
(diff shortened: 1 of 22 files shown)


kenisteward:
Your winter might break others due  to the nature of linting I do not
suggest this pe be merged .  Just lint iit in your own project

kenisteward:
What o truly suggest is not putting the code into you project but building
it and serving it as a bundle .  The code is generated and shouldn't really
be changed by hand.

kenisteward:
Another suggestion would be to use the exclude option for tslint to ignore
the generated code. https://palantir.github.io/tslint/usage/configuration/

wing328:
So I think what @kenisteward suggested is to use customized templates to meet different linting requirement and use the `-t` option when generating the code.

kenisteward:
@wing328 that is definitely one way of going about it. I honestly forgot about that! haha.  I was trying to give a solution without them having to learn the inner workings of mustache stuff.

@gupbeheer while your changes meet your (and mostly my) ts lint restrictions there are some of them that are preferential and definitely may not (e.g. forcing single quotes instead of double is based on preference not necessity)

You could definitely create your own templates that fit your coding style but to be honest linting mostly covers keeping code made by developers under check to make sure we write similar code (in my opinion).  Since this is generated, you can just ignore linting rules on this folder as  you (should) never be working directly on this code.

sebastianhaas:
I go with @kenisteward here, I would recommend not putting the auto-generated sources in your project but rather include it per npm and use it as a dependency.

gupbeheer:
No problem. I saw the code I committed in #6751 and the linting errors that came up in the files I committed were 90% my new code. So I made it more 'as the other code' and made the linting pass. I thought I should contribute this. That is the reason for this PR. 
In our own codebase we exempt the generated code from linting and we generate service clients and push them to our NPM registry by using the Typescript specific properties:  `--additional-properties npmName="",npmVersion="",npmRepository=""`

kenisteward:
@gupbeheer great to hear!  If you could close this issue then that'd be great :)

PR #6955, opened 2017-11-14
[Typescript][Angular]Issue #6902 Return Client Response for Angular 4.3 and above with Observe string

### LINKED ISSUE #6902: [Typescript][Angular] Response headers are not available.
##### Description
I want to access the response headers for an E-tag. For my current swagger yaml file, it doesn't generate anything to access it. [This is the same issue](https://github.com/swagger-api/swagger-codegen/issues/3999) it is supposedly fixed, but in my current generated code base I cannot find the withHttpInfo methods, are they removed in a newer version? Or am I doing something wrong?
@wing328

##### Swagger-codegen version

master (commit f78d958 ))

```yaml
path:
    parameters:
    ............
    get:    
      .............
      produces:
      - "application/json"
      .............
      responses:
        200:          
          headers:
            ETag:
              type: string
          schema:
            ............         
```

```typescript
 public get(......params..): Observable<any> {
        if (parm=== null || param=== undefined) {
            throw new Error('error.');
        }
        let headers = this.defaultHeaders;

        return this.httpClient.get<any>(`${this.basePath}...path...}`,
            {
                headers: headers,
                withCredentials: this.configuration.withCredentials,
            }
        );
    }
```

#### Suggestion

There is a way to [fix it](https://stackoverflow.com/a/45506099/8836665) with the new httpclient.
@kenisteward

kenisteward:
### PR checklist

### Description of the PR

Fixes #6902 with some added bonuses
@bedag-moo @HermenOtter Gonna need you 2 to test this as i'm not currently running angular ^4.3 Will be upgrading to 5 very soon though.

Note: For overloads to work with proper intellisense the first overload must always be for 'body'.  if any changes are made to the overload that rearranges them intellisense will break and confuse consumers

### Techies!
@TiFu @taxpon  @sebastianhaas  @Vrolijkx

### DIFF
--- a/samples/client/petstore/typescript-angular-v2/default/api/pet.service.ts
+++ b/samples/client/petstore/typescript-angular-v2/default/api/pet.service.ts
@@ -235,7 +235,7 @@ export class PetService {
 
         let requestOptions: RequestOptionsArgs = new RequestOptions({
             method: RequestMethod.Post,
-            headers,
+            headers: headers,
             body: body == null ? '' : JSON.stringify(body), // https://github.com/angular/angular/issues/10612
             withCredentials:this.configuration.withCredentials
         });
@@ -288,7 +288,7 @@ export class PetService {
 
         let requestOptions: RequestOptionsArgs = new RequestOptions({
             method: RequestMethod.Delete,
-            headers,
+            headers: headers,
             withCredentials:this.configuration.withCredentials
         });
         // https://github.com/swagger-api/swagger-codegen/issues/4037
@@ -341,7 +341,7 @@ export class PetService {
 
         let requestOptions: RequestOptionsArgs = new RequestOptions({
             method: RequestMethod.Get,
-            headers,
+            headers: headers,
             search: queryParameters,
             withCredentials:this.configuration.withCredentials
         });
@@ -395,7 +395,7 @@ export class PetService {
 
         let requestOptions: RequestOptionsArgs = new RequestOptions({
             method: RequestMethod.Get,
-            headers,
+            headers: headers,
             search: queryParameters,
             withCredentials:this.configuration.withCredentials
         });
@@ -441,7 +441,7 @@ export class PetService {
 
         let requestOptions: RequestOptionsArgs = new RequestOptions({
             method: RequestMethod.Get,
-            headers,
+            headers: headers,
             withCredentials:this.configuration.withCredentials
         });
         // https://github.com/swagger-api/swagger-codegen/issues/4037
@@ -495,7 +495,7 @@ export class PetService {
 
         let requestOptions: RequestOptionsArgs = new RequestOptions({
             method: RequestMethod.Put,
-            headers,
+            headers: headers,
             body: body == null ? '' : JSON.stringify(body), // https://github.com/angular/angular/issues/10612
             withCredentials:this.configuration.withCredentials
         });
@@ -569,7 +569,7 @@ export class PetService {
 
         let requestOptions: RequestOptionsArgs = new RequestOptions({
             method: RequestMethod.Post,
-            headers,
+            headers: headers,
             body: convertFormParamsToString ? formParams.toString() : formParams,
             withCredentials:this.configuration.withCredentials
         });
@@ -645,7 +645,7 @@ export class PetService {
 
         let requestOptions: RequestOptionsArgs = new RequestOptions({
             method: RequestMethod.Post,
-            headers,
+            headers: headers,
             body: convertFormParamsToString ? formParams.toString() : formParams,
             withCredentials:this.configuration.withCredentials
         });
--- a/samples/client/petstore/typescript-angular-v2/default/api/pet.service.ts
+++ b/samples/client/petstore/typescript-angular-v2/default/api/pet.service.ts
@@ -15,6 +15,7 @@ import { Inject, Injectable, Optional }                      from '@angular/core
 import { Http, Headers, URLSearchParams }                    from '@angular/http';
 import { RequestMethod, RequestOptions, RequestOptionsArgs } from '@angular/http';
 import { Response, ResponseContentType }                     from '@angular/http';
+import { CustomQueryEncoderHelper }                          from '../encoder';
 
 import { Observable }                                        from 'rxjs/Observable';
 import '../rxjs-operators';
@@ -24,7 +25,6 @@ import { Pet } from '../model/pet';
 
 import { BASE_PATH, COLLECTION_FORMATS }                     from '../variables';
 import { Configuration }                                     from '../configuration';
-import { CustomQueryEncoderHelper }                          from '../encoder';
 
 
 @Injectable()
@@ -196,6 +196,7 @@ export class PetService {
      * Add a new pet to the store
      * 
      * @param body Pet object that needs to be added to the store
+     
      */
     public addPetWithHttpInfo(body: Pet, extraHttpRequestParams?: RequestOptionsArgs): Observable<Response> {
         if (body === null || body === undefined) {
@@ -234,7 +235,7 @@ export class PetService {
 
         let requestOptions: RequestOptionsArgs = new RequestOptions({
             method: RequestMethod.Post,
-            headers: headers,
+            headers,
             body: body == null ? '' : JSON.stringify(body), // https://github.com/angular/angular/issues/10612
             withCredentials:this.configuration.withCredentials
         });
@@ -251,6 +252,7 @@ export class PetService {
      * 
      * @param petId Pet id to delete
      * @param apiKey 
+     
      */
     public deletePetWithHttpInfo(petId: number, apiKey?: string, extraHttpRequestParams?: RequestOptionsArgs): Observable<Response> {
         if (petId === null || petId === undefined) {
@@ -286,7 +288,7 @@ export class PetService {
 
         let requestOptions: RequestOptionsArgs = new RequestOptions({
             method: RequestMethod.Delete,
-            headers: headers,
+            headers,
             withCredentials:this.configuration.withCredentials
         });
         // https://github.com/swagger-api/swagger-codegen/issues/4037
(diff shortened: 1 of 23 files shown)


samples/client/petstore/typescript-angular-v2/default/api/pet.service.ts
```diff
@@ -338,7 +341,7 @@ export class PetService {
 
         let requestOptions: RequestOptionsArgs = new RequestOptions({
             method: RequestMethod.Get,
-            headers: headers,
+            headers,
```
macjohnny: I suggest not to remove the explicit `headers:` part, since without it, refactoring variable names can easily lead to problems

kenisteward: I'm not exactly sure hat you mean.  Please elaborate so i can understand.  Thanks!

macjohnny: @kenisteward when creating objects like this:
```
{
    something
}
```
instead of
```
{
    something: something
}
```
the name of the property of the object changes when refactoring variable names, leading to potential problems if one does not notice it.

kenisteward: Ahh.  Unfortunately, mostly due to how our templates are setup, that will be a big breaking change either way.  

For the 4.3 client and the non 4.3 client they both use the same name for a variable however if someone changed the name for 1 client thinking it would be fine that would also cause errors.  

I'm also assuming you mean refactoring via our mustache templates as one should not be refactoring their code directly as new versions should be automatically generated.

HermenOtter:
@kenisteward When generating with Angular typescript 4.3 typescript returns this error: 

```
severity: 'Error'
message: 'Argument of type '{ withCredentials: boolean; headers: HttpHeaders; observe: "body" | "response" | "events"; report...' is not assignable to parameter of type '{ headers?: HttpHeaders; observe?: "body"; params?: HttpParams; reportProgress?: boolean; respons...'.
  Types of property 'observe' are incompatible.
    Type '"body" | "response" | "events"' is not assignable to type '"body"'.
      Type '"response"' is not assignable to type '"body"'.'
source: 'ts'
```

HermenOtter:
@kenisteward It seems to be the order of the parameters.

When I order them like this, I got 1 problem less:
```typescript
 return this.httpClient.get<InlineResponse200>(`${this.basePath}/subj/v1/subjecten/`,
            {
                headers,
                observe,
                params: queryParameters,
                reportProgress,
                withCredentials: this.configuration.withCredentials,
            }
        );
```
Now it complains about credentials:

kenisteward:
@HermenOtter turns out i missed up the mustache template for when there are no previous fileds but we still need observe / report progress looking into this now.

HermenOtter:
@kenisteward Did you push a fix to the PR?

kenisteward:
@HermenOtter Not yet.  It turns out that was a bug but the actual bug causing your problem was because it doesn't like typing the function as  'body' | 'response' | 'events' for correlation with passing it to the httpclient.  I'm going to switch to any right now and try again.  I'm not getting any errors locally with any and we get type safety with the overloads

kenisteward:
@macjohnny I think i''ll change to what you suggested for clarity in the future for others.

kenisteward:
@HermenOtter Fixed!

HermenOtter:
@kenisteward @macjohnny I can confirm it works as expected for my application. However I did not test all combinations.

Thank you for your effort.

kenisteward:
@HermenOtter thanks for testing! Yay new features! If there are no other issues @wing328  please merge at you leisure. 

@bedag-moo  would love for you to test as well if you have time!

samples/client/petstore/typescript-angular-v4.3/npm/api/pet.service.ts
```diff
@@ -156,8 +171,13 @@ export class PetService {
...
      * Multiple status values can be provided with comma separated strings
      * @param status Status values that need to be considered for filter
+     * @param observe set whether or not to return the data Observable as the body, response or events. defaults to returning the body.
+     * @param reportProgress flag to report request and response progress.
      */
-    public findPetsByStatus(status: Array<string>): Observable<Array<Pet>> {
+    public findPetsByStatus(status: Array<string>, observe?: 'body', reportProgress?: boolean): Observable<Array<Pet>>;
+    public findPetsByStatus(status: Array<string>, observe?: 'response', reportProgress?: boolean): Observable<HttpResponse<Array<Pet>>>;
+    public findPetsByStatus(status: Array<string>, observe?: 'events', reportProgress?: boolean): Observable<HttpEvent<Array<Pet>>>;
+    public findPetsByStatus(status: Array<string>, observe: any = 'body', reportProgress: boolean = false ): Observable<any> {
```
bedag-moo: The type of observe seems too permissive here, because it would allow invocations like:

```
findPetsByStatus(status, observe: HttpResponse)
findPetsByStatus(status, observe: "Body")
```

Angular's HttpClient restricts this parameter to permitted values:
```
observe?: HttpObserve,
```
where
```
/**
 * @stable
 */
export type HttpObserve = 'body' | 'events' | 'response';
```

So we should be able to import that type, and use it ourselves? Is there a reason we can't follow angular's lead here?

kenisteward: you can't import it.  I tried it myself and it did not work.  You can try it in your project if you'd like and tell me if i'm doing something wrong.

kenisteward: also it doens't allow those invocations.  you are only allowed to pass thoes specific things cause it's the string type.

kenisteward: the last observe: any is not counted for the invocations.  only the overloads before it.

bedag-moo:
I'm afraid I won't have time for tests in the near future, but if it works for Hermann Otter, it very likely works for me, too.

samples/client/petstore/typescript-angular-v4.3/npm/api/pet.service.ts
kenisteward: @bedag-moo from angular public_api.d.ts
export { HttpClient } from './src/client';

they don't export HttpObserve

bedag-moo: My apologies, I didn't know that typescript treats does not consider the last signature part of the overload list. Your code is fine then, sorry for not checking.

bedag-moo:
This PR looks good to me.

HermenOtter:
@kenisteward @bedag-moo I had more time and I came across a few warnings and a problem.
TSLint is complaining about the use of let's instead of const, as well ` if (httpContentTypeSelected != undefined) {` != should be !== and `[tslint] " should be ' (quotemark)`

@kenisteward Since you changed the implementation to <any> in the method overload, I doesn't return `HttpResponse<Subject>` anymore, but instead `HttpResponse<any>`. I am not sure if it has something to do with my version of typescript or my intellisense.

I might be able to find some time to commit fixes to those warnings, but I don't know why the typings don't work.

kenisteward:
In general , Please don't commit changes for tslint as everyone's linting
can be different .

In this case changing the != To !== Is programmatically different and
correct but others like the quote marks are preferential.
I'll fix that in my changes but don't worry about linting errors as much .
Ingneral they are for code you touch (imo) not generated code. You can
either exclude the project from linting or simply compile and include as a
dependency from a repo

HermenOtter:
@kenisteward Thanks for your reply, did you read the second paragraph of my answer? It seems like the typings are broken, while it should not be possible, did you experience the same problem?

kenisteward:
@HermenOtter I did and I'm wondering what's going on.  The only reason HttpResponse<Subject> would be returned is if Subject is not applied as a response type on the return.  The template that generates the returntype is this:
```
{{#returnType}}{{{returnType}}}{{#isResponseTypeFile}}|undefined{{/isResponseTypeFile}}{{/returnType}}{{^returnType}}any{{/returnType}}
```

which would mean that returnType is not set for that operation.  

Could i get a peak at the endpoint yaml you are using to confirm whether my hypothesis is correct?

kenisteward:
@HermenOtter I also have a sneaking suspicion I've introduced a bug with the return type for file based requests.  Do you have any endpoints that return files by any chance?

HermenOtter:
@kenisteward  I remember it generating the method overloads with Subject as type, I am confused. When I change the types to <Subject> it actually works. I do not have endpoints that return files, so I can't help you with that. I have sent you the yaml of the endpoint.

kenisteward:
@hermenotter what do you mean when you change the types it works? You have an example?

HermenOtter:
@kenisteward In a previous version it seemed to have worked, but it did not causing my confusing. Could it be an allof is causing my problem:
```yaml
responses:
        200:
          description: "A subject"
          headers:
            ETag:
              type: string
          schema:
            type: "object"
            allOf:
            - $ref: "#/definitions/Subject"
            - type: "object"
              properties:
                links:
                  $ref: 'https://api.swaggerhub.com/...
```

HermenOtter:
The allOf problem seems to be an issue outside the scope of this PR. I think that if @kenisteward can confirm the returning of a file in an endpoint works, the PR can be merged.

wing328:
I'll merge it tomorrow (Thur) if no one has further feedback or question on this PR.

kenisteward:
@wing328 thanks!

HermenOtter:
Thank you @wing328!

kenisteward:
@HermenOtter As a note, the master branch currently only supports features for OAS 2.0.  allOf, oneOf, anyOf, not are all OAS 3.0 features.  That should be implemented on the 3.0.0 branch if we were to support it.

PR #6959, opened 2017-11-14
[CppRest] Support optional parameters

fvarose:
### PR checklist

### Description of the PR

This PR adds support for optional parameters (those marked as `required: false` in a swagger spec) using `boost::optional`.
Users of the client code can now pass `boost::none` to API calls for any optional parameters they are not interested in using.

Notes:
- `in: path` params are not covered because they cannot be optional (see [OpenAPI spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameterObject)).
- `in: body` params are also not covered, as the implementation would be less trivial. The current behaviour for those params doesn't change, i.e. they are always required to make a call.
- This doesn't add any new linking dependencies as `boost::optional` is header-only, and `boost` itself is already required to compile the generated client code.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/cpprest/api-header.mustache
+++ b/modules/swagger-codegen/src/main/resources/cpprest/api-header.mustache
@@ -14,6 +14,8 @@
 {{#imports}}{{{import}}}
 {{/imports}}
 
+#include <boost/optional.hpp>
+
 {{#apiNamespaceDeclarations}}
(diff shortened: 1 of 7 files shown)


fvarose:
cc @stkrwork @ravinikam

wing328:
@fvarose thanks for the enhancemnet.

@ravinikam thanks for reviewing the change.

Issue #6907, opened 2017-11-07
Test endpoint returns invalid data


tomplus:
##### Description

http://petstore.swagger.io which is used by integration tests, returns invalid data, for example pets without name (name is required). There are a lot of such records.

##### Command line used for generation

```
curl -X GET "http://petstore.swagger.io/v2/pet/-4227537848496351315" -H  "accept: application/xml"

<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
  <Pet>
    <id>-4227537848496351315</id>
    <photoUrls/>
    <status>sold</status>
    <tags/>
  </Pet>
```

##### Suggest a fix/enhancement

Can I delete invalid records ? Is it possible to add such records by API ? Can we reset data periodically ? 

Thanks.

wing328:
@tomplus yes, the public petstore server contains invalid data as anyone can insert invalid records since there's no validation.

The CIs run a local copy of the Petstore server to avoid issues due to bad data.

tomplus:
@wing328 Thanks, it's good to know. I see that some Python tests use public API, I'll fix it.

tomplus:
Ok, I got it, CIs have /etc/hosts overwritten https://github.com/swagger-api/swagger-codegen/blob/master/.travis.yml#L31 and tests can have public server configured.

PR #6968, opened 2017-11-15
[python] fix tests, tornado ssl fix

tomplus:
### PR checklist

### Description of the PR

Fixes related to:

1. Travis doesn't know that tests fail because an exit codes from bash scripts are always zero - fixed
2. Tornado client doesn't work in python < 2.7.9 - fixed
3. Some tests are skipped because public PetStore endpoint has incorrect data. Travis use local instance of PetStore server and all test can run. I removed `@skip` and also force endpoint name to `localhost`. IMO it'll remind developers about starting a local instance. #6907 

Please take a look: @wing328 @taxpon @frol @mbohlool @cbornet @kenjones-cisco

### DIFF
--- a/modules/swagger-codegen/src/main/resources/python/tornado/rest.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/tornado/rest.mustache
@@ -6,14 +6,7 @@ import io
 import json
 import logging
 import re
-
 import ssl
-try:
-    from ssl import create_default_context
-except ImportError:
-    # backward capability create_default_context is on Python 3.4+, 2.7.9+
-    def create_default_context():
-        return ssl.SSLContext(ssl.PROTOCOL_SSLv23)
 
 import certifi
 # python 2 and python 3 compatibility library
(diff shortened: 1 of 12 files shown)


frol:
> Travis doesn't know that tests fail because an exit codes from bash scripts are always zero - fixed

Does this mean that flake8 failure will be also ignored? Should we use `set -e` or run the scripts with explicit `bash -e ./test_python2_and_3.sh`?

tomplus:
Travis sees an exit code from a script but it's exit code from last command. So now we have situation where python test failed but flake8 "covers" it and returns 0. Now I exit immediately if python tests failed. flake8 is a last command so its code will be interpreted correctly. If flake return non-zero code, Travis will notice it.

I consider using `set -e` but I think it would be too big change in this quite simple case.

tomplus:
> Tornado client doesn't work in python < 2.7.9 - fixed

Finally I had to drop compatibility with older Python. It's more complicated and I'll fix it in next PR.

frol:
Looks good to me!

wing328:
cc @toumorokoshi as well

PR #6985, opened 2017-11-16
Match rust integer type to minimum and maximum constraints

fralalonde:
@frol @farcaller 

In rust-server, use rust integer types best matching the `format`, `minimum` and `maximum` schema constraints.

Matching rules cover all rust standard int size, from 8 to 64bits, signed and unsigned, including `size` types.

- `minimum` is used to determine if the int type should be signed. Default is signed.
- `maximum` is used to determine the int size. Default is `isize` or `usize`.
- Standard `formats` `int32` and `int64` can be used to force int size.
- Existing custom formats `uint32` and `uint64` have been preserved, even though they are superseded by new code.
- `minimum` and `maximum` constraints also respects their `exclusive` flags if specified.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RustServerCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RustServerCodegen.java
@@ -1,13 +1,6 @@
 package io.swagger.codegen.languages;
 
-import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.JsonSerializer;
-import com.fasterxml.jackson.databind.SerializerProvider;
-import com.fasterxml.jackson.databind.module.SimpleModule;
-import com.google.common.collect.ArrayListMultimap;
-import com.google.common.collect.Lists;
-import com.google.common.collect.Multimap;
 import io.swagger.codegen.*;
 import io.swagger.models.*;
 import io.swagger.models.parameters.BodyParameter;
@@ -21,8 +14,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
-import java.io.IOException;
-import java.math.BigDecimal;
 import java.util.*;
 import java.util.Map.Entry;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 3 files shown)


jaytiar:
The logic as you've described it (assuming I've understood) doesn't seem to cope with cases where the valid range is -(something large) to +(something small). You'll correctly pick a signed quantity, but the size will only be big enough for the positive max, not the negative max. I think you need to use the maximum and minimum to determine the size needed.

fralalonde:
You are right. I changed the calculation to use `Long.numberOfLeadingZeros` to determine the required int size (acounting signedness), which should cover that case.

As a sensible default, I made `usize` and `isize` be used when the minimum is unspecified or requires less than 32 bits to be handled.

I also added a test for the min / max constraints to int sizing logic, you can look at it, it makes the choices clearer.

wing328:

fralalonde:
Thank you for noticing this. I fixed my local git config for the future, amended the commit _and_ added my secondary email to my github account for good measure.

fralalonde:
Although the commit is now to my primary email, I noticed that my avatar picture for the commit is small, and the generic octocat is still shown behind. Let me know if this is normal or if it's something I should fix.

modules/swagger-codegen/src/test/java/io/swagger/codegen/languages/RustServerCodegenTest.java
```diff
@@ -0,0 +1,48 @@
...
+        assertEquals(RustServerCodegen.matchingIntType(false, null, 127L), "i8");
+        assertEquals(RustServerCodegen.matchingIntType(false, null, 128L), "i16");
+        assertEquals(RustServerCodegen.matchingIntType(false, null, (long) Short.MAX_VALUE), "i16");
+        assertEquals(RustServerCodegen.matchingIntType(false, null, (long) Short.MAX_VALUE + 1), "i32");
+        assertEquals(RustServerCodegen.matchingIntType(false, null, (long) Integer.MAX_VALUE), "i32");
+        assertEquals(RustServerCodegen.matchingIntType(false, null, (long) Integer.MAX_VALUE + 1), "i64");
+        assertEquals(RustServerCodegen.matchingIntType(false, null, Long.MAX_VALUE), "i64");
+    }
+
+}
```
bjgill: Nice

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RustServerCodegen.java
```diff
@@ -269,7 +262,7 @@ public String toApiName(String name) {
 
     /**
      * Escapes a reserved word as defined in the `reservedWords` array. Handle escaping
-     * those terms here.  This logic is only called if a variable matches the reserved words
+     * those terms here.  This logic is only called if a variable matches the reseved words
```
bjgill: `reserved`?

modules/swagger-codegen/src/test/java/io/swagger/codegen/languages/RustServerCodegenTest.java
fralalonde: Thanks

wing328:
@fralalonde thanks for the PR, which has been merged into master.

@bjgill thanks for reviewing the change.

PR #6994, opened 2017-11-19
[TypeScript-Fetch] Support for non-JSON body params

### LINKED ISSUE #6940: [Typescript-Fetch] Routes with non-JSON body parameters still get sent JSON
##### Description

In the TypeScript-Fetch client generator, payloads sent to any routes with body parameters are unconditionally tagged with a `Content-Type` of `application/json` and `JSON.stringify`-ed before being sent. Routes that expect content such as tar archives are broken by this behavior.

##### Swagger-codegen version

2.3.0-snapshot

https://raw.githubusercontent.com/moby/moby/master/api/swagger.yaml

See for example the `PutContainerArchive` operation route, which `consumes: ["application/x-tar", "application/octet-stream"]`.

```
generate -i /moby/api/swagger.yaml -l typescript-fetch -o /package
```

##### Steps to reproduce

Generate the client with the linked spec (or use the `docker-client` package from npm, which is generated in this way), inspect the code for `putContainerArchive` and observe that it adds a `Content-Type` `application/json` header and invokes `JSON.stringify` on the content. If you actually try this against a Docker instance you'll get an invalid tar archive error.

The codegen should look at the route's `consumes` attribute to figure out the content type header, and if it isn't JSON, send it through without JSON stringify-ing.

masaeedu:
### PR checklist

### Description of the PR
This change adds support for non-"application/json" MIME types in a body parameter. The logic has some hairy bits I'm not entirely happy with, but I'm trying to follow what the other language generators do:

 - Instead of always emitting a default `"Content-Type"` of `"application/json"`, emit the first MIME-type from the operation's `consumes` array if available, and fall back to `"application/json"` if this is not present.

   This is a little ugly; if the user needs to use a MIME-type from the `consumes` array that isn't the first one, they need to know to pass it in `options.headers["Content-Type"]`, which isn't very visible. I'd like to make the content type an explicit, required parameter on operations where there's more than one  entry in the `consumes` array, but for now I've just followed what I've seen most of the other language clients. 

 - In cases where the body parameter has datatype `string` and the content type isn't `application/json`, simply set the request body to the parameter directly. Otherwise, `JSON.stringify` the parameter as before. This doesn't take into consideration other any other possible `dataType` values that may represent raw body streams; I am unaware of any such datatypes but if you know of any please point them out.

Hopefully some of the reviewers can provide feedback on these points.

Fixes #6940.

### Technical committee reviewers

### DIFF
--- a/modules/swagger-codegen/src/main/resources/typescript-fetch/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-fetch/api.mustache
@@ -212,7 +212,12 @@ export const {{classname}}FetchParamCreator = function (configuration?: Configur
 
     {{/hasFormParams}}
     {{#bodyParam}}
+            {{^consumes}}
             localVarHeaderParameter['Content-Type'] = 'application/json';
+            {{/consumes}}
+            {{#consumes.0}}
+            localVarHeaderParameter['Content-Type'] = '{{mediaType}}';
+            {{/consumes.0}}
 
     {{/bodyParam}}
             localVarUrlObj.query = Object.assign({}, localVarUrlObj.query, localVarQueryParameter, options.query);
@@ -223,7 +228,8 @@ export const {{classname}}FetchParamCreator = function (configuration?: Configur
             localVarRequestOptions.body = localVarFormParams.toString();
     {{/hasFormParams}}
     {{#bodyParam}}
-            localVarRequestOptions.body = JSON.stringify({{paramName}} || {});
+            const needsSerialization = (<any>"{{dataType}}" !== "string") || localVarRequestOptions.headers['Content-Type'] === 'application/json';
+            localVarRequestOptions.body =  needsSerialization ? JSON.stringify({{paramName}} || {}) : ({{paramName}} || "");
     {{/bodyParam}}
 
             return {
--- a/samples/client/petstore/typescript-fetch/builds/default/api.ts
+++ b/samples/client/petstore/typescript-fetch/builds/default/api.ts
@@ -360,7 +360,8 @@ export const PetApiFetchParamCreator = function (configuration?: Configuration) 
             // fix override query string Detail: https://stackoverflow.com/a/7517673/1077943
             delete localVarUrlObj.search;
             localVarRequestOptions.headers = Object.assign({}, localVarHeaderParameter, options.headers);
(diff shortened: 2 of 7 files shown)


masaeedu:
Note: all the `Response` changes are unrelated to the PR, running `./run-in-docker.sh mvn clean package && ./bin/typescript-fetch-petstore-all.sh` on a clean clone of master will result in the same changes. I believe the `Response` -> `any` diff was mistakenly introduced in #6739.

modules/swagger-codegen/src/main/resources/typescript-fetch/api.mustache
```diff
@@ -212,7 +212,12 @@ export const {{classname}}FetchParamCreator = function (configuration?: Configur
 
     {{/hasFormParams}}
     {{#bodyParam}}
+            {{^consumes}}
             localVarHeaderParameter['Content-Type'] = 'application/json';
+            {{/consumes}}
+            {{#consumes.0}}
+            localVarHeaderParameter['Content-Type'] = '{{mediaType}}';
```
wing328: @masaeedu Please use {{{mediaType}}} to use the original value fo the media type as I remember there are edge cases causing issues with the HTML-escaped values.

masaeedu: @wing328 Good catch. Should we be doing this for things like `{{className}}`, `{{paramName}}` etc. as well?

wing328:
Yes please.

2017年11月27日 下午9:49，"Asad Saeeduddin" <notifications@github.com>寫道：

modules/swagger-codegen/src/main/resources/typescript-fetch/api.mustache
wing328: @masaeedu yes please

wing328: FYI. I've filed https://github.com/swagger-api/swagger-codegen/pull/7073 to clean up {{mediaType}} in other templates.

wing328:
@masaeedu PR merged into master. I'll fix the {{mediaType}} later.

wing328:
@masaeedu fixed via #7210

PR #7031, opened 2017-11-22
[Python] Resolve ref object before other primitive types for model

### LINKED ISSUE #7030: [Python] Cannot resolve model correctly when $ref file name collide with primitive data type name
##### Description

When I try to define an object named array.yaml and reference it in another object, the reference won't be correctly resolved. Swagger-code-gen would think I'm referencing just an native array, thus it'll be translated to list type for python instead of my own `array` class. 

##### Swagger-codegen version
2.2.3

Below is a sample yaml
```yaml
type: object
allOf:
  - $ref: ./models/object-response.yaml
  - properties:
      items:
         type: array
         items:
           $ref: './models/array.yaml'
         description: a list of array objects
```
In the generated python code for this class, it has below field,

```python
    swagger_types = {
        'pagination_info': 'PaginationInfo',
        'items': 'list[list]'
    }
```
which I would like it to be 

```python
    swagger_types = {
        'pagination_info': 'PaginationInfo',
        'items': 'list[array]'
    }
```
while `array` is the object I defined in array.yaml.

The reason for this to happen is that when parsing the yaml spec, the `$ref: './models/array.yaml'` line will be parsed as `#/definitions/array`, which to me indicate a user defined object instead of primitive type like `array`, so I think code-gen should be able to resolve it correctly. One fix I can think of is to change `io.swagger.codegen.languages.PythonClientCodegen#getTypeDeclaration` to resolve ref type property before primitive properties. If this sounds ok I can work on a pull request.

zhengl7:
### PR checklist

- [] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

This change allows ref file to be named as primitive data types like array, and will be resolved correctly.

Fix #7030

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PythonClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PythonClientCodegen.java
@@ -356,7 +356,22 @@ public class PythonClientCodegen extends DefaultCodegen implements CodegenConfig
 
     @Override
     public String getTypeDeclaration(Property p) {
-        if (p instanceof ArrayProperty) {
+        if (p instanceof RefProperty) {
+            String datatype;
+            try {
+                RefProperty r = (RefProperty) p;
+                datatype = r.get$ref();
+                if (datatype.indexOf("#/definitions/") == 0) {
+                    datatype = datatype.substring("#/definitions/".length());
+                    return toModelName(datatype);
+                }
+            } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


mbohlool:
The change looks good to me, but I wonder why nobody else replied here for 21 days? (I didn't because I was on vacation :)). I would wait for at least one more LGTM before merge this.

zhengl7:
Hi, @taxpon (2017/07) @frol (2017/07) @mbohlool (2017/07) @cbornet (2017/09) @kenjones-cisco (2017/11), sorry about the ping after long time, but is this good to be merged?

Issue #6806, opened 2017-10-25
Enum definition of type integer generates invalid Java code


andrewharrison135:
##### Description
The output of the following swagger generates invalid Java code. 
The code that is generated by default with the Java language uses Google GSON JsonReader for parsing JSON objects. The generated code uses 
```java
jsonReader.nextInteger();
```
 for the read method but it should be 
```java
jsonReader.nextInt(); 
```
as per the GSON docs: https://google.github.io/gson/apidocs/com/google/gson/stream/JsonReader.html#nextInt--

I'm not sure what the actual issue is, it could be an issue with a old GSON version. 
Any help would be appreciated

##### Swagger-codegen version
2.3.3

##### Swagger declaration file content or url
Swagger doc:
```yaml
swagger: '2.0'
paths: {}
definitions:
  suit:
    type: integer
    description: A description of suit types.
    format: int32
    x-enumNames:
      - hearts
      - spades
      - clubs
      - diamonds
    enum:
      - 0
      - 1
      - 2
      - 3
```

Generated Java Code:
```java
public enum Suit {
  
  NUMBER_0(0),
  NUMBER_1(1),
  NUMBER_2(2),
  NUMBER_3(3);

  private Integer value;

  Suit(Integer value) {
    this.value = value;
  }

  public Integer getValue() {
    return value;
  }

  ......

  public static class Adapter extends TypeAdapter<Suit> {
    @Override
    public void write(final JsonWriter jsonWriter, final Suit enumeration) throws IOException {
      jsonWriter.value(enumeration.getValue());
    }

    @Override
    public Suit read(final JsonReader jsonReader) throws IOException {
      Integer value = jsonReader.nextInteger();
      return Suit.fromValue(String.valueOf(value));
    }
  }
}
```

##### Command line used for generation
java -jar swagger-codegen-cli-2.2.3.jar generate -i test.yaml -l java

Using:
gson-2.8.1-sources.jar

craffael:
I can confirm this. I've noticed that the corresponding `src/main/java/res/Java/modelEnum.mustache` already accounts for this: Line 63 in this file reads as

    {{{dataType}}} value = jsonReader.{{#isInteger}}nextInt(){{/isInteger}}{{^isInteger}}next{{{dataType}}}(){{/isInteger}};

This means that the template already makes a distinction between Integer and non-integer data types. However, the problem here is the variable `isInteger` is not part of the `CodegenModel.java`.

[...]

PR #7051, opened 2017-11-24
[Java] Fix error for integer enum model.

craffael:
(cherry picked from commit 07c328aab4af4e4337871b644cf7b9bf7f3c22f0)

### PR checklist

- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language. @bbdouglas, @JFCote, @jfiala @lukoyanov @cbornet 

### Description of the PR

Fixes Issue #6806 .

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenModel.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenModel.java
@@ -39,7 +39,7 @@ public class CodegenModel {
     public Set<String> allMandatory;
 
     public Set<String> imports = new TreeSet<String>();
-    public boolean hasVars, emptyVars, hasMoreModels, hasEnums, isEnum, hasRequired, hasOptional, isArrayModel, hasChildren;
+    public boolean hasVars, emptyVars, hasMoreModels, hasEnums, isEnum, hasRequired, hasOptional, isArrayModel, hasChildren, isInteger;
     public boolean hasOnlyReadOnly = true; // true if all properties are read-only
     public ExternalDocs externalDocs;
 
@@ -131,6 +131,8 @@ public class CodegenModel {
             return false;
         if (isEnum != that.isEnum)
             return false;
+        if(isInteger != that.isInteger)
+            return false;
         if (externalDocs != null ? !externalDocs.equals(that.externalDocs) : that.externalDocs != null)
             return false;
         if (!Objects.equals(hasOnlyReadOnly, that.hasOnlyReadOnly))
@@ -177,6 +179,7 @@ public class CodegenModel {
         result = 31 * result + (hasMoreModels ? 13:31);
         result = 31 * result + (hasEnums ? 13:31);
         result = 31 * result + (isEnum ? 13:31);
+        result = 31 * result + (isInteger ? 13:31);
         result = 31 * result + (externalDocs != null ? externalDocs.hashCode() : 0);
         result = 31 * result + (vendorExtensions != null ? vendorExtensions.hashCode() : 0);
         result = 31 * result + Objects.hash(hasOnlyReadOnly);
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -1438,6 +1438,9 @@ public class DefaultCodegen {
             if (impl.getType() != null) {
                 Property p = PropertyBuilder.build(impl.getType(), impl.getFormat(), null);
                 m.dataType = getSwaggerType(p);
+                if(m.dataType.equals("Integer")) {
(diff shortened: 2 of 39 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenModel.java
```diff
@@ -39,7 +39,7 @@
     public Set<String> allMandatory;
 
     public Set<String> imports = new TreeSet<String>();
-    public boolean hasVars, emptyVars, hasMoreModels, hasEnums, isEnum, hasRequired, hasOptional, isArrayModel, hasChildren;
+    public boolean hasVars, emptyVars, hasMoreModels, hasEnums, isEnum, hasRequired, hasOptional, isArrayModel, hasChildren, isInteger;
```
JFCote: It doesn't make sense to have `isInteger` in the `CodegenModel`. What about float? double? long? boolean?

`isInteger` is part of `CodegenProperty` and `CodegenParameter` and this is where it belong IMHO.

I didn't check the rest of the commit because I think this is a major problem.

craffael: I agree with you that my implementation is not optimal. As you say, the problem is really that most `CodegenProperty`s can appear as a `CodegenModel` by redefining inline properties as definitions in the swagger file (cf. [Swagger Spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject) ). I can think of two solutions that would be a bit nicer but also a lot more work:
1) Factor most fields (e.g. isInteger, isDouble, minimum, minLength,...) of `CodegenProperty` into a base class, `BASE_CLASS`, and make `CodegenModel` inherit from `BASE_CLASS`.
2) Instead of making `CodegenModel` inherit from `BASE_CLASS`, add a field to `CodegenModel` of type `BASE_CLASS` that contains the `isInteger`, `isLong`, etc. attributes. This field would be mostly of importance if `CodegenModel.isEnum==true` and could be zero otherwise.

JFCote:
@craffael : I did not validate everything below but I'm pretty sure it will help you figure it out.

If you check in the JavaPlayFramework generator, you will see that they use "isInteger" in the enum declaration. See here:

The model mustache: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/JavaPlayFramework/model.mustache#L14

The enumOuterClass: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/JavaPlayFramework/enumOuterClass.mustache#L9

The enumClass: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/JavaPlayFramework/enumClass.mustache#L5

As you can see, they use variable like `gson`, `allowableValues` and `enumVars` to have access to isInteger, isDouble, etc...

I'm pretty sure you can do the same since it's all already in CodegenModel. In the worst case, maybe you need to add a little bit of code in you generator itself.

I suggest you close your previous PR and propose a new one using this solution instead.

Hope that help!

JFCote:
Like I said in #6806 , I propose that you close this PR and use a more standard solution used in other generator.

Thanks :)

craffael:
@JFCote Thanks very much for your valuable input. I read it Friday evening but didn't have time then to implement it all through. Today I've taken a deep look at how the java-play-framework implementation handles integer enum values but unfortunately that doesn't help much:

As far as I can see it, the `java-play-framework` generator does not support the GSON serializer and uses jackson instead. I know that some of the templates (e.g. https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/JavaPlayFramework/enumClass.mustache#L5) distinguish between GSON and Jackson serializers, but running `java -jar swagger-codegen.jar config-help -l java-play-framework` did not reveal any command line switch that would enable the GSON serializer. As far as I can see it, the only (unsupported) way to enable GSON with the play framework would be to use `--additional-properties gson=true`. But this produces invalid code. For the swagger example of issue #6806 , the generated java code of the enum model looks as follows:

```java
public enum Suit {
  
  @SerializedName(0)
  NUMBER_0(0),
  
  @SerializedName(1)
  NUMBER_1(1),
  
  @SerializedName(2)
  NUMBER_2(2),
  
  @SerializedName(3)
  NUMBER_3(3);
```

this is clearly invalid as `@SerializedName` expects a string and not an integer as argument. (i.e. the `isInteger` variable that is used in the mustache template https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/JavaPlayFramework/enumOuterClass.mustache#L9 is actually not present, even though it's an integer.

craffael:
I have a much nicer fix implemented in PR https://github.com/swagger-api/swagger-codegen/pull/7059

Issue #6892, opened 2017-11-06
[Kotlin] select date library


magiepooh:
##### Description

select date library like Java Generator.

##### Swagger-codegen version

2.2.3+

##### Related issues/PRs

https://github.com/swagger-api/swagger-codegen/issues/5730

##### Suggest a fix/enhancement

`generate -i swagger.yaml -l kotlin  -o out -DdateLibrary=threetenbp`

wing328:
@magiepooh thanks for the feedback. May I know if you've time to contribute the enhancement?

magiepooh:
I'll work on this!

magiepooh:
Related PR: https://github.com/swagger-api/swagger-codegen/pull/7516

wing328:
Thanks again for the PR. Closing this.

PR #7054, opened 2017-11-25
[kotlin] support selection of datelibrary

magiepooh:
### PR checklist

### Description of the PR

support selection of date library. (`java8`, `threetenbp`, `string`)

```bash
# no option
ags="generate -t modules/swagger-codegen/src/main/resources/kotlin-client -i modules/swagger-codegen/src/test/resources/2_0/petstore.yaml -l kotlin --artifact-id kotlin-petstore-client -D -o samples/client/petstore/kotlin $@"

# java8
ags="generate -t modules/swagger-codegen/src/main/resources/kotlin-client -i modules/swagger-codegen/src/test/resources/2_0/petstore.yaml -l kotlin --artifact-id kotlin-petstore-client -D dateLibrary=java8 -o samples/client/petstore/kotlin $@"

# no option
ags="generate -t modules/swagger-codegen/src/main/resources/kotlin-client -i modules/swagger-codegen/src/test/resources/2_0/petstore.yaml -l kotlin --artifact-id kotlin-petstore-client -D dateLibrary=string -o samples/client/petstore/kotlin $@"

# no option
ags="generate -t modules/swagger-codegen/src/main/resources/kotlin-client -i modules/swagger-codegen/src/test/resources/2_0/petstore.yaml -l kotlin --artifact-id kotlin-petstore-client -D dateLibrary=threetenbp -o samples/client/petstore/kotlin $@"
```

related Issue: https://github.com/swagger-api/swagger-codegen/issues/6892

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/KotlinClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/KotlinClientCodegen.java
@@ -16,6 +16,8 @@ import java.util.Map;
 public class KotlinClientCodegen extends DefaultCodegen implements CodegenConfig {
     static Logger LOGGER = LoggerFactory.getLogger(KotlinClientCodegen.class);
 
+    public static final String DATE_LIBRARY = "dateLibrary";
+
     protected String groupId = "io.swagger";
     protected String artifactId = "kotlin-client";
     protected String artifactVersion = "1.0.0";
@@ -24,6 +26,20 @@ public class KotlinClientCodegen extends DefaultCodegen implements CodegenConfig
     protected String apiDocPath = "docs/";
     protected String modelDocPath = "docs/";
     protected CodegenConstants.ENUM_PROPERTY_NAMING_TYPE enumPropertyNaming = CodegenConstants.ENUM_PROPERTY_NAMING_TYPE.camelCase;
+
+    protected String dateLibrary = DateLibrary.JAVA8.value;
+
+    public enum DateLibrary {
+        STRING("string"),
+        THREETENBP("threetenbp"),
+        JAVA8("java8");
+
+        public final String value;
+
+        DateLibrary(String value) {
+            this.value = value;
+        }
+    }
 
     /**
      * Constructs an instance of `KotlinClientCodegen`.
@@ -181,6 +197,13 @@ public class KotlinClientCodegen extends DefaultCodegen implements CodegenConfig
         cliOptions.add(new CliOption(CodegenConstants.GROUP_ID, "Client package's organization (i.e. maven groupId).").defaultValue(groupId));
         cliOptions.add(new CliOption(CodegenConstants.ARTIFACT_ID, "Client artifact id (name of generated jar).").defaultValue(artifactId));
         cliOptions.add(new CliOption(CodegenConstants.ARTIFACT_VERSION, "Client package version.").defaultValue(artifactVersion));
+        CliOption dateLibrary = new CliOption(DATE_LIBRARY, "Option. Date library to use");
+        Map<String, String> dateOptions = new HashMap<>();
+        dateOptions.put(DateLibrary.THREETENBP.value, "Threetenbp");
+        dateOptions.put(DateLibrary.STRING.value, "String");
+        dateOptions.put(DateLibrary.JAVA8.value, "Java 8 native JSR310");
+        dateLibrary.setEnum(dateOptions);
+        cliOptions.add(dateLibrary);
 
         CliOption enumPropertyNamingOpt = new CliOption(CodegenConstants.ENUM_PROPERTY_NAMING, CodegenConstants.ENUM_PROPERTY_NAMING_DESC);
         cliOptions.add(enumPropertyNamingOpt.defaultValue(enumPropertyNaming.name()));
@@ -218,6 +241,10 @@ public class KotlinClientCodegen extends DefaultCodegen implements CodegenConfig
         this.sourceFolder = sourceFolder;
     }
 
+    public void setDateLibrary(String library) {
+        this.dateLibrary = library;
+    }
+
     @Override
     public void processOpts() {
         super.processOpts();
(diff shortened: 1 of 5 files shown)


wing328:
@magiepooh thanks for the PR 👍 

cc @jimschubert @cbornet

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/KotlinClientCodegen.java
```diff
@@ -264,6 +291,26 @@ public void processOpts() {
...
+            importMapping.put("LocalDate", "org.threeten.bp.LocalDate");
+            importMapping.put("LocalDateTime", "org.threeten.bp.LocalDateTime");
+        } else if (DateLibrary.STRING.value.equals(dateLibrary)) {
+            additionalProperties.put(DateLibrary.STRING.value, "true");
+            typeMapping.put("date-time", "kotlin.String");
+            typeMapping.put("date", "kotlin.String");
+            typeMapping.put("Date", "kotlin.String");
+            typeMapping.put("DateTime", "kotlin.String");
+        } else if (DateLibrary.JAVA8.value.equals(dateLibrary)) {
+            additionalProperties.put(DateLibrary.JAVA8.value, "true");
```
jimschubert: `additionalProperties` should be `true` here rather than `"true"` if it's intended to be used for switching in templates.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/KotlinClientCodegen.java
```diff
@@ -264,6 +291,26 @@ public void processOpts() {
             LOGGER.warn(CodegenConstants.INVOKER_PACKAGE + " with " + this.getName() + " generator is ignored. Use " + CodegenConstants.PACKAGE_NAME + ".");
         }
 
+        if (additionalProperties.containsKey(DATE_LIBRARY)) {
+            setDateLibrary(additionalProperties.get(DATE_LIBRARY).toString());
+        }
+
+        if (DateLibrary.THREETENBP.value.equals(dateLibrary)) {
+            additionalProperties.put(DateLibrary.THREETENBP.value, "true");
```
jimschubert: additionalProperties should be true here rather than "true" if it's intended to be used for switching in templates.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/KotlinClientCodegen.java
```diff
@@ -264,6 +291,26 @@ public void processOpts() {
...
+            typeMapping.put("date", "LocalDate");
+            typeMapping.put("DateTime", "LocalDateTime");
+            importMapping.put("LocalDate", "org.threeten.bp.LocalDate");
+            importMapping.put("LocalDateTime", "org.threeten.bp.LocalDateTime");
+        } else if (DateLibrary.STRING.value.equals(dateLibrary)) {
+            additionalProperties.put(DateLibrary.STRING.value, "true");
```
jimschubert: I think having just `string` as an additional property would be confusing.

For instance, in a template this would lead to:

```
{{#string}}string date logic{{/string}}{{^string}}other logic{{/string}}
```

This seems like an unlikely use case, until it starts getting used and we can't change it without a breaking change. Also, I think this may cause confusion on some property/params which may have `string` properties as well.

additionalProperties should be true here rather than "true" if it's intended to be used for switching in templates.

Any objection to using `isStringDate` or something? iirc, Mustache will consider the string "true" for properties prefixed with `is` as a boolean. I could be wrong, but I prefer to set booleans explicitly. At one point, we were able to set them as strings but I don't know if that code was changed or if the Mustache dependency was updated but as of a few weeks ago they had to be explicit booleans.

jimschubert:
@magiepooh I don't know if you saw my comments? If you'd like, I can make the changes and open a PR against your fork.

tfujita-folio:
Thank you for reviewing! (I was not aware.)

> additionalProperties should be true here rather than "true" if it's intended to be used for switching in templates.

I'll fix it.

> https://github.com/swagger-api/swagger-codegen/pull/7054#discussion_r153106067

I think that selection regarding dates should be managed with a unified option.

`isStringDate` option and `dateLibrary` option may confuse the user. What do you think?

magiepooh:
Thank you for reviewing! (I was not aware.)

> additionalProperties should be true here rather than "true" if it's intended to be used for switching in templates.

I'll fix it.

> #7054 (comment)

I think that selection regarding dates should be managed with a unified option.

Multiple option (isStringDate option and dateLibrary option ) may confuse the user. What do you think?

jimschubert:
@maggiepooh I thought about it a little more. There's a `isString` and `isDate` on vars and params. Which of these would be true in this situation? I think it would be equally as confusing (maybe moreso) to have isString=false, isDate=true, and also have `string=true` available from additional properties. I think whatever is chosen will be confusing to consumers, mostly because `String` isn't a library for handling dates.

Maybe a better approach would be to set a private flag if the additional property is set (and remove it from additional properties to avoid any confusion), and post process variables and parameters to change from dates to strings. After all, that's what this option intends to do, correct? This way, it wouldn't interfere with templates that have `isDate` blocks.

I haven't looked at other generators to see if/how they handle this. @wing328 is there a common approach to forcing date/date-time to strings?

wing328:
For PHP, we use our own serializer to handle the datetime to string conversion. For Java, we rely on 3rd-party libraries.

For Kotlin, can we follow Java's approach? (just throwing a suggestion but I've not looked into the Kotlin client code)

magiepooh:
> I thought about it a little more. There's a isString and isDate on vars and params. Which of these would be true in this situation? I think it would be equally as confusing (maybe moreso) to have isString=false, isDate=true, and also have string=true available from additional properties. I think whatever is chosen will be confusing to consumers, mostly because String isn't a library for handling dates.

I agree. 

@jimschubert Should I fix it as follows?
`additionalProperties.put(DateLibrary.STRING.value, "true");`
↓
`additionalProperties.put("isStringDate", "true");`

magiepooh:
I added additionalProperties for retrofit (HTTP client for Android).
However, now I thought that the string property is not necessary when creating a Retrofit adapter. So, I removed the string property.

wing328:
@magiepooh CircleCI reports the following issue:
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileTestKotlin'.
> Could not resolve all dependencies for configuration ':testCompile'.
   > Could not resolve org.hamcrest:hamcrest-core:1.3.
     Required by:
         io.swagger:kotlin-petstore-client:1.0.0 > io.kotlintest:kotlintest:2.0.2 > com.sksamuel.koors:koors:0.90.0 > junit:junit:4.12
      > Could not HEAD 'http://repo1.maven.org/maven2/org/hamcrest/hamcrest-core/1.3/hamcrest-core-1.3.pom'.
         > Connection reset

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED
```

I wonder if you can take a look when you've time.

magiepooh:
@wing328 

Can you help me? I don't know why CI failure.
My change is only `additionalProperties`. ( https://github.com/swagger-api/swagger-codegen/compare/f39dc67...8d0f661 )

wing328:
I did a test with the option `threetenbp` but looks like there are some issues:
- build.gradle does not have the following to import the module:
```
compile group: 'org.threeten', name: 'threetenbp', version: '1.3.6'
```
- the following import statement is incorrect (in Order.kt)
```
import io.swagger.client.models.org.threeten.bp.LocalDateTime
```

I think I've solutions to these problems and will squeeze some time this weekend to work on it.

boylenssen:
Why is legacy not supported here? For Java, this sets it to `Date`. I would like this option for Kotlin too.

I used the snapshot of 2.4.0 (feb 10) and threetenbp works fine there. Legacy gives me Java8 datetime

PR #7056, opened 2017-11-25
adding fluent api options to spring java codegen

UltimateDogg:
this adds two boolean flags simpleFluentPattern and
collectionFluentPattern for the spring language generator. Both default to true which is the current
behavior today. This is probably applicable to anything that is java related but there seems to be a lot of disparate classes that deal with Java. feedback is welcome.

I added these flags as some versions of jackson has issues with multiple setters that take the same input. 

@bbdouglas  @JFCote @sreeshas  @jfiala @lukoyanov  @cbornet

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SpringCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/SpringCodegen.java
@@ -33,6 +33,8 @@ public class SpringCodegen extends AbstractJavaCodegen
     public static final String SPRING_CLOUD_LIBRARY = "spring-cloud";
     public static final String IMPLICIT_HEADERS = "implicitHeaders";
     public static final String SWAGGER_DOCKET_CONFIG = "swaggerDocketConfig";
+    public static final String COLLECTION_FLUENT_PATTERN = "collectionFluentPattern";
+    public static final String SIMPLE_FLUENT_PATTERN = "simpleFluentPattern";
 
     protected String title = "swagger-petstore";
     protected String configPackage = "io.swagger.configuration";
@@ -45,6 +47,8 @@ public class SpringCodegen extends AbstractJavaCodegen
     protected String responseWrapper = "";
     protected boolean useTags = false;
     protected boolean useBeanValidation = true;
+    protected boolean simpleFluentPattern = true;
+    protected boolean collectionFluentPattern = true;
     protected boolean implicitHeaders = false;
     protected boolean swaggerDocketConfig = false;
     protected boolean useOptional = false;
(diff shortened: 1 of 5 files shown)


cbornet:
@UltimateDogg can you detail the problem you have with Jackson ?

UltimateDogg:
Sure @cbornet. Error was com.fasterxml.jackson.databind.JsonMappingException: Can not construct instance of x.x.x.x.L2SwitchInterface: no String-argument constructor/factory method to deserialize from String value ('WAN1') the error seemed to be coming from the fact that there is a field like

  @JsonProperty("name")
  @JacksonXmlProperty(localName = "name")
  private String name = null;

and getter and setters of 

  public L2SwitchInterface name(String name) {
    this.name = name;
    return this;
  }

  public String getName() {
    return name;
  }

  public void setName(String name) {
    this.name = name;
  }


since there are two setters it seems that jackson is having issues picking which one is the setter. removing the one helped.

cbornet:
I don’t understand : there’s a mix of name and L2SwitchInterfaceName. Can you provide your swagger spec and the command line used for generation ?

UltimateDogg:
here is the stripped down area that caused issue with version jackson 2.8.5, but works ok in latest 2.9
```
package com.github.test.model;

import java.util.Objects;
import com.fasterxml.jackson.annotation.JsonProperty;

/**
 * L2SwitchInterfaces
 */
@javax.annotation.Generated(value = "io.swagger.codegen.languages.SpringCodegen", date = "2017-11-27T15:02:32.772-05:00")
public class L2SwitchInterfaces   {
  @JsonProperty("name")
  private String name = null;

  public L3ucpeApiUniversalcpeftUniversalcpeftL2SwitchInterfaces name(String name) {
    this.name = name;
    return this;
  }

  public String getName() {
    return name;
  }

  public void setName(String name) {
    this.name = name;
  }
}
```
basically when i passed WAN1 in the name field it would fail unless i removed the 
```
  public L3ucpeApiUniversalcpeftUniversalcpeftL2SwitchInterfaces name(String name) {
    this.name = name;
    return this;
  }
```
so i wrote some flags for this

cbornet:
I just tried the Spring Petstore sample and didn't reproduce your issue. The Pet model has a string property called name just as your snippet. So your issue is something else. Could you share a minimal project reproducing your issue on a github repo ?

UltimateDogg:
i will try and see about setting up a github project with the exact details. I know that my swagger is about 8 levels deep so i dont know if that had something to do with it.

UltimateDogg:
i was unable to make a test case. It may be the combination of version / libraries I used or it may have been a transitive issue that was due to environment. The features added do work however so i will leave it up to you if you think there is any value. I can also always reopen if i am able to reproduce in the future.

PR #7061, opened 2017-11-26
[python/tornado] ssl improvements

tomplus:
### PR checklist

### Description of the PR

In my last PR (https://github.com/swagger-api/swagger-codegen/pull/6968) I had to add some hack to disable SSL in older versions of Python. I looked into the implementation and I found that Tornado also didn't support older version of SSL in Python for security reasons (https://github.com/tornadoweb/tornado/pull/2177). Moreover it implements everything we need and also there is a notice about pycurl which doesn't work with SSLContext. Finally I switched to their implementation and it's simple and works correctly with SimpleAsyncHTTPClient and CurlAsyncHTTPClient.

Question: Do we have something to test SSL in CI pipeline ?

Please take a look: @wing328 @taxpon @frol @mbohlool @cbornet @kenjones-cisco @toumorokoshi

### DIFF
--- a/modules/swagger-codegen/src/main/resources/python/tornado/rest.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/tornado/rest.mustache
@@ -6,9 +6,7 @@ import io
 import json
 import logging
 import re
-import ssl
 
-import certifi
 # python 2 and python 3 compatibility library
 import six
 from six.moves.urllib.parse import urlencode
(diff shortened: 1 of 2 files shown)


wing328:
> Question: Do we have something to test SSL in CI pipeline ?

Nope I don't think so.

modules/swagger-codegen/src/main/resources/python/tornado/rest.mustache
```diff
@@ -41,30 +39,10 @@ class RESTClientObject(object):
...
         # maxsize is number of requests to host that are allowed in parallel
-        # ca_certs vs cert_file vs key_file
-        # http://stackoverflow.com/a/23957365/2985775
 
-        # ca_certs
-        if configuration.ssl_ca_cert:
-            ca_certs = configuration.ssl_ca_cert
-        else:
-            # if not set certificate file, use Mozilla's root certificates.
-            ca_certs = certifi.where()
```
wing328: @tomplus shall we keep the default certificate `certifi.where()`?

modules/swagger-codegen/src/main/resources/python/tornado/rest.mustache
```diff
@@ -41,30 +39,10 @@ class RESTClientObject(object):
...
-
-        if hasattr(ssl, 'create_default_context'):
-            # require Python 2.7.9+, 3.4+
-            self.ssl_context = ssl.create_default_context()
```
kenjones-cisco: Does the actual tornado client handle creating the ssl_context now, thereby making this redundant?

tomplus: Yes, Tornado create context: https://github.com/tornadoweb/tornado/blob/v4.5.2/tornado/simple_httpclient.py#L248

modules/swagger-codegen/src/main/resources/python/tornado/rest.mustache
kenjones-cisco: This is also handled by the tornado client it seems based on the link provided.

kenjones-cisco:
LGTM

modules/swagger-codegen/src/main/resources/python/tornado/rest.mustache
tomplus: Latest versions of Tornado use `certifi`. But it's worth to mention that unstable version from `master` doesn't have this dependency, certificates from system will be used. A developer will still be able to add `certifiy` if she/he really needs to do it.

Issue #7074, opened 2017-11-29
[java] Gson inheritance support


craffael:
##### Description

JavaClients that rely on GSON to serialize a polymorphic type do not set the discriminator field. At the moment this field must be set manually by the user. 
##### Swagger-codegen version

2.3 Snapshot

##### Swagger declaration file content or url

```yaml
swagger: "2.0"
info:
  version: "1.0.0"
  title: Demo


paths:
  /foo:
    post:
      summary: "post a pet"
      parameters:
        - name: body
          in: body
          required: true
          schema: 
            $ref: '#/definitions/Pet'
      responses:
        201:
          description: successful.
  
          
definitions:
  Pet:
    discriminator: type
    type: object
    required:
      - type
    properties:
      type:
        type: string
      name:
        type: string
        
  Cat:
    allOf:
      - $ref: '#/definitions/Pet'
      - properties: 
          huntingSkil:
            type: string
```

##### Command line used for generation

    generate -l java --additional-properties library=retrofit2

##### Steps to reproduce

Run the above command line and open the generated model classes.

##### Related issues/PRs

As far as I know, there is no related issue for this...

##### Suggest a fix/enhancement

Change the default constructor of the Model with the discriminator such that it sets the discriminator field automatically. E.g. for the example above I suggest to add the following default constructor to the `Pet` class:

```java
public Pet() {
    this.setType(this.getClass().getSimpleName());
  }
```

PR #7075, opened 2017-11-29
[Java] Fix serialization of the discriminator field when using GSON

craffael:
### PR checklist

- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language. @bbdouglas @JFCote @sreeshas @jfiala @lukoyanov @cbornet

### Description of the PR

This PR fixes the small issue #7074 with the serialization of the discriminator field when using GSON. Until now, the value of the discriminator field has to be set manually by the user. This is suboptimal since the (unofficial) standard seems to be that the discriminator field is set to the name of the schema (e.g. Cat/Dog/Pet). Currently, the **De**serialization logic in `JSON.java` relies on this convention to detect the actual type of a polymorphic object. I think we should apply the same convetion to serialization.

I propose to set the value of the discriminator field in the constructor of the root model. The user can still override the value of this field if needed, but if he doesn't specify anything else, it will use the name of the underlying model (e.g. Cat or Dog or Pet).


modules/swagger-codegen/src/main/resources/Java/pojo.mustache
```diff
@@ -66,7 +66,11 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#parcela
   private {{{datatypeWithEnum}}} {{name}} = {{{defaultValue}}};
   {{/isContainer}}
 
-  {{/vars}}
+  {{/vars}}{{#gson}}{{#discriminator}}
+  public {{classname}}() {
+    this.{{discriminator}} = this.getClass().getSimpleName();
+  }
+  {{/discriminator}}{{/gson}}
```
wing328: Better do it as follows to avoid empty lines wth spaces:
```
  {{/vars}}
  {{#gson}}
  {{#discriminator}}
......
......
  {{/discriminator}}
  {{/gson}}
```

craffael: Thanks for the tip. I just realized that swagger codegen uses JMustache...
I've pushed a new commit with this change :)

PR #7091, opened 2017-12-01
Added support for Operation EXAMPLE value in ASP.NET Core Generator.

knom:
RE: Issue #7092, #5423
@mandrean @jimschubert 
Added support for Operation EXAMPLE value in ASP.NET Core Generator.

Now exampleJson is currectly generated.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java
@@ -422,6 +422,20 @@ public abstract class AbstractCSharpCodegen extends DefaultCodegen implements Co
                         }
                     }
 
+                    if (operation.examples != null){
+                        for (Map<String, String> example : operation.examples)
+                        {
+                            for (Map.Entry<String, String> entry : example.entrySet())
+                            {
+                                // Replace " with \", \r, \n with \\r, \\n
+                                String val = entry.getValue().replace("\"", "\\\"")
+                                    .replace("\r","\\r")
+                                    .replace("\n","\\n");
+                                entry.setValue(val);
+                            }
+                        }
+                    }
+
                     processOperation(operation);
                 }
             }
--- a/samples/server/petstore/aspnetcore/src/IO.Swagger/Controllers/PetApi.cs
+++ b/samples/server/petstore/aspnetcore/src/IO.Swagger/Controllers/PetApi.cs
@@ -92,6 +92,8 @@ namespace IO.Swagger.Controllers
             // return StatusCode(400);
 
             string exampleJson = null;
+            exampleJson = "<Pet>\n  <id>123456789</id>\n  <name>doggie</name>\n  <photoUrls>\n    <photoUrls>aeiou</photoUrls>\n  </photoUrls>\n  <tags>\n  </tags>\n  <status>aeiou</status>\n</Pet>";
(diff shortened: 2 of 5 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java
```diff
@@ -422,6 +422,20 @@ private void postProcessEnumRefs(final Map<String, Object> models) {
...
                     }
 
+                    if (operation.examples != null){
+                        for (Map<String, String> example : operation.examples)
+                        {
+                            for (Map.Entry<String, String> entry : example.entrySet())
+                            {
+                                // Replace " with \", \r, \n with \\r, \\n
+                                String val = entry.getValue().replace("\"", "\\\"")
+                                    .replace("\r","\\r")
```
jimschubert: There are a few issues with this approach that I see…

1. This will replace unexpected results. For instance, what if my example was:
   ```
    { 
        "example_commands": [
            "\run", "\nip", "\yap", "\rollover"
        ],
        "fast_commands": [
            "\\run", "\\rollover"
         ],
        "omg_slow_down_puppy": [
             "\\\run", "\\\rollover"
         ],
        "search_command" : "\\"
    }
    ````
2. This will lead to failed compilation if we don't escape other things like double-quotes and format placeholders like `{0}`, `{0,15}`, `{0:t}`, etc.
3. afiak this will result in multiline strings, regardless of how many times you escape these characters. I could be wrong on that, though.

The amount of effort to resolve this is, I don't think, worth the benefit. 

Some of this could be resolved by replacing only on boundaries, for example a regex like `\b\\r\b` which would catch probably most cases outside of text occurrences:

```
this\r"matches
this\rdoesn't
```

wing328:

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractCSharpCodegen.java
knom: Great suggestions @jimschubert!
On the double quotes.. I do replace these as well..

I'm not that fluent in Java..
How would I rewrite the code to do regex replace instead?

Thanks!

wing328:
@knom seems like the rebase failed as the PR now contains commits not authored by you.

I'll create another PR by cherry-picking the commits owned by you.

samples/server/petstore/aspnetcore/src/IO.Swagger/Controllers/PetApi.cs
```diff
@@ -92,6 +92,8 @@ public virtual IActionResult FindPetsByStatus([FromQuery][Required()]List<string
             // return StatusCode(400);
 
             string exampleJson = null;
+            exampleJson = "<Pet>\n  <id>123456789</id>\n  <name>doggie</name>\n  <photoUrls>\n    <photoUrls>aeiou</photoUrls>\n  </photoUrls>\n  <tags>\n  </tags>\n  <status>aeiou</status>\n</Pet>";
```
knom: Any way to ONLY pick JSON examples here?

knom: @wing328  @jimschubert

knom:
@wing328 I've just redone the REBASE.. now it should work!

samples/server/petstore/aspnetcore/src/IO.Swagger/Controllers/PetApi.cs
jimschubert: Probably a hack, but iterate responses in the generator code and remove them if they don't start with a "{". If you go that route, is suggest logging a warning any time an example is dropped by the generator in this way.

Mustache doesn't really have logic. It does allow custom filters, so maybe we could write a filter that will only write a value of it is valid json. An issue with that approach is that you can't pick "the first json example* like you could in The generator processing.

wing328: There was a discussion to add an `isJson` tag before but no one seems to have time to implement that yet.

jimschubert: why a tag rather than Mustache filter? just curious.

jimschubert: oh. I get what you're saying. It would be weird to use a tag to get a single json object from ab list of mixed possible types, I think.

Issue #7094, opened 2017-12-01
Add support for SwaggerResponse[] Attributes with no response data


knom:
RE: ASPNET Core Server Generator
@mandrean @jimschubert

[SwaggerResponse] attributes are only added for operation responses that return DATA.
If a response just is a STATUS CODE, now data, the generator ommits the SwaggerResponse.
```
        /// <summary>
        /// 
        /// </summary>
        /// <remarks>Submit a new parcel to the logistics service. </remarks>
        /// <param name="newParcel"></param>
        /// <response code="200">Successfully submitted the new parcel</response>
        /// <response code="500">The operation failed due to an error.</response>
        [HttpPost]
        [Route("/api/parcel")]
        [SwaggerOperation("SubmitParcel")]
        [SwaggerResponse(200, type: typeof(NewParcelInfo))]
        // ??? Where is the 500 result?
        public virtual IActionResult SubmitParcel([FromBody]Parcel newParcel)
```

Thanks!

PR #7096, opened 2017-12-01
Updated SwaggerResponse Attributes so they use the actual dataType instead of the common returnType

knom:
RE: Issue #7094
@mandrean @jimschubert

I've modified the server generator, so that responses with no data are excluded, but multiple responses with their dataType are honored..

```
        [Route("/api/parcel/{trackingId}")]
        [ValidateModelState]
        [SwaggerOperation("TrackParcel")]
        [SwaggerResponse(statusCode: 200, type: typeof(TrackingInformation), description: "Parcel exists.")]
        [SwaggerResponse(statusCode: 500, type: typeof(Error), description: "The operation failed due to an error.")]
        public virtual IActionResult TrackParcel([FromRoute]string trackingId)
```

### DIFF
--- a/modules/swagger-codegen/src/main/resources/aspnetcore/controller.mustache
+++ b/modules/swagger-codegen/src/main/resources/aspnetcore/controller.mustache
@@ -32,8 +32,8 @@ namespace {{packageName}}.Controllers
         [{{httpMethod}}]
         [Route("{{{basePathWithoutHost}}}{{{path}}}")]
         [ValidateModelState]
-        [SwaggerOperation("{{operationId}}")]{{#responses}}{{#returnType}}
-        [SwaggerResponse({{code}}, typeof({{&returnType}}), "{{message}}")]{{/returnType}}{{/responses}}
+        [SwaggerOperation("{{operationId}}")]{{#responses}}{{#dataType}}
+        [SwaggerResponse(statusCode: {{code}}, type: typeof({{&dataType}}), description: "{{message}}")]{{/dataType}}{{^dataType}}{{/dataType}}{{/responses}}
         public virtual IActionResult {{operationId}}({{#allParams}}{{>pathParam}}{{>queryParam}}{{>bodyParam}}{{>formParam}}{{>headerParam}}{{#hasMore}}, {{/hasMore}}{{/allParams}})
         { {{#responses}}
 {{#dataType}}
--- a/samples/server/petstore/aspnetcore/src/IO.Swagger/Controllers/PetApi.cs
+++ b/samples/server/petstore/aspnetcore/src/IO.Swagger/Controllers/PetApi.cs
@@ -81,8 +81,7 @@ namespace IO.Swagger.Controllers
         [Route("/v2/pet/findByStatus")]
         [ValidateModelState]
         [SwaggerOperation("FindPetsByStatus")]
-        [SwaggerResponse(200, typeof(List<Pet>), "successful operation")]
-        [SwaggerResponse(400, typeof(List<Pet>), "Invalid status value")]
+        [SwaggerResponse(statusCode: 200, type: typeof(List<Pet>), description: "successful operation")]
         public virtual IActionResult FindPetsByStatus([FromQuery][Required()]List<string> status)
         { 
             //TODO: Uncomment the next line to return response 200 or use other options such as return this.NotFound(), return this.BadRequest(..), ...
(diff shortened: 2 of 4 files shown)


modules/swagger-codegen/src/main/resources/aspnetcore/controller.mustache
```diff
@@ -31,9 +31,10 @@ namespace {{packageName}}.Controllers
         [{{httpMethod}}]
         [Route("{{{basePathWithoutHost}}}{{{path}}}")]
         [ValidateModelState]
-        [SwaggerOperation("{{operationId}}")]{{#responses}}{{#returnType}}
-        [SwaggerResponse({{code}}, typeof({{&returnType}}), "{{message}}")]{{/returnType}}{{/responses}}
-        public virtual {{#returnType}}IActionResult{{/returnType}}{{^returnType}}void{{/returnType}} {{operationId}}({{#allParams}}{{>pathParam}}{{>queryParam}}{{>bodyParam}}{{>formParam}}{{>headerParam}}{{#hasMore}}, {{/hasMore}}{{/allParams}})
+        [SwaggerOperation("{{operationId}}")]{{#responses}}{{#dataType}}
+        [SwaggerResponse(statusCode: {{code}}, type: typeof({{&dataType}}), description: "{{message}}")]{{/dataType}}{{^dataType}}
```
jimschubert: This (and everywhere using `dataType` rather than `returnType` in the responses array) would still need to be `typeof({{&returnType}})` as the desired return type and the actual data type may not be the same (e.g. returnType = ICollection, dataType = something that isn't ICollection).

modules/swagger-codegen/src/main/resources/aspnetcore/controller.mustache
```diff
@@ -31,9 +31,10 @@ namespace {{packageName}}.Controllers
...
+        [SwaggerResponse(statusCode: {{code}}, description: "{{message}}")]{{/dataType}}{{/responses}}
+        public virtual IActionResult {{operationId}}({{#allParams}}{{>pathParam}}{{>queryParam}}{{>bodyParam}}{{>formParam}}{{>headerParam}}{{#hasMore}}, {{/hasMore}}{{/allParams}})
```
jimschubert: It's not really intuitive here, but swagger specs documented without a return type will cause a compilation error if you specify `IActionResult` rather than `void`. At least it did when I wrote the template. Can you verify with a spec that has responses with void (undefined) return types.

modules/swagger-codegen/src/main/resources/aspnetcore/controller.mustache
knom: Isn't returnType the overall return type of the operation?
I'm looking for the type of a certain response..

modules/swagger-codegen/src/main/resources/aspnetcore/controller.mustache
knom: Not sure I get your comment right..?

Surely you have to return *something*, or throw an exception..

E.g.
```
        [HttpGet]
        public IActionResult Get()
        {
            return NoContent(); // 200
        }
        [HttpGet]
        public IActionResult Get2()
        {
            throw new NotImplementedException();
        }
```

Having VOID as return type limits people from returning ActionResults (which map to Statuscodes).. Working with beginners a lot that confused them (how do they return anything if the return value is void..)

Let me know if this is what you mean.. :) Thanks!

knom:
Any update on this thread?
Is this targeted for 2.3.0 as well?

wing328:
@knom please rebase on the latest master to help resolve the merge conflicts when you've time.

Current master is 2.3.0

modules/swagger-codegen/src/main/resources/aspnetcore/controller.mustache
knom: Hey @jimschubert: I just realized that the XML Comments do some of this anyways..
Can you confirm?

We could remove line 36 essentially?

knom:
@wing328: I've changed the code once more.. Because I learned that Swashbuckle is generating empty responses from XML comments anyways..

jimschubert:
Sorry it took a while to review. I've been busy with the holidays and two small kids.

The last commit looks good. I don't think the extra empty block needs to be changed, unless you want to change it.

knom:
Thanks for reviewing!
I agree on your suggestions..

What's the "swagger default"?

Thanks.

jimschubert:
Swagger spec allows a "default" response to be defined alongside others. So for example, of you define a 200 with a success response type, and a "default" with a ref to an error type, the idea is that you have a defaulted response type for all error types. Again, I'm not sure it is used much, so may not be worth investigating. It was more a general question about whether we should support it.

See docs: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#responseObject

PR #7112, opened 2017-12-05
[Typescript All] Prevent usage of reserved keyword for model fields

JFCote:
### PR checklist

### Description of the PR

Reserved keywords were escaped everywhere but for this. I just add it in the `AbstractTypescriptClientCodegen`. I run all the typescript sample I could find. All the changes are not related to my change because the petstore doesn't contain any reserved keyword.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractTypeScriptClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractTypeScriptClientCodegen.java
@@ -405,6 +405,10 @@ public abstract class AbstractTypeScriptClientCodegen extends DefaultCodegen imp
                 if (Boolean.TRUE.equals(var.isEnum)) {
                     var.datatypeWithEnum = var.datatypeWithEnum.replace(var.enumName, cm.classname + "." + var.enumName);
                 }
+
+                if (isReservedWord(var.name)) {
+                    var.name = escapeReservedWord(var.name);
+                }
             }
         } 
 
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractTypeScriptClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractTypeScriptClientCodegen.java
@@ -147,28 +147,34 @@ public abstract class AbstractTypeScriptClientCodegen extends DefaultCodegen imp
 
     @Override
     public String toParamName(String name) {
-        // replace - with _ e.g. created-at => created_at
-        name = name.replaceAll("-", "_"); // FIXME: a parameter should not be assigned. Also declare the methods parameters as 'final'.
-
-        // if it's all uppper case, do nothing
-        if (name.matches("^[A-Z_]*$"))
-            return name;
-
-        // camelize the variable name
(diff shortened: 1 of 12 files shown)


TiFu:
Just to check that I understand the issues this PR solves correctly: does this change escape Enum values, which are reserved keywords? e.g. escapes `await` in an enum with value `await`

JFCote:
I think enum were already escaped (I did not validate). I'm escaping model members which are reserved keywords.

For example, if you have a model called `Dog` with a boolean named `default`, it will escape `default` with `_default` in the final generated code.

Without this escape, the typescript is not compiling.

Does it make more sense? :)

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractTypeScriptClientCodegen.java
```diff
@@ -405,6 +405,10 @@ public String toEnumName(CodegenProperty property) {
                 if (Boolean.TRUE.equals(var.isEnum)) {
                     var.datatypeWithEnum = var.datatypeWithEnum.replace(var.enumName, cm.classname + "." + var.enumName);
                 }
+
+                if (isReservedWord(var.name)) {
```
wing328: @JFCote Shall we do it in `toVarName` instead similar to other generators? e.g. https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RubyClientCodegen.java#L453

JFCote: @wing328 As you can see here: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractTypeScriptClientCodegen.java#L169

In typescript, the toVarName is simply returning the variable name itself without transformation. Should I change it to the same thing as javascript? 
https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java#L490

Let me know and I'll do the changes!

wing328: Please follow the JS approach, which is used by other generators (e.g. Ruby, PHP, etc)

JFCote: Ok! Will try to commit a fix this week! Thanks for the review!

JFCote:
@wing328 Changes are done! Let me know if something is still missing!

wing328:
@JFCote saw your change. Do we still need the reserved word check in `postProcessModels`?

JFCote:
@wing328 Do you think it's called from somewhere else automatically?

JFCote:
@wing328 I've just test it and you are right, it seems to be called from somewhere, so no need in the `postProcessModels`. I've just submit a new version.

wing328:
@JFCote looks good to me 👍

kenisteward:
Thanks for this.

gennadiylitvinyuk:
The config Parameter modelPropertyNaming is now ignored

Issue #2886, opened 2016-05-16
C# generated code should use ASP.NET core project structure


kbrooks:
We should add an option to the csharp generator to use the new ASP.NET core project structure.

@wing328

wing328:
@kbrooks thanks for the suggestion. Is there any official guide from MS on the ASP.NET core project structure?

I did a search and found this: http://zablo.net/blog/post/modular-asp-net-core-project-structure (which is not official)

Also to be clear, you're referring to C# API client but not ASP.NET5 server stub, right?

Recently, there's a change to the folder structure via https://github.com/swagger-api/swagger-codegen/pull/2752 so please pull the latest master and give it a try if you've not done so.

kbrooks:
The official docs appear to be down right now, but in case they come back up: https://docs.asp.net/en/latest/conceptual-overview/understanding-aspnet5-apps.html.

The main difference is the use of project.json files instead of the old xml configuration files.
The new project structure will be useful to anyone developing on mono or .NET core.

The ASP.NET5 server stub actually already uses the ASP.NET core project structure, so we could use that as a guide.

wing328:
Yup, I got a 404. Will check it out later.

jimschubert:
@kbrooks @wing328  I'd put this on the back burner for now. There's talk of dropping the `project.json` and moving back to MSBuild based .csproj after RC2.

See: https://twitter.com/davidfowl/status/730219570783363073 and https://youtu.be/P9HqMZviaMg?t=1712. There was also a pretty [lengthy discussion on Reddit](https://www.reddit.com/r/programming/comments/4iveez/projectjson_is_dead_aspnet_core_goes_back_to/) about a week ago.

Since Xamarin Studio is free to use on OS X and Linux and Visual Studio Community Edition is also free to use, there's nothing blocking anyone from developing on mono on any OS. Both of these support targeting current .NET 4.6.1.

The only thing this blocks is from consuming .NET Core assemblies, which I think would be fairly minor to incorporate after all the csproj stuff dies down.

kdc415:
Looks like MS says they're going back to MSBuild

[https://blogs.msdn.microsoft.com/dotnet/2016/05/23/changes-to-project-json/](url)

I've just downloaded my first swagger generated api server stub. I downloaded core RC1 to try to run it and it hosed my VS 2015 installation. 

[...]

PR #7114, opened 2017-12-05
[AspNetCore SERVER] Added support for ASP.NET Core 2.0

knom:
@jimschubert, @mandrean, @wing328 

Somehow related to #2886, but for the SERVER
Moved the project structure, mustache files, .. to ASP.NET Core 2.0.

Let me know what you think!


jimschubert:
This looks good, but should probably go into 2.3.0. I can't see the target branch on mobile.

wing328:
@jimschubert it's targeting current master (which is 2.3.0)

wing328:
@knom if I understand correctly, we'll no longer support ASP.NET CORE 1.0 after merging this PR into master so I've marked this PR as a "Breaking Change" (without fallback).

knom:
@wing328 that is right!
There's three changes..
- Moved away from project.json --> .csproj
  (see https://www.stevejgordon.co.uk/project-json-replaced-by-csproj)
- Updated all the dependencies to .NET Core 2
- Removed some of the setup code, as it's now built-in in .NET Core 2 anyways and was a duplicate

jimschubert:
@wing328 Thanks. I haven't looked at the version in master in a while, I didn't realize it was promoted to 2.3.0. I didn't check the compile to verify, but @knom has been making quality contributions and I don't see any reason it would not work as expected.

knom:
Thanks for the quick merge :-)

PR #7120, opened 2017-12-06
Added Base Path support as well as PathParameterValidation

knom:
@jimschubert @wing328 
- BasePath Support:
   The basePath property of the Swagger document is being honored and URLs are relatively created.

- Path Parameter Validation:
   [Required], [Range], .. attributes are being generated for PATH and QUERY parameters by codegen.
   These are also being rendered out by Swashbuckle.SwaggerGen through a filter.


wing328:
@knom the change looks good to me. Thanks again for the contribution.

jimschubert:
Looking good

PR #7125, opened 2017-12-07
[Rust-Server] use status codes as the variants of the response enums instead of descriptions

### LINKED ISSUE #7106: [Rust] rust-server crashes when the description of response is non-ascii strings
##### Description

rust-server crashes when the description of response is non-ascii strings

##### Swagger-codegen version

master ( 61e706ba6f6b75d68651028d1796a2b12e2034f2 )

[this](https://gist.github.com/KeenS/0e539a9cb71c23d178729e2a41ecc376)

```console
$ java -jar  ~/compile/swagger-codegen/modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate  -v -l rust-server -i swagger.json  -o server
```

##### Steps to reproduce

just run the command above

##### Related issues/PRs

None

This is caused because the description is used for identifiers. According to the [specification](https://swagger.io/specification/#responseObject), the description can be arbitrary common mark strings, so it is not suitable for identifiers.

I have created a [patch](https://github.com/KeenS/swagger-codegen/commit/5f003746ce27306e4bf1cec9df1c019036ea3015) that replaces the uses of messages with `StatusXxx`s.
If someone likes this, I'll submit a  PR.

KeenS:
### PR checklist

@frol (2017/07) @farcaller (2017/08)
### Description of the PR

this fixes #7106 . `camelize(rsp.message.split("[^A-Za-z ]")[0].replace(" ", "_"));` produces an empty array when a non-ascii string is given.


bjgill:
Thanks for the PR. #7106 certainly looks problematic. 

Unfortunately, I'm not sure this is the correct solution:
* One of the goals of `rust-server` is to provide a nice, human-friendly API. As [an example](https://github.com/swagger-api/swagger-codegen/pull/7125/files#diff-f094cc6efde03ba4e0e1d46c92d247f1L180), `LoginUserResponse::InvalidUsername` is more meaningful than `Status200`. I'd be somewhat reluctant to lose this usability.
* rust-server is intended to be agnostic as to the transport mechanism. This would tie the Rust code more strongly to HTTP, which might be undesirable in the longer term.

The current code is a heuristic that, as you discovered, can fail quite spectacularly. I wonder if we should introduce a `responseId` extension (similar to the current `operationId`), which would allow us to side-step this problem.

KeenS:
> One of the goals of rust-server is to provide a nice, human-friendly API

I see.

> `LoginUserResponse::InvalidUsername` is more meaningful than `Status200`

I agree. Even if you used `StatusOk` or like that, the resolution of information is low.

> I wonder if we should introduce a `responseId` extension

I think it's a good idea. Then, I have some questions

* What if the `responseId` field is absent? Just raise an error, or fall back to `Status200` ?
* How to document it? Are there any per language documentation?

KeenS:
I'll implement 3. Thank you.

samples/server/petstore/rust-server/src/client.rs
```diff
@@ -1649,7 +1628,7 @@ impl Api for Client {
                 0 => {
 
 
-                    Ok(CreateUsersWithListInputResponse::SuccessfulOperation)
+                    Ok(CreateUsersWithListInputResponse::Status0)
```
bjgill: This file contains the errant `Statusxxx` I mentioned earlier.

In any case, this `Status0` worries me - it looks as if we're not handling the `default` case (see https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#responses-object).

samples/server/petstore/rust-server/src/lib.rs
```diff
@@ -35,173 +35,221 @@ pub use swagger::{ApiError, Context, ContextWrapper};
 
 #[derive(Debug, PartialEq)]
 pub enum TestSpecialTagsResponse {
+    /// successful operation
```
bjgill: 😄

modules/swagger-codegen/src/main/resources/rust-server/client.mustache
```diff
@@ -269,8 +269,8 @@ impl Api for Client {
                     let response_{{name}} = response.headers.get::<Response{{nameInCamelCase}}>().ok_or_else(|| "Required response header {{baseName}} for response {{code}} was not found.")?;
 {{/headers}}
 
-{{#dataType}}                    Ok({{operationId}}Response::{{message}}{{^headers}}(body){{/headers}}{{#headers}}{{#-first}}{ body: body, {{/-first}}{{name}}: response_{{name}}.0.clone(){{^-last}}, {{/-last}}{{#-last}} }{{/-last}}{{/headers}})
-{{/dataType}}{{^dataType}}                    Ok({{operationId}}Response::{{message}}{{#headers}}{{#-first}}{ {{/-first}}{{^-first}}, {{/-first}}{{name}}: response_{{name}}.0.clone(){{#-last}} }{{/-last}}{{/headers}})
+{{#dataType}}                    Ok({{operationId}}Response::{{#vendorExtensions}}{{x-responseId}}{{/vendorExtensions}}{{^headers}}(body){{/headers}}{{#headers}}{{#-first}}{ body: body, {{/-first}}{{name}}: response_{{name}}.0.clone(){{^-last}}, {{/-last}}{{#-last}} }{{/-last}}{{/headers}})
+{{/dataType}}{{^dataType}}                    Ok({{operationId}}Response::Status{{code}}{{#headers}}{{#-first}}{ {{/-first}}{{^-first}}, {{/-first}}{{name}}: response_{{name}}.0.clone(){{#-last}} }{{/-last}}{{/headers}})
```
bjgill: I suspect this might be the source of the errant `Statusxxx` I've seen further down. Should this be `{{#vendorExtensions}}{{x-responseId}}{{/vendorExtensions}}` instead of `Status{{code}}`?

KeenS: I was careless, thanks for pointing out.

bjgill:
Looks good, thanks. It'd be nice to have special-case handling for the `default` code (i.e. `0`), but not essential.

Not sure what the CI failure is - looks unrelated to any of your changes.

```
Failed tests:   testCreateAndGetPetAsync(io.swagger.client.api.PetApiTest)

Tests run: 56, Failures: 1, Errors: 0, Skipped: 2

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.12:test (default-test) on project swagger-petstore-okhttp-gson: There are test failures.
```

KeenS:
I'm trying to fix the problem but even codes of master fails too on my environment.
It'd be better to wait for a while.

KeenS:
I missed the failing test is against JDK 7... It'll take time to setup the environment.

wing328:
@frol @farcaller @bjgill please review when you've time.

frol:
LGTM

PR #7142, opened 2017-12-09
Spelling fixes

scop:


wing328:
@scop thanks for the fix, which looks good to me.

The upcoming PRs will pick up the change in the petstore samples.

PR #7149, opened 2017-12-11
Fixed generating class for properties with empty properties

korkut89:
Codegen doesn't generate class when an object defined with an empty
properties list. When an object named risk property defined with empty properties list generator defines it as java.lang.Object. But when reading the object with ObjectMapper it doesn't read `risk : {}`.

### PR checklist

### Description of the PR

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/InlineModelResolver.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/InlineModelResolver.java
@@ -48,7 +48,7 @@ public class InlineModelResolver {
                                     if (model instanceof ModelImpl) {
                                         ModelImpl obj = (ModelImpl) model;
                                         if (obj.getType() == null || "object".equals(obj.getType())) {
-                                            if (obj.getProperties() != null && obj.getProperties().size() > 0) {
+                                            if (obj.getProperties() != null) {
                                                 flattenProperties(obj.getProperties(), pathname);
                                                 String modelName = resolveModelName(obj.getTitle(), bp.getName());
                                                 bp.setSchema(new RefModel(modelName));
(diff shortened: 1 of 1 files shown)


wing328:
> Codegen doesn't generate class when an object defined with an empty
properties list

Can you share more about the use case with an empty properties list?

korkut89:
I added an example yaml file @wing328

wing328:
@korkut89 sorry I mean why an object is defined without properties? I think only a handful of programming languages support empty object (an object without property)

What does the corresponding JSON payload look like?

korkut89:
```
{
    "Data" : {
        "Greeting" : "Hi"
    },
    "Risk": {}
}
```

It should look something like that.

wing328:
@korkut89 thanks for sharing more. Given that the JSON payload is well-defined, why define it as an empty object?

korkut89:
@wing328 
https://github.com/OpenBankingUK/account-info-api-spec/blob/master/dist/v1.1/account-info-swagger.yaml
We are using this spec. Since this spec defined as standard for open banking we must use this. I am not sure why defined it as an empty object.

wing328:
@korkut89 I think the spec owner will need to update the spec to have a proper definition of the model with valid properties instead.

cbornet:
@wing328 well, although weird, this spec is valid. So we should generate something correct.
The filter on the properties size was introduced by #2297 so the impact of changing it has to be carefully studied as regards to the issue it was solving.

korkut89:
I think this spec is likely to be updated in the long run. But can't say for certain that it is going to change. I agree with @cbornet this is a valid spec.

wing328:
I agreed with you guys the spec is valid according to the specification.

My point is without knowing how the object (JSON, XML) is defined, how can any developers figure out how to process the object?

The issue is nothing swagger-codegen-specified. Someone who is manually writing the client will still encounter the same issue.

cbornet:
To me, Risk should be a class with no attributes, no ?

PR #7160, opened 2017-12-12
Added apiNameSuffix parameter for JaxRsSpec code generator

ndurette:
### PR checklist

- [ ] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be fold .und in `.\bin\windows\`.

### Description of the PR

I added an additional property to specify an api name suffix instead of always using "Api".
I only changed the JaxRsSpec generator but it will also affect generator that are not reimplementing "toApiName()".

I did run the script as mentioned but I didn't include to change as it changed unrelated stuff. 
You might want to run all the script on master.

### DIFF
--- a/modules/swagger-codegen-cli/src/main/java/io/swagger/codegen/cmd/Generate.java
+++ b/modules/swagger-codegen-cli/src/main/java/io/swagger/codegen/cmd/Generate.java
@@ -74,6 +74,10 @@ public class Generate implements Runnable {
             description = CodegenConstants.API_PACKAGE_DESC)
     private String apiPackage;
 
+    @Option(name = {"--api-name-suffix"}, title = "api name suffix",
+            description = CodegenConstants.API_NAME_SUFFIX_DESC)
+    private String apiNameSuffix;
+
     @Option(name = {"--model-package"}, title = "model package",
             description = CodegenConstants.MODEL_PACKAGE_DESC)
     private String modelPackage;
@@ -215,6 +219,10 @@ public class Generate implements Runnable {
 
         if (isNotEmpty(apiPackage)) {
             configurator.setApiPackage(apiPackage);
+        }
+
+        if (isNotEmpty(apiNameSuffix)) {
+            configurator.setApiNameSuffix(apiNameSuffix);
         }
 
         if (isNotEmpty(modelPackage)) {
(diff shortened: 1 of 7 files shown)


JFCote:
@ndurette I think it would be nice to see the changes on at least one generator (run the shell script). Also, I would be nice to add a `.sh` file that use your new parameter so we could the the effect it have compared to the default.

ndurette:
@JFCote I added a new script to generate a new server using my feature. I also ran the previous "jaxrs-spec-petstore-server" scripts.

modules/swagger-codegen-cli/src/main/java/io/swagger/codegen/cmd/Generate.java
```diff
@@ -74,6 +74,10 @@
             description = CodegenConstants.API_PACKAGE_DESC)
     private String apiPackage;
 
+    @Option(name = {"--api-name-suffix"}, title = "api name suffix",
```
JFCote: I don't think this should be here at all. If I'm not mistaken, these are the general swagger-codegen wide parameters that all generator should support. I think you only need to add you parameters to your specific codegen. See this for an example: 

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaPlayFrameworkCodegen.java#L22

JFCote: In fact, I check all your code, it seems you are indeed trying to create an "all generators" parameter... I'm not sure it is really safe to add that kind of parameters. I will add @wing328  and @cbornet to the conversation to have their advices.

ndurette:
In my case adding the property in JavaJAXRSSpecServerCodegen would do it. 
Do you think it can be valuable to have this additional property in AbstractJavaJAXRSServerCodegen?

The idea behind this modification was to copy the "model-name-suffix" feature. But I agree that it might not fit for every generator. 

I'll wait for comments and then make the changes as needed.

wing328:
> The idea behind this modification was to copy the "model-name-suffix" feature. But I agree that it might not fit for every generator.

Given that it's only targeting a specified generator, please add the option to JavaJAXRSSpecServerCodegen for the time being. If there is more demand to add this feature to other generators, we'll add it one by one.

PR #7103, opened 2017-12-02
Added support for VERSION, APPNAME and CONTACT from Swagger file into Startup.cs

knom:
#7102 
@jimschubert
Version and AppName are now used for `AddSwaggerGen(..)`, resulting in the Swagger generated being more like the original one.

PR #7168, opened 2017-12-13
Added Contact NAME to Startup.cs

knom:
Follow-up to #7103


modules/swagger-codegen/src/main/resources/aspnetcore/Startup.mustache
```diff
@@ -61,6 +61,7 @@ namespace {{packageName}}
                         Description = "{{#appName}}{{{appName}}}{{/appName}}{{^appName}}{{packageName}}{{/appName}} (ASP.NET Core 2.0)",
                         Contact = new Contact()
                         {
+                           Name = "{{#infoName}}{{{infoName}}}{{/infoName}}",
```
wing328: @knom What about defaulting to "Swagger Codegen Contributors" if `infoName` (name in contact) is not defined?

The goal is to do a bit of growth hacking for Swagger Codegen.

Same for `infoUrl`: defaulting to "https://github.com/swagger-api/swagger-codegen" if `infoURL` is not define.

knom: Sure thing!
Can you edit the commit?
I shutdown my build VM..

knom: Nice idea!
Let me update..

Issue #7159, opened 2017-12-12
[AspnetCore Server] Include original swagger.json as a file in the package


knom:
@wing328 @jimschubert 
Can we include the original JSON passed into Swagger-Codegen as a file in the generated ZIP package?

Would be great to compare with the Swashbuckle generated JSON..

jimschubert:
In think this may cause some licensing issues.

I wonder if we could default the option to false and opt-in to copying the swagger definition to the .swagger folder in the output directory.

wing328:
Some other server generators are generating/including the original OpenAPI/Swagger spec in JSON/YAML format.

e.g. https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ErlangServerCodegen.java#L125

Template: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/erlang-server/swagger.mustache

PR #7170, opened 2017-12-13
[AspnetCore Server] Added Original swagger

knom:
Resolving issue #7159 
Adding the original swagger to the package.

@wing328 please review!
I followed your guidance..

Thanks.

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AspNetCoreServerCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AspNetCoreServerCodegen.java
@@ -1,15 +1,20 @@
 package io.swagger.codegen.languages;
 
+import com.fasterxml.jackson.core.JsonGenerator;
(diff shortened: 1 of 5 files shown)


wing328:
@knom looks good to me. Will merge once the CI tests pass

cc @mandrean @jimschubert

jimschubert:
Looks good, although, I don't know that there's a lot of value providing the static file or Swashbuckle option, unless it's configured as an option in the generator that conditionally includes Swashbuckle generation.

I would think a "sane default" of using Swashbuckle would be appropriate. We could then include the original in the .swagger directory with instructions in the README for those who prefer the static file. Reason is, this would be simpler the maintain than the inevitable options around dynamic or static specification.

knom:
@wing328 do you know when any of the pull requests is going to show up on EDITOR.SWAGGER.IO?
They still use the old swagger-codegen version..

thx,

knom:
@wing328?

PR #7194, opened 2017-12-17
[Qt5] plug mem-leaks

antis81:
The generated Qt5 code produces some serious memory leaks. What I noticed is, that one has to use the API for example like this in a real environment:

```c++
// "this" is assumed a valid QObject
void my_func() {
SWGDefaultApi* api = new SWGDefaultApi();
// actually "getSomethingSignalE" should be used always (!) instead - just for the example this is shorter to write
QObject::connect(api, &SWGDefaultApi::getSomethingSignal, this, [api](QList<SWGMyObject*>* list) {
  // do something with the list:
  // the following sample loop produces a definite (and serious) mem-leak
  for (auto swg_my_obj : *list) {
    qDebug() << swg_my_obj->asJson();
  }

  // manual cleanup is needed to free memory
  qDeleteAll(*list);
  delete list;
  delete api;
});

// finally call the api
api->getSomething();
}
```

While this looks very "unfamiliar" in the Qt world, it is also cumbersome and error prone in usage.

This PR is meant to fix those leaks and also should make maintenance of the templates more simple. This breaks compatibility though, because it uses Qt mechanisms (COW & smart-pointers instead of e.g. pointers to QList etc.). OTOH it plugs the most serious mem-leaks.

Unfortunately I had to make changes "blind" without a functional test or anything. So, if someone could actually check it, that would be great.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/qt5cpp/HttpRequest.cpp.mustache
+++ b/modules/swagger-codegen/src/main/resources/qt5cpp/HttpRequest.cpp.mustache
@@ -47,7 +47,7 @@ void {{prefix}}HttpRequestInput::add_file(QString variable_name, QString local_f
     qsrand(QDateTime::currentDateTime().toTime_t());
 
     manager = new QNetworkAccessManager(this);
-    connect(manager, SIGNAL(finished(QNetworkReply*)), this, SLOT(on_manager_finished(QNetworkReply*)));
+    connect(manager, &QNetworkAccessManager::finished, this, &HttpRequestWorker::on_manager_finished);
 }
 
 {{prefix}}HttpRequestWorker::~{{prefix}}HttpRequestWorker() {
--- a/modules/swagger-codegen/src/main/resources/qt5cpp/api-body.mustache
+++ b/modules/swagger-codegen/src/main/resources/qt5cpp/api-body.mustache
@@ -98,10 +98,10 @@ void
 
     {{#bodyParams}}
     {{#isContainer}}
-    auto {{paramName}}_jobj = new QJsonObject();
-    toJsonArray((QList<void*>*){{paramName}}, {{paramName}}_jobj, QString("body"), QString("{{prefix}}User*"));
+    QJsonObject {{paramName}}_jobj;
+    toJsonArray((QList<void*>){{paramName}}, {{paramName}}_jobj, QString("body"), QString("{{prefix}}User*"));
 
-    QJsonDocument doc(*{{paramName}}_jobj);
+    QJsonDocument doc({{paramName}}_jobj);
     QByteArray bytes = doc.toJson();
 
     input.request_body.append(bytes);
@@ -174,9 +174,9 @@ void
     QJsonObject obj = doc.object();
 
     foreach(QString key, obj.keys()) {
-        qint32* val;
-        setValue(&val, obj[key], "{{returnBaseType}}", "");
-        output->insert(key, *val);
+        qint32 val;
+        setValue(&val, obj[key], "{{returnBaseType}}", QString());
+        output->insert(key, val);
     }
     {{/isMapContainer}}
     {{^isMapContainer}}
--- a/modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
+++ b/modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
@@ -366,7 +366,7 @@ toJsonArray(QList<void*> value, QJsonObject& output, QString innerName, QString 
         for(void* obj : value) {
             {{prefix}}Object *{{prefix}}object = reinterpret_cast<{{prefix}}Object *>(obj);
             if({{prefix}}object != nullptr) {
-                outputarray.append(*({{prefix}}object->asJsonObject()));
+                outputarray.append({{prefix}}object->asJsonObject());
             }
         }
     }
@@ -409,7 +409,7 @@ toJsonArray(QList<void*> value, QJsonObject& output, QString innerName, QString 
         for(double obj : *(reinterpret_cast<QList<double>*>(&value)))
             outputarray.append(QJsonValue(obj));
     }
-    output->insert(innerName, outputarray);
+    output.insert(innerName, outputarray);
 }
 
 void
--- a/modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
+++ b/modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
@@ -413,78 +413,12 @@ toJsonArray(QList<void*> value, QJsonObject& output, QString innerName, QString 
 }
 
 void
-toJsonMap(QMap<QString, void*>* value, QJsonObject* output, QString innerName, QString innerType) {
-    if((value == nullptr) || (output == nullptr)) {
-        return;
+toJsonMap(QMap<QString, void*> value, QJsonObject& output, QString innerName, QString innerType) {
+    for (auto itemkey: value.keys()) {
+        QJsonObject mapobj;
+        ::{{cppNamespace}}::toJsonValue(itemkey, value.value(itemkey), &mapobj, innerType);
+        output.insert(innerName, mapobj);
     }
-    QJsonObject mapobj;
-    if(innerType.startsWith("{{prefix}}")){
-        auto items = reinterpret_cast< QMap<QString, {{prefix}}Object*> *>(value);
-        for(auto itemkey: items->keys()) {
-            ::{{cppNamespace}}::toJsonValue(itemkey, items->value(itemkey), &mapobj, innerType);
-        }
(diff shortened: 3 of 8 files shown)


antis81:
@ravinikam @stkrwork @fvarose Please review, thanks!

wing328:
@antis81 when you've time, please rebase on the latest master to resolve the merge conflicts.

antis81:
@wing328 Done. Please double check the "complex types" (QList, QMap), as this may lead to data loss. However everything rather gets better than worse. So the risk is worth this. :)

modules/swagger-codegen/src/main/resources/qt5cpp/api-body.mustache
```diff
@@ -98,8 +98,8 @@ void
 
     {{#bodyParams}}
     {{#isContainer}}
-    auto {{paramName}}_jobj = new QJsonObject();
-    toJsonArray((QList<void*>*){{paramName}}, {{paramName}}_jobj, QString("body"), QString("{{prefix}}User*"));
+    QJsonArray {{paramName}}_jobj;
+    toJsonArray((QList<void*>){{paramName}}, {{paramName}}_jobj, QString("body"), QString("SWGUser*"));
```
antis81: Minor whoopsie with "SWGUser" -> will fix and push.

antis81:
@wing328 Seems something went wrong; also seems not to be on "my end"…??? Maybe typo in version number?

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'kotlin-petstore-client'.
> Could not resolve all dependencies for configuration ':classpath'.
   > Could not resolve commons-lang:commons-lang:2.6.
     Required by:
         :kotlin-petstore-client:unspecified > org.jetbrains.kotlin:kotlin-gradle-plugin:1.1.2
      > Could not HEAD 'http://repo1.maven.org/maven2/commons-lang/commons-lang/2.6/commons-lang-2.6.pom'.
         > Connection reset
```

wing328:
>          > Connection reset

No worry. I'll restart the job, which should clear the error.

wing328:
cc @etherealjoy @ravinikam @stkrwork @fvarose

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
```diff
@@ -371,115 +301,112 @@ setValue(void* value, QJsonValue obj, QString type, QString complexType) {
...
-    }
+toJsonArray(QList<void*> value, QJsonObject& output, QString innerName, QString innerType) {
+
     QJsonArray outputarray;
     if(innerType.startsWith("{{prefix}}")){
-        for(void* obj : *value) {
+        for(void* obj : value) {
             {{prefix}}Object *{{prefix}}object = reinterpret_cast<{{prefix}}Object *>(obj);
             if({{prefix}}object != nullptr) {
                 outputarray.append(*({{prefix}}object->asJsonObject()));
```
etherealjoy: _asJsonObject_ returns an object instead of pointer

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
```diff
@@ -557,12 +484,7 @@ toJsonMap(QMap<QString, void*>* value, QJsonObject* output, QString innerName, Q
         }
     }
     output->insert(innerName, mapobj);
```
etherealjoy: output is a reference not a pointer anymore.

etherealjoy:
I think this rework is a good step forward to plug the mem leaks.
But when I generated with antis81/patch-1 branch and the code does not compile anymore. There were signature mismatch in many places.

wing328:
Thanks @etherealjoy for reviewing 👍

etherealjoy:
The Memory Leaks from the Qt Slots when the HTTP response is received is a major concern if the user is not aware of what is going on in the generated code. After connected to the signals, the user has to delete the pointer received on the slots.
One one is to use references wherever possible. I am interested to hear.

antis81:
@etherealjoy Thanks for the review, awesome! Qt has fine mechanisms builtin to do that, so from the public API side there is no need to use pointers or references at all. In almost any case it is cheaper to just copy the (swagger/json) object. Btw. it also can reduce memory consumption greatly. As written in the description, it is true this PR breaks API compatibility. Over the years i became am a fan of breaking things, if that "fixes stuff big time" - sorry. :innocent:

antis81:
@etherealjoy I added your points from the review. However for `toJsonMap` it did not work out that simple, so I replaced that code block altogether with a working solution. Also I will rebase this on master again, so nobody has to bother about merge conflicts. :)

antis81:
@wing328 Could you please restart the build on circle-ci? Looks like another plugin related problem.

modules/swagger-codegen/src/main/resources/qt5cpp/HttpRequest.cpp.mustache
```diff
@@ -47,7 +47,7 @@ void {{prefix}}HttpRequestInput::add_file(QString variable_name, QString local_f
     qsrand(QDateTime::currentDateTime().toTime_t());
 
     manager = new QNetworkAccessManager(this);
-    connect(manager, SIGNAL(finished(QNetworkReply*)), this, SLOT(on_manager_finished(QNetworkReply*)));
+    connect(manager, &QNetworkAccessManager::finished, this, &HttpRequestWorker::on_manager_finished);
```
etherealjoy: Missing {{prefix}} before HttpRequestWorker

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-header.mustache
```diff
@@ -11,11 +11,10 @@ namespace {{this}} {
...
 
     void setValue(void* value, QJsonValue obj, QString type, QString complexType);
-    void toJsonArray(QList<void*>* value, QJsonObject* output, QString innerName, QString innerType);
-    void toJsonValue(QString name, void* value, QJsonObject* output, QString type);
-    void toJsonMap(QMap<QString, void*>* value, QJsonObject* output, QString innerName, QString innerType);
+    void toJsonArray(QList<void*> value, QJsonObject& output, QString innerName, QString innerType);
+    void toJsonValue(QString name, void* value, QJsonObject& output, QString type);
+    void toJsonMap(QMap<QString, void*> value, QJsonObject& output, QString innerName, QString innerType);
     bool isCompatibleJsonValue(QString type);
-    QString stringValue(QString* value);
```
etherealjoy: QString stringValue(QString* value) is used in the code, strangely it compiles. But when I am running the SW the server throws error saying ?true in Query Path because for whatever reason Qt has used the bool version of the function overload. Very weird.

etherealjoy:
@antis81 
I took over the antis81:patch-1 and compiled and test on my machine, but I had to made changes.
- modelFactory.mustache --> fixed one mem leak
- model-body.mustache --> Some piece of code is generated only when model has arrays, so corrected here
- HttpRequest.cpp.mustache --> Missing some token.
- helpers-header.mustache -->adapted the signature and added again 
```C++QString stringValue(QString* value);
```
- helpers-body.mustache --> adapted the signature

Main problem with the patch is that 
All setValue from inside the model uses the address of the members so for pointer variables address of a pointer is passed and therefore the changes in helpers-body.mustache is incorrect inside the setValue function
toJsonArray and toJsonMap needs a pointer variable of the Address of the Map/List so that reinterpretation will be possible. Passing a List directly is not possible.
Changes are here
https://github.com/etherealjoy/swagger-codegen/tree/qt5cpp_rework_antis81_patch-1

antis81:
Unfortunately I do not have enough time at the moment to review or contribute stuff. Thanks @etherealjoy for taking over!

etherealjoy:
@wing328 How can I take the changes of antis81 keeping his contribution intact?

wing328:
@etherealjoy please try the following in your "swagger-codegen" (folked) repo:
```
git remote add antis81  https://github.com/antis81/swagger-codegen.git
git pull 
git checkout patch-1
git log (to make sure you see the commits in this PR) 
```
Then you can continue to work on this branch (and push it to your own repo) or start a new branch based on `patch-1`

antis81:
Here's what I usually do in such situation:

```
# just like wing328 wrote, add a new git remote…
git remote add antis81  https://github.com/antis81/swagger-codegen.git

# …now fetch the remote branch
git fetch antis81 patch-1

# (just in case, make sure your local branch is the active one)
git checkout qt5cpp_rework_antis81_patch-1

# now simply cherry-pick the commit on top of your branch
git cherry-pick antis81/patch-1

# resolve possible conflicts and done
```

etherealjoy:
@antis81 @wing328 
We can close this now. #7695 has been merged.

antis81:
Yeah, thanks for the good work!

PR #7204, opened 2017-12-18
[C++][Qt5] use QJsonObject instead of QJsonArray for maps

etherealjoy:
### PR checklist

- [X] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [X] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [X] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [X] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

(details of the change, additional tests that have been done, reference to the issue for tracking, etc)

Generated code for Docker API version 1.32. Client code communicates with dockerd with the parameters successfully interpreted by dockerd.

cc @stkrwork @ravinikam @fvarose

PR #7206, opened 2017-12-18
[qt5cpp] Disable creation of empty json fields and handle prefix in model generation

### LINKED ISSUE #7117: [qt5cpp] sigsegv when working with multiple models
##### Description

When two models are generated for qt5cpp from two yaml files, there is a sigsev error due to the wrong symbol being used from the duplicate symbols generated after compilation due to the mustache files not respecting modelnameprefix.
Also when prefix is used not always correctly handled.
##### Swagger-codegen version

2.3.0

Just use two yaml files in one binary.

--additional-properties cppNamespace=docker --additional-properties modelNamePrefix=Docker -l qt5cpp
--additional-properties cppNamespace=k8s    --additional-properties modelNamePrefix=K8S -l qt5cpp

##### Steps to reproduce 

create static or shared libs and link to exe when running the symbols from the other lib gets executed instead since all models have same class names.
namespace also did not help

### LINKED ISSUE #7205: [qt5cpp] Primitive Types are generated with default json values even when not set
##### Description

When the Model data is not being updated, the primitive types are already initialized with default values.
This is a problem because there is no check if the data is updated and when sent to the server is an issue because they are valid data.

##### Swagger-codegen version
2.3.0

Checks are being done to see if the data is updated before serialization.

etherealjoy:
### PR checklist

### Description of the PR

Fixes #7117 
Fixes #7205

Generated code for Docker API version 1.32. Client code communicates with dockerd with the parameters successfully interpreted by dockerd.

cc @stkrwork @ravinikam 

Easier to merge after #7204


etherealjoy:
#7117
The prefix handling allows the generated models to have different symbols and therefore resolves to unique symbol avoiding sigsegv in due to incompatibility.

etherealjoy:
#7205 
Primitive Types when not updated will not generated avoiding false data to be sent.
Strings will not be sent when Empty. 
Containers will not be sent when empty.
Objects will not be serialized when members are not set.

etherealjoy:
Will conflict will #7194.

modules/swagger-codegen/src/main/resources/qt5cpp/HttpRequest.h.mustache
```diff
@@ -20,7 +20,7 @@ enum HttpRequestVarLayout {NOT_SET, ADDRESS, URL_ENCODED, MULTIPART};
 namespace {{this}} {
 {{/cppNamespaceDeclarations}}
 
-class SWGHttpRequestInputFileElement {
+class {{prefix}}HttpRequestInputFileElement {
```
stkrwork: What about the header guards here?

etherealjoy:
Tests

- [X] PetStore Updated and passed.

- [X] Docker API version 1.32 client generated and used to pull an image.

modules/swagger-codegen/src/main/resources/qt5cpp/HttpRequest.h.mustache
etherealjoy: done

wing328:
FYI. I've resolved the merge conflicts via https://github.com/swagger-api/swagger-codegen/pull/7206/commits/dcb308a5cfd066197a7d9834abc55020a667e531

PR #7212, opened 2017-12-19
[Rust-Server] Use string literal based mime initialization

### LINKED ISSUE #7126: [Rust-Server] it generates broken codes in some cases around mimetypes
##### Description

when the mime type is not a single token, like `application/vnd.github.v3+json`, it produces invalid rust code

##### Swagger-codegen version

e75523a6846f0aa584ada33674744f395b781a7f 

https://gist.github.com/KeenS/195f694ce0e0215dc76541f799ba713d

```
$ java -jar wagger-codegen-cli.jar generate  -v -l rust-server -i swagger.json   -o server
```

##### Steps to reproduce

run the command above and run

##### Related issues/PRs

None

use `"mime/type".parse().unwrap()`. I'll do it.

KeenS:
### PR checklist

- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.  @frol (2017/07) @farcaller (2017/08) @bjgill (2017/12)

### Description of the PR

fixes #7126.


frol:
LGTM

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/RustServerCodegen.java
```diff
@@ -528,9 +528,8 @@ public CodegenOperation fromOperation(String path, String httpMethod, Operation
         // if "consumes" is defined (per operation or using global definition)
         if (consumes != null && !consumes.isEmpty()) {
             List<Map<String, String>> c = new ArrayList<Map<String, String>>();
-            for (String key : consumes) {
+            for (String mimeType : consumes) {
                 Map<String, String> mediaType = new HashMap<String, String>();
-                String mimeType = processMimeType(key);
 
                 if (mimeType.startsWith("Application/Xml")) {
```
bjgill: I suspect there may be problems - it looks as if later bits of code assume that the `mimetype` has been generated by `processMimeType` - hence this checking against `Application/Xml`.

According to the [docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types), mime types are supposed to be case insensitive, but are generally lower case.

Now that we've no longer got `processMimeType`, do we need to be a bit more clever about how we detect the XML mime type?

Otherwise, it looks sensible.

KeenS: I think canonicalizing to lowercase will do.

Issue #7225, opened 2017-12-20
[scala] stripPackageName should be configurable


westse:
Needed because some APIs (e.g. Kubernetes) include version information in packages. Stripping causes many generated API and model classes to be named the same, causing a last-one-wins situation which is unacceptable.

1. Version of swagger-codegen used: latest from master branch
1. Language: scala

##### Steps to reproduce

1. Example input spec to produce the problem: https://raw.githubusercontent.com/kubernetes-client/java/master/kubernetes/swagger.json
1. In generated output, notice multiple versions from input spec have been lost due to package stripping.

##### Suggest a fix/enhancement

Make stripping package prefix optional.

PR #7226, opened 2017-12-20
[scala] make stripPackage name configurable via property

westse:
See https://github.com/swagger-api/swagger-codegen/issues/7225

Needed because some APIs (e.g. Kubernetes) include version information in packages. Stripping causes many generated API and model classes to be named the same, causing a last-one-wins situation which is unacceptable.

@clasnake  @jimschubert

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractScalaCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractScalaCodegen.java
@@ -27,6 +27,7 @@ public abstract class AbstractScalaCodegen extends DefaultCodegen {
     protected String modelPropertyNaming = "camelCase";
     protected String invokerPackage = "io.swagger.client";
     protected String sourceFolder = "src/main/scala";
+    protected boolean stripPackageName = true;
 
     public AbstractScalaCodegen() {
         super();
@@ -56,6 +57,10 @@ public abstract class AbstractScalaCodegen extends DefaultCodegen {
 
         if (additionalProperties.containsKey(CodegenConstants.SOURCE_FOLDER)) {
             this.setSourceFolder((String) additionalProperties.get(CodegenConstants.SOURCE_FOLDER));
+        }
+        if (additionalProperties.containsKey(CodegenConstants.STRIP_PACKAGE_NAME) &&
+                "false".equalsIgnoreCase(additionalProperties.get(CodegenConstants.STRIP_PACKAGE_NAME).toString())) {
+            this.stripPackageName = false;
         }
     }
 
@@ -197,7 +202,7 @@ public abstract class AbstractScalaCodegen extends DefaultCodegen {
     }
 
     protected String stripPackageName(String input) {
-        if (StringUtils.isEmpty(input) || input.lastIndexOf(".") < 0)
+        if (!stripPackageName || StringUtils.isEmpty(input) || input.lastIndexOf(".") < 0)
             return input;
 
         int lastIndexOfDot = input.lastIndexOf(".");
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractScalaCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractScalaCodegen.java
@@ -61,6 +61,9 @@ public abstract class AbstractScalaCodegen extends DefaultCodegen {
         if (additionalProperties.containsKey(CodegenConstants.STRIP_PACKAGE_NAME) &&
                 "false".equalsIgnoreCase(additionalProperties.get(CodegenConstants.STRIP_PACKAGE_NAME).toString())) {
             this.stripPackageName = false;
+            additionalProperties.put(CodegenConstants.STRIP_PACKAGE_NAME, false);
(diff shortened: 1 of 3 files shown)


westse:
The "Shippable" check is failing, but for an unrelated reason:  `Package 'esl-erlang' has no installation candidate`. Closing/reopening this PR to see if re-running the check helps.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractScalaCodegen.java
```diff
@@ -27,6 +27,7 @@
     protected String modelPropertyNaming = "camelCase";
     protected String invokerPackage = "io.swagger.client";
     protected String sourceFolder = "src/main/scala";
+    protected boolean stripPackageName = true;
```
ramzimaalej: this option should be false by default to maintain backward compatibility. Not everyone is deploying to Kubernetes.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractScalaCodegen.java
```diff
@@ -57,6 +58,9 @@ public void processOpts() {
         if (additionalProperties.containsKey(CodegenConstants.SOURCE_FOLDER)) {
             this.setSourceFolder((String) additionalProperties.get(CodegenConstants.SOURCE_FOLDER));
         }
+        if (additionalProperties.get("stripPackageName") != null) {
```
ramzimaalej: "stripPackageName" should be added as constant in CodegenConstants

Use additionalProperties.containsKey() to checkout if the property exists in the map.
I would do it this way
`
if (additionalProperties.containsKey(CodegenConstants.STRIP_PACKAGE_NAME) &&
 "true".equalsIgnoreCase(additionalProperties.get(CodegenConstants.STRIP_PACKAGE_NAME)) {
this. stripPackageName = true;
}
`

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractScalaCodegen.java
westse: `stripPackageName = true` *is* retaining backward compatibility. I'm just trying to make it configurable so it can be set to `false`

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractScalaCodegen.java
westse: Regarding `CodegenConstants`, this is currently a scala-only feature, as far as I'm aware. Should language specific options be put there?

westse: Regarding the conditional, I would agree if the default for `stripPackageName` was false, but it is true (as noted above), so that code snippet won't work.  But I'm happy to change the code as long as the logic is correct.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractScalaCodegen.java
ramzimaalej: oh, sorry I misunderstood it.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractScalaCodegen.java
ramzimaalej: The plan is to migrate all scala generators into one. 

so the code snippet could be like this.
`
if (additionalProperties.containsKey(CodegenConstants.STRIP_PACKAGE_NAME) && "false".equalsIgnoreCase(additionalProperties.get(CodegenConstants.STRIP_PACKAGE_NAME)) { this. stripPackageName = false; }
`

westse:
@ramzimaalej , ok PR updated. Look ok now?

ramzimaalej:
Thanks @westse!

@wing328 LGTM

westse:
Can this be merged?

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractScalaCodegen.java
```diff
@@ -57,6 +58,10 @@ public void processOpts() {
         if (additionalProperties.containsKey(CodegenConstants.SOURCE_FOLDER)) {
             this.setSourceFolder((String) additionalProperties.get(CodegenConstants.SOURCE_FOLDER));
         }
+        if (additionalProperties.containsKey(CodegenConstants.STRIP_PACKAGE_NAME) &&
+                "false".equalsIgnoreCase(additionalProperties.get(CodegenConstants.STRIP_PACKAGE_NAME).toString())) {
+            this.stripPackageName = false;
```
jimschubert: This should probably also set `additionalProperties` with the boolean value of `false` so it can be used in templates. Otherwise, it will remain `"false"` and this can be confusing in templates. There are helpers in the super class.

jimschubert:
Latest commit addresses my comments. Thanks!

wing328:
@westse thanks for the PR, which has been merged into 2.4.0

@ramzimaalej @jimschubert @clasnake thanks for reviewing the change.

PR #7238, opened 2017-12-21
[Typescript jQuery] Added optional extra JQueryAjaxSettings at the class and function level

JFCote:
### PR checklist

### Description of the PR

The jQuery generator was missing the possibility to specify options for the request. I added this feature like the other generators + the possibility to specify it at the class level which I think is a must (for example, if you want to use `withCredentials` everywhere.


kenisteward:
I will look at this when I get to a computer. I want to make sure we are
not giving the right to specify urls or anything and only stuff like with
credentials because the generator should handle everything else

JFCote:
@kenisteward Maybe we could only pass `xhrFields` in this case? But we must see my changes as "additional" or "override" parameters. The basic parameters are handled by the generators.

PR #7245, opened 2017-12-22
[typescript-angular] Add option to generate tagged unions

trenneman:
Using the option `taggedUnions` will create a union type for each parent
type instead of extending interfaces. The union types are tagged by using
the discriminator values.

And also:
* Add support for aliases;
* Add support for read-only properties.

### PR checklist

### Description of the PR

This PR adds an option enabling [Tagged union](https://en.wikipedia.org/wiki/Tagged_union) types for [TypeScript](https://www.typescriptlang.org/docs/handbook/advanced-types.html). For the [pet-example](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md) this outputs the following models:

```
// pet.ts

export type Pet = Cat | Dog;

// cat.ts

/**
 * A representation of a cat
 */
export interface Cat { 
    name: string;
    petType: 'Cat';
    /**
     * The measured skill for hunting
     */
    huntingSkill: Cat.HuntingSkillEnum;
}

export namespace Cat {
    export type HuntingSkillEnum = 'clueless' | 'lazy' | 'adventurous' | 'aggressive';
    export const HuntingSkillEnum = {
        Clueless: 'clueless' as HuntingSkillEnum,
        Lazy: 'lazy' as HuntingSkillEnum,
        Adventurous: 'adventurous' as HuntingSkillEnum,
        Aggressive: 'aggressive' as HuntingSkillEnum
    }
}

// dog.ts

/**
 * A representation of a dog
 */
export interface Dog { 
    name: string;
    petType: 'Dog';
    /**
     * the size of the pack the dog is from
     */
    packSize: number;
}
```

as opposed to

```
// pet.ts

export interface Pet { 
    name: string;
    petType: string;
}

// cat.ts

/**
 * A representation of a cat
 */
export interface Cat extends Pet { 
    /**
     * The measured skill for hunting
     */
    huntingSkill: Cat.HuntingSkillEnum;
}
export namespace Cat {
    export type HuntingSkillEnum = 'clueless' | 'lazy' | 'adventurous' | 'aggressive';
    export const HuntingSkillEnum = {
        Clueless: 'clueless' as HuntingSkillEnum,
        Lazy: 'lazy' as HuntingSkillEnum,
        Adventurous: 'adventurous' as HuntingSkillEnum,
        Aggressive: 'aggressive' as HuntingSkillEnum
    }
}

// dog.ts

/**
 * A representation of a dog
 */
export interface Dog extends Pet { 
    /**
     * the size of the pack the dog is from
     */
    packSize: number;
}
```


TiFu:
This solution assumes that the parent type is abstract and cannot be returned by the API.
Is returning the parent class prohibited by the openAPI specification if a discriminator field is present? 
From reading the spec i'm not sure that is the case.

kenisteward:
@tifu you beat me to it. Also I thought discriminators weren't fully
supported for less than 3.0

kenisteward:
Also what is the benefit of using unions ? If you have a base type wouldn't
it be preferred to actually have the type generated.

trenneman:
@TiFu this indeed assumes the parent type is abstract and as far as I understand this is what the [spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#discriminatorObject) suggests: "If the discriminator value does not match an implicit or explicit mapping, no schema can be determined and validation SHOULD fail."

But even if this was not the case, you can always decide not to enable tagged unions.

@kenisteward I know the support for discriminators has been improved in the 3.0 spec. Nevertheless they are already there in 2.0, at least for Spring & Elm (and probably others). Currently, this still requires the vendor extension `x-discriminator-value` instead of the mapping introduced in 3.0.

To name a few benefits of tagged unions:

* Types are guaranteed and distinctable with tagged unions. That is
    ```
    const cat: Cat = {
        name: 'Hachi',
        petType: 'Dog',
        huntingSkill: 'clueless'
    }
    ```
    would be fine with inheritance, not with tagged unions. It's like [making impossible state impossible](https://www.youtube.com/watch?v=IcgmSRJHu_8); 
* For TypeScript, there is no difference when using generics or types, e.g. `function clone<T extends Pet>(pet: T): T {...}` will work in either case. There are no classes here;
* As described by [typescriptlang.org](https://www.typescriptlang.org/docs/handbook/advanced-types.html) or even clearer in this [blog post](https://blog.mariusschulz.com/2016/11/03/typescript-2-0-tagged-union-types) (see `describePaymentMethod`), it is very convient to switch over all types (and to ensure you actually process all types) and to have TypeScript know what type you are working with without adding type annotations.

TiFu:
Correct me if I'm wrong, but the part you quoted does not prohibit an API Endpoint returning a Pet (just a Pet, not a subclass of Pet). Maybe @wing328 can clarify the interpretation of the OpenAPI spec..

If this is indeed the case I wouldn't feel very comfortable adding this option to the TypeScript code generator. Generating invalid code due to an option is very bad from a user standpoint, because it violates the contract defined by the code generator: valid json in, valid code out. 
If this is added, it will need very good documentation describing exactly in which cases it breaks.

trenneman:
@TiFu to me the spec seems to describe just that having this generic Pet is not allowed. Also, having a strict set of sub-types makes the most sense to me. But I agree it is not specified very clearly. What do you think, @wing328 ?

If this is not the case I agree it should be made very clear what the impact of enabling this option is. However, I do feel it can be a great addition to handling your remote data with TypeScript.

trenneman:
@wing328 Can you please help us out?

I was thinking we could also make the option allow two modes: one with only tagged unions and one where an additional type allows for any value, like
```
export type Pet = Cat | Dog | OtherPet;

// where

interface Cat {
  petType: 'cat';
  ...
}

interface Dog {
  petType: 'dog';
  ...
}

interface OtherPet {
  petType: string; // <-- this allows for any pet
  ...
}
```

wing328:
@trenneman I'll take a look by this weekend and reply back

wing328:
My take is to add the option given that it's defaulted to false and as you guys have pointed out, the support of discriminator in  OpenAPI 2.0 was not very mature (but that should not prevent us from adding additional feature to the generator)

trenneman:
Thanks, @wing328. Would this option still be legitimate in OpenAPI 3.0 then?

@TiFu what do you think of @wing328 's suggestion? Does defaulting to false and adding documentation seem feasible to you?

TiFu:
I am fine with that. @kenisteward what about you?

We just have to remember, that we might have to remove it again for OpenAPI Spec 3.0. Let's see if William knows more about that.

kenisteward:
Looks like I was late to the party :( @tifu I think I would have preferred
with everything under 3 not having the feature since sub 3 doesn't support
descriptors

This should have been a feature written for 3.0.0 with that support in mind

kenisteward:
With 3.0.0 we wouldn't even need an option because it should be known as
default when you have a descriminator

TiFu:
`since sub 3 doesn't support descriptors` 
@kenisteward Descriptors or discriminators? The latter is included in Open API Spec 2

Does 3.0.0 clarify the issues with non-abstract base types? My understanding is that it still doesn't address this issue.

trenneman:
That's a quick merge all of the sudden. I do not know the details of 3.0, but I agree this should only be in if it complies to that spec. And if so, we probably want to add the option there as well.

kimamula:
@wing328 Is this to be included in the next release?

In addition to the concern described above, I found another problem regarding `readOnly`.
`readOnly` in OpenAPI "[means that it MAY be sent as part of a response but MUST NOT be sent as part of the request](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-13)", while declaring properties as `readonly` in TypeScript has no effect on preventing them from being sent as part of the request.
Therefore, the change introduces unnecessary restriction only to annoy users.

PR #7254, opened 2017-12-26
[haskell-http-client] handle */* mimetype correctly &  split api modules & allowNonUniqueOperationIds

jonschoning:
1. This change allows the user to choose any any content type for serialization when the spec consumes `*/*`.  (The current behavior forces the user to use `MimeAny` which does does not provide a method of serialization)

2. Also, generate separate api modules, for each child api. (similar to how the other generators split out the APIS into separate files)    
Instead of putting all the API operations into one file, each API gets its own module/file under API/ (e.g. API/Store.hs)

3. Add add a cliOption "allowNonUniqueOperationIds" (default False).  When cli option "allowNonUniqueOperationIds" is true, allows *different* API modules to contain the same operationId, and then each API must be imported qualified.


guoshimin:
Thanks for doing this, @jonschoning! How hard would it be to break up Model.hs into submodules as well? Can the same technique apply?

jonschoning:
The same technique cannot be naively applied to create a module per model, because unlike APIs, models often depend on other models, and there can even be circular depdencies between models.   

In theory one could figure out the imports for each model, but Haskell definitely doesn't allow circular imports, and Kubernetes has those too.

Notice other languages put all models into the same namespace, just with separate files. I don't think they try to figure out the imports needed to put each model in it's own namespace.

guoshimin:
I see. No big deal then.

PR #7257, opened 2017-12-26
[erlang-client] Erlang request utils

tsloughter:
### Description of the PR

The main change here is simplifying work on the core http request logic. This will make it easier for me, or anyone else who decides to work on it, to begin adding features around authorization and more complex handling of responses.

I've also updated the packageName for the petstore sample.


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ErlangClientCodegen.java
```diff
@@ -117,6 +117,24 @@ public String getSwaggerType(Property p) {
         return type;
     }
 
+    public static String underscore2(String word) {
```
wing328: @tsloughter Is there any issue with `underscore`? If that's the case, I would suggest fixing `underscore` instead of creating `underscore2`.

tsloughter: Ah yea, bad name there. It isn't an issue that can be fixed in the original. The problem is `underscore` replaces `.` with `/`. Makes sense for creating paths in languages that utilize directories but for Erlang with its flat src directory and namespace we need to replace `.` with `_` as well.

wing328: @tsloughter What about using `undercore(str.replaceAll(".", "_"))` instead?

tsloughter: Ah, good idea, fixing.

tsloughter:
Updated per your comment @wing328

PR #7258, opened 2017-12-26
[Go] Fix operation files clobbering model files.

### LINKED ISSUE #6942: master branch does not produce compilable go code.
##### Description

Trying to build a go client creates a go library that does not compile. Tried on master. When trying to build on 2.2.3, it produces working go code that does not have the missing API declaration.

##### Swagger-codegen version

Master

https://sftp.solace.com/download/VMR_SEMPV2_SCHEMA_YAML

```
java -jar ./swagger-codegen-cli.jar generate -i https://sftp.solace.com/download/VMR_SEMPV2_SCHEMA_YAML -l go -DpackageName=semp_client -o semp_client
```
##### Steps to reproduce

``` shell
$ java -jar ./swagger-codegen-cli.jar generate -i https://sftp.solace.com/download/VMR_SEMPV2_SCHEMA_YAML -l go -DpackageName=semp_client -o semp_client
$ cd semp_client
$ go get -v 
./about_api_response.go:15:8: undefined: AboutApi
```

Clearly AboutApi is not created here, even though there are references of it everywhere.

antihax:
### PR checklist

### Fix operation files clobbering model files.
Prepend `model_` to all model file names to operation groups with the same name from clobbering the model files. This should be a non-breaking change. This resolves #6942


bvwells:
@antihax I was wondering if this completely fixes the issue or just makes it more unlikely? It feels that perhaps the right fix is to separate the models and operations into separate sub-packages, but this would obviously be a more intrusive fix. Any thoughts on this?

antihax:
I had thought of both. 

It think it would be better to keep them in the main package so users can easily access the models directly. I do this in a few cases where I wish to embed results with other data in a struct before encoding it to json or some other format.

It shouldn't interfere much as the operations end up on the client, and the models remain in the package.

bvwells:
My worry was that the issue is still there if an operation is called ModelAboutApi and a model is called AboutApi. Although this is perhaps a rather perverse example. 

Personally I'm quite used to having the operations and models separated having used go-swagger and other swagger-codegen generators such as the typescript-angular client. 

If we go with this fix, it might be something we look to get feedback from users.

My only other comment is that this change also effects the go-server so it might be worth running the script to update the go-server example.

antihax:
Should still work, the model file will be `model_aboutapi.go` and the operation group would be `modelaboutapi.go`. It's somewhat rare this should happen; the operation group gets defined by the tag, in the case below its `about` which ends us up with an `AboutApi` struct holding all the operations for this tag.

```
  /about/api:
    get:
      description: 'Gets an API Description object.


        A SEMP client authorized with a minimum access scope/level of "global/none"
        is required to perform this operation.


        This has been available since 2.2.0.'
      operationId: getAboutApi
      parameters: []
      produces:
      - application/json
      responses:
        '200':
          description: The API Description object's attributes, and the request metadata.
          schema:
            $ref: '#/definitions/AboutApiResponse'
        default:
          description: Error response
          schema:
            $ref: '#/definitions/SempMetaOnlyResponse'
      security:
      - basicAuth: []
      summary: Gets an API Description object.
      tags:
      - about
```

bvwells:
Wouldn't you end up with a file called model_about_api.go if the tag was called 'model_about' and the same filename if the model definition was called AboutAPI?

antihax:
Sure would. Let me see if there is a better way of separating these two then.

bvwells:
Definitely an edge case, but I always think these cases come back to bite at some point. Thanks for taking a look.

antihax:
I tried putting them into a sub directory, however this is no longer considered in the package so a new package would have to be created for the models which is going to complicate things a bit.

Going to see if i can find a different solution before trying to go down that road.

bvwells:
Yes, I think the APIs and models would have to be in their own packages. It might be that something equivalent to a GOPATH would need to be passed to the generator to correctly sort out the paths of the packages.

antihax:
Currently I renamed all the operation files to `api_<operationtag>.go` and the models remaining as `model_<model>.go` which should resolve any possible name conflicts and remains a non-breaking change for users. It also helps organize the folder a little better.

Still a bit leery about separating them into packages, it would require a lot of template changes and further discussions about what those packages should be named as that determines how a user imports them.

For example:
```
import (
	"github.com/someapi/generatedpackage"
	"github.com/someapi/generatedpackage/model"
)

type UsersModel {
	RequestID 	int64
	Output		[]model.ReturnedModel
}

func main() {
	ret := apiClient.GetSomeModel(reqID)
	doSomething(&UsersModel { 
		RequestID: reqID, 
		Output: ret,
	})
}
```

Where as currently:

```
import (
	"github.com/someapi/generatedpackage"
)

type UsersModel {
	RequestID 	int64
	Output		[]generatedpackage.ReturnedModel
}

func main() {
	ret := apiClient.GetSomeModel(reqID)
	doSomething(&UsersModel { 
		RequestID: reqID, 
		Output: ret,
	})
}
```

bvwells:
Looks good. Thanks for taking another look at the issue. 

My personal feeling is that it would be good to separate the models and APIs into separate packages, but I can understand your reservations about this. Especially as it is a breaking change. Having the models and APIs in the same package means that it is still possible (however quite unlikely) that model and API struct names can clash. Separating them would fix this.

Good to go from my point of view!

wing328:
@antihax when you've time, I wonder if you can rebase on the latest 2.4.0 as I merged another PR, which contains code formatting enhancements.

@bvwells thanks for reviewing the change.

antihax:
I created #7337. Something went horribly wrong with my local repo.

PR #7260, opened 2017-12-27
Java8 fix

amcginn:
As of 2.3.0 release, a change had been made that used Java 8 methods. This PR changes those calls to the equivalent in Java 7. Also reverts the quick-fix of updating the minimum required version.


wing328:
@amcginn thanks for the PR. I've also tested with CI (travis) to confirm it's working for Java7.

We'll use Travis rather than CircleCI to catch this moving forward.

PR #7275, opened 2017-12-29
[Scala] Fix inconsistency between *Api and *ApiAsyncHelper default parameters

### LINKED ISSUE #7273: [Scala] Inconsistency between *Api and *ApiAsyncHelper default parameters
##### Description

In #6538, we defaulted non-required Option parameters to `None` for both sync and async methods in the *Api classes. However, the helper methods in *ApiAsyncHelper (which the methods in *Api simply wrap) default to `defaultValue` which is inconsistent and also error prone (will open a separate issue for this, but see https://github.com/swagger-api/swagger-codegen/pull/6538#discussion_r140950422 for context /cc @wing328).

##### Swagger-codegen version

2.3.0

gmarz:
### PR checklist

- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language. (/cc @clasnake @jimschubert)

### Description of the PR

Closes #7273


gmarz:
> Unless I'm reading this incorrectly, the change removes support for default values for types that are optionally required and have a default defined.

That's correct. However, I believe there are > 1 issues with defaulting to `defaultValue` at the moment (in addition to https://github.com/swagger-api/swagger-codegen/pull/6538#discussion_r140950422, I also think `defaultValue = "someString"` ends up as `Some(someString)` which is also wrong...but I need to test that again to confirm).

More importantly, the `*Api` methods at the moment default to `None` so this change at least makes `*ApiAsyncHelper` consistent with that behavior until we fix the default values.

jimschubert:
@gmarz please see my pr #7286, where I added integration tests referring to this PR and changing from the `None` default change from the other linked PR back to supporting default values.

I don't understand your comment that a spec defining `defaultValue = "some string"` resulting in the generated code ending up at `Some(some string)` is wrong. Are you saying that the string isn't quoted? If so, that's an easy enough fix, and I can make the change with an appropriate test modification in PR #7286.  I think string is the only primitive value that would have this issue in Scala (if it's not be properly quoting).

Sorry, I meant to tag you in that other pr as it directly relates to and partially conflicts with this pr.

gmarz:
> Are you saying that the string isn't quoted? 

Yep, sorry for not being clear. This is what I observed during my testing of 2.3.0, but would be great if you could confirm as part of #7286 and even better if you can add a fix!

> I think string is the only primitive value that would have this issue in Scala (if it's not be properly quoting).

Sounds right.

jimschubert:
@gmarz I'm working on a fix. I have the string issue fixed locally, but there were issues with string types having formats (date, date-time, byte, binary). I'll try to have those fixed and added to that PR within a day or two.

Thanks for raising that as an issue, because it was a pretty big one.

jimschubert:
I updated #7286 to support all previously missing functionality, and this should allow for full support of default values. Please review, as I think that PR would supersede this one.

wing328:
Closed via https://github.com/swagger-api/swagger-codegen/pull/7286 instead

Issue #7276, opened 2017-12-29
[akka-scala] exception template FileNotFoundException


shijinkui:
1. replace the template path from `File.separator` to `/` to resolve template not found exception.
2. change `object` API to `class`, more easy use for java developer

##### Description

```
java.lang.RuntimeException: Could not generate model 'xxxx'
        at io.swagger.codegen.DefaultGenerator.generateModels(DefaultGenerator.java:409)
        at io.swagger.codegen.DefaultGenerator.generate(DefaultGenerator.java:721)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
        at java.lang.reflect.Method.invoke(Unknown Source)
Caused by: java.io.FileNotFoundException: /akka-scala/\model.mustache
        at com.github.jknack.handlebars.io.URLTemplateLoader.sourceAt(URLTemplateLoader.java:70)
        at com.github.jknack.handlebars.Handlebars.compile(Handlebars.java:357)
        at com.github.jknack.handlebars.Handlebars.compile(Handlebars.java:343)
        at io.swagger.codegen.DefaultGenerator.getHandlebars(DefaultGenerator.java:1012)
        at io.swagger.codegen.DefaultGenerator.processTemplateToFile(DefaultGenerator.java:737)
        at io.swagger.codegen.DefaultGenerator.generateModels(DefaultGenerator.java:396)
        at io.swagger.codegen.DefaultGenerator.generate(DefaultGenerator.java:721)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
        at java.lang.reflect.Method.invoke(Unknown Source)

```

##### Swagger-codegen version

I use api 3.0 and swagger codegen 3.0-snapshot, `akka-scala`

```
  "io.swagger" % "swagger-parser-v3" % "2.0.0-rc1",
  "io.swagger" % "swagger-codegen" % "3.0.0-SNAPSHOT"
```

##### Swagger declaration file content or url

![image](https://user-images.githubusercontent.com/648508/34430344-efae3364-ec9e-11e7-94af-d574a6975cd2.png)

PR #7277, opened 2017-12-29
[ISSUE-7276][akka-scala] exception template FileNotFoundException

shijinkui:
### PR checklist
1. replace the template path from `File.separator` to `/` to resolve template not found exception.
2. change `object` API to `class`, more easy use for java developer

### Description of the PR

https://github.com/swagger-api/swagger-codegen/issues/7276

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/AbstractGenerator.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/AbstractGenerator.java
@@ -108,7 +108,7 @@ public abstract class AbstractGenerator {
         }
             
         // Fall back to the template file embedded/packaged in the JAR file...
-        return config.embeddedTemplateDir() + File.separator + templateFile;
+        return config.embeddedTemplateDir() + "/" + templateFile;
     }
 
     public String readResourceContents(String resourceFilePath) {


shijinkui:
ping @farcaller @JackDanger

farcaller:
I don't really know scala enough to lgtm that.

shijinkui:
> I don't really know scala enough to lgtm that.
Thanks.

ping @clasnake  @jimschubert

modules/swagger-codegen/src/main/java/io/swagger/codegen/AbstractGenerator.java
```diff
@@ -108,7 +108,7 @@ public String getFullTemplateFile(CodegenConfig config, String templateFile) {
         }
             
         // Fall back to the template file embedded/packaged in the JAR file...
-        return config.embeddedTemplateDir() + File.separator + templateFile;
```
wing328: @shijinkui the reason why we use File.separator is to make sure it works with Windows, Linux and Mac.

We do have `bin/windows/akka-scala-petstore.bat` to cover `akka-scala` in the Windows platform.

Are you encountering the issue with 2.3.0 JAR or a local JAR built from the latest master?

jimschubert:
Rather than change object to class, can we create a Java API here? objects are basically singletons in Scala, so high traffic apis would be creating many object instances with this change.

modules/swagger-codegen/src/main/java/io/swagger/codegen/AbstractGenerator.java
shijinkui: latest master: 3.0 snapshot

shijinkui:
> Rather than change object to class, can we create a Java API here? objects are basically singletons in Scala, so high traffic apis would be creating many object instances with this change.

Create New Java API maybe let users confuse. Keep only one API may be better. `new API()` and `API` have less different for usage even the runtime memory location. `object` is util class.

wing328:
I second with @shijinkui to avoid singletons in the auto-generated Scala client.

jimschubert:
It's pretty common for Scala libs to expose Java interfaces. The reason I asked is because you'll still have an issue with the Headers object needing to be referenced with the `$` for object access, right? You'll also have to account for accessing Scala collections and Option types.

As for object vs class, I usually don't disagree there. object is not only meant for a utility type. It is a Scala practice to have pure types and to be functional in nature. An object allows for importing member functions statically, and any state will be shared by all consumers. With a class, you have new state per instance and, although you can import members from an instantiated type, you have no guarantees that state remains the same. This can be an issue in Akka. So I don't think the change from object to class would be desirable for most Scala consumers as it would break existing static imports and may introduce concurrency bugs with Akka (I would have to run some tests to verify the last statement).

jimschubert:
Please note... adding companion object may alleviate my concerns.

That is, changing this to trait rather than class and doing something like

```
object {{className}} extends {{className}}
```

Then your type is Java accessible and you continue to have the object API.

shijinkui:
hi, @jimschubert Maybe there're some probably bugs for akka if don't use `object`.Trait is `interface`,  can't be used directly. 

I delete the object changing, and will create new issue for discuss.

modules/swagger-codegen/src/main/java/io/swagger/codegen/AbstractGenerator.java
wing328: I just couldn't the issue with missing file in the customized template folder: https://ci.appveyor.com/project/WilliamCheng/swagger-codegen-wh2wu/build/test_windows_template-7192

Branch: https://github.com/swagger-api/swagger-codegen/tree/test_windows_template

wing328:
cc @ramzimaalej  as well

modules/swagger-codegen/src/main/java/io/swagger/codegen/AbstractGenerator.java
wing328: We cannot merge this as this change in abstract generator will impact all generators

ramzimaalej:
I agree with what @jimschubert said. I'm concerned about concurrency, and object is the best option to avoid them.

shijinkui:
> I agree with what @jimschubert said. I'm concerned about concurrency, and object is the best option to avoid them.

Thanks. Had reset object. This PR Only fix the runtime error.

modules/swagger-codegen/src/main/java/io/swagger/codegen/AbstractGenerator.java
ramzimaalej: @shijinkui I was not able to reproduce the issue. I'm on Mac. Also as @wing328 mentioned your solution is not platform independent.

shijinkui:
hi, @wing328 @jimschubert 
Is this PR OK?

ramzimaalej:
@shijinkui we cannot merge your proposed solution as it's not os independent. If you can tell me the steps to reproduce, I will take a closer look.

shijinkui:
@ramzimaalej #7276  I will provide a example project later.

PR #7295, opened 2018-01-02
Split out a RequestOptions function.

brendandburns:
### Description of the PR
Split each API method in half to add a public facing _operation_RequestOptions() method. The purpose of this method is to queue everything up ready to execute, but not actually execute it.

This allows the user to intercept the request, make modifications, and execute it on their own. This is useful (for example) when you want to use a WebSockets protocol that nonetheless re-uses all of the auth and everything else from a RESTful API, or when a RESTful API has some methods that return raw streams of data instead of JSON.

Typescript technical committee:


wing328:
@brendandburns thanks for the PR. From what I can tell, it's not a breaking change.

If no concern from @TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx, I'll merge it into 2.4.0 this weekend.

TiFu:
I can't seem to find the change in this PR which removes the ObjectSerializer from api.ts (the diff which is hidden by default because it's too big).  
Another issue is that the version in VERSION changed from 2.3.0 to 2.2.3.

@wing328 Do you know to which PR these changes are related or if the changes aren't an issue?

wing328:
Looks like @brendandburns is not using the latest master when creating the PR as all the localVar variable prefix has been removed.

Please rebase on the latest master and update the TS node samples.

Issue #7262, opened 2017-12-27
[Angular-typescript] Regression/bug: Parameters with an _ ignore naming convention 


HermenOtter:
##### Description
Parameters with an _ ignore naming convention since a few days. This bug occurred a few months ago as well and has been fixed at the time.

##### Swagger-codegen version

master(54ce4caafc3909056630d97b428b2e8a9022a6f9 and  8ae11847d3e2300bc79cc9a08c20c187cd2d4c15) are not working as expected.

##### Swagger declaration file content or url
```yaml
/sadr/v1/soorten_adres/{code_soort_adres}:
    parameters:
    - name: "code_soort_adres"
      in: "path"
      required: true
      type: "string"
```
##### Command line used for generation

```java -jar swagger-codegen-cli.jar generate -i https://api.swaggerhub.com/apis/Centric/Belastingen_Subjecten/0.2.4 -l typescript-angular -c=./SCGConf.md```

**SCGConf.md**
`{"modelPropertyNaming": "original"}`

**Code generation**
Generates `code_soort_adres` as parameter for the method, but when used in that same method it is called as `codeSoortAdres` which obviously doesn't work. 

`public getSoortenAdresByCode(code_soort_adres: string,`
`return this.httpClient.get<any>(${this.basePath}/.../${encodeURIComponent(String(codeSoortAdres))}`

##### Related issues/PRs

[7201, regression](https://github.com/swagger-api/swagger-codegen/issues/7201)

##### Suggest a fix/enhancement

I believe it is has something to do with issue 7201.

@wing328 @JFCote @kenisteward

meisterlampe:
I have the same problem after upgrading to v2.3.0

JFCote:
@HermenOtter @meisterlampe I'm the one responsible for this regression (I think). It was fixing other stuff and it is working like the other generators but there seems to be something missing. I can assure you that it was tested against our petstore samples and it was tested by a user who had problem with that.

Anyway, I will have a look at this today and try to come up with a fix as soon as possible. I'll keep you updated here.

JFCote:
@HermenOtter  @meisterlampe 
I have debugged this and I am relieved to see that my code is not responsible for this regression. This problem is created by this very weird loop:
https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngularClientCodegen.java#L268

[...]

Issue #6632, opened 2017-10-07
When is 'modelPropertyNaming' config option necessary?


will-beta:
I have some APIs coded with c#, conforming to PascalCase naming conversion for the property of classes (also in the swagger document).
After I generated a Typescript-Node client library with [swagger-online-editor](https://swagger.io/swagger-editor/) , to my surprise I found that the generated properties of classes are camelCase.
So after some researching, I found the 'modelPropertyNaming' config option, which can be set as camelCase | PascalCase | Original.
My question is: In what kind of scenarios, is 'modelPropertyNaming' config option necessary?

Thanks in advance !

wing328:
@will-beta Some generators (e.g. JavaScript) do not have a proper way to map between JSON key names and object property names (JSON serialization/deserialization) and therefore we introduced the 'modelPropertyNaming" option to let the users specify the naming convention instead of following the style guide of the respective language.

If I remember correctly, the community has added proper JSON serialization/deserialization to TS-node so please give it a try and let us know if you find any issue.

PR #7313, opened 2018-01-04
Fix issue #7262 

JFCote:
The problem was that camelCase naming was forced only in this part of the code when everywhere else it is configurable.

### PR checklist

### Description of the PR

Fix issue #7262

This bug seems to have been introduced by this commit: https://github.com/swagger-api/swagger-codegen/commit/a63e3f14f58cdce57be0f1091905d8653eb26626
So I'm tagging its author for review: @defmonk0 

In summary, I've made sure that the parameters names in the path received the same treatment everywhere else instead of applying a camelCase by default.


defmonk0:
The issue I see with this change is it essentially reverts to the original functionality before a63e3f1 (it would simply use the loop instead of the original regex method). This would simply cause the issue I was attempting to fix.

As described in [my response](https://github.com/swagger-api/swagger-codegen/issues/7262#issuecomment-355565524) to issue #7262, it seems we need to change this code to be compatible with the `modelPropertyNaming` option. The functionality of this option is described in #6632. 

My change is compatible when the option is not set (defaults to `camelCase`, I believe, per Angular style guide). This change is compatible when set to `original`, as described in #7262. We are completely missing compatibility with the option set to `PascalCase`.

Pinging @wing328, since he merged my changes initially and is active on every mention of this option I can find. ;P I imagine he'll have a better notion of how we're supposed to be supporting this, since I didn't even know it existed before now. Lol.

JFCote:
@defmonk0 My change is compatible with the naming option since it uses `toVarName` which is used in all other generators to take care of this:
See here: https://github.com/swagger-api/swagger-codegen/pull/7313/files#diff-ad60e100e050d5ef52943803a2adc5fdR279

By calling this, it will check what is the prefered naming convention and replace the param name by the good naming in the path.

defmonk0:
@JFCote Holy crap, I completely missed that function call nested inside the `.append`. My bad. Lol.

With that there, yes, I agree, this should work perfectly well. Looks fine to me.

wing328:
@JFCote thanks for the quick fix.

@defmonk0 thanks for reviewing the change.

PR #7323, opened 2018-01-05
Make generated services conform to tslint rules set by angular-cli

manbearwiz:
### PR checklist

### Description of the PR

Make generated services conform to tslint rules set by angular-cli; particularly:

- [`import-spacing`](https://palantir.github.io/tslint/rules/import-spacing/)
- [`max-line-length`](https://palantir.github.io/tslint/rules/max-line-length/)
- [`no-trailing-whitespace`](https://palantir.github.io/tslint/rules/no-trailing-whitespace/)
- [`prefer-const`](https://palantir.github.io/tslint/rules/prefer-const/)
- [`quotemark`](https://palantir.github.io/tslint/rules/quotemark/)
- [`semicolon`](https://palantir.github.io/tslint/rules/semicolon/)

```json
{
  "import-spacing": true,
  "max-line-length": [
    true,
    140
  ],
  "no-trailing-whitespace": true,
  "prefer-const": true,
  "quotemark": [
    true,
    "single"
  ],
  "semicolon": [
    true,
    "always"
  ]
}
```


kenisteward:
I've discussed this before and this should not be merged. Different
projects have different rules and therefore we cannot as a library
generator confo to them all.  @sebastianhaas agrees.

The idea for ts lint is to make sure code confirmes to rules so when
multiple people are writing it it's all written the same way .  In this
case it is generated so no ok should be changing g that code.  Best
practice is to import it as a module external tot he project.

What happens when someone else comes along and has Mac chars 80 instead of
the angular xli standard? If we merge stuff like this the changes would
never end of different people with different preferences. There is no hard
or even soft correct answer when it comes to these types of changes.

TiFu:
Good point, I agree.
I'll keep that in mind for the next lint change PR.

manbearwiz:
The fact that different people have preferences it valid; however, this library should be consistent with itself. As an example, as it is today, quotemarks are mixed so no matter if I have quotemarks set to "single" or "double" I will receive linting errors. Other rules like `prefer-const` are [just considered best practice](https://www.typescriptlang.org/docs/handbook/variable-declarations.html#let-vs-const) and, as far as I know, has no one opposing them. In the end, this library should have a style that it follows, and by far the most popular tslint rules for a typescript-angular project are the ones that come with angular-cli.

kenisteward:
@manbearwiz You make a valid point in the generated code should be consistent with itself. I can agree with this being an effort to get the generated code conforming to best practice TS rules but if that is the case we should always in the future only approve pull requests that conform to those rules to stay consistent.

I'd be fine with approving this if that is the game plan but we'll have to be extremely vigilant in reviews from now on.  At the very least we should add a test that runs TSLint with the said rules and does not accept a change if it doesn't conform to the proper coding rules.

@TiFu @sebastianhaas @Vrolijkx @taxpon thoughts?

TiFu:
I think there are certain advantages to enforcing a style guide. Like @manbearwiz  said, certain best practices like prefer const are not controversial at all.  If we want to enforce this, we definitely need to add tslint to the tests to make it easy to verify. It would also make reading the generated code a bit easier. 

One issue with this approach is, that changes like max line length might not work for every swagger configuration. E.g. for api endpoint methods the line length might depend on the number of input parameters. This inconsistency could become a pain point, if people start to complain about their generated code not honoring tslint. Some other options like white-spacing rules would be useful, because we can then guarantee that those rules are also honored in the generated code. 
What's your opinion on this? 

We should also take into account that we have 3 (or more?) other TS clients. IMO we should use one style for all of them. I would propose to stick to one of the styles used by Microsoft e.g. [TS Contrib](https://github.com/Microsoft/tslint-microsoft-contrib) or [the config used for the TS compiler ](https://github.com/Microsoft/TypeScript/blob/master/tslint.json). Open for discussion though, I'd assume that most of the other styles (angular-cli, ...) are very similar.

manbearwiz:
@TiFu you make a good point about the max line length  for function definitions. We could add something like `// tslint:disable-next-line:max-line-length` above these functions in the template to [disable that rule for specific lines](https://palantir.github.io/tslint/usage/rule-flags/). Alternatively, we could put each input parameter on a new line but I think that would look worse.

As for their being many different tslint configs out there you're totally right.

- [Microsoft/tslint-microsoft-contrib](https://raw.githubusercontent.com/Microsoft/tslint-microsoft-contrib/master/tslint.json)
- [Microsoft/TypeScript](https://raw.githubusercontent.com/Microsoft/TypeScript/master/tslint.json)
- [angular/angular-cli](https://raw.githubusercontent.com/angular/angular-cli/master/tslint.json)
- [angular-cli output](https://raw.githubusercontent.com/angular/angular-cli/master/tests/e2e/assets/1.0.0-proj/tslint.json)
- [angular/angular](https://raw.githubusercontent.com/angular/angular/master/tslint.json)
- [palantir/tslint](https://raw.githubusercontent.com/palantir/tslint/master/tslint.json)

At first look, a lot of these seem to contradict; however, I believe we can satisfy the majority of them by going with the most strict version of a rule. For example, some want line lengths of 140 some want 100 but 100 satisfys both. Most use the `no-console` rules but some have different functions they allow; by disallowing all console functions you satisfy everyone. Because of this, I think it would make sense to start with a rule set and allow changes to it as long as it is more strict and does not contradict the rest of the rules.

itsTeknas:
Hello, I'm new here. I just imported the generated module in my project and saw the TSlint errors (loads of them). A Simple `ng lint --fix` fixed a lot of them directly. Even I was headed to make a contribution to fix these errors and then I stumbled upon this thread.

From a user perspective, when the editor.swagger.io websites lets you generate a client for `typescript-angular` shouldn't we safely assume that the user intends to import this module into an Angular project?

If so then we must strive to make the user experience better by conforming to the **Default** angular specific TSlint style guides. People are always allowed to have project-specific style guides. But the majority of developers will tend to stick to the style guides provided by the angular framework.

Just a suggestion.

TiFu:
I'm not sure if this is possible (I don't know how angular works at all),
but ideally you want to add the generated module as a dependency only.

If that's what you did ignore this comment. Currently travelling and don't
have my laptop with me to look into this more.

It's quite hard to satisfy all rules reliably (e.g. line length) and as
others already mentioned different people might use different rule sets.
I also don't think it's worth the effort to conform strictly to arbitrary
rules in automatically generated code as long as the code is readable.

Which style errors did you encounter?

Am 16.03.2018 16:26 schrieb "Sanket Berde" <notifications@github.com>:

Hello, I'm new here. I just imported the generated module in my project and
saw the TSlint errors (loads of them). A Simple ng lint --fix fixed a lot
of them directly. Even I was headed to make a contribution to fix these
errors and then I stumbled upon this thread.

From a user perspective, when the editor.swagger.io websites lets you
generate a client for typescript-angular shouldn't we safely assume that
the user intends to import this module into an Angular project?

If so then we must strive to make the user experience better by conforming
to the *Default* angular specific TSlint style guides. People are always
allowed to have project-specific style guides. But the majority of
developers will tend to stick to the style guides provided by the angular
framework.

Just a suggestion.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/swagger-api/swagger-codegen/pull/7323#issuecomment-373639068>,
or mute the thread

EvAlex:
The discussion is misleading. Current PR should absolutely be merged. It makes generated code much better. There are some code style rules that are not subjective.

PR #7329, opened 2018-01-06
[Akka-scala] Clean unused dependencies such swagger-core

shijinkui:
Clean unused dependencies such swagger-core

generate petstore by akka-scala.


wing328:
@shijinkui thanks for cleaning up the dependencies.

cc @clasnake  @jimschubert @shijinkui

jimschubert:
Change looks fine.

Removal of Joda converters is a potential breaking change for anyone who has modified ApiInvoker and excluded those changes from generation. It may also affect users passing mappings via command line. We should probably note that in release notes, but this is preferable to including generally unused artifacts. Similarly, users could add the artifact back to the desired build script, and exclude the change in .swagger-codegen-ignore if the artifact is indeed needed.

shijinkui:
@wing328 there is a continuous-integration, it didn't merged in. Please show me the stack, let me check whether is the code fault.

shijinkui:
ping @wing328 resolved break changed.

PR #7338, opened 2018-01-08
[Rust] Implement minimal auth support

euank:
### PR checklist

- [x] Ran the shell script under `./bin/` to update Petstore sample
- [x] Filed the PR against the correct branch: `master`
- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) (pending this being ready or review)

### Description of the PR

This PR implements the minimum required to implement oauth or basic auth on the client side

### DIFF
--- a/modules/swagger-codegen/src/main/resources/rust/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/rust/api.mustache
@@ -2,6 +2,7 @@
 use std::rc::Rc;
 use std::borrow::Borrow;
 use std::borrow::Cow;
+use std::collections::HashMap;
 
 use hyper;
 use serde_json;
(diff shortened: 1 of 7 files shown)


euank:
I've manually verified Basic Auth works as I expect via generating a hydra client (based on [this definition](https://github.com/ory/hydra/blob/v0.10.10/docs/api.swagger.json)) and calling the `.introspect_o_auth2_token` api with correct and incorrect basic-auth credentials.

Next up, oauth2 via more hydra apis.

euank:
cc @frol @farcaller @bjgill, I think this is ready for review.

I've been using this code for an oauth2 flow with success so far

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -39,26 +40,84 @@ impl<C: hyper::client::Connect>{{classname}} for {{classname}}Client<C> {
...
+            let mut query = ::url::form_urlencoded::Serializer::new(String::new());
+{{#queryParams}}
+            query.append_pair("{{baseName}}", &{{paramName}}{{#isListContainer}}.join(","){{/isListContainer}}.to_string());
+{{/queryParams}}
+{{#hasAuthMethods}}
+            for (key, val) in &auth_query {
+                query.append_pair(key, val);
+            }
+{{/hasAuthMethods}} 
+            format!("?{}", query.finish())
```
frol: Since the `?` will end up even for an empty query string, I would just put in the `format!` below and just return `query.finish()` here.

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -39,26 +40,84 @@ impl<C: hyper::client::Connect>{{classname}} for {{classname}}Client<C> {
...
+        {
+            match configuration.basic_auth {
+                Some(ref auth_conf) => {
+                    let auth = hyper::header::Authorization(
+                        hyper::header::Basic {
+                            username: auth_conf.0.to_owned(),
+                            password: auth_conf.1.to_owned(),
+                        }
+                    );
+                    auth_headers.insert("Authorization".to_owned(), format!("{}", auth));
```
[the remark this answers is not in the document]
euank: `Authorization` doesn't have a `to_string` method, just a `fmt::Display` impl. I couldn't find a cleaner way to do this.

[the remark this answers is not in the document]
frol: > This trait is automatically implemented for any type which implements the `Display` trait. As such, `ToString` shouldn't be implemented directly: `Display` should be implemented instead, and you get the `ToString` implementation for free.

https://doc.rust-lang.org/stable/std/string/trait.ToString.html

There might be some black magic involved, but as far as I know, it should have `.to_string()` since it implements the Display trait.

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -39,26 +40,84 @@ impl<C: hyper::client::Connect>{{classname}} for {{classname}}Client<C> {
...
         let configuration: &configuration::Configuration<C> = self.configuration.borrow();
 
+{{#hasAuthMethods}}
+        let mut auth_headers = HashMap::<String, String>::new();
+        let mut auth_query = HashMap::<String, String>::new();
+{{#authMethods}}
+{{#isApiKey}}
+        match configuration.api_key {
+            None => (),
+            Some(ref apikey) => {
```
bjgill: A minor style point - might this be clearer as `if let Some(ref apikey) = configuration.api_key { ...`?

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -39,26 +40,84 @@ impl<C: hyper::client::Connect>{{classname}} for {{classname}}Client<C> {
...
+                let val = match apikey.prefix {
+                    Some(ref prefix) => format!("{} {}", prefix, key),
+                    None => key,
+                };
+            {{#isKeyInHeader}}
+                auth_headers.insert("{{keyParamName}}".to_owned(), val);
+            {{/isKeyInHeader}}
+            {{#isKeyInQuery}}
+                auth_query.insert("{{keyParamName}}".to_owned(), val);
+            {{/isKeyInQuery}}
```
bjgill: I assume `isKeyInHeader` and `isKeyInQuery` are guaranteed to be mutually exclusive?

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -39,26 +40,84 @@ impl<C: hyper::client::Connect>{{classname}} for {{classname}}Client<C> {
...
+                Some(ref auth_conf) => {
+                    let auth = hyper::header::Authorization(
+                        hyper::header::Basic {
+                            username: auth_conf.0.to_owned(),
+                            password: auth_conf.1.to_owned(),
+                        }
+                    );
+                    auth_headers.insert("Authorization".to_owned(), format!("{}", auth));
+                }
+                None => {}
```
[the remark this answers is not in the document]
bjgill: Or indeed `if let ...` (as above)

modules/swagger-codegen/src/main/resources/rust/configuration.mustache
```diff
@@ -1,10 +1,23 @@
 {{>partial_header}}
 use hyper;
+use std::collections::HashMap;
 
 pub struct Configuration<C: hyper::client::Connect> {
   pub base_path: String,
   pub user_agent: Option<String>,
   pub client: hyper::client::Client<C>,
+
```
bjgill: Would be nice not to have this empty line here.

modules/swagger-codegen/src/main/resources/rust/configuration.mustache
```diff
@@ -1,10 +1,23 @@
 {{>partial_header}}
 use hyper;
+use std::collections::HashMap;
```
bjgill: Is this not going to yield warnings? It might be nice to allow the warning or only add this `use` statement if necessary.

modules/swagger-codegen/src/main/resources/rust/api.mustache
euank: As I understand it yes, but these parameters aren't really well documented so I could be wrong.

modules/swagger-codegen/src/main/resources/rust/configuration.mustache
euank: It would, but at this point we're too far gone on warnings.

I would like to do a pass dedicated to squashing warnings at some point, which I think will be easiest with a certain amount of refactoring. For now, I'm okay with living with another warning in the sea of them.

euank: To be specific, I think refactoring the shared code (the whole `fn operationId` bit) to be one generic function which takes configurable things as parameters would result in more readable code that would be easier to eliminate warnings from... if only because all the code would be there and just not be called necessarily. Each operation would then just be a short function calling the big generic "do_call" style function with appropriate args.

I recall some of the other clients take this approach, but I don't have references handy because I dug into this some months ago.

modules/swagger-codegen/src/main/resources/rust/api.mustache
euank: Thank you for the suggestion. I agree that's cleaner and I'm not quite sure why I didn't do that the first time.

modules/swagger-codegen/src/main/resources/rust/api.mustache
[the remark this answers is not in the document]
euank: Worked just fine, thanks for the suggestion/pointer!

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -39,26 +40,84 @@ impl<C: hyper::client::Connect>{{classname}} for {{classname}}Client<C> {
...
+                auth_headers.insert("{{keyParamName}}".to_owned(), val);
+            {{/isKeyInHeader}}
+            {{#isKeyInQuery}}
+                auth_query.insert("{{keyParamName}}".to_owned(), val);
+            {{/isKeyInQuery}}
+            }
+        };
+{{/isApiKey}}
+{{#isBasic}}
+        {
```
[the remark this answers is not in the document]
euank: No reason it turns out. I think I was worried about some lifetime thing, but it compiles fine without them on stable. Removed.

euank:
Addressed comments (excepting the fact that there are still warnings, I'd like to punt on that per my reply)

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -39,26 +40,71 @@ impl<C: hyper::client::Connect>{{classname}} for {{classname}}Client<C> {
...
+        };
+{{/isBasic}}
+{{#isOAuth}}
+        if let Some(ref token) = configuration.oauth_access_token {
+            let auth = hyper::header::Authorization(
+                hyper::header::Bearer {
+                    token: token.to_owned(),
+                }
+            );
+            auth_headers.insert("Authorization".to_owned(), format!("{}", auth));
```
bjgill: Minor niggle - should this be `auth.to_string()` for consistency with the previous block?

frol: Should this `format!` also become `auth.to_string()`?

wing328:
@euank did you have a chance to review the feedback by @bjgill and @frol?

euank:
Thanks for the ping; fixed that last nit, sorry for not getting it properly the first time. Thanks for the review!

frol:
LGTM

@euank Thanks for the improvement!

PR #7335, opened 2018-01-07
[Rust] Recommend style based on 'rustfmt' defaults

euank:
Rustfmt is widely accepted in the rust community. The actual style it
enforces is not completely stable, but running a reasonably recent
rustfmt and reading the short rfc-based style guide seems better than
nothing by a long shot.

### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [N/A] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.

cc @frol @farcaller @bjgill

PR #7339, opened 2018-01-08
[Rust] Format example with rustfmt

euank:
This was handled via 'cargo fmt' on the current nightly rust. I also
moved a few lines around and deleted an old comment.

### PR checklist

- [x] Filed the PR against the correct branch: `master`

### Description of the PR

Since #7335 merged, this updates the example to match that style in the sample example.

I'll go through and update the style in more places once I don't have other PRs which would merge-conflict in flight.

cc @frol @farcaller @bjgill

### DIFF
--- a/samples/client/petstore/rust/examples/client.rs
+++ b/samples/client/petstore/rust/examples/client.rs
@@ -1,35 +1,39 @@
-extern crate petstore_client;
-extern crate hyper;
-extern crate tokio_core;
 extern crate futures;
+extern crate hyper;
+extern crate petstore_client;
+extern crate tokio_core;
 
 use hyper::Client;
 use hyper::client::HttpConnector;
 use tokio_core::reactor::Core;
-use futures::{Future};
+use futures::Future;
 
 fn main() {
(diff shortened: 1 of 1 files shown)


frol:
LGTM

bjgill:
Nice. Do you not need to update the mustache template that generates this, instead, though?

euank:
@bjgill the 'examples' folder isn't generated from mustache templates; it only lives in that location as far as I can tell.

I would like to update the moustache template's style too, but I'm trying to avoid merge conflicts by doing that after the other in-flight PRs I have land.

wing328:
@euank thanks for the PR, which has been merged into 2.4.0.

What about running "cargo fmt" automatically in the output folder after generating the Rust code?

euank:
@wing328 I like that idea, but I assumed that depending on cargo would be an issue. I noticed that the go generated client doesn't meet `gofmt` standards and `gofmt` isn't automatically run, so I sorta assumed rust shouldn't be calling tools like that either.

wing328:
@euank I've some ideas. Will try to implement one later and collect feedback from you guys.

PR #7340, opened 2018-01-08
Qt5cpp Add support for nested containers

### LINKED ISSUE #7250: [qt5cpp] Segfault when accessing/deleting object after calling API containing an object as body param
##### Description

The generated API passed the object param in the request body as a copy. When the API returns the copy is deleted and the pointer references inside also get deleted. After this attempting to delete or access the object causes segfault since the pointers inside the object have been deleted when the copy is destroyed on API return.
e.g.
```C++
void configCreate(DockerConfigSpec body);
```
instead of 
```C++
void configCreate(DockerConfigSpec& body);
```
##### Swagger-codegen version

2.3.0
3.0.0

Any swagger file containing Objects with dynamically allocated members as part of the request body.

Passing the body Param by reference solves the problem.

etherealjoy:
### PR checklist

### Description of the PR

cc @fvarose @stkrwork @ravinikam 

- Fixes #7250
- Support for nested containers, only to two levels.
- Addition of Helper function to handle maps similar to how arrays are handled.

- [X] PetStore Examples Passed
- [X] Docker API version 1.32 Container Creation passed.


etherealjoy:
I made the PR in 2.4.0 because I see master does not have all the changes from 2.3.0

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
```diff
@@ -122,124 +121,253 @@ setValue(void* value, QJsonValue obj, QString type, QString complexType) {
...
-            for (auto item : **val) {
-                delete item;
+                (*output)->append(val);
             }
-            delete *val;
-            *val = output;
         }
         else if(QStringLiteral("qint32").compare(complexType) == 0) {
-            QList<qint32> **output = reinterpret_cast<QList<qint32> **> (value);
+            auto output = reinterpret_cast<QList<qint32> **> (value);
```
stkrwork: Why use auto here?

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
```diff
@@ -122,124 +121,253 @@ setValue(void* value, QJsonValue obj, QString type, QString complexType) {
...
-            delete *val;
+            if(*val != nullptr) delete *val;
             *val = so;
         }
     }
     else if(type.startsWith("QList") && QString("").compare(complexType) != 0 && obj.isArray()) {
         // list of values
         if(complexType.startsWith("{{prefix}}")) {
-            QList<{{prefix}}Object*>* output = new QList<{{prefix}}Object*>();
+            auto output = reinterpret_cast<QList<{{prefix}}Object *> **> (value);
```
stkrwork: Why use auto here?

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
```diff
@@ -122,124 +121,253 @@ setValue(void* value, QJsonValue obj, QString type, QString complexType) {
...
             for (const QJsonValue & jval : arr){
                 qint64 val;
-                setValue(&val, jval, QStringLiteral("qint64"), QStringLiteral(""));
+                ::{{cppNamespace}}::setValue(&val, jval, QStringLiteral("qint64"), QStringLiteral(""));
                 (*output)->push_back(val);
             }
         }
         else if(QStringLiteral("bool").compare(complexType) == 0) {
-            QList<bool> **output = reinterpret_cast<QList<bool> **> (value);
+            auto output = reinterpret_cast<QList<bool> **> (value);
```
stkrwork: Why use auto here?

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
```diff
@@ -122,124 +121,253 @@ setValue(void* value, QJsonValue obj, QString type, QString complexType) {
...
             for (const QJsonValue & jval : arr){
                 bool val;
-                setValue(&val, jval, QStringLiteral("bool"), QStringLiteral(""));
+                ::{{cppNamespace}}::setValue(&val, jval, QStringLiteral("bool"), QStringLiteral(""));
                 (*output)->push_back(val);
             }
         }
         else if(QStringLiteral("float").compare(complexType) == 0) {
-            QList<float> **output = reinterpret_cast<QList<float> **> (value);
+            auto output = reinterpret_cast<QList<float> **> (value);
```
stkrwork: Why use auto here?

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
```diff
@@ -122,124 +121,253 @@ setValue(void* value, QJsonValue obj, QString type, QString complexType) {
...
             for (const QJsonValue & jval : arr){
                 float val;
-                setValue(&val, jval, QStringLiteral("float"), QStringLiteral(""));
+                ::{{cppNamespace}}::setValue(&val, jval, QStringLiteral("float"), QStringLiteral(""));
                 (*output)->push_back(val);
             }
         }
         else if(QStringLiteral("double").compare(complexType) == 0) {
-            QList<double> **output = reinterpret_cast<QList<double> **> (value);
+            auto output = reinterpret_cast<QList<double> **> (value);
```
stkrwork: Why use auto here?

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
```diff
@@ -122,124 +121,253 @@ setValue(void* value, QJsonValue obj, QString type, QString complexType) {
...
             for (const QJsonValue & jval : arr){
                 double val;
-                setValue(&val, jval, QStringLiteral("double"), QStringLiteral(""));
+                ::{{cppNamespace}}::setValue(&val, jval, QStringLiteral("double"), QStringLiteral(""));
                 (*output)->push_back(val);
             }
         }
         else if(QStringLiteral("QString").compare(complexType) == 0) {
-            QList<QString*> **output = reinterpret_cast<QList<QString*> **> (value);
+            auto output = reinterpret_cast<QList<QString*> **> (value);
```
stkrwork: Why use auto here?

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
```diff
@@ -122,124 +121,253 @@ setValue(void* value, QJsonValue obj, QString type, QString complexType) {
...
             for (const QJsonValue & jval : arr){
                 QString * val = new QString();
-                setValue(&val, jval, QStringLiteral("QString"), QStringLiteral(""));
+                ::{{cppNamespace}}::setValue(&val, jval, QStringLiteral("QString"), QStringLiteral(""));
                 (*output)->push_back(val);
             }
         }
         else if(QStringLiteral("QDate").compare(complexType) == 0) {
-            QList<QDate*> **output = reinterpret_cast<QList<QDate*> **> (value);
+            auto output = reinterpret_cast<QList<QDate*> **> (value);
```
stkrwork: Why use auto here?

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
```diff
@@ -122,124 +121,253 @@ setValue(void* value, QJsonValue obj, QString type, QString complexType) {
...
             for (const QJsonValue & jval : arr){
                 QDate * val = new QDate();
-                setValue(&val, jval, QStringLiteral("QDate"), QStringLiteral(""));
+                ::{{cppNamespace}}::setValue(&val, jval, QStringLiteral("QDate"), QStringLiteral(""));
                 (*output)->push_back(val);
             }
         }
         else if(QStringLiteral("QDateTime").compare(complexType) == 0) {
-            QList<QDateTime*> **output = reinterpret_cast<QList<QDateTime*> **> (value);
+            auto output = reinterpret_cast<QList<QDateTime*> **> (value);
```
stkrwork: Why use auto here?

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
```diff
@@ -122,124 +121,253 @@ setValue(void* value, QJsonValue obj, QString type, QString complexType) {
...
-                setValue(&val, jval, QStringLiteral("QDateTime"), QStringLiteral(""));
+                ::{{cppNamespace}}::setValue(&val, jval, QStringLiteral("QDateTime"), QStringLiteral(""));
                 (*output)->push_back(val);
             }
         }
     }
+    else if(type.startsWith("QMap") && QString("").compare(complexType) != 0 && obj.isObject()) {
+        // list of values
+        if(complexType.startsWith("{{prefix}}")) {
+            auto output = reinterpret_cast<QMap<QString, {{prefix}}Object*> **> (value);
```
stkrwork: Why use auto here?

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
```diff
@@ -122,124 +121,253 @@ setValue(void* value, QJsonValue obj, QString type, QString complexType) {
...
+                for(auto itemkey : varmap.keys() ){
+                    auto  val = ({{prefix}}Object*)::{{cppNamespace}}::create(complexType);
+                    auto  jsonval = QJsonValue::fromVariant(varmap.value(itemkey));
+                    ::{{cppNamespace}}::setValue(&val, jsonval, complexType, complexType);
+                    (*output)->insert(itemkey, val);
+                }
+            }
+        }
+        else if(QStringLiteral("qint32").compare(complexType) == 0) {
+            auto output = reinterpret_cast<QMap<QString, qint32> **> (value);
```
stkrwork: Why use auto here?

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
```diff
@@ -122,124 +121,253 @@ setValue(void* value, QJsonValue obj, QString type, QString complexType) {
...
+                for(auto itemkey : varmap.keys() ){
+                    qint32 val;
+                    auto  jsonval = QJsonValue::fromVariant(varmap.value(itemkey));
+                    ::{{cppNamespace}}::setValue(&val, jsonval, QStringLiteral("qint32"), QStringLiteral(""));
+                    (*output)->insert( itemkey, val);
+                }
+            }
+        }
+        else if(QStringLiteral("qint64").compare(complexType) == 0) {
+            auto output = reinterpret_cast<QMap<QString, qint64> **> (value);
```
stkrwork: Why use auto here?

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
```diff
@@ -302,54 +430,133 @@ toJsonValue(QString name, void* value, QJsonObject* output, QString type) {
 }
 
 void
-toJsonArray(QList<void*>* value, QJsonArray* output, QString innerName, QString innerType) {
+toJsonArray(QList<void*>* value, QJsonObject* output, QString innerName, QString innerType) {
+    if((value == nullptr) || (output == nullptr)) {
+        return;
+    }
+    QJsonArray outputrray;
```
stkrwork: typo?

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
```diff
@@ -302,54 +430,133 @@ toJsonValue(QString name, void* value, QJsonObject* output, QString type) {
...
+}
+
+void
+toJsonMap(QMap<QString, void*>* value, QJsonObject* output, QString innerName, QString innerType) {
+    if((value == nullptr) || (output == nullptr)) {
+        return;
+    }
+    QJsonObject mapobj;
+    if(innerType.startsWith("{{prefix}}")){
+        auto items = reinterpret_cast< QMap<QString, {{prefix}}Object*> *>(value);
```
stkrwork: Why use auto here?

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-body.mustache
fvarose: IMO using `auto` is a much better alternative to specifying the full type here, which would be redundant and decrease readability. I don't see any reason for this project to shy away from writing modern C++ code.
@etherealjoy could you please restore the auto keywords?

etherealjoy:
@stkrwork @fvarose : I have added the auto keyword again.

samples/client/petstore/qt5cpp/client/SWGUserApi.h
```diff
@@ -35,14 +35,14 @@ class SWGUserApi: public QObject {
...
-    void createUsersWithListInput(QList<SWGUser*>* body);
+    void createUser(SWGUser& body);
+    void createUsersWithArrayInput(QList<SWGUser*>*& body);
+    void createUsersWithListInput(QList<SWGUser*>*& body);
     void deleteUser(QString* username);
     void getUserByName(QString* username);
     void loginUser(QString* username, QString* password);
     void logoutUser();
-    void updateUser(QString* username, SWGUser body);
+    void updateUser(QString* username, SWGUser& body);
```
fvarose: Did you make this parameter a reference to avoid copies, or to be able to modify it inside the functions? If it's to avoid copies, please make it a `const` reference.
(same remark for createUser() above)

etherealjoy: Yes . a const reference would be better. I will check it.

samples/client/petstore/qt5cpp/client/SWGUserApi.h
```diff
@@ -35,14 +35,14 @@ class SWGUserApi: public QObject {
     QString basePath;
     QMap<QString, QString> defaultHeaders;
 
-    void createUser(SWGUser body);
-    void createUsersWithArrayInput(QList<SWGUser*>* body);
-    void createUsersWithListInput(QList<SWGUser*>* body);
+    void createUser(SWGUser& body);
+    void createUsersWithArrayInput(QList<SWGUser*>*& body);
+    void createUsersWithListInput(QList<SWGUser*>*& body);
```
fvarose: References to pointers are a bit odd, their only purpose is to be able to modify the pointer itself in the function. Is that the intent here?

fvarose: Side note, a pointer is already a cheap copy, and doesn't benefit from passing it as a reference to pointer. You could make this a const pointer instead to state that this is a read-only input.

etherealjoy: Yes I agree with the pointer stuff. But the Mustache is a quite big and a bit ugly that is why  I did not add the code to do it that way. But I will check if we can do it cleanly.

fvarose: Right, I see. Nevermind my comment then, just use `const &` regardless of the type.
Agreed on the ugliness, afaict that whole qt code is in dire need of refactoring anyway, way too many raw pointer handling for my taste 😳

etherealjoy: Indeed. In fact I have removed all memory reallocation in the helper class. There were too many memory leaks. I will just add the const.

etherealjoy: Ok, it is not so small change... The object is accessed and it has to have a const specifier in the methods being invoked as well. So the changes are cascading. If you agree I will drop the const too. 
Right now I need to prevent the copy and the reference is working. Maybe I do this after I make the other PRs.

fvarose: Ok, it's definitely outside the scope of this PR then. The non-const reference will do for now :)

fvarose:
PR looks good to me 👍

PR #6760, opened 2017-10-19
[JAVA] equals and hashCode for models with byte[] vars

bmordue:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming langauge.

### Description of the PR

For models with a byte array field, the generated `equals` and `hashCode` methods are not correct. This change modifies the template to use `Array.equals` and `Arrays.hash` for byte array variables.

I think this goes at least some way to being a fix https://github.com/swagger-api/swagger-codegen/issues/3731

cc @bbdouglas @JFCote @sreeshas @jfiala @lukoyanov @cbornet

PR #7341, opened 2018-01-08
[JAVA] equals and hashCode for models with byte[] and binary strings

### LINKED ISSUE #3731: [JAVA] array equals() fails findbugs
##### Description

Using swagger codegen to generate objects with arrays fails findbugs: http://findbugs.sourceforge.net/bugDescriptions.html#EC_BAD_ARRAY_COMPARE
##### Swagger-codegen version

2.2.2-SNAPSHOT

Create a property in a swagger definition like this:

```
byte:
     type: string
     format: byte
```

   Generates this code in equals()
`Objects.equals(this._byte, awesomeSnapshot._byte)`

bmordue:
### PR checklist

### Description of the PR

For models with string fields in "byte" or "binary" format, the generated equals and hashCode methods are not correct. This change modifies the template to use Array.equals and Arrays.hash for byte array and binary variables.

Fixes #3731
This PR replaces/supersedes #6760 

cc @bbdouglas @JFCote @sreeshas @jfiala @lukoyanov @cbornet
cc @brentryan (reporter of #3731)


modules/swagger-codegen/src/main/resources/Java/pojo.mustache
```diff
@@ -140,15 +140,15 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#parcela
       return false;
     }{{#hasVars}}
     {{classname}} {{classVarName}} = ({{classname}}) o;
-    return {{#vars}}Objects.equals(this.{{name}}, {{classVarName}}.{{name}}){{#hasMore}} &&
+    return {{#vars}}{{#isByteArray}}Arrays{{/isByteArray}}{{#isBinary}}Arrays{{/isBinary}}{{^isByteArray}}{{^isBinary}}Objects{{/isBinary}}{{/isByteArray}}.equals(this.{{name}}, {{classVarName}}.{{name}}){{#hasMore}} &&
```
wing328: @bmordue One suggestion: What about using `{{#isBinary}}java.util.Arrays{{/isBinary}}` to avoid unused import when no property is defined as byte array or binary?

bmordue: Maybe I've misunderstood, but I thought `isBinary` and `isByteArray` apply to a single CodegenProperty and would only be relevant inside `models.model.vars` in the template?

wing328: My point is in https://github.com/swagger-api/swagger-codegen/pull/7341/files#diff-cc3e8eb9a2b26248932102a87608fc60R7, you're hard-coding "import java.util.Arrays;".

My suggestion is to avoid the hard-coding the import

wing328: @cbornet what's your take on this?

cbornet: Yes there's a risk of having an unused import. But note that it's the case in a lot of other places of our templates.
Managing imports is really something not easy with code generators.

wing328: Yup, we'll try to take care of the unused import in the future so this PR is good to merge.

Issue #7261, opened 2017-12-27
[Java] Bug: boolean getter used for non-primitive Boolean


macjohnny:
##### Description

When generating a model from a swagger file with
```
  myProperty:
    type: boolean
```
the generated model file will contain
```
  @JsonProperty("myProperty")
  private Boolean myProperty= null;

  @ApiModelProperty()
  public Boolean isMyProperty() {
    return myProperty;
  }

  public void setMyProperty(Boolean myProperty) {
    this.myProperty= myProperty;
  }
```
The problematic line is the following:
```
  public Boolean isMyProperty() {
```
since the [JavaBeans Specification - section 8.3.2](https://edelivery.oracle.com/otn-pub/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/beans.101.pdf) states that
> In addition, for boolean properties, we allow a getter method to match the pattern:
> ```
> public boolean is<PropertyName>();
> ```
> This “is<PropertyName>” method may be provided instead of a “get<PropertyName>” method,
or it may be provided in addition to a “get<PropertyName>” method.

And in the example above, the `isMyProperty()` method is generated for a non-primitive `Boolean` property, which should not be the case.

First, this is a breaking change that was introduced with `2.3.0`, which could easily have been avoided by adding the `isMyProperty()` in addition to a `getMyProperty()` method.

Second, some bean property mappers may not recognized the property as a readable attribute, e.g. 
in the Spring BeanWrapper: https://jira.spring.io/browse/SPR-9482 

This change was introduced in https://github.com/swagger-api/swagger-codegen/pull/6177

##### Swagger-codegen version

affected: 2.3.0
type: regression, used to work in 2.2.3

##### Related issues/PRs
https://github.com/swagger-api/swagger-codegen/pull/6177

##### Suggest a fix/enhancement

I suggest to have the `isMyProperty()` method in addition to the `getMyProperty()` method instead of replacing it.

macjohnny:
cc @wing328

wing328:
> I suggest to have the isMyProperty() method in addition to the getMyProperty() method instead of replacing it.

Your suggestion sounds good to me. I didn't realize it would break Spring BeanWrapper usage.

Please file a PR so that we can review.

macjohnny:
@wing328 would it be possible to revert the change at first instance and release a patch version `2.3.1` and later on implement the additional `isMyProperty()` method, since adding the additional method requires some work

wing328:
[...]

PR #7344, opened 2018-01-08
[Java] allow setting boolean getter (is, has, get) in templates

wing328:
### PR checklist

### Description of the PR

To fix https://github.com/swagger-api/swagger-codegen/issues/7261

If this fix looks good, I'll apply the same change to other mustache templates:
```
modules/swagger-codegen/src/main/resources//MSF4J/pojo.mustache
modules/swagger-codegen/src/main/resources//JavaJaxRS/spec/pojo.mustache
modules/swagger-codegen/src/main/resources//JavaJaxRS/pojo.mustache
modules/swagger-codegen/src/main/resources//JavaJaxRS/resteasy/eap/pojo.mustache
modules/swagger-codegen/src/main/resources//JavaJaxRS/resteasy/pojo.mustache
modules/swagger-codegen/src/main/resources//JavaJaxRS/cxf-cdi/pojo.mustache
modules/swagger-codegen/src/main/resources//JavaJaxRS/cxf/pojo.mustache
modules/swagger-codegen/src/main/resources//JavaSpring/pojo.mustache
modules/swagger-codegen/src/main/resources//JavaInflector/pojo.mustache
modules/swagger-codegen/src/main/resources//Java/pojo.mustache
modules/swagger-codegen/src/main/resources//JavaPlayFramework/pojo.mustache
modules/swagger-codegen/src/main/resources//undertow/pojo.mustache
modules/swagger-codegen/src/main/resources//JavaVertXServer/pojo.mustache
modules/swagger-codegen/src/main/resources//java-pkmst/pojo.mustache
```


modules/swagger-codegen/src/main/resources/Java/pojo.mustache
```diff
@@ -118,9 +118,16 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#parcela
 {{#vendorExtensions.extraAnnotation}}
   {{{vendorExtensions.extraAnnotation}}}
 {{/vendorExtensions.extraAnnotation}}
+  {{#isBoolean}}
```
macjohnny: @wing328 I don't understand where you could configure to use either `getMyBooleanProperty()` or `isMyBooleanProperty()`.
Am I missing something?

wing328: You can use the -t option with customized templates as follows to have `getMyBooleanProperty`:
```
  {{#isBoolean}}
  public {{{datatypeWithEnum}}} get{{getter}}() {
    return {{name}};
  }
  {{/isBoolean}}
  {{^isBoolean}}
  public {{{datatypeWithEnum}}} {{getter}}() {
    return {{name}};
  }
  {{/isBoolean}}
```

macjohnny: don't you think it would be easier to have a configuration option to choose between the two possibilities?
as a developer I wouldn't want to have to copy and customize a swagger mustache template to simply configure a standard-variant. Customizing requires additional efforts when e.g. upgrading. 
Moreover, the using `is...()` with a non-primitive `Boolean` instead of a primitive `boolean` is not specified and can thus lead to discussions concerning support in e.g. data mappers like here: https://jira.spring.io/browse/SPR-9482

macjohnny: @cbornet what is your opinion on that?

wing328: > don't you think it would be easier to have a configuration option to choose between the two possibilities?

Agreed that adding an option is one way but personally I try to avoid too many options as there're already quite a lot of options for Java client generator already.

Besides `is`, `get`, another possible choice is `has`.

macjohnny: @wing328 trying to keep the number of options as small as possible is a good approach. however, as the option in this case would allow using standard-behavior of data mappers, (which used to work before), I consider it worth adding instead of requiring customizations.
Personally, I think it is one of the strengths of swagger codegen to be able to generate standard-code from a swagger-file and directly use it without customizations.
And using the default data mapper of Spring Boot definitely is such a standard-setup.

svenhuber:
Since this is standard behavior in Spring Boot I would prefer having an option too. If you don't need it, the additional option will do no harm to users.

wing328:
Thanks for the feedback.

I'll merge this into master so that we can release v2.3.1.

I'll create another task to track the enhancement adding another CLI option (booleanGetterPrefix) for Java generator.

William

PR #7351, opened 2018-01-09
[csharp] Generate SDK based on .netstandard 2.0

bzglm:
Being able to generate a C# SDK based on the .netstandard 2.0 framework would be a very useful upgrade as it would allows a much larger number of projects to be compatible with the generated SDK, in opposition with the currently available 1.3 version.


wing328:
@bzglm I notice that this PR is about merging master into 3.0.0.

I believe @HugoMario will do it when he has time so closing this one.

For your enhancement to C# generator to support .Net standard 2.0, please submit another PR and we'll review accordingly.

PR #7366, opened 2018-01-11
Feature/7365 angular request param enum

macjohnny:
### PR checklist

### Description of the PR

Allows request params of type enum to be represented as enum in the generated angular client code.

Before:
```
public findPetsByStatus(status: Array<string>, observe?: 'body', reportProgress?: boolean): Observable<Array<Pet>>;
```
after:
```
public findPetsByStatus(status: Array<'available' | 'pending' | 'sold'>, observe?: 'body', reportProgress?: boolean): Observable<Array<Pet>>;
```


macjohnny:
cc @TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @wing328 @bedag-moo

macjohnny:
@wing328 could you include this PR in 2.3.1 please?

macjohnny:
the code of this PR could also be moved to `AbstractTypeScriptClientCodegen` to allow using this feature with any TypeScript client. what do you think?

macjohnny:
cc @JohannesHoppe

kenisteward:
@macjohnny i can agree with you there.  I also approve of this change.

macjohnny:
@kenisteward would you mind reviewing the code?

macjohnny:
would anyone of the technical comitee mind please reviewing the code?

cc @TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @wing328 @bedag-moo

bedag-moo:
Nice improvement :-)

Might this be beneficial for all typescript generators? If so, the super class AbstractTypeScriptClientCodegen might be a better home for this code.

macjohnny:
@bedag-moo this feature could easily be made available to all typescript generators by moving the code to `AbstractTypeScriptClientCodegen` as you suggested. However, I don't know whether this would be accepted. We could also first merge it as it is and in a second step discuss about moving it to the abstract typescript generator.

bedag-moo:
I think we should be able to find out whether it would be accepted, since their authors are all in the technical committee :-)

Personally, I am happy either way, I just thought now we be a good time to decide where we put this code.

TiFu:
Wasn't this already implemented in #6233? 

Also note, that there was quite a long discussion about literal unions in this PR after it was merged.

kenisteward:
@TiFu this is different.

#6233 was the typescript code representation of what the enum will be in model code.

This PR sets those would be values in the service code instead of things like String so we have strong types for calling service names.

Note: in the descirption it shows this

```
public findPetsByStatus(status: Array<string>, observe?: 'body', reportProgress?: boolean): Observable<Array<Pet>>;
```

becomes

```
public findPetsByStatus(status: Array<'available' | 'pending' | 'sold'>, observe?: 'body', reportProgress?: boolean): Observable<Array<Pet>>;
```

so now 
```
service.findPetsByStatus('someString');
```
 would throw a compiler error which helps the developer know the server won't accept that value.

I think all the clients would benefit from this. My only thing about the code was I believe enum values can't be anything but numbers or string  so the check for Date/DateTime shouldn't be needed.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngularClientCodegen.java
```diff
@@ -174,9 +169,87 @@ public String getTypeDeclaration(Property p) {
...
+    /**
+     * Converts a list of numbers to a literal union for representing enum values as a type.
+     * Example: 3 | 9 | 55
+     *
+     * @param values
+     * @return
+     */
+    private String numericEnumValuesToEnumTypeUnion(List<Number> values) {
+        List<String> stringValues = new ArrayList<>();
+        for (Object value: values) {
```
TiFu: Code Style: Change `Object` to  `Number`? This would make it clearer that it's used for numbers only.

samples/client/petstore/typescript-angular-v2/default/.swagger-codegen/VERSION
```diff
@@ -1 +1 @@
-2.3.0
\ No newline at end of file
+2.3.1-SNAPSHOT
```
TiFu: Is this because someone forgot to update the samples? Are you sure that your branch is based on current master?

kenisteward: This should be for 2.3.1 now shouldn't it @TiFu .  I thought at some point I talked to @wing328  about a patch release but I can't remember exactly.

macjohnny: @kenisteward @TiFu the branch is based on the master of some 5 days ago, but I will rebase and generate the code again.

macjohnny:
@kenisteward concerning the allowed types of enums:
I tested with the following modification of the `petstore.yml`
```
/pet/findByStatus:
    get:
      tags:
        - pet
      summary: Finds Pets by status
      description: Multiple status values can be provided with comma separated strings
      operationId: findPetsByStatus
      produces:
        - application/xml
        - application/json
      parameters:
        - name: status
          in: query
          description: Status values that need to be considered for filter
          required: true
          type: array
          items:
            type: string
            enum:
              - available
              - pending
              - sold
            default: available
          collectionFormat: csv
        - name: status2
          in: query
          description: Status values that need to be considered for filter
          required: true
          type: array
          items:
            type: integer
            format: int32
            enum:
              - 3
              - 4
              - 5
          collectionFormat: csv
        - name: status3
          in: query
          description: Status values that need to be considered for filter
          required: true
          type: string
          enum:
            - available
            - pending
            - sold
        - name: status4
          in: query
          description: Status values that need to be considered for filter
          required: true
          type: string
          format: date
          enum:
            - 2012-02-04
            - 2012-02-05
            - 2014-02-04
        - name: status5
          in: query
          description: Status values that need to be considered for filter
          required: true
          type: number
          format: float
          enum:
            - 5.4
            - 6.5
            - 7.9

      responses:
        '200':
          schema:
            type: array
            items:
              $ref: '#/definitions/Pet'
        '400':
          description: Invalid status value
      security:
        - petstore_auth:
            - 'write:pets'
            - 'read:pets'
```

It seems that `float`, `date`, etc. are valid enum types, too.
Looking at their corresponding property classes in https://github.com/swagger-api/swagger-core/tree/master/modules/swagger-models/src/main/java/io/swagger/models/properties this seems to work as intended.

So far I found one limitation of the enum implementation: When using
```
          items:
            type: integer
            enum:
              - 3
              - 4
              - 5
```
without specifying e.g. `format: int32`, the resulting type is a string without the enum values, since it will be an [`AbstractNumericProperty`](https://github.com/swagger-api/swagger-core/blob/master/modules/swagger-models/src/main/java/io/swagger/models/properties/AbstractNumericProperty.java) which does not define enum values.

macjohnny:
there is an issue when generating models with
```
definitions:
  EnumArrays:
    type: object
    properties:
      just_symbol:
        type: string
        enum:
          - ">="
          - "$"
      array_enum:
        type: array
        items:
          type: string
          enum:
            - fish
            - crab
```
which results in 
```
export interface EnumArrays {
    justSymbol?: EnumArrays.JustSymbolEnum;
    arrayEnum?: Array<'fish' | 'crab'>;
}
export namespace EnumArrays {
    export type JustSymbolEnum = '>=' | '$';
    export const JustSymbolEnum = {
        GreaterThanOrEqualTo: '>=' as JustSymbolEnum,
        Dollar: '$' as JustSymbolEnum
    }
    export type ArrayEnumEnum = 'fish' | 'crab';
    export const ArrayEnumEnum = {
        Fish: 'fish' as ArrayEnumEnum,
        Crab: 'crab' as ArrayEnumEnum
    }
}
```

The problem is that 
```
    arrayEnum?: Array<'fish' | 'crab'>;
```
should be
```
    arrayEnum?: Array<ArrayEnumEnum>;
```
in order to use the generated enum types instead of directly typing it as a literal union of the allowed values. 

Somehow this needs to be fixed by allowing to distinguish between model properties and API client request parameters. Anyone's got an idea?

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngularClientCodegen.java
macjohnny: @TiFu done

macjohnny:
@kenisteward @Vrolijkx @TiFu @sebastianhaas @taxpon
I prepared a separate branch where I moved the code to the `AbstractTypeScriptClientCodegen`
https://github.com/macjohnny/swagger-codegen/tree/feature/7365-typescript-request-param-enum
I needed to disable two equality assertions in the unit tests of the `TypeScriptFetchModelTest`, because they failed due to the issue with model enums being replaced with literal unions as described in https://github.com/swagger-api/swagger-codegen/pull/7366#issuecomment-358349226

macjohnny:
@SamuelBeliveau do you have an idea for the issue described here https://github.com/swagger-api/swagger-codegen/pull/7366#issuecomment-358349226 ?

macjohnny:
here is my proposal to resolve the issue:
we allow for a custom implementation of the generation of the datatype of the parameter here:
https://github.com/swagger-api/swagger-codegen/pull/7366/commits/61b22ae9b24ed744e5a3f7cae0d83826c3d9fbdf#diff-0c1b6d04d7c2c41f87f7c710a5610d88R2511
and can then implement it in the `TypeScriptAngularCodegen` or `AbstractTypeScriptCodegen` here:
https://github.com/swagger-api/swagger-codegen/pull/7366/commits/61b22ae9b24ed744e5a3f7cae0d83826c3d9fbdf#diff-ad60e100e050d5ef52943803a2adc5fdR175

I also updated the https://github.com/macjohnny/swagger-codegen/tree/feature/7365-typescript-request-param-enum branch, which should now also work correctly, allowing to have this feature for all typescript clients.

what do you guys think about it?

macjohnny:
I filed a new PR https://github.com/swagger-api/swagger-codegen/pull/7433 for the general implementation of this feature for all TypeScript clients

macjohnny:
@wing328 this PR can be closed if https://github.com/swagger-api/swagger-codegen/pull/7433 is merged, enabling this feature for all typescript clients.

PR #7383, opened 2018-01-12
Generate Boolean getX getters rather than isX getters

sandyscoffable:
For Tapestry compatibility so models can be used directly in Forms


sandyscoffable:
Sorry this was supposed to be on a fork ... closing.

PR #7385, opened 2018-01-12
This wraps all http requests in hystrix by default

sandyscoffable:
See the following screenshot of stacktrace from the Admin panel of this working (with generated code):
screen shot 2018-01-12 at 10 14 04


sandyscoffable:
Again, this was supposed to go to a fork. FFS.

Issue #7218, opened 2017-12-19
[CSharp] Problems with code creation


karussell:
##### Description

I have no windows here and so I tried to build the c# version of [our client library](https://github.com/graphhopper/directions-api-clients/tree/master/csharp) under Ubuntu 16.04, but this fails with:

```bash
[INFO] Target framework: net45
[INFO] Download nuget and packages
File ‘nuget.exe’ already there; not retrieving.

Mozilla Roots Importer - version 5.4.1.6
Download and import trusted root certificates from Mozilla's MXR.
Copyright 2002, 2003 Motus Technologies. Copyright 2004-2008 Novell. BSD licensed.

WARNING: mozroots is deprecated, please move to cert-sync instead.

Downloading from 'https://hg.mozilla.org/releases/mozilla-release/raw-file/default/security/nss/lib/ckfw/builtins/certdata.txt'...
Importing certificates into user store...
1 new root certificates were added to your trust store.
Import process completed.

[INFO] Copy DLLs to the 'bin' folder
[INFO] Run 'mcs' to build bin/IO.Swagger.dll
warning CS8001: SDK path could not be resolved
src/IO.Swagger/Model/ResponseCoordinatesArray.cs(30,58): error CS0305: Using the generic type `System.Collections.Generic.List<T>' requires `1' type argument(s)
/usr/lib/mono/4.5/mscorlib.dll (Location of the symbol related to previous error)
src/IO.Swagger/Api/GeocodingApi.cs(128,29): error CS0104: `Configuration' is an ambiguous reference between `IO.Swagger.Client.Configuration' and `IO.Swagger.Model.Configuration'
src/IO.Swagger/Client/Configuration.cs(23,18): (Location of the symbol related to previous error)
src/IO.Swagger/Model/Configuration.cs(30,26): (Location of the symbol related to previous error)
src/IO.Swagger/Api/GeocodingApi.cs(167,16): error CS0104: `Configuration' is an ambiguous reference between `IO.Swagger.Client.Configuration' and `IO.Swagger.Model.Configuration'
src/IO.Swagger/Client/Configuration.cs(23,18): (Location of the symbol related to previous error)
src/IO.Swagger/Model/Configuration.cs(30,26): (Location of the symbol related to previous error)
src/IO.Swagger/Api/GeocodingApi.cs(167,16): error CS0104: `Configuration' is an ambiguous reference between `IO.Swagger.Client.Configuration' and `IO.Swagger.Model.Configuration'
[...]

PR #7400, opened 2018-01-14
[csharp] Support arrays of arrays for properties and models

jimschubert:
### PR checklist

### Description of the PR

Previous implementation didn't support multiple levels of array with
array items as OpenAPI spec supports. This means an object defined as
type: array with items = type: array|items=double (which is common in
GIS) would not be possible.

This implementation assumes generics in the nested type definitions, so
the above would generate List<List<double?>> for model parent types as
well as property type declarations.

This should resolve multiple issues as presented in #7218:
* (cherry-picked from #7399) `ToJson` method wasn't working for both composition and polymorphism
* (cherry-picked from #7394) `Configuration` references in api.template weren't fully qualified, causing conflicts with user defined `Configuration` models
* Arrays of Arrays of some type were not supported

As part of this change, I've also removed the case-sensitive comparison for reserved word evaluation. User has a `Break` model which was being renamed to `ModelBreak`. This shouldn't be necessary in C#, especially because we can escape reserved words in C# 3.x+ by preceding the reserved word with an `@`. I haven't done this in this commit since it seems that all generated examples with reserved words use the title cased representation, but this should be a simple fix if it is required for parameter names.

/cc @mandrean 

@wing328  I've opened this PR so @karussell might evaluate all fixes in one. It may be easier to merge this and close the other two.


karussell:
I can confirm our library now compiles with this on Linux via mono and also the created tests do pass.

theason:
For my Swagger based API I am trying to generate a C# client for, it has a root type for "Configuration", and I noticed that even with this branch you've committed the code still fails to build. Can you review the enclosed json file that I am trying to generate against please?
[final.json.txt](https://github.com/swagger-api/swagger-codegen/files/1645138/final.json.txt)

?

theason:
Update, it looks like the json uses a parameter called "Type" over and over...

"parameters": [
          {
            "in": "query",
            "description": "A list of words to search for, separated by spaces. Prefer complete words or the start of words.",
            "name": "q",
            "type": "string"
          },
          {
            "in": "query",
            "description": "A list of item types to search among. If no types are given all item types will be included in the search.",
            "name": "type",
            "type": "array",
            "items": {
              "type": "string"
            },
            "collectionFormat": "multi"
          }
        ],

jimschubert:
@theason I've looked at the swagger definition you shared.

The reason you're getting `Type` repeated is because your spec has a string of `allOf` property inclusions where `type` is getting duplicated.

Here's an example:

`SwitchGear` is defined as:

```
    "SwitchGear": {
      "allOf": [
        {
          "required": [
            "type"
          ],
          "properties": {
            "floorMounted": {
              "$ref": "#/definitions/FloorMounted"
            },
            "type": {
              "type": "string"
            }
          },
          "discriminator": "type"
        },
        {
          "$ref": "#/definitions/DataCenterEquipment"
        }
      ]
    },
```

The allOf includes one schema defining `name=type, type=string`, then continues to reference `DataCenterEquipment`, defined as:


```
"DataCenterEquipment": {
      "allOf": [
        {
          "required": [
            "type"
          ],
          "properties": {
            "modelName": {
              "type": "string"
            },
            "serialNumber": {
              "type": "string"
            },
            "partNumberId": {
              "type": "integer",
              "format": "int32"
            },
            "weight": {
              "type": "number",
              "format": "double"
            },
            "description": {
              "type": "string"
            },
            "partNumber": {
              "type": "string"
            },
            "type": {
              "type": "string"
            },
            "barcode": {
              "type": "string"
            },
            "manufacturer": {
              "type": "string"
            },
            "height": {
              "type": "number",
              "format": "double"
            }
          },
          "discriminator": "type"
        },
        {
          "$ref": "#/definitions/dataCenterItem"
        }
      ]
    }
```

This also contains a property defined as `name=type, type=string`.

`discriminator` is defined in Open API 2.0 as:

> Adds support for polymorphism. The discriminator is the schema property name that is used to differentiate between other schema that inherit this schema. The property name used MUST be defined at this schema and it MUST be in the required property list. When used, the value MUST be the name of this schema or any schema that inherits it.

`discriminator` is used to basically create a base type with a named type that identifies subtypes. I've never seen it used in multiple `allOf` schemas, possibly because I mainly support C# which doesn't support multiple inheritance.

To check, I've also compiled the Java generator. It doesn't seem to have a similar issue with compilation because there are no constructors enforcing the required arguments. It also doesn't appear to me that it handles the `type` property issue as would be expected because it results in a `type` property in the super type and another in the sub type which shadows the super type's property. In C#, there are additional constraints for shadowing which would require us to use `override` or `new` explicitly.

What are your expectations for when your model defines type in two different models, both as discriminators?  Since I don't know what the correct behavior should be, and your spec doesn't match how I've seen discriminators done elsewhere, you may want to open a new issue for this.

To address your issue with `Configuration` conflicting, you could either modify the template and change the `Configuration` property in api class to something else, or you could use `--model-name-suffix=Model` to append `Model` to all model names. You could also use a `using` alias on the imported Configuration model and refer to your model by alias:

```
using ConfigurationModel = MyApi.Model.Configuration
```

theason:
Thank you so much for reviewing my problem! Unfortunately, this is a REST API that I am trying to consume from a 3rd party product I have no control over. I had a hunch that their system didn't quite conform to the OpenAPI standard. I won't call them out by name here; suffice to say they are a multibillion dollar organization that I have almost no sway over (but if you search the json for a URL you'll find it). 

It would seem that auto-generating a client for this API isn't a possibility given the disconnected decisions they made.

karussell:
@wing328 will it go into the next release 2.4.0 or only 3.0.0?

karussell:
Ups, I can now see the milestone v2.4.0, please ignore.

mikeclemson:
I'm consuming JSON from a third party, and I receive the same error for Configuration and for OperatingSystem (C# and PowerShell clients). By using this PR and making new builds, I fix the Configuration issue, but the OperatingSystem ambiguous reference remains. Should this be part of this issue, a new issue, or is there another way to solve it?

jimschubert:
I think that issue is related to a current feature request for fully qualified models. The issue in this pr fix was a conflict with a property in the API class. OperatingSystem would be a conflict with imports, so just a little different.

You could workaround this by passing `--model-name-prefix=Model` or `--model-name-suffix=Model`. Check the CLI help to verify these are the correct switches for your version of swagger codegen.

You could also look into extending the templates. I think OperatingSystem is under System, and that may be difficult to work around without fully qualified model names.

Issue #7386, opened 2018-01-12
[JAXRS-SPEC] Lowercase enums sent as uppercase


alessiofachechi:
##### Description

When sending enums with lowercase 'value' attributes the uppercase enum name seems to be sent instead.

##### Swagger-codegen version

2.3.0

##### Related issues/PRs

#4045

Issue #4045, opened 2016-10-20
[JAXRS-CXF] Lowercase enums sent as uppercase


markus-wa:
##### Description

When sending enums with lowercase 'value' attributes the uppercase enum name seems to be sent instead.
##### Swagger-codegen version

2.2.2-SNAPSHOT
##### Swagger declaration file content or url

http://petstore.swagger.io/v2/swagger.json
##### Command line used for generation

java -jar swagger2java/swagger-codegen-cli-2.2.2-SNAPSHOT.jar generate -i swagger.json -c config.json -l jaxrs-cxf -o out
##### Steps to reproduce

Example (petstore Pet.status) - note the lowercase values for the enum:

``` json
"status":{"type":"string","description":"pet status in the store","enum":["available","pending","sold"]}}
```

Generates this enum in the model (https://github.com/swagger-api/swagger-codegen/blob/master/samples/server/petstore/jaxrs-cxf-cdi/src/gen/java/io/swagger/model/Pet.java#L44):

``` java
@XmlType(name="StatusEnum")
@XmlEnum
public enum StatusEnum {

    AVAILABLE(String.valueOf("available")), PENDING(String.valueOf("pending")), SOLD(String.valueOf("sold"));


    private String value;

    StatusEnum (String v) {
        value = v;
    }

    public String value() {
        return value;
    }

    public static StatusEnum fromValue(String v) {
        return valueOf(v);
    }
}
```

Implementation of getPetById():

``` java
    @Override
    public Response getPetById(Long arg0) {
        Pet result = new Pet();
        result.setId(arg0);
        result.setStatus(StatusEnum.AVAILABLE);
        result.setName("Test");
        return Response.ok().entity(result).type(MediaType.APPLICATION_JSON_TYPE).build();
    }
```

Results in uppercase status:

``` json
{
  "id": 5,
  "name": "Test",
  "status": "AVAILABLE"
}
```

The java client then expects a lowercase value (as specified) which results in a null value for status.
##### Suggest a Fix

I assume this happens because the enum's toString() method should be overridden to return the result of value() which isn't done, so it returns the name of the enum, which is uppercase.

triptec:
I think this is a problem in jaxrs-jersey aswell

markus-wa:
While I haven't tested it, neither jersey1 nor jersey2 would be affected if it is indeed just the missing toString() method.

See:
[...]

PR #7401, opened 2018-01-14
fix of issue #7386

alessiofachechi:
### Description of the PR

This PR contains the fix for the issue #7386 that is related to the issue #4045.


wing328:
@canadaduane thanks for the PR. I checked other Java server generators, which already contain similar fix.

PR #7408, opened 2018-01-15
Use supportsES6 flag in ts compilation for language typescript-angular

sake:
We are using the client generated for language typescript-angular in a NativeScript project which can't deal with es6 modules. This fix uses the supportsES6 option to set appropriate values in the TypeScript compiler options.


wing328:

Issue #7444, opened 2018-01-18
Object definition returned in nested array/map is not added to the imports causing compiling problems in many languages


gurubook:
##### Description

Object definition returned in nested array/map is not added to the imports causing compiling problems in many languages

##### Swagger-codegen version

2.4.0-SNAPSHOT and earlier

##### Swagger declaration file content or url

https://gist.github.com/gurubook/9116e6d036bc43dbd9b06e3ad6a897dd#file-nested_collection_return-json

##### Steps to reproduce

generate java client with the attached schema -> compile

gurubook:
Added revised pull request #7580.

egwepas:
This at least still exists on 3.0.12 on java generator, I'm facing that same issue.

Is there any workaround ?

PR #7445, opened 2018-01-18
Add object definition returned in nested array/map to the imports

gurubook:
### Description of the PR
Code generation on operation that return object in "nested" array/map does not add the required definition to the import list.
This patch address the issue saving the contained object type during the recursion.
This should solve issue #7444


wing328:

wing328:
Please use spaces intead of tab.

Also can you share a spec to reproduce the issue?

gurubook:
Fixed indentation issue and added return-nested.yaml

PR #7464, opened 2018-01-22
[scala-clienrt]Unified build tool to sbt

magnolia-k:
Since it is a project of scala, it is better to unify it to sbt

### PR checklist

### Description of the PR

Unified build tool to sbt

 - deleted file with gradle relation
 - added sbt command
 - added build.properties to specify version of sbt
 - deleted the README and LICENSE templates that have not generated anything
 - regenerated sample directory

please review
@clasnake  @jimschubert  @shijinkui


jimschubert:
Plenty of people prefer Gradle over SBT. Many teams _have_ to use Maven (my team at Expedia is one of few that uses SBT, other teams using Scala build with Maven).

I think a good solution would be to add a generator option for which build tool to generate, and maybe default to sbt.

wing328:
@magnolia-k thanks for the PR.

We'd a discussion on that before.

One option is to add yet another option on which build file (maven, gradle, sbt) to generate.

Agreed with you that sbt is a lot more popular among Scala developers (but I also found a few Scala developers using gradle in their Scala project)

My view is to produce all 3 of them by default and let the developers choose which one they prefer.

One way to omit other bulid files (e.g. pom.xml) is to whiltelist them in the `.swagger-codege-ignore` file.

magnolia-k:
understood

There are modules that do not support gradle, such as Scalatra, but should you add support for gradle there?

jimschubert:
when you say the module doesn't support Gradle, do you mean our Scalatra generator? If there isn't a pom.xml or Gradle files there, we should probably include them just for consistency.

One reason why I personally prefer the option over generating everything and then excluding in the ignore file... is that custom templates will have to include empty files for those the user isn't generating. At least, I believe this was the case in a recent Scala client. With an option, one could pass a config json and simplify a little.

magnolia-k:
Oh, for that reason, is there a pom.xml in Scalatra's module directory?

PR #7473, opened 2018-01-22
Issue 4824 fix gson deserialize format byte

### LINKED ISSUE #4824: [JAVA] Codegen - swagger type=string format=byte not serialised correctly
##### Description

I am currently creating frameworks for the various supported languages using Codegen. Generated clients for Python, PHP, Ruby and C# work as expected, as do other 3rd party systems that support Swagger files (e.g. Microsoft Flow). However, the JSON generated by the Codegen Java client is not correct, or at least not in line with the JSON generated by the other languages.

Related snippet from the Swagger file

```json
        "source_file_content": {
          "description": "Content of the file to convert",
          "x-ms-summary": "Source file content",
          "type": "string",
          "format": "byte",
          "uniqueItems": false
        }
```

The Java code that is generated (snippets from the same model file):

```java
  @SerializedName("source_file_content")
  private byte[] sourceFileContent = null;

  public ConvertData sourceFileContent(byte[] sourceFileContent) {
    this.sourceFileContent = sourceFileContent;
    return this;
  }

  @ApiModelProperty(example = "null", required = true, value = "Content of the file to convert")
  public byte[] getSourceFileContent() {
    return sourceFileContent;
  }

  public void setSourceFileContent(byte[] sourceFileContent) {
    this.sourceFileContent = sourceFileContent;
  }
```

When I make the REST call this is serialised to the following JSON

```json
    {
        "use_async_pattern":false,
        "source_file_name":"test.txt",
        "source_file_content":[83,71,86,108,98,71,56,61],
        "output_format":"PDF",
        "copy_metadata":false,
        "fail_on_error":true
    }
```

As you can see the byte array is serialised as the individual bytes in the array,

However, the JSON generated by the clients for the other languages serialise to a string, which is what we need. (Please ignore that the string is base64 encoded, that is not done by the client framework, but rather by the application that calls the framework)

```json
    {
         "use_async_pattern":false,
         "source_file_name":"test.txt",
         "source_file_content":" UEsDBBQAB[string truncated for brevity]==",
         "output_format":"PDF",
         "copy_metadata":false,
         "fail_on_error":true
    }
```

When I change the `byte[]` types in the generated model to `String` then all works well.

```java
  @SerializedName("source_file_content")
  private String sourceFileContent = null;

  public ConvertData sourceFileContent(String sourceFileContent) {
    this.sourceFileContent = sourceFileContent;
    return this;
  }

  @ApiModelProperty(example = "null", required = true, value = "Content of the file to convert")
  public String getSourceFileContent() {
    return sourceFileContent;
  }

  public void setSourceFileContent(String sourceFileContent) {
    this.sourceFileContent = sourceFileContent;
  }
```

##### Swagger-codegen version

I have tried both the current stable (2.2.1) as well as the Master. Both exhibit the same behaviour.

The Swagger file can be found at https://api.muhimbi.com/api-docs/v1/swagger.json

```bash
   -i https://api.muhimbi.com/api-docs/v1/swagger.json \
   -l java \
   -o java-client-new
```

##### Steps to reproduce

1. Generate client framework as per the above command.
2. Open the model/ConvertData.java file and inspect the `sourceFileContent` related properties.

##### Related issues

Issue #669 (Support binary data in body). It appears that this may actually be the cause of this bug, though not 100% sure.

##### Suggest a Fix

I am afraid that my knowledge of Java is limited, so I am in no position to fix this myself. However, based on the behaviour described above, as well as the way other codegen clients behave, when type=string and format=byte the data type of the generated Java property should be `String`, not `byte[]`.

When I manually make this change in the generated model, the correct JSON is generated and our API works well when used from Java. There is no need to change the serialisation logic, just the data types.

Interestingly, looking at the C# code generated by Codegen, that also uses the `byte[]` type for these properties. As that client generates the correct JSON, the serialiser must behave differently.

If this issue results in a change I am able (and happy) to test it.

benrobot:
The java client based off of okhttp-gson libraries failed to serialized and deserialize `type=string`, `format=byte`

This PR aims to add byte[] serialization/deserialization functionality by adding a new Gson type adapter which employs the Base64 class available starting with Java 8

Unit testing showed that the default gson setup was also escaping the = (equal signs) commonly found at the end of a base64 string therefore, the disableHtmlEscaping option has been set via the gson builder

Fix #4824 
cc: @bbdouglas, @JFCote, @sreeshas, @jfiala, @lukoyanov, @cbornet, @jeff9finger


modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/JSON.mustache
```diff
@@ -37,6 +37,7 @@ import java.text.ParsePosition;
 import java.time.LocalDate;
 import java.time.OffsetDateTime;
 import java.time.format.DateTimeFormatter;
+import java.util.Base64;
```
cbornet: Okhttp has a Base64 class that can be used instead of the Java8 one. This way android users can enjoy the fix.

modules/swagger-codegen/src/test/resources/2_0/petstore.json
```diff
@@ -917,6 +917,10 @@
             "pending",
             "sold"
           ]
+        },
+        "profilePhoto": {
```
cbornet: This will impact all generators. So all examples shall be regenerated

modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/JSON.mustache
```diff
@@ -55,6 +56,9 @@ public class JSON {
     private OffsetDateTimeTypeAdapter offsetDateTimeTypeAdapter = new OffsetDateTimeTypeAdapter();
     private LocalDateTypeAdapter localDateTypeAdapter = new LocalDateTypeAdapter();
     {{/jsr310}}
+    {{#java8}}
+    private LocalByteArrayAdapter localByteArrayAdapter = new LocalByteArrayAdapter();
```
cbornet: Why not just ByteArrayAdapter ?

modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/JSON.mustache
```diff
@@ -105,6 +109,10 @@ public class JSON {
             .registerTypeAdapter(OffsetDateTime.class, offsetDateTimeTypeAdapter)
             .registerTypeAdapter(LocalDate.class, localDateTypeAdapter)
             {{/jsr310}}
+            {{#java8}}
+            .disableHtmlEscaping(){{! Added to prevent base64 = (equal sign) from being escaped }}
```
cbornet: The gson default is to escape html (it is the contrary for Jackson) so this should be a choice left to the SDK user.

samples/client/petstore/java/okhttp-gson/src/test/java/io/swagger/client/api/PetApiTest.java
```diff
@@ -1,393 +1,170 @@
...
+     * Add a new pet to the store
+     *
+     * 
+     *
+     * @throws ApiException
+     *          if the Api call fails
+     */
     @Test
-    public void testApiClient() {
-        // the default api client is used
```
cbornet: Something is wrong. The tests shouldn’t be deleted

modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/JSON.mustache
benrobot: Removed disableHtmlEscaping.  Changed test to compensate.

modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/JSON.mustache
benrobot: Agreed. I thought I was following the naming convention of the other type adapters but only the DateAdapter was prepended with Local.

modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/JSON.mustache
benrobot: Done

samples/client/petstore/java/okhttp-gson/src/test/java/io/swagger/client/api/PetApiTest.java
benrobot: I will put these back.

modules/swagger-codegen/src/test/resources/2_0/petstore.json
benrobot: Will do

wing328: We want to keep petstore.json, petstore.yaml the same as the Petstore server (https://petstore.swagger.io) so please undo the change in these files.

I noticed that you already added the additional test case in petstore-with-fake-endpoints-models-for-testing.yaml (there's an similar test case here: https://github.com/benrobot/swagger-codegen/blob/53018c475ad6bbc097f29d47b11a2bc55b5142ea/modules/swagger-codegen/src/test/resources/2_0/petstore-with-fake-endpoints-models-for-testing.yaml#L1202)

benrobot: Thank you.  I'll undo the addition of Pet.profilePhoto and use the existing format_test.byte.

benrobot: Just to be clear.  I am ***not*** regenerating all the clients because I've undone the changes instead.

samples/client/petstore/java/okhttp-gson/src/test/java/io/swagger/client/api/PetApiTest.java
benrobot: The deleted tests have been undeleted

wing328:
@benrobot thanks for the PR, which has been merged into master.

brendandburns:
@wing328 we're using the 3.0.0 branch but I'd like to have this fix. Is there a good way to do that? Can I propose a cherry-pick of this PR back into the `3.0.0` branch?

Thanks!
--brendan

webron:
@HugoMario ping

bakshiabhisek:
@wing328 Looks like this fix is still not in 3.0.11?

tadschik:
Seems its missing here:
https://github.com/swagger-api/swagger-codegen-generators/blob/master/src/main/resources/handlebars/Java/libraries/okhttp-gson/JSON.mustache

PR #7494, opened 2018-01-24
Fix little bug in csharp READMEs

kaasasolut:
I've skipped the PR checklist because I've only edited README files a little bit.


wing328:
@kaasasolut thanks for the PR, which has been merged into master.

PR #7497, opened 2018-01-24
Upgraded swagger-parser and swagger-core versions

EtonDoze:
`swagger-parser` released a new version on 1/23/2018. It has lots of fixes for `allOf` notation. Also updating `swagger-core versions` since it came as transitive dependency of `swagger-parser`.
The code diff is only version number change throughout the codebase. No other files/classes/tests changed here.

Ran all tests and they passed with flying colors.

Not sure if I need to..
* Create a new issue for this upgrade PR
* Include all the sample files that changes due after running `bin/java-petstore-all.sh`


wing328:
@EtonDoze thanks for the PR, which looks good to me.

EtonDoze:
Thanks @wing328 for the merge. I am really looking forward to use these fixes in the released version. I wonder what is the process/plan to release next version of `swagger-codegen`? How do I trigger that?

wing328:
We hope to do a quarterly release but there's no guarantee.

Issue #7070, opened 2017-11-28
[C#] Method signatures use incorrect casing when generated since PR #6305


jamesbar2:
##### Description

Since Issue #6281 and PR https://github.com/swagger-api/swagger-codegen/pull/6305, the parameters in model constructors are not using the [recommended style rules from Microsoft](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-parameters).  This creates breaking changes for our users who are using this formatting:

###### Old:
`new Model(argument1: "foo", argument3: "bar")` 

###### New:
`new Model(Argument1: "foo", Argument3: "bar")` 

**This does not apply to the API method parameters, only model constructor parameters.**

##### Swagger-codegen version

2.3.0

###### JSON Language Configuration File:
```
{
    "packageName": "MyApi.Client",
    "apiPackage": "Api"
}
```

Also, attempting adding `"modelPropertyNaming": "camelCase"`, this provided the desired result for the parameters, but then the models were also named in `camelCase`.

##### Command line used for generation
```
docker run --rm -v /swagger_codegen:/local swaggerapi/swagger-codegen-cli generate -i /v1.json -l csharp -o /local/generated/csharp -c /local/csharp.json
```

##### Related issues/PRs

#6281 
PR https://github.com/swagger-api/swagger-codegen/pull/6305

##### Suggest a fix/enhancement

Switch back to using camelCase on the model constructors, per the style guidelines from Microsoft.

jimschubert:
@jamesbar2 Sorry I've just seen this issue. My mother passed in September, and I took a hiatus from open source contributions for a while.

I'll try and have a fix for this by tonight or tomorrow night. I agree that, regardless of property naming, constructor parameters should follow Microsoft's naming guidelines.  The fix I have in mind will cause constructor parameters to always be generated in camelCase, so if you were to use PascalCase property names or snake_case… it should never affect named constructor args.

However, if your users are doing object initializers, any non-PascalCase change to property naming convention would be a client breaking change.

jimschubert:
@jamesbar2 I've opened #7519 to fix this bug. I was wondering if you'd have a minute to evaluate the branch?

jamesbar2:
@jimschubert I'll check it out now!  Sorry to hear about your mom, thanks so much for taking a look a this!

PR #7519, opened 2018-01-28
[csharp] ctor params should always be camelCase

jimschubert:
After PR #6305, var names defaulted to PascalCase results in constructor
arguments also being PacalCase. Model properties and constructor
arguments have no reason to be the same case, and in fact may cause
issues (`name = name` will result in a compilation error).

This commit forces all constructor params in models to lowerCase.

This is a necessary change, for instance, if client SDK consumers assign
using named args:

```
var a = new Model(first = "", second = "")
```

The PacalCase default and update to constructor arg casing will break
existing consumers of the client.

See #7070 for more details and discussion.

### PR checklist

/cc @mandrean


jimschubert:
NOTE: This change may be a breaking change for consumers using named constructor parameter assignments for 2.3.x releases. This PR reverts behavior to pre-2.3.x camel case args for constructors, regardless of model property naming strategy.

jimschubert:
AppVeyor failed because I'll need to account for escaping reserved word. But the approach can be reviewed. I'll comment here when it's fixed.

jimschubert:
I've updated with a parameter-specific lambda for camel casing words. This lambda accounts for rules in escaping reserved words as well as applying some additional logic related to parameter names (e.g. names can't start with numbers). These are all provided by the calling generator.

Issue #7541, opened 2018-01-31
[python][model] Should not import self


wy-z:
##### Description

![image](https://user-images.githubusercontent.com/4213483/35605132-95240598-0681-11e8-8551-7fc7c7dd5b0d.png)

##### Swagger-codegen version

2.3.1

##### Swagger declaration file content or url

```json
"PlacementNode": {
    "type": "object",
    "title": "PlacementNode",
    "properties": {
        "create": {
            "type": "string",
            "format": "date-time"
        },
        "id": {
            "type": "integer",
            "format": "int64"
        },
        "name": {
            "type": "string"
        },
        "parent": {
            "$ref": "#/definitions/PlacementNode"
        },
        "type": {
            "type": "string"
        },
        "update": {
            "type": "string",
            "format": "date-time"
        }
    }
}
```

##### Command line used for generation

generate python sdk

gorlins:
ran into this too... did not occur in v2.2.3

jimmyjames:
Seeing this as well on 2.3.0. Maybe related to #6839? (looking at the generated `# noqa: F401,E501` comments introduced there).

jacobweber:
Same issue....had to manually remove one of the circular references before building.

tomplus:
In my opinion we should remove these imports from models. It makes a lot of problems with circular dependencies between 2 or more models (model A has an attribute with model B and model B has a optional attribute with model A). These imported objects are not used at all and we know about it - we added F401 to disable warning from pylint :/

macks22:
Didn't have time to make any contributions to this PR, but I made a gist that can be used to comment out the problematic lines after the code is generated. Hopefully this can be useful for those still dealing with this problem.

https://gist.github.com/macks22/a087cce914090b4a5d3a16e8ab3c28aa

erraggy:
> In my opinion we should remove these imports from models. It makes a lot of problems with circular dependencies between 2 or more models (model A has an attribute with model B and model B has a optional attribute with model A).
## YES
Please remove the imports for parent/child referenced types as they do in fact cause runtime errors from circular imports!
![PLEASE](https://media.giphy.com/media/l0NwNrl4BtDD7JCx2/giphy.gif)

Sarek-Wyona:
We currently have data models that use other data types as part of their definition. We are seeing circular imports in this case. 

[...]

PR #7542, opened 2018-01-31
 [python][client] Comment out useless imports

wy-z:
### PR checklist

### Description of the PR

- [python][client] Comment out useless imports to avoid circular importation https://github.com/swagger-api/swagger-codegen/issues/7541


cbornet:
These imports are at least used for the docstrings so I don’t think they can be removed like this.

wy-z:
See #7541 , any PR is welcome~

wing328:
@wy-z I think a proper way to do it is to loop through the `imports` in `postProcessModels` and remove it if it's the same as the model name.

wing328:
Sorry we've to close this one. Please open a new one with the suggested fix.

tomplus:
@cbornet AFAIK docstrings don't require imports. Python complains about these imports too (it's the reason why there is F401 directive for pylint).

@wing328 removing self-import was implemented by removing the first item from the list. We can do it better as you suggest, but the biggest problem is with circular dependencies. I have no idea how to fix it - it'll simpler to delete it.

wing328:
@tomplus right. Let's go with removing it first.

@wy-z If there's a minimal spec to reproduce the issue, please share as it can help verify the fix later.

PR #7551, opened 2018-02-01
[JAVA][Client] New object instead of null for empty POST request

bmordue:
### PR checklist

### Description of the PR

It's safer to initialize the request body to an empty `Object` instead of `null` if there are no body parameters. In particular, this allows Jackson to successfully serialize a POST request with no body parameters when the Content-Type header is set to "application/json, which is set by default.

cc @bbdouglas @JFCote @sreeshas @jfiala @lukoyanov @cbornet @jeff9finger


JFCote:
@bmordue The more I think about it, I'm pretty sure that a body is mandatory in a POST (and PUT) call. So I'm not sure we want to start doing "defensive programming". But I may be wrong. Maybe we should ask the community what they think of this. @wing328 , any thoughs on that?

wing328:
@JFCote there was a use case before in which the POST operation allows empty body (in other words, the body parameter is optional), and the API server will create an object automatically and return it back to the client.

bmordue:
@JFCote Thanks for your feedback. I had a read through the HTTP spec and couldn't see anything that states that POST (or PUT) requests must have a request body. I also found this thread on the IETF mailing list about a similar question: http://lists.w3.org/Archives/Public/ietf-http-wg/2010JulSep/0272.html

In my specific case, the entity that is being POSTed is described by header and query parameters, which is why there's no body in the request. 

From my point of view, part of the problem is that swagger-codegen applies a "Content-Type: application/json" header if none is specified. This means that my service is attempting to parse an empty body as JSON, and fails. The two approaches to resolve this with a generated client seem to be either don't apply the content-type header, or make sure that the request body is parseable as JSON even if I don't have data to send in the body. The second approach seemed to require fewer changes in swagger-codegen with potentially wide effect.

JFCote:
@wing328 @bmordue I understand better now! Thanks for the clarification.

You changes looked good but I wasn't sure about all the generated samples with what looked like regression. Maybe it's because sample didn't get updated in a while but I found strange that the version are going back.

Maybe we should wait for another reviewer to validate.

wing328:
@bmordue CircleCI reports the following issues with updated samples:
```
Tests in error: 
  testUpdatePetWithForm(io.swagger.client.api.PetApiTest): Cannot have body and form params
  testUploadFile(io.swagger.client.api.PetApiTest): Cannot have body and form params
  testDeletePet(io.swagger.client.api.PetApiTest): com.fasterxml.jackson.databind.JsonMappingException: No serializer found for class java.lang.Object and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS) )
  testDeleteOrder(io.swagger.client.api.StoreApiTest): com.fasterxml.jackson.databind.JsonMappingException: No serializer found for class java.lang.Object and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS) )

Tests run: 44, Failures: 0, Errors: 4, Skipped: 3
```

Issue #6375, opened 2017-08-25
Adding custom logging intercetpor throws java.io.IOException: Stream closed


mladendinev:
I generated a java client code from my own swagger definition using the swagger-codegen-cli and _resttemplate_  as codegen library, which worked flawlessly in the beginning. However, I wanted to add my own Logging interceptor which I've done at the initialization stage (Bean initialization) in my application: 

```
@Bean
public DefaultApi defaultApi() {
RestTemplate restTemplate = new RestTemplate(new BufferingClientHttpRequestFactory(new SimpleClientHttpRequestFactory()));
List<ClientHttpRequestInterceptor> interceptors = new ArrayList<ClientHttpRequestInterceptor>();
interceptors.add(new CustomLoggingInterceptor());
restTemplate.setInterceptors(interceptors);
return new ApiClient(restTempalte);
```

When I make a request through the client, I receive a very deeply nested **java.io.IOException: Stream closed.**

wing328:
cc @nbruno (resttemplate creator) to see if he's any clue.

nvbruno:
What's the full stacktrace? Also, your posted code has a typo (restTempalte) and is probably not compilable (DefaultApi is not assignable to ApiClient). Is this the actual code that produces the error?

nbruno:
What's the full stacktrace? Also, your posted code has a typo (restTempalte) and is probably not compilable (DefaultApi is not assignable to ApiClient). Is this the actual code that produces the error?

Chexpir:
@nbruno I reproduce the problem. it's very straightforward to reproduce, I'd say we need to reopen it. it happens on 2.3.1

2018-02-01 17:52:42.267  INFO 2195 --- [nio-8086-exec-3] piClient$ApiClientHttpRequestInterceptor : URI: http://localhost:18088/...-key/2134953542
2018-02-01 17:52:42.267  INFO 2195 --- [nio-8086-exec-3] piClient$ApiClientHttpRequestInterceptor : HTTP Method: GET
2018-02-01 17:52:42.267  INFO 2195 --- [nio-8086-exec-3] piClient$ApiClientHttpRequestInterceptor : HTTP Headers: Accept=[*/*],Content-Type=[application/json],User-Agent=[Java-SDK],Content-Length=[0]
2018-02-01 17:52:42.267  INFO 2195 --- [nio-8086-exec-3] piClient$ApiClientHttpRequestInterceptor : Request Body: 
[...]

PR #7558, opened 2018-02-02
#6375 - Fix I/O Exception. Closed stream.

Chexpir:
Fix attempted read from closed stream, as response.getBody() was reading an InputStream that needs to be read later.

CC @jeff9finger


Chexpir:
As per the conversation in #6375 I am closing the PR, as there is no defect per se, and there is a documented workaround to read twice a request or a response body.

PR #7577, opened 2018-02-04
Feature/unify response handling

jmglsn:
### PR checklist

- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) @jebentier (2017/07) @dkarlovi (2017/07) @mandrean (2017/08) @jfastnacht (2017/09) @ackintosh (2017/09) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

This changes should uniform the response data deserialization process. I recently faced the problem that the expected response model for non default response codes like 404 where not exposed inside the ApiException. This was caused by the different handling of the initial data.
But as you can see there are some other changes, maybe we can see this as a suggestion - that we could discus.

- The `async` functions use the same response handling as the `sync` implementations now. This makes sure that we don't face different response data. The previous implementation missed any exception model mapping.
- The client is using `php 7.1` syntax, I know this is another breaking change, but php 5 is _almost_ dead.
- We have different exceptions, `ApiException` as base and `MappedApiException` for responses that are defined by specification. 
- I also changed the signature of `ApiException`.
- I dropped the dev requirements for phpcs and phpcs-fix, as this are more or less just tools that we might suggest to adapt the code, but for the generated client itself, they are not required.
- phpUnit is now version 6 / 7.

I see some more possible things that I would like to discus with you.

- Can we avoid static calls? Maybe inject the `ObjectSerializer`?
- Should be export / generate interfaces?
- What about another package for the shared client logic? So we could focus more on model / interface generation - and provide different client implementations, not only based on guzzle.
- What about the tests that we generate? Some of them are broken - not by this change. How can we avoid this in future?

I'm sorry for the size of the PR and the BC breaks, but please see this as an suggestion.


wing328:
> The client is using php 7.1 syntax, I know this is another breaking change, but php 5 is almost dead.

Would it be possible to keep the PHP client working with PHP 5.x? This is not to say we encourage PHP developers not to upgrade to 7.x but seems like a lot of users are still using PHP 5.x

For Java, there's an option `supportJava6` to make the Java API client compatible with JDK6 so that developers still have the option to run it in their environment if they can't JDK to the latest version for whatever reason.

ackintosh:
Thank you for your suggestion!
I think that we should keep the PHP client works with  [currently supported versions](http://php.net/supported-versions.php).

Issue #5494, opened 2017-04-26
[Typescript][Fetch] configurable BASE_PATH


mulles3008:
Hi,

I wonder if there is an easy way of having a configurable `BASE_PATH` as I need to switch between different hosts depending on the current environment (production and dev). This should be a common scenario in my opinion.

I realize that there are (at least) 2 solutions:
1. Adjusting the `host` and/or `basePath` fields in the definition file
2. Passing the desired base url to constructor function of each concrete API.

But I believe both options are far from ideal.
1. This means that one needs multiple definition files (i.e. one for production, one for dev) and these need be kept in sync... Also one needs to switch between the production- and the dev-version of the generated clients depending on the environment. 
2. Is OK but tedious nonetheless (especially for large APIs) as one has to repeat the base url whenever one instantiates a concrete API.

Configuring the base url only once in centralized fashion would be less error-prone and much more convenient.

Thanks for any suggestions!

wing328:
I think the following PR is related to your suggestion:

https://github.com/swagger-api/swagger-codegen/pull/3403

mulles3008:
Oh... thx! Will definitely have a look at it!

zwenza:
@wing328 this is still a issue in the typescript-fetch templates.

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/typescript-fetch/api.mustache#L9

I found no way to override the base-path for all APIs, instead you currently need to configure each API in each usage on their own.

Is there any plan to change this?

wing328:
@zwenza the configuration object also contains the `basePath` parameter: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/typescript-fetch/configuration.mustache#L46

Is that what you're looking for?

PR #7609, opened 2018-02-07
[Typescript][Fetch] added option to override default basePath

zwenza:
### PR checklist

### Description of the PR

I joined the discussion in Issue #5494 because i would need the possibility to configure a custom default basePath for all generated API's with the `typescript-fetch` template.

Currently afaik you can only update the base-path for a specific api via the configuration object in the api constructors. But instead of writing this in every API creation i would like to specify a default basePath used by all APIs.

This PR would add a function to the API to set the default base-path `BASE_PATH` to another value. 

I am very open to other ideas how we could tackle this problem. This would be my first working suggestion.


zwenza:
@macjohnny would you like to review this change?

macjohnny:
@zwenza this requires setting the base path for every api, which is ok for now, but we could improve it to use a single configuration class like https://github.com/swagger-api/swagger-codegen/blob/v2.3.0/modules/swagger-codegen/src/main/resources/typescript-angular/configuration.mustache#L10 with static parameters that can be changed. 

Would you like to give it a try? You would also need to include that file in the generation similar to https://github.com/swagger-api/swagger-codegen/blob/v2.3.0/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngularClientCodegen.java#L92 in order to have the `configuration.mustache` processed once.

zwenza:
Sure I can do that!
Just to check if i understand you correct: you want to add a additional moustache for that global configs, not in the existing `configuration.moustache` ?

macjohnny:
@zwenza you are right, we could add a `GlobalConfiguration` class containing static variables only in the existing https://github.com/swagger-api/swagger-codegen/blob/v2.3.0/modules/swagger-codegen/src/main/resources/typescript-fetch/configuration.mustache and since it is already included in the api files, we simply need to use the base path from that static variables instead of the constant defined in the template.

zwenza:
Great, I will add this today

zwenza:
thanks 👍 should i do something further to get this merged, or will someone just pick this up sometime ?

macjohnny:
@zwenza the PR should be ready to be merged. @wing328 should eventually merge it if no issues arise.

wing328:
Is this similar to the singleton approach to have global properties (e.g. basePath)?

We started with the singleton approach for some clients (e.g. Python) and moved away due to various issues (e.g. multi-threading)

wing328:
cc @TiFu (2017/07) @taxpon (2017/07) @sebastianhaas (2017/07) @kenisteward (2017/07) @Vrolijkx (2017/09) @macjohnny (2018/01)

TiFu:
Looks like it is @wing32 . 

I agree with your concerns. I'd prefer an object based approach and just using a 'default' object for the instantiation of the api. This is far more flexible and doesn't run into any multi threading issues (Webworkers are a thing).

kenisteward:
@tifu @wing238. I agree. Object based approach would be best .

I'd even love to see this in the base typescript but I'd need to go back and see how it would flow.

zwenza:
I agree that the current solution could lead into such problems. 
If i understand you right, you suggest to provide a configurable default object which i then would pass each manually at instantiation ?

TiFu:
@zwenza yeah. The default configuration object should be passed by default, maybe similar to portableFetch and GlobalConfiguration.basePath in this sample:

`  constructor(configuration?: Configuration, protected basePath: string = GlobalConfiguration.basePath, protected fetch: FetchAPI = portableFetch) {`

zwenza:
@TiFu so you mean instead of passing only the basePath (which is the current state in my solution) you would pass the whole configuration object?

How would this fix the multi threading issues?

TiFu:
Maybe I wasn't clear enough what my idea was.

1. The configuration object should no longer be `static`, so that you can create multiple objects (e.g. one for each thread), so that they can be changed independently. This solves the multi threading issue.

The second idea was to pass the configuration object directly, so that someone can easily add more options e.g. auth settings to it and use them in the api. Anyway, I just noticed that there's a `Configuration` object already, which contains the [basepath](https://github.com/zwenza/swagger-codegen/blob/e5e47caef158c9f817b2d90dc312ebfd038660b0/samples/client/petstore/typescript-fetch/builds/with-npm-version/configuration.ts#L57)?
This object is used in [L. 49 of api.mustache](https://github.com/swagger-api/swagger-codegen/pull/7609/files#diff-5b00d4f9d86ecef17023b02beda653eaL49) .

Why do we not provide a default object for `configuration` (using `configuration: Configuration = defaultObject`)? 
The constructor of configuration could set appropriate default values (e.g. the [base path contained in GlobalConfiguration l. 5](https://github.com/swagger-api/swagger-codegen/pull/7609/files#diff-341ffefa541963a8916cf40ded6ded47R5). A value passed by the parameter `param` of `Configuration#constructor` should take precedence over the default values. 

This would allow changing all available settings, without passing all the parameters manually. No idea why I didn't notice that earlier.

zwenza:
If i make the basePath non-static wouldn't i have the problem again that i cannot set the basePath global to another value?

TiFu:
If you only ever use the same object instance of GlobalConfiguration (or `Configuration` e.g. a default instance which could be exported by `configuration.mustache`) that wouldn't be a problem. 

Let me make the idea a bit clearer with a gist: [https://gist.github.com/TiFu/e5efe5a920dfd1c54231d98cc9bfc8d8](https://gist.github.com/TiFu/e5efe5a920dfd1c54231d98cc9bfc8d8)

Would this solve your issue or am I missing something obvious?

zwenza:
@TiFu thanks i understand now! :)

zwenza:
@TiFu i implemented your proposed solution. Can you review the changes again? 

It seems like the circleci build failed because it wasn't able to fetch a dependency but i cannot retrigger the build.

zwenza:
@wing328 is this ready to merge? I cannot retrigger the circleci build, maybe you have the permission to do that ?

Issue #7612, opened 2018-02-07
[MAVEN PLUGIN] NullPointerException when not specifying configOptions


matthewmcgarvey:
##### Description

When specifying additionalProperties, reservedWordMappings,etc. without specifying configOptions the plugin throws a NullPointerException.

##### Swagger-codegen version
2.3.1 and master (2.4.0-SNAPSHOT)

##### Swagger declaration file content or url

```xml
<plugin>
    <groupId>io.swagger</groupId>
    <artifactId>swagger-codegen-maven-plugin</artifactId>
    <version>2.3.1</version>
    <executions>
        <execution>
            <goals>
                <goal>generate</goal>
            </goals>
            <configuration>
                <inputSpec>src/main/resources/api.yaml</inputSpec>
                <language>java</language>
                <additionalProperties>
                    <additionalProperty>key=value</additionalProperty>
                </additionalProperties>
            </configuration>
        </execution>
    </executions>
</plugin>
```

##### Steps to reproduce

Specify additionalProperties without configOptions

wing328:
Close

PR #7613, opened 2018-02-07
[MAVEN PLUGIN] Checking for null configOptions before looking for property

matthewmcgarvey:
### Description of the PR
Issue  #7612

Checks if configOptions is null before checking if it has a property

### DIFF
--- a/modules/swagger-codegen-maven-plugin/src/main/java/io/swagger/codegen/plugin/CodeGenMojo.java
+++ b/modules/swagger-codegen-maven-plugin/src/main/java/io/swagger/codegen/plugin/CodeGenMojo.java
@@ -469,32 +469,32 @@ public class CodeGenMojo extends AbstractMojo {
         }
 
         //Apply Instantiation Types
-        if (instantiationTypes != null && !configOptionsNotNullAndContainsKey("instantiation-types")) {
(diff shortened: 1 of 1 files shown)


wing328:
@matthewmcgarvey thanks for the PR.

What about putting [the code](https://github.com/lampo/swagger-codegen/blob/8b363c4d576679c5eb2dbcf9e8e5cf9ff31894a4/modules/swagger-codegen-maven-plugin/src/main/java/io/swagger/codegen/plugin/CodeGenMojo.java#L470-L499) inside the [if block](https://github.com/lampo/swagger-codegen/blob/8b363c4d576679c5eb2dbcf9e8e5cf9ff31894a4/modules/swagger-codegen-maven-plugin/src/main/java/io/swagger/codegen/plugin/CodeGenMojo.java#L434-L469)?

matthewmcgarvey:
@wing328 I inlined the method calls but it wouldn't go inside that if block because the properties should be applied if configOptions is null as well as when the properties aren't on configOptions.
I was negating `configOptionsNotNullAndContainsKey` in all instances I was using it which could have been confusing.

wing328:
@matthewmcgarvey ah now I understand what you're trying to fix 👍

Issue #7620, opened 2018-02-08
[JAVA][feature] add jackson option to retrofit2 library


matthewmcgarvey:
##### Description
[The retrofit2 library hardcodes Gson as the mapping library](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java#L228).  Allow for a Jackson option while leaving Gson as the default.

##### Swagger-codegen version
Master (2.4.0-SNAPSHOT)

PR #7621, opened 2018-02-08
[JAVA][feature] add jackson support to retrofit2 library 

matthewmcgarvey:
### PR checklist

- [] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR
Issue #7620 

Allows for the retrofit2 library to be used with jackson instead of gson.  Could have solved with templates but gson was hardcoded in the JavaClientCodegen as well as adding a template that depended on gson.

Sidenote: Has there been thought to reorganizing the library structure?  There are a lot of optional properties in JavaClientCodegen that have to do with specifically retrofit


modules/swagger-codegen/src/main/resources/Java/libraries/retrofit2/ApiClient.mustache
```diff
@@ -47,10 +45,15 @@ public class ApiClient {
...
+  {{#gson}}
+  private JSON json = new JSON();
+  {{/gson}}
+  {{#jackson}}
+  private ObjectMapper mapper = new ObjectMapper();
+  {{/jackson}}
 
   public ApiClient() {
-    apiAuthorizations = new LinkedHashMap<String, Interceptor>();
+    apiAuthorizations = new LinkedHashMap<>();
```
cbornet: The code should be compatible with JDK6 (for Android)

modules/swagger-codegen/src/main/resources/Java/libraries/retrofit2/ApiClient.mustache
```diff
@@ -154,40 +156,8 @@ public class ApiClient {
...
-    return this;
-  }
-
-  public ApiClient setLocalDateFormat(DateTimeFormatter dateFormat) {
-    this.json.setLocalDateFormat(dateFormat);
-    return this;
-  }
-
-  {{/jsr310}}
+{{#gson}}{{>gson/mappingConverterFactory}}{{/gson}}
```
cbornet: Can we put this into another class/file instead ?

cbornet: Eg : the JSON class

matthewmcgarvey: I have fixed the java6 issue, but I don't believe that this change will work.  The JSON class for the gson library was necessary for all the gson configuration but Jackson doesn't need any configuration in this circumstance, just to replace the gson logic in this class which is tied to the ApiClient

cbornet: There is also mapper configuration for Jackson that should be moved to a JSON class.

davidyscott:
It looks like matthewmcgarvey put a lot of work into this change, which I would also greatly appreciate seeing.  What's happened with this PR?

PR #7623, opened 2018-02-08
Kotlin httpclient access

kgilmer:
### PR checklist

### Description of the PR

This change causes the internal static OkHttpClient in ApiClient to be passed in the constructor rather than using a private static instance.  This is desired so that HTTP debug logging can be configured.  The constructor change propagates to DefaultApi class.  A default value is constructed which can alternatively be supplied by the client.

This was the simplest way that occurred to me to allow this access but there may be other approaches that are better.  

Testing: 
* I've consumed my forked repo branch via jitpack on an internal project and have used the feature successfully.
* Unit tests pass.

CC @jimschubert

### DIFF
--- a/modules/swagger-codegen/src/main/resources/kotlin-client/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/kotlin-client/api.mustache
@@ -5,6 +5,7 @@ package {{apiPackage}}
 {{/imports}}
 
 import {{packageName}}.infrastructure.*
+import okhttp3.OkHttpClient
 {{#threetenbp}}
 import org.threeten.bp.LocalDateTime
 {{/threetenbp}}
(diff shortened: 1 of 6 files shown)


jimschubert:
Thanks for this PR.

Do you see any issues with having an instance per API? I think it looks fine, but I haven't run any tests on the change.

kgilmer:
Hi Jim, well in my cases I only have one API per swagger client, so no.  But there may be other use cases I'm unaware of.  Its certainly less efficient to have an OkHttpClient instance per Api.  The Java Swagger codegen I think generates another class to hold configuration that can span Apis.  Perhaps something like that could be utilized here but it would be a much bigger change to introduce another class than the few lines I have in this PR.  In any case, for those wishing to share instances of OkHttpClient they can create it in their code and supply it to each Api constructor explicitly.

kgilmer:
Closing this PR in favor of https://github.com/swagger-api/swagger-codegen/pull/7628 due to issue w/ commit email address.

PR #7628, opened 2018-02-09
[kotlin] Allow clients to specify OkHttpClient instance for configura…

kgilmer:
…tion use cases.

### PR checklist

### Description of the PR

This change causes the internal static OkHttpClient in ApiClient to be passed in the constructor rather than using a private static instance.  This is desired so that HTTP debug logging can be configured.  The constructor change propagates to DefaultApi class.  A default value is constructed which can alternatively be supplied by the client.

This was the simplest way that occurred to me to allow this access but there may be other approaches that are better.  

I had a git configuration issue with my work computer, this is a fixed PR.  The original: https://github.com/swagger-api/swagger-codegen/pull/7623.  The only difference is my email address associated with the commit.

Testing: 
* I've consumed my forked repo branch via jitpack on an internal project and have used the feature successfully.
* Unit tests pass.

CC @jimschubert

### DIFF
--- a/modules/swagger-codegen/src/main/resources/kotlin-client/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/kotlin-client/api.mustache
@@ -5,12 +5,13 @@ package {{apiPackage}}
 {{/imports}}
 
 import {{packageName}}.infrastructure.*
+import okhttp3.OkHttpClient
 {{#threetenbp}}
 import org.threeten.bp.LocalDateTime
 {{/threetenbp}}
 
 {{#operations}}
-class {{classname}}(basePath: kotlin.String = "{{{basePath}}}") : ApiClient(basePath) {
+class {{classname}}(basePath: kotlin.String = "{{{basePath}}}", httpClient: OkHttpClient = OkHttpClient()) : ApiClient(basePath, httpClient) {
 
     {{#operation}}
     /**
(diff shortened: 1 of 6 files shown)


kgilmer:
Just following up on this PR...  Are there any concerns with this change?

jimschubert:
I don't see any issues (looks the same as the previous closed PR).

modules/swagger-codegen/src/main/resources/kotlin-client/api.mustache
```diff
@@ -5,12 +5,13 @@ package {{apiPackage}}
...
 
 import {{packageName}}.infrastructure.*
+import okhttp3.OkHttpClient
 {{#threetenbp}}
 import org.threeten.bp.LocalDateTime
 {{/threetenbp}}
 
 {{#operations}}
-class {{classname}}(basePath: kotlin.String = "{{{basePath}}}") : ApiClient(basePath) {
+class {{classname}}(basePath: kotlin.String = "{{{basePath}}}", httpClient: OkHttpClient = OkHttpClient()) : ApiClient(basePath, httpClient) {
```
chriswininger, line 14: have you considered putting the httpClient as the first arg? That would make it so that you can override the httpClient without having to supply the url

chriswininger: I also wonder if it would be better to user OkHttpClient.Builder().build()? I'm not super versed in OkHttp internals so not sure if there's a real difference

PR #7626, opened 2018-02-09
Kotlin empty post body

kgilmer:
…agger-codegen generates clients from Api models without bodies for these methods.

### PR checklist

- [X] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [X] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [X] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [X Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

This change prevents an NPE in the case that a POST/PUT/PATCH request did not specify a model.  The Java client allows empty models, so I believe this is a valid change.

See also: https://github.com/square/okhttp/issues/751

Testing:

I've consumed my forked repo branch via jitpack on an internal project and have used the feature successfully.
Unit tests pass.

CC @jimschubert

PR #7629, opened 2018-02-09
[kotlin] Fix NPE for POST/PUT/PATCH with empty request models.

kgilmer:
…agger-codegen generates clients from Api models without bodies for these methods.

### PR checklist

- [X Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

This change prevents an NPE in the case that a POST/PUT/PATCH request did not specify a model.  The Java client allows empty models, so I believe this is a valid change.

See also: https://github.com/square/okhttp/issues/751

This is a resubmission of https://github.com/swagger-api/swagger-codegen/pull/7626 due to an issue w/ email address associated with commit.

Testing:

I've consumed my forked repo branch via jitpack on an internal project and have used the feature successfully.
Unit tests pass.

CC @jimschubert

### DIFF
--- a/modules/swagger-codegen/src/main/resources/kotlin-client/infrastructure/ApiClient.kt.mustache
+++ b/modules/swagger-codegen/src/main/resources/kotlin-client/infrastructure/ApiClient.kt.mustache
@@ -85,9 +85,9 @@ open class ApiClient(val baseUrl: String) {
             RequestMethod.DELETE -> Request.Builder().url(url).delete()
             RequestMethod.GET -> Request.Builder().url(url)
             RequestMethod.HEAD -> Request.Builder().url(url).head()
-            RequestMethod.PATCH -> Request.Builder().url(url).patch(safeRequestBody(body, contentType))
-            RequestMethod.PUT -> Request.Builder().url(url).put(safeRequestBody(body, contentType))
-            RequestMethod.POST -> Request.Builder().url(url).post(safeRequestBody(body, contentType))
+            RequestMethod.PATCH -> Request.Builder().url(url).patch(requestBody(body, contentType))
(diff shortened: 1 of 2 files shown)


jimschubert:
@kgilmer Thanks. We obviously don't want to prevent people from using the client, so it makes sense to support this use case.

I have a question, though. What is the reason for dropping content-type in the fallback of the elvis operator? Is this required, or can it be `RequestBody.create(null, contentType)`? I ask because, similar to the discussion in the linked okhttp thread, there are plenty of APIs that rely on content negotiation and I'd worry that emptying the content type would result in a 404 or otherwise not touching the expected endpoint.

kgilmer:
@jimschubert at the time it seemed that a message of 0 size would not have a content-type (by definition) but I see your point.  I can update it to pass the content-type.

jimschubert:
@kgilmer yeah I don't think we can solve every edge case. And for those in which an empty content type is preferable, the template can always be customized. I know for the APIs that I regularly use (ASP.NET, Finatra, some Spring) the content types hello define our endpoints. The empty body parsing would be affected by the serializers in use on the server.

kgilmer:
I am not sure about the ci/circleci check failing.  Is this something I need to look into?  Are there other items to be addressed?

wing328:
@kgilmer the CircleCI test failure was due to maven repo throttling the connection to download the dependencies. I've restarted the job and the all tests look good now.

wing328:
@kgilmer thanks for the PR, which has been merged into master. (sorry for the delay in merging due to Chinese New Year celebration)

kgilmer:
Yay!  Happy year of the 🐕!

PR #7635, opened 2018-02-10
7616 correct body params filename

### LINKED ISSUE #7616: Wrong feeder filenames in gatling tests
##### Description

Gatling language generates wrongly named `bodyParams` feeder file.
The generator generates e.g. `xyz-BodyParams.csv` while the generated gatling test expects `xyz-bodyParams.csv` which is also in line with other feeder files e.g. `-headerParams.csv`.

##### Swagger-codegen version

Latest master

Not relevant

`...l scala-gatling...`

##### Steps to reproduce

Use a yaml spec with body parameters and generate the gatling project. 

##### Related issues/PRs

None found.

Change

```java
 FileUtils.writeStringToFile(new File(outputFolder + File.separator + dataFolder + File.separator + operation.getOperationId() + "-" + "BodyParams.csv"), StringUtils.join(bodyFeederParams, ","));
```

to

```java
 FileUtils.writeStringToFile(new File(outputFolder + File.separator + dataFolder + File.separator + operation.getOperationId() + "-" + "bodyParams.csv"), StringUtils.join(bodyFeederParams, ","));
```

I can do a PR for that.

atomfrede:
### PR checklist

@clasnake @jimschubert @shijinkui 

### Description of the PR

The `-BodyParams.csv` file must be named `-bodyParams.csv` as that file is used in the gatling test defnition. Futhermore all other files are also using lowercase names (e.g. `-headerParams.csv`).

close #7616

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ScalaGatlingCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/ScalaGatlingCodegen.java
@@ -252,7 +252,7 @@ public class ScalaGatlingCodegen extends AbstractScalaCodegen implements Codegen
                             operation.setVendorExtension("x-gatling-body-feeder", operation.getOperationId() + "BodyFeeder");
                             operation.setVendorExtension("x-gatling-body-feeder-params", StringUtils.join(sessionBodyVars, ","));
                             try {
-                                FileUtils.writeStringToFile(new File(outputFolder + File.separator + dataFolder + File.separator + operation.getOperationId() + "-" + "BodyParams.csv"), StringUtils.join(bodyFeederParams, ","));
+                                FileUtils.writeStringToFile(new File(outputFolder + File.separator + dataFolder + File.separator + operation.getOperationId() + "-" + "bodyParams.csv"), StringUtils.join(bodyFeederParams, ","));
                             } catch (IOException ioe) {
                                 LOGGER.error("Could not create feeder file for operationId" + operation.getOperationId(), ioe);
                             }
--- a/modules/swagger-codegen/src/main/resources/scala-gatling/build.gradle
+++ b/modules/swagger-codegen/src/main/resources/scala-gatling/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id 'com.github.lkishalmi.gatling' version '0.4.1'
+    id 'com.github.lkishalmi.gatling' version '0.7.1'
 }
 
 repositories {


wing328:
@atomfrede thanks for the PR.

cc @andavis84 for review.

atomfrede:
@wing328 The sample files are supposed to be commited too, right?

jimschubert:
Is this issue platform it environment specific? Like would the previous work only on Windows, or in shells with opts enabled for case insensitive filename matching?

I agree with the fix, just wondering why it would work for some and not others.

atomfrede:
I stumbled across that using linux, didn't tried it on windows.

jimschubert:
I was wondering because I ran the original in my old Mac with HFS+ filesystem. My new Mac is APFS, which I believe is also case insensitive but case preserving. I've had issues with changing filename casing for files stored in git, but never had case issues with tools like this.

When I have a chance this weekend, I'll run again locally and then within an Alpine docker image with Scala and report back here if I find differences. It would be a good point for reviewers to double check filename case sensitivity.

jimschubert:
I tested this out a bit, and I couldn't reproduce the behavior I remembered seeing from my review of the initial generator. I ran each time with `gradle gatlingRun`.

So I looked at how Gatling defines a [classpath resource](https://github.com/gatling/gatling/blob/a48532cca8f230035332ff5b0e4d3901e086941f/gatling-core/src/main/scala/io/gatling/core/util/Resource.scala#L33):

```
private object ClasspathResource {
    def unapply(location: Location): Option[Validation[Resource]] =
      Option(getClass.getClassLoader.getResource(location.path.replace('\\', '/'))).map { url =>
        url.getProtocol match {
          case "file" => FileResource(url.jfile).success
          case "jar"  => ArchiveResource(url).success
          case _      => s"$url is neither a file nor a jar".failure
        }
      }
}
```

And I tested this logic locally with a file under resources and named in all lowercase as `a.txt`. Passing `A.txt` causes the extractor to fail as expected-- I'd expect the JVM to be case sensitive for resources. Shell commands, however, consider the filename case insensitively:

```
$ ls
a.txt
$ cat a.txt
Hello, World!
$ cat A.txt
Hello, World!
```

Which is expected on the case insensitive APFS variant file system.

I'm wondering if it's possible that I ran gatling via gradle in some way that used file based resolution rather than classpath? I'm not that familiar with Gatling to even recall how I ran it previously.  I tested this out using a `FileInputStream` which is what's used in Gatling, and with the same lowercase filename, the following reads the contents when passed as `A.txt`:

```
import java.io._
import scala.io._

object Main extends App {
  val file = new File("/path/to/src/main/resources/A.txt")
  val contents = Source.fromInputStream(new FileInputStream(file)).mkString
  println(contents)
}
```

Obviously this is to be expected because my file system is case insensitive. It's just somewhat weird that in Gatling I could potentially succeed on a case insensitive file system in one path of execution but fail when the same file is loaded via classpath.

Anyway, the fix in this PR is obviously the way to go for consistency.

PR #7652, opened 2018-02-13
[Swift] Add public initializer for modelObject.

ejithon:
### PR checklist

### Description of the PR

https://github.com/swagger-api/swagger-codegen/issues/6941

Add explicit public initializer again for modelObject.mustache.

Initializers are no longer see from other Xcode targets after merged the above issue. Swift struct can omit explicit initializer, but is necessary if reference it from outer packages.

In my case, I usually generate a CocoaTouch Framework from generated codes and reuse it via Carthage.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/swift4/modelObject.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift4/modelObject.mustache
@@ -22,6 +22,14 @@ public struct {{classname}}: Codable {
 {{/isEnum}}
 {{/allVars}}
 
+    {{#hasVars}}
+    public init({{#vars}}{{name}}: {{{datatypeWithEnum}}}{{^required}}?{{/required}}{{#hasMore}}, {{/hasMore}}{{/vars}}) {
+        {{#vars}}
+        self.{{name}} = {{name}}
+        {{/vars}}
+    }
+    {{/hasVars}}
+
 {{#additionalPropertiesType}}
     public var additionalProperties: [String:{{{additionalPropertiesType}}}] = [:]
 
--- a/modules/swagger-codegen/src/main/resources/swift4/modelObject.mustache
+++ b/modules/swagger-codegen/src/main/resources/swift4/modelObject.mustache
@@ -22,13 +22,13 @@ public struct {{classname}}: Codable {
 {{/isEnum}}
 {{/allVars}}
 
-    {{#hasVars}}
+{{#hasVars}}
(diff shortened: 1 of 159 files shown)


wing328:
@ejithon thanks for the PR. cc technical committee for review:

@jgavris (2017/07) @ehyche (2017/08) @Edubits (2017/09) @jaz-ah (2017/09)

(the CircleCI failure can be ignored)

fl034:
I found the same issue while testing 2.4.0 with the new struct models.

[From the Apple Docs: ](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/AccessControl.html#//apple_ref/doc/uid/TP40014097-CH41-ID3)
> **Default Memberwise Initializers for Structure Types**
> The default memberwise initializer for a structure type is considered private if any of the structure’s stored properties are private. Likewise, if any of the structure’s stored properties are file private, the initializer is file private. Otherwise, the initializer has an access level of internal.
> As with the default initializer above, if you want a public structure type to be initializable with a memberwise initializer when used in another module, you must provide a public memberwise initializer yourself as part of the type’s definition.

Thus we have to provide an initializer. Memberwise synthesized initializers are set to `internal` which leads to the fact that they're invisible if the generated client is not in the same module.

fl034:
I tested your changes and they're pretty well working 🎉 . Thanks!

d-date:
I think this is make sense. when converted from class to struct, I forget that memberwise initializer in public struct disappear. unless we merged, they cannot initialize struct from outside.

@wing328 This PR should be merged ASAP. Could you merge this PR?

d-date:
cc @jgavris @ehyche @Edubits @jaz-ah

modules/swagger-codegen/src/main/resources/swift4/modelObject.mustache
```diff
@@ -22,6 +22,14 @@ public struct {{classname}}: Codable {
 {{/isEnum}}
 {{/allVars}}
 
+    {{#hasVars}}
```
d-date: NIT: should be as below for indent

```mustache
{{/allVars}}

{{#hasVars}}
    public init({{#allVars}}{{name}}: {{{datatypeWithEnum}}}{{^required}}?{{/required}}{{#hasMore}}, {{/hasMore}}{{/allVars}}) {
        {{#allVars}}
        self.{{name}} = {{name}}
        {{/allVars}}
    }
{{/hasVars}}
{{#additionalPropertiesType}}
```

wing328:
The change looks reasonable to me. Merging it into master.

PR #7665, opened 2018-02-15
Fix for Visual Studio Project GUID

brabidou:
It appears that ASP Core is now using a new GUID for the project in visual studio. The old one {8BB2217D-0F2D-49D1-97BC-3654ED321F3B} does not load in VS2017. We should now be using {9A19103F-16F7-4668-BE54-9A1E7A4F7556}

Not sure if  @mandrean or @jimschubert maintain aspnetcore or not?

[CSharpProjectSystemPackage from dotnet github page ](https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.CSharp.VS/Packaging/CSharpProjectSystemPackage.cs)

```
    internal class CSharpProjectSystemPackage : AsyncPackage
    {
        public const string ProjectTypeGuid = "9A19103F-16F7-4668-BE54-9A1E7A4F7556";
        public const string LegacyProjectTypeGuid = "FAE04EC0-301F-11d3-BF4B-00C04F79EFBC";
        public const string PackageGuid = "860A27C0-B665-47F3-BC12-637E16A1050A";
        private const string ProjectTypeGuidFormatted = "{" + ProjectTypeGuid + "}";
...
```

### DIFF
--- a/modules/swagger-codegen/src/main/resources/aspnetcore/Solution.mustache
+++ b/modules/swagger-codegen/src/main/resources/aspnetcore/Solution.mustache
@@ -2,7 +2,7 @@
 # Visual Studio 15
 VisualStudioVersion = 15.0.26114.2
 MinimumVisualStudioVersion = 10.0.40219.1
-Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "{{packageName}}", "src\{{packageName}}\{{packageName}}.csproj", "{{packageGuid}}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "{{packageName}}", "src\{{packageName}}\{{packageName}}.csproj", "{{packageGuid}}"
 EndProject
 Global
         GlobalSection(SolutionConfigurationPlatforms) = preSolution


jimschubert:
It's this for ASP.NET Core 2.0, or is 1.0 also affected? I seem to recall someone saying that 1.0 wouldn't open in VS 2017 as well. I had issues with my VM running Windows and wasn't able to test against Visual Studio bugs for a bit, but I've sorted it out and should be able to look into this soon.

modules/swagger-codegen/src/main/resources/aspnetcore/Solution.mustache
```diff
@@ -2,7 +2,7 @@
 # Visual Studio 15
 VisualStudioVersion = 15.0.26114.2
 MinimumVisualStudioVersion = 10.0.40219.1
-Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "{{packageName}}", "src\{{packageName}}\{{packageName}}.csproj", "{{packageGuid}}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "{{packageName}}", "src\{{packageName}}\{{packageName}}.csproj", "{{packageGuid}}"
```
jimschubert: I think this will have to be conditional based on some flag. We support Xamarin Studio (Visual Studio on Mac, now) as well as Rider.

brabidou: Makes sense, most frustrating part is VS just fails with no real indication of why it couldn't load the project. Maybe the documentation can be updated to let others know about this issue?

thanks for looking into it.

wing328: What about generating 2 solutions files for Xamarin and VS Studio respectively? e.g.
```
IO.Swagger.xamarin.sln
IO.Swagger.vs.sln
```
Developers can open the one based on their preference.

jimschubert: I don't know if that's a good approach as it could be very confusing which to open in other editors like VS Code or Rider.

Because this is dependent on VS 2017, I think it would make sense to switch on that, as there's been discussion about other VS 2017 issues. Or, if there's a project version in the new sln format, that would be ideal, because that format would likely be used for multiple future versions of VS. That is, passing `vs=2017` then opening in VS 2019 could be confusing.

brabidou: What about making the project GUID itself a option to be passed in?

PR #7685, opened 2018-02-19
Fix performance linting problem with maps in java ApiClient template

dennistruemper:
I just fixed a little map performance Problem, because our linter keeps finding this and we'd like to use upstream template and not custom ones.


wing328:
@canadaduane tests (with updated samples) passed via https://travis-ci.org/swagger-api/swagger-codegen/builds/344730950

dennistruemper:
@wing328 Hi, are you sure with Client: Python? All I changed was Java

wing328:
@dennistruemper I deleted the comment, which is for another PR.

I started testing this PR. Will let you know if I've any question.

wing328:
@dennistruemper thanks for the PR, which has been merged into master

FYI. Java Petstore samples updated via 724a8fb so that CIs can verify the change.

Issue #7622, opened 2018-02-08
3.0.0 server-code generation always failing due to missing /<backend>/model.mustache


asnowfix:
##### Description

I am trying to generate server-side code out of a open api 3.0 specification. I tried the 2 backends that seems to be available in 3.0.0-rc0 (jaxrs & nodejs-server), with the same output error pattern: a missing Mustache.io file.

```
Caused by: java.io.FileNotFoundException: /JavaJaxRS/model.mustache
```

##### Swagger-codegen version

I am using 3.0.0-rc0 downloaded using wget. This is not a regression, as far as I can tell.

##### Swagger declaration file content or url

https://github.com/openservicebrokerapi/servicebroker/blob/master/openapi.yaml

##### Command line used for generation

```
java -jar swagger-codegen-cli.jar generate --lang=jaxrs --output=. --input-spec=api/openapi.yaml --group-id=com.hp.gdrs --artifact-id=aws-cf-service-broker --api-package=com.
hp.gdrs.osb.api --model-package=com.hp.gdrs.osb.model
```

##### Steps to reproduce

1. Download the Open Service Broker specification as YAML
2. Run swagger-codegen on the above YAML

##### Related issues/PRs

Related issue might be completion of the support of OpenAPI 3.0 support in https://github.com/swagger-api/swagger-codegen/issues/6598

HugoMario:
Hey @asnowfix sorry for delay response.

Right now `swagger-codegen` 3.0.0 is supporting `java` language for client and `inflector` language for server. That's the main reason the template is not found.

We can keep this issue open till `jaxrs` be implemented.

jmini:
Is there a description of the necessary steps to convert a Swagger 2.0 generator to an OpenAPI 3.0 generator? Even a raw TODO list with the main steps would be sufficient to start.

We would also like to have a `jaxrs` backend for our application. For the moment we are using a OpenAPI 3.0.1 Spec, but for code-generation and runtime we have converted it back to Swagger 2.0.

HugoMario:
@jmini no really, i mean there is not a formal document about it, but right now the steps are:

- create (or copy and paste) class from `swagger-codegen` from `3.0.0` branch to `swagger-codegen-generator`. The swagger 2.0 classes have been already replaced by OAS 3 classes.
- create templates on `swagger-codegen-generator`. Keep in mind that these templates work with handlebars.
[...]

PR #7718, opened 2018-02-24
[3.0.0] Rename "jaxrs" to "jaxrs-jersey"

jmini:
In the `JavaJerseyServerCodegen` class there is a `TODO` indicating that `jaxrs` should be renamed to `jaxrs-jersey`.

With the switch from version 2 to version 3, such a change can be made.

Of course on the long term, JaxRS generators will be moved to `swagger-codegen-generator`. See #7622

cc: @HugoMario


HugoMario:
thanks @jmini !

Issue #7557, opened 2018-02-02
[CPPRest] [C++]Object without properties results in segmentation fault in generated code


violetelephant12:
##### Description

When generating code using a `swagger.yaml` file that includes an object without properties in the definitions like so:

```yaml
attributes:
  type: object
```

the generated model file contains this fromJson function:

```cpp
void Attributes::fromJson(web::json::value& val)
{
    if(val.has_field(utility::conversions::to_string_t("attribute1")))
    {
        if(!val[utility::conversions::to_string_t("attribute1")].is_null())
        {
            std::shared_ptr<Object> newItem(nullptr);
            newItem->fromJson(val[utility::conversions::to_string_t("attribute1")]);
            setAttribute1( newItem );
        }
    }
}
```

As you can see a shared_ptr is used that points to `nullptr` therefore every time I run my program I get a segmentation fault when `newItem->fromJson()` is called because it's trying to access the function from the nullptr which can't work.

When I tried adding some dummy properties to the object the generation seemed to work correctly. However, my service return the contents of a JSON file which I can't know the structure of so I can't add any correct properties to workaround this problem. Am I doing something wrong? Is there another way to describe the response? When using the same YAML the generation of an Java server and client works without problems.

##### Swagger-codegen version

I've encountered this problem when trying Codegen version 2.2.1, 2.2.3, 2.3.0 and 2.3.1.

##### Swagger declaration file content or url
This YAML results in the same problem. In my personal swagger file the `attributes` object has more than this one property but that doesn't seem to be an issue.

```yaml
swagger: '2.0'
info:
  title: Test API
  description: test api
  version: "1.0.0"
schemes:
  - https
  - http
tags:
  - name: test
produces:
  - application/json
paths:
  /test:
    get:
      tags:
        - test
      summary: Test path
      operationId: testId
      produces:
        - application/json
      responses:
        '200':
          description: Attributes of the system
          schema:
            $ref: '#/definitions/attributes'
        '400':
          description: Error

definitions:
  attributes:
    type: object
    properties:
      attribute1:
        type: object
```

##### Command line used for generation

I'm using

```
java -jar ~/Downloads/swagger-codegen-cli.jar generate -i ./test.yaml -l cpprest
```

to generate the code.

##### Steps to reproduce

1. Put the above mentioned YAML into a new `test.yaml` file.
2. Download the current codegen jar (version 2.3.1)
3. Use the above mentioned command to generate the client.
[...]

PR #7742, opened 2018-02-28
[CPPRest] Adapt templates to correctly generate code regarding the "Object" class

violetelephant12:
Modify model-source.mustache and model-header.mustache to correctly deal with Object objects by removing use of nullptr and instead creating a new Object object which eliminates the segmentation fault described in issue #7557. This is done by creating methods for each possible datatype that returns a new object of the corresponding type.

### PR checklist

- [] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.

### Description of the PR

Based on issue #7557 this pull request fixes problems when using the "Object" class in C++/CPPRest. Prior to this objects of type "Object" were created in ``fromJson`` function using a ``nullptr`` which led to a segmentation fault upon every execution. Now a new object is generated and used so the error is fixed. This is done by generating a method for each possible datatype that returns a new instance of said type. Also the ``fromJson`` function can now deal with a JSON object containing other objects compared to the prior version which would throw an ``not an array`` error in the same situation.

[Issue #7557 ](https://github.com/swagger-api/swagger-codegen/issues/7557)

### DIFF
--- a/modules/swagger-codegen/src/main/resources/cpprest/model-header.mustache
+++ b/modules/swagger-codegen/src/main/resources/cpprest/model-header.mustache
@@ -5,12 +5,13 @@
  * {{description}}
  */
 
-#ifndef {{modelHeaderGuardPrefix}}_{{classname}}_H_
-#define {{modelHeaderGuardPrefix}}_{{classname}}_H_
+#ifndef {{classname}}_H_
+#define {{classname}}_H_
 
 {{^parent}}
 {{{defaultInclude}}}
-#include "../ModelBase.h"
+#include "ModelBase.h"
+#include "Object.h"
 {{/parent}}
 
 {{#imports}}{{{this}}}
@@ -50,11 +51,24 @@ public:
     /// {{description}}
     /// </summary>
     {{^isNotContainer}}{{{datatype}}}& {{getter}}();
-    {{/isNotContainer}}{{#isNotContainer}}{{{datatype}}} {{getter}}() const;
+    {{/isNotContainer}}{{#isNotContainer}}{{{datatype}}} {{getter}}() const;    
     {{/isNotContainer}}{{^required}}bool {{nameInCamelCase}}IsSet() const;
     void unset{{name}}();
     {{/required}}
     void {{setter}}({{{datatype}}} value);
+    
+    {{#isString}}
+    {{{datatype}}}* getNew{{nameInCamelCase}}Instance();
+    {{/isString}}
+    {{#isContainer}}
+    {{{datatype}}}* getNew{{nameInCamelCase}}Instance();
+    {{/isContainer}}
+    {{^isString}}
+    {{^isContainer}}
+    {{baseType}}* getNew{{nameInCamelCase}}Instance();
+    {{/isContainer}}
+    {{/isString}}
+       
     {{/isInherited}}
     {{/vars}}
 
@@ -72,6 +86,7 @@ protected:
 }
 {{/modelNamespaceDeclarations}}
 
-#endif /* {{modelHeaderGuardPrefix}}_{{classname}}_H_ */
+#endif /* {{classname}}_H_ */
 {{/model}}
 {{/models}}
+
--- a/modules/swagger-codegen/src/main/resources/cpprest/model-source.mustache
+++ b/modules/swagger-codegen/src/main/resources/cpprest/model-source.mustache
@@ -2,6 +2,7 @@
 {{#models}}{{#model}}
 
 #include "{{classname}}.h"
+#include "Object.h"
 
 {{#modelNamespaceDeclarations}}
 namespace {{this}} {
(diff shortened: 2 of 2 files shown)


violetelephant12:
@ravinikam @stkrwork @fvarose

modules/swagger-codegen/src/main/resources/cpprest/model-header.mustache
```diff
@@ -5,12 +5,13 @@
  * {{description}}
  */
 
-#ifndef {{modelHeaderGuardPrefix}}_{{classname}}_H_
-#define {{modelHeaderGuardPrefix}}_{{classname}}_H_
+#ifndef {{classname}}_H_
```
stkrwork: why change this?

modules/swagger-codegen/src/main/resources/cpprest/model-header.mustache
```diff
@@ -50,11 +51,24 @@ public:
     /// {{description}}
     /// </summary>
     {{^isNotContainer}}{{{datatype}}}& {{getter}}();
-    {{/isNotContainer}}{{#isNotContainer}}{{{datatype}}} {{getter}}() const;
+    {{/isNotContainer}}{{#isNotContainer}}{{{datatype}}} {{getter}}() const;
```
stkrwork: why add whitespace at the end?

modules/swagger-codegen/src/main/resources/cpprest/model-source.mustache
```diff
@@ -2,6 +2,7 @@
 {{#models}}{{#model}}
 
 #include "{{classname}}.h"
+#include "Object.h"
```
stkrwork: why reinclude it? you already included it in the header of this

stkrwork: Also, does every single model really need an Object?

modules/swagger-codegen/src/main/resources/cpprest/model-header.mustache
```diff
@@ -50,11 +51,24 @@ public:
...
     {{^isNotContainer}}{{{datatype}}}& {{getter}}();
-    {{/isNotContainer}}{{#isNotContainer}}{{{datatype}}} {{getter}}() const;
+    {{/isNotContainer}}{{#isNotContainer}}{{{datatype}}} {{getter}}() const;    
     {{/isNotContainer}}{{^required}}bool {{nameInCamelCase}}IsSet() const;
     void unset{{name}}();
     {{/required}}
     void {{setter}}({{{datatype}}} value);
+    
+    {{#isString}}
+    {{{datatype}}}* getNew{{nameInCamelCase}}Instance();
```
stkrwork: Why do you need those? Also who frees the memory that you have requested with these functions?

modules/swagger-codegen/src/main/resources/cpprest/model-source.mustache
```diff
@@ -251,13 +286,15 @@ void {{classname}}::fromJson(web::json::value& val)
         {{/isDateTime}}
         {{^isDateTime}}
         if(!val[utility::conversions::to_string_t("{{baseName}}")].is_null())
-        {
-            {{{datatype}}} newItem({{{defaultValue}}});
-            newItem->fromJson(val[utility::conversions::to_string_t("{{baseName}}")]);
-            {{setter}}( newItem );
+        {        	
+           	{{{datatype}}} newItem(getNew{{nameInCamelCase}}Instance());
```
stkrwork: this will produce memory leaks

modules/swagger-codegen/src/main/resources/cpprest/model-header.mustache
violetelephant12: Copy and paste mistake from an older version of the file, I corrected it

modules/swagger-codegen/src/main/resources/cpprest/model-header.mustache
violetelephant12: Same as above

modules/swagger-codegen/src/main/resources/cpprest/model-source.mustache
violetelephant12: You're right, I missed that, I removed the unnecessary include. Also I haven't found a way to only include the Object header under certain circumstances.

modules/swagger-codegen/src/main/resources/cpprest/model-source.mustache
violetelephant12: I don't see why but I changed this part anyway so it (hopefully) is okay with you.

modules/swagger-codegen/src/main/resources/cpprest/model-header.mustache
violetelephant12: The memory is freed exactly as with the prior version in other generated parts of the code. I need these functions to be able to correctly handle the generation of the shared_ptr with an Object. It might seem stupid but this is the only way I found which generates the code correctly without using the prior nullptr (-> segmentation fault) as well as it doesn't touch any other generated code. Every other possible solution broke the build in classes which don't use any Object.

modules/swagger-codegen/src/main/resources/cpprest/model-header.mustache
```diff
@@ -72,6 +86,6 @@ protected:
 }
 {{/modelNamespaceDeclarations}}
 
-#endif /* {{modelHeaderGuardPrefix}}_{{classname}}_H_ */
+#endif /* {{classname}}_H_ */
```
stkrwork: Still wrong

modules/swagger-codegen/src/main/resources/cpprest/model-source.mustache
```diff
@@ -251,13 +285,14 @@ void {{classname}}::fromJson(web::json::value& val)
         {{/isDateTime}}
         {{^isDateTime}}
         if(!val[utility::conversions::to_string_t("{{baseName}}")].is_null())
-        {
-            {{{datatype}}} newItem({{{defaultValue}}});
-            newItem->fromJson(val[utility::conversions::to_string_t("{{baseName}}")]);
-            {{setter}}( newItem );
+        {
```
stkrwork: spaces are used here not tabs

stkrwork: also why is there unnecessary whitespace?

modules/swagger-codegen/src/main/resources/cpprest/model-source.mustache
```diff
@@ -251,13 +285,14 @@ void {{classname}}::fromJson(web::json::value& val)
         {{/isDateTime}}
         {{^isDateTime}}
         if(!val[utility::conversions::to_string_t("{{baseName}}")].is_null())
-        {
-            {{{datatype}}} newItem({{{defaultValue}}});
-            newItem->fromJson(val[utility::conversions::to_string_t("{{baseName}}")]);
-            {{setter}}( newItem );
+        {        	
+           	{{{datatype}}} newItem = getNew{{nameInCamelCase}}Instance();
```
stkrwork: did you test this?

modules/swagger-codegen/src/main/resources/cpprest/model-header.mustache
violetelephant12: Whops, sorry, missed that one. Should be fixed now.

modules/swagger-codegen/src/main/resources/cpprest/model-source.mustache
violetelephant12: I removed all tabs and (hopefully) all whitespace.

modules/swagger-codegen/src/main/resources/cpprest/model-header.mustache
```diff
@@ -10,7 +10,8 @@
 
 {{^parent}}
 {{{defaultInclude}}}
-#include "../ModelBase.h"
+#include "ModelBase.h"
+#include "Object.h"
```
stkrwork: does every model require an object?

stkrwork: does every model require an object?

stkrwork:
Another option to rework the default Value of an Object property is to add an else if here:
https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/CppRestClientCodegen.java#L350

```
} else if (p instanceof ObjectProperty) {
  //...
}
```

PR #7749, opened 2018-03-01
[3.0.0] Fix FileNotFound Exception /Xxxxxxxx/model.mustache

jmini:
In the `swagger-codegen` jar, the `io.swagger.codegen.languages.XxxxCodegen` classes are no longer working, because the moustache templates have been removed and need to be migrated.

When users try to generate code for those languages (using one of the broken Codegen classes) they get following error:

```
Caused by: java.io.FileNotFoundException: /Xxxxxxxx/model.mustache
```

This is not really user friendly.

By unregistering broken classes from the service registry (goal of this pull request), the user will no longer see the generators as available.

```
Can't load config class with name xxxx Available: 
java
inflector
```

Of course the list will grow as soon as additional generators are migrated to the new repository. See: [Swagger Codegen migration (swagger codegen generators repository)](https://github.com/swagger-api/swagger-codegen/wiki/Swagger-Codegen-migration-(swagger-codegen-generators-repository))

In addition, I developpers work on migrating a language to swagger v3, they will have no conflict with old Codegen classes, because they are no longer registered.


jmini:
If this pull request if approved the clean-up section of this [wiki page](https://github.com/swagger-api/swagger-codegen/wiki/Swagger-Codegen-migration-(swagger-codegen-generators-repository)) needs to be modified.

> Remove the old generators from the service registry of the `swagger-codegen` module. … `io.swagger.codegen.CodegenConfig` 

=> This is no longer necessary.

An this needs to be added:

> Have a look at the tests scripts:
>
> * `/bin/*.sh`
> * `/bin/windows/*.bat`
>
> Find the one testing your generator (you can search for “`-l xxx `” where `xxx` is the name of your language)
>
> The java command there needs to be uncommented in order to activate the test back.
> The arguments might need to be updated.

HugoMario:
LGTM thanks @jmini !!

PR #7757, opened 2018-03-05
[3.0.0] Remove unused method from CodegenConfig interface

jmini:
3.0.0 is now using handlebar instead of mustache. The method `processCompiler` is no longer used in `CodegenConfig`. 

Removing it from the interface will allow to remove the `jmustache` dependency in the `swagger-codegen-generators` repository.

cc: @HugoMario


jmini:
Some background about this change:

The dependency is not entirely removed, because it is still used in the `io.swagger.codegen.languages` package.
But this package is no longer used at runtime. The code that is no longer used and present for legacy reasons until the migration is over.

The idea of this change is to be able to use `swagger-codegen-generators` like this:

```
<dependency>
  <groupId>io.swagger</groupId>
  <artifactId>swagger-codegen-generators</artifactId>
  <version>1.0.0-SNAPSHOT</version>
  <exclusions>
    <exclusion>
      <groupId>com.samskivert</groupId>
      <artifactId>jmustache</artifactId>
    </exclusion>
    <!-- ... additional exclusions, depending on the use case -->
  </exclusions>
</dependency>
```

jmini:
@HugoMario: I have merged the master branch into this pull request to fix the CI builds.

HugoMario:
thanks @jmini, this totally makes sense to me, i'll update [generators](https://github.com/swagger-api/swagger-codegen-generators) project based on this change.

jmini:
I had already the change on my fork: https://github.com/swagger-api/swagger-codegen-generators/pull/32

HugoMario:
oh, sorry about it, i think i can merge it anyway.

PR #7766, opened 2018-03-06
[3.0.0] Migrate guava logic to java 8

jmini:
With https://github.com/swagger-api/swagger-codegen-generators/issues/5 we have started to remove guava from `swagger-codegen-generators`.

This is useless if the code used by `swagger-codegen-generators` in `swagger-codegen` is still using Guava.

This pull request removes the guava imports from the code, except in the `io.swagger.codegen.languages` package (that contains already dead code => code that is no longer used and present for legacy reason until the migration is over).

The goal is to be able to use `swagger-codegen-generators` in a project configured as follow:

    <dependency>
      <groupId>io.swagger</groupId>
      <artifactId>swagger-codegen-generators</artifactId>
      <version>1.0.0-SNAPSHOT</version>
      <exclusions>
        <exclusion>
          <groupId>com.google.guava</groupId>
          <artifactId>guava</artifactId>
        </exclusion>
        
      </exclusions>
    </dependency>

### DIFF
--- a/modules/swagger-codegen/src/main/java/config/Config.java
+++ b/modules/swagger-codegen/src/main/java/config/Config.java
@@ -1,7 +1,6 @@
 package config;
 
-import com.google.common.collect.ImmutableMap;
-
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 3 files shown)


jmini:
I have rebased the change, build is now OK. (cc @HugoMario)

HugoMario:
thanks @jmini !

PR #7769, opened 2018-03-06
Check date value before calling to_iso8601

garte:
When deserializing a date value the value has to be a string when calling to_iso8601. Otherwise it fails with a match error due to a is_binary() guard.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/elixir/deserializer.ex.mustache
+++ b/modules/swagger-codegen/src/main/resources/elixir/deserializer.ex.mustache
@@ -24,7 +24,7 @@ defmodule {{moduleName}}.Deserializer do
     value = Map.get(model, field)
     case is_binary(value) do
       true -> case DateTime.from_iso8601(value) do
-                {:ok, datetime} ->
+                {:ok, datetime, _offset} ->
                   Map.put(model, field, datetime)
                 _ ->
                   model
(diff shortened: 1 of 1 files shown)


wing328:
@garte thanks for the PR. 

cc @niku @chingor13 for review.

modules/swagger-codegen/src/main/resources/elixir/deserializer.ex.mustache
```diff
@@ -21,11 +21,15 @@ defmodule {{moduleName}}.Deserializer do
...
   def deserialize(model, field, :date, _, _options) do
-    case DateTime.from_iso8601(Map.get(model, field)) do
-      {:ok, datetime} ->
-        Map.put(model, field, datetime)
-      _ ->
-        model
+    value = Map.get(model, field)
+    case is_binary(value) do
+      true -> case DateTime.from_iso8601(value) do
+                {:ok, datetime} ->
```
chingor13: Another bug is that [`DateTime.from_iso8601`](https://hexdocs.pm/elixir/DateTime.html#from_iso8601/2) returns a tuple of size 3.

So it should actually be `{:ok, datetime, _offset}`

garte: Latest commit should fix this, would that be ok?

wing328:
@garte thanks for the PR, which has been merged into master.

@chingor13 thanks for reviewing the change

Issue #6443, opened 2017-09-06
[nodejs] generated service.js files are placed in the wrong folder


sandro97git:
##### Description

When generating nodejs server code with swagger-codegen-cli, the generated service.js files are placed in the controllers folder but they should be in a folder called service. This folder is not created.

##### Swagger-codegen version

swagger-codegen from master branch. commit b8183fb

##### Swagger declaration file content or url

```yaml
swagger: "2.0"
info:
  description: "This is a sample server Petstore server.  You can find out more about     Swagger at [http://swagger.io](http://swagger.io) or on [irc.freenode.net, #swagger](http://swagger.io/irc/).      For this sample, you can use the api key `special-key` to test the authorization     filters."
  version: "1.0.0"
  title: "Swagger Petstore"
  termsOfService: "http://swagger.io/terms/"
  contact:
    email: "apiteam@swagger.io"
  license:
    name: "Apache 2.0"
    url: "http://www.apache.org/licenses/LICENSE-2.0.html"
host: "petstore.swagger.io"
basePath: "/v2"
tags:
- name: "pet"
  description: "Everything about your Pets"
  externalDocs:
    description: "Find out more"
    url: "http://swagger.io"

schemes:
- "http"
paths:
  /pet/findByStatus:
    get:
      tags:
      - "pet"
      produces:
      - "application/xml"
      - "application/json"
      responses:
        '200':
          headers:
            MyHeader:
              type: string
              default: "*"
          description: "successful operation"
```

##### Command line used for generation

java -DdebugOpeations -jar swagger-codegen-cli.jar generate -i swagger.yaml -l nodejs-server -o out

##### Suggest a fix/enhancement

Probably something has to be corrected in [NodeJSServerCodegen.java](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NodeJSServerCodegen.java), so that the service files are correctly placed in the service folder.

sandro97git:
I never faced the issue with the online code generator, which I think is because it uses an older release which does not have this bug.

Also I never faced the issue with the docker image of the codegen-cli from docker hub. I don't know why the issue isn't present there. (I think the docker image has always latest source code from master, right?). Maybe this is the case because the docker image runs in linux and not windows? (Path syntax or somethin like this?)

DmyMi:
The issue is present only on windows machines with 2.3.0 snapshot codegen. Linux machines are working as intended. 
[...]

PR #7808, opened 2018-03-09
Fix nodejs-server path issue in windows platform

wing328:
### PR checklist

### Description of the PR

To fix https://github.com/swagger-api/swagger-codegen/issues/6443

cc @CodeNinjai (2017/07) @frol (2017/07) @cliffano (2017/07)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/NodeJSServerCodegen.java
```diff
@@ -164,7 +164,7 @@ public String apiFilename(String templateName, String tag) {
     }
 
     private String implFileFolder(String output) {
-        return outputFolder + "/" + output + "/" + apiPackage().replace('.', '/');
+        return (outputFolder + File.separator + output + File.separator + apiPackage()).replace('.', File.separatorChar);
```
frol: 1. Is there a file path API for Java to avoid such an error-prone path concatenation implementation?
2. Why should all the dots (`.`) be replaced with a path separator? What if I have a folder with a dot in its name?

wing328: 1. Not that I'm aware of and I agree the current solution is error-prone so I'm open to new ideas for a better implementation.

2. I do not know the reason. The logic was added before I joined this project. My suggestion is to keep the current behavior and try to come up with workaround/solution if users file an issue about it.

frol: 1. OK
2. If you want to keep the current behavior, you should probably drop those parentheses to limit the replacement only for the `apiPackage()` string instead of the whole result:

    ```
    outputFolder + File.separator + output + File.separator + apiPackage().replace('.', File.separatorChar)
    ```

wing328: 2. Done via e10c09d

frol:
LGTM

wing328:
@dlarr thanks for the review. I think my approach should still work. 

Given that it does not impact Mac/Linux, I'll merge it and ask my friends to try it in their Windows box.

Issue #7809, opened 2018-03-09
[Golang] In case of HTTP error, response body is inaccessible


marians:
##### Description

When a swagger-generated go client performs a request and gets some sort of HTTP error (HTTP response code `>= 300`), the response body is read and placed in the returned error string.

This prevents any later access to the response body.

##### Swagger-codegen version

2.3.1

##### Swagger declaration file content or url

https://github.com/giantswarm/gsclientgen/blob/swagger-codegen-2.3.1/api-spec/oai-spec.yaml

##### Command line used for generation

https://github.com/giantswarm/gsclientgen/blob/swagger-codegen-2.3.1/Makefile#L7-L14

##### Steps to reproduce

The API I am trying to consume has a difference in the response body structure in case of HTTP success than it has in case of an error.

Example for success:

```
HTTP/1.1 200 OK
{
  "general": {...},
  "workers": {...}
}
```

Example for a bad request:

```
HTTP/1.1 200 OK
{
  "code": "BAD_ARGUMENT",
  "workers": "The URL parameter 'num' must not be zero"
}
```

In order to get the details about the error (`code` and `message`) I need to obtain the response body JSON and parse/unmarshal it in my own function.

Unfortunately, the body is no longer accessible at that point (from [source](https://github.com/giantswarm/gsclientgen/blob/swagger-codegen-2.3.1/default_api.go#L614-L618)):

```golang
        defer localVarHttpResponse.Body.Close()
	if localVarHttpResponse.StatusCode >= 300 {
		bodyBytes, _ := ioutil.ReadAll(localVarHttpResponse.Body)
		return successPayload, localVarHttpResponse, reportError("Status: %v, Body: %s", localVarHttpResponse.Status, bodyBytes)
	}
```

The fact that the body has been read once means that it cannot be read again. An attempt to do so yields an error

    http: read on closed response body

I'll add that I think the error returned should not contain the response body. For example, if the error response is large, this may totally obscure any logs the error is written to.

##### Related issues/PRs

None yet.

##### Suggest a fix/enhancement

Do not read the response body in case of an HTTP error, thus leave handling of the error response body to the caller.

antihax:
I have a commit on my repo to handle this. Take a look and let me know how this looks?
https://github.com/antihax/swagger-codegen/commit/2003a2ff7a16997c99644f8dae7d28ec8f64b978#diff-7782decde9080f5a41d4c37aeb579302R61

[...]

PR #7810, opened 2018-03-09
Remove reading the response body in case of HTTP error

marians:
### PR checklist

- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change.

### Description of the PR

Addresses #7809 by not reading the response body in case of an HTTP status larger than / equal to 300, to make the body accessible to callers.

### Unrelated changes

Refreshing the generated code for the petstore example yielded some changes which don't seem to be related to the template change in this PR (EnumStringRequired etc.)

CC @antihax @bvwells

### DIFF
--- a/modules/swagger-codegen/src/main/resources/go/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/go/api.mustache
@@ -209,10 +209,8 @@ func (a *{{{classname}}}Service) {{{nickname}}}(ctx context.Context{{#hasParams}
 	if err != nil || localVarHttpResponse == nil {
 		return {{#returnType}}successPayload, {{/returnType}}localVarHttpResponse, err
 	}
-	defer localVarHttpResponse.Body.Close()
 	if localVarHttpResponse.StatusCode >= 300 {
-		bodyBytes, _ := ioutil.ReadAll(localVarHttpResponse.Body)
-		return {{#returnType}}successPayload, {{/returnType}}localVarHttpResponse, reportError("Status: %v, Body: %s", localVarHttpResponse.Status, bodyBytes)
+		return {{#returnType}}successPayload, {{/returnType}}localVarHttpResponse, reportError("Status: %v", localVarHttpResponse.Status)
 	}
 {{#returnType}}
 	{{#withXml}}
--- a/samples/client/petstore/go/go-petstore/api/swagger.yaml
+++ b/samples/client/petstore/go/go-petstore/api/swagger.yaml
@@ -107,11 +107,11 @@ paths:
         type: "array"
         items:
           type: "string"
+          default: "available"
           enum:
           - "available"
           - "pending"
           - "sold"
-          default: "available"
         collectionFormat: "csv"
         x-exportParamName: "Status"
       responses:
@@ -528,7 +528,7 @@ paths:
       parameters:
       - name: "username"
         in: "path"
-        description: "The name that needs to be fetched. Use user1 for testing. "
+        description: "The name that needs to be fetched. Use user1 for testing."
         required: true
         type: "string"
         x-exportParamName: "Username"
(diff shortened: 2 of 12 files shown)


marians:
The CircleCI test fails with this error:

```
# _/home/ubuntu/swagger-codegen/samples/client/petstore/go/go-petstore
go-petstore/api_another_fake.go:14:2: imported and not used: "io/ioutil"
go-petstore/api_fake.go:14:2: imported and not used: "io/ioutil"
go-petstore/api_fake_classname_tags123.go:14:2: imported and not used: "io/ioutil"
go-petstore/api_store.go:14:2: imported and not used: "io/ioutil"
go-petstore/api_user.go:14:2: imported and not used: "io/ioutil"
FAIL	_/home/ubuntu/swagger-codegen/samples/client/petstore/go [build failed]
[ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:1.2.1:exec (go-test) on project Goswagger: Command execution failed. Process exited with an error: 2 (Exit value: 2) -> [Help 1]
[ERROR] 
[ERROR] 
[ERROR] 
[ERROR]   mvn <goals> -rf :Goswagger
```

In the generated petstore sample client the package `io/ioutil` is imported, but not used, in several files. This is in fact related to the change I provided (remove `ioutil.ReadAll` on the response body).

However I am not sure what to do. https://github.com/giantswarm/swagger-codegen/blob/6e89f5f1ca564ea41f2ed4fea8248153eff9cbf0/modules/swagger-codegen/src/main/resources/go/api.mustache#L157 contains another potential use of `ioutil`.

marians:
@bvwells Thanks! You wrote what I kind of understood already (see my comment above).

How can I make the import conditional? Can I simply apply the same wrapping of `{{#hasFormParams}} ... {{#formParams}} ... {{#isFile}}` etc. to the import line?

marians:
Doesn't seem to work. Import block now looks like this:

```nohighlight
import (
{{#hasFormParams}}
{{#formParams}}
{{#isFile}}
	"io/ioutil"
{{/isFile}}
{{/formParams}}
{{/hasFormParams}}
	"net/http"
	"net/url"
	"strings"
	"golang.org/x/net/context"
{{#imports}}	"{{import}}"
{{/imports}}
)
```

go vet:

```nohighlight
go vet ./...
# _/Users/marian/git/swagger-codegen/samples/client/petstore/go/go-petstore
./api_pet.go:509:13: undefined: ioutil
```

`hasFormParams` etc. are used inside a loop of `operations/operation`. Could that be the reason?

Maybe we can make use of that extra import here?

```
{{#imports}}	"{{import}}"
{{/imports}}
```

How would that work?

bvwells:
Hi @marians, sorry for the delay in getting back to you. 

I suspect that you will have to bind a variable into the mustache template. Have a look at the option ```withXml``` in ```api.mushache''' and https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractGoCodegen.java.

You will probably have to loop over all the operations to determine whether any operation has form parameters.

Hope this helps.

arekkas:
There is a serious issue with this PR: Requests are no longer automatically closed. Developers have to make sure to use `.Close()` on every response coming from this API. **This will cause memory leaks** and should be considered as a breaking change. Even more so, we will cause a nil pointer access (= fatal exception) when trying to close the connection on an empty response object, as returned [here](https://github.com/swagger-api/swagger-codegen/pull/7810/files#diff-293df8239a8ee15d445676a48a96a143R205).

While I generally agree with the concept of having the response body in case of errors available, I think this should be solved by return the response body directly using `ioutils.ReadAll` on error, and close the connection accordingly.

The only other solution I see is to not close the connection in case of a response code > 300. Here, we could access the response body but would still be closing connections, preventing memory leaks. This in turn means that on error we have to close the connection manually. So it has the same breaking changes as discussed in the first paragraph.

I think we need to rework the whole template which is being discussed in https://github.com/swagger-api/swagger-codegen-generators/issues/34

marians:
Closing this PR in the hope that the issue here will be sorted out in https://github.com/swagger-api/swagger-codegen-generators/issues/34 .

Issue #7699, opened 2018-02-21
[C++] [pistache] Code generation for Pistache leads to uncompilable code


yvesdm3000:
##### Description

I can generate a server with 1 API with json model classes which is fine, but when I add the code snippet put in later section of the bugreport, the generated .cpp file doesn't compile:

>In file included from /server/impl/DefaultApiImpl.h:28:0,
>                 from /server/DefaultApiMainServer.cpp:16:
>/server/api/DefaultApi.h:33:38: error: ‘model’ is not a namespace-name
> using namespace io::swagger::server::model;
>                                      ^~~~~
>/server/api/DefaultApi.h:33:43: error: expected namespace-name before ‘;’ token
> using namespace io::swagger::server::model;
>                                           ^
>In file included from /server/DefaultApiMainServer.cpp:16:0:
>/server/impl/DefaultApiImpl.h:36:38: error: ‘model’ is not a namespace-name
> using namespace io::swagger::server::model;
>                                      ^~~~~
>/server/impl/DefaultApiImpl.h:36:43: error: expected namespace-name before ‘;’ token
> using namespace io::swagger::server::model;
>                                           ^
>make[2]: *** [server/CMakeFiles/default_api_server.dir/DefaultApiMainServer.cpp.o] Error 1
>make[1]: *** [server/CMakeFiles/default_api_server.dir/all] Error 2

From what I gather, this 'default' handler that should return HTML doesn't use any model and therefore the header for the model isn't included, but the "using namespace" is however generated (which is a bad idea for a header file BTW, but that is a different issue)

##### Swagger-codegen version

2.4.0-SNAPSHOT

##### Swagger declaration file content or url
```yaml
  '/':
    get:
      summary: Returns the webapp
      produces:
        - text/html
      responses:
        200:
          description: Returns the webapp
          examples:
            text/html:
              <html><body>Main webapp</body></html>
```
##### Command line used for generation

java -jar ../swagger-codegen/modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -v -i swagger.yaml -l pistache-server -o server

##### Suggest a fix/enhancement

Probably an issue in the mustache templates ?

PR #7816, opened 2018-03-11
[C++][pistache-server] Optional integer query parameter is not supported

yvesdm3000:
### PR checklist

- [OK] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [OK] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [OK] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [OK] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

The generated pistache code doesn't support an optional query parameter that is declared as being an integer. This code in the mustache templates fixes this issue.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/pistache-server/api-impl-source.mustache
+++ b/modules/swagger-codegen/src/main/resources/pistache-server/api-impl-source.mustache
@@ -7,7 +7,7 @@
 namespace {{this}} {
 {{/apiNamespaceDeclarations}}
 
-using namespace {{modelNamespace}};
+//using namespace {{modelNamespace}};
 
 {{classname}}Impl::{{classname}}Impl(Pistache::Address addr)
     : {{classname}}(addr)
--- a/modules/swagger-codegen/src/main/resources/pistache-server/api-source.mustache
+++ b/modules/swagger-codegen/src/main/resources/pistache-server/api-source.mustache
@@ -7,7 +7,20 @@
 namespace {{this}} {
 {{/apiNamespaceDeclarations}}
 
-using namespace {{modelNamespace}};
+//using namespace {{modelNamespace}};
+
+template<typename T>
+Pistache::Optional<T> ConvertQueryParam(const Pistache::Optional<std::string>& str)
+{
+    if (str.isEmpty())
+        return Pistache::Optional<T>();
+    return Pistache::Optional<T>(Pistache::Some((T)std::stoi(str.get())));
+}
+
+Pistache::Optional<std::string> ConvertQueryParam(const Pistache::Optional<std::string>& str)
+{
+    return str;
+}
 
 {{classname}}::{{classname}}(Pistache::Address addr)
     : httpEndpoint(std::make_shared<Pistache::Http::Endpoint>(addr))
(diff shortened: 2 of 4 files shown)


modules/swagger-codegen/src/main/resources/pistache-server/api-source.mustache
```diff
@@ -61,7 +74,7 @@ void {{classname}}::{{operationIdSnakeCase}}_handler(const Pistache::Rest::Reque
     {{/hasBodyParam}}{{#hasQueryParams}}
     // Getting the query params
     {{#queryParams}}
-    auto {{paramName}} = request.query().get("{{baseName}}");
+    auto {{paramName}} = ConvertQueryParam<{{dataType}}>(request.query().get("{{baseName}}"));
```
stkrwork: So everything gets treated as an optional now?

modules/swagger-codegen/src/main/resources/pistache-server/api-impl-source.mustache
```diff
@@ -7,7 +7,7 @@
 namespace {{this}} {
 {{/apiNamespaceDeclarations}}
 
-using namespace {{modelNamespace}};
+//using namespace {{modelNamespace}};
```
stkrwork: why remove the namespace?

modules/swagger-codegen/src/main/resources/pistache-server/api-source.mustache
yvesdm3000: Oops yes you're right, I missed that,I'll correct that

modules/swagger-codegen/src/main/resources/pistache-server/api-impl-source.mustache
yvesdm3000: I have this issue where no model is used, so no included headerfile, and this line is like a hardcoded namespace declaration that breaks the build in that situation

https://github.com/swagger-api/swagger-codegen/issues/7699

I admit I forgot to remove that in this commit...

modules/swagger-codegen/src/main/resources/pistache-server/api-source.mustache
yvesdm3000: Oops, yes, I'll correct that.

yvesdm3000: I'm returning back on this. request.query().get() allready returns an optional all the time. So it is allready treated as an optional at all times. Fixing this will involve changing pistache.

yvesdm3000:
```c++
namespace Uri {
    typedef std::string Fragment;

    class Query {
    public:
        Query();
        Query(std::initializer_list<std::pair<const std::string, std::string>> params);

        void add(std::string name, std::string value);
        Optional<std::string> get(const std::string& name) const;
        bool has(const std::string& name) const;

        void clear() {
            params.clear();
        }

    private:
        std::unordered_map<std::string, std::string> params;
    };
} // namespace Uri
```

modules/swagger-codegen/src/main/resources/pistache-server/api-impl-source.mustache
stkrwork: well, as soon as you fix this one, i can approve it.

PR #7840, opened 2018-03-14
[qt5cpp] delete callback data allocated before signal emission

etherealjoy:
### PR checklist

### Description of the PR
Before emitting the signals declared in the API class when the callback from HTTP Worker object is received, there is memory allocation being made which is unknown to user code.

This PR adds a wrapper to any memory allocation being done and schedule deleteLater so that once the signal callback has been invoked, the object can be scheduled to be deleted.
Ideal case would be all memory allocation done by the generated code will be deallocated inside it, and user allocated memory will be done on user code.

Tests
- [X] PetStore 
- [X] Docker API tests with Valgrind

There is a limitation with the current PetStore tests because when the callback is received the loop is exited which prevents deleteLater from running. So the PetStore still have the memory Leak visible.

@stkrwork @fvarose @ravinikam


etherealjoy:
@wing328 How do I remove this annoying old baseline which keeps popping up even after I have resync my fork with upstream/master?

modules/swagger-codegen/src/main/resources/qt5cpp/helpers-header.mustache
```diff
@@ -21,6 +22,19 @@ namespace {{this}} {
     QString stringValue(qint64 value);
     QString stringValue(bool value);
 
+    template <typename ObjPtrT>
```
stkrwork: maybe move this class to its own file

etherealjoy: Do you want to suggest a more appropriate class name also?

etherealjoy: OK, I came up with object wrapper which is what it is..., :)

modules/swagger-codegen/src/main/resources/qt5cpp/QObjectWrapper.h.mustache
```diff
@@ -0,0 +1,28 @@
...
+#define {{prefix}}_QOBJECT_WRAPPER_H
+
+#include <QObject>
+
+{{#cppNamespaceDeclarations}}
+namespace {{this}} {
+{{/cppNamespaceDeclarations}}
+
+    template <typename ObjectPtrT>
+    class QObjectWrapper : public QObject {
```
stkrwork: Maybe also add the prefix to this class name as well

Issue #7866, opened 2018-03-17
[typescript-angular] Incorrect handling of model-name-prefix and model-name-suffix


univac490:
##### Description

When the --model-name-prefix or --model-name-suffix are used, the model filename
includes the prefix/suffix, but the import filename includes them twice.

For example, if we use "--model-name-prefix swg" and the model is "pet", the resulting
model filename is "SwgPet.ts", but corresponding import is:

import { SwgSwgPet} from '../model/SwgSwgPet';

##### Swagger-codegen version

ref: refs/heads/master

##### Steps to reproduce

java -jar swagger-codegen-cli.jar generate -l typescript-angular --model-name-prefix swg

##### Related issues/PRs

 #7865

colepalm-netapp:
@univac490 Has there been any resolution to this? Are you using a workaround?

I'm having the same issue and am debating how to solve it.

univac490:
@colepalm-netapp Please see 44e2b014b64f075d242387c95259dc1b2d1a11be. That's what I'm using.

colepalm-netapp:
@univac490 Awesome! Appreciate it.

ruhkopf:
This is still happening in `3.0.8`. Prefix and Suffix produce incorrect import statements. Any chance this could be fixed also in the 3 branch?

mrpaaske:
This is still an issue with version 2.4.7

thomas-lucas1985:
This is supposed to be fixed in 2.4.9?

I just used it and I'm still ending up with double prefix in the import statements e.g.

import {  IICostCenters } from '../model/iICostCenters' - i is used as prefix obviously.
Anyone else still has this issue? Or am I missing something?

GregAttack76:
The issue is still existing in 2.4.9:
for example: <modelNameSuffix>Suffix</modelNameSuffix> generates: apiResponse**SuffixSuffix**

as import in services

dimitriy-k:
I have issue in 3.0.8 as well, please fix a.s.a.p.

sebge2emasphere:
I have the same with with 3.0.20 (I tried suffix and prefix).

baltie:
So, this was fixed in September 2019 and then reverted back to the faulty behavior in October 2019, without any comments or discussion whatsoever?

fpaaske:
@frantuma @HugoMario What's happening to this one?

cgrabmann:
Any news on this one?
[...]

PR #7868, opened 2018-03-18
Fix #7865: Incorrect handling of --model-name-prefix/suffix (python-f…

### LINKED ISSUE #7865: [Python] Incorrect handling of model-name-prefix and model-name-suffix
##### Description

When the --model-name-prefix or --model-name-suffix are used, the model filename
includes the prefix/suffix, but the import filename includes them twice.

For example, if we use "--model-name-prefix swg" and the model is "pet", the resulting
model filename is "swg_pet.py", but corresponding import is:

from swagger_server.models.swg_swg_pet import SwgPet

Also, reserved word escaping should be done after the prefix/suffix has been added. 

##### Swagger-codegen version

ref: refs/heads/master

##### Steps to reproduce

java -jar swagger-codegen-cli.jar generate -l python-flask --model-name-prefix swg

univac490:
### Description of the PR

Fix import in *_controller.py when using --model-name-prefix or --model-name-suffix

java -jar swagger-codegen-cli.jar generate -l python-flask --model-name-prefix swg 

Before:

from swagger_server.models.swg_swg_error import SwgError

After:

from swagger_server.api_models.swg_error import SwgError

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/FlaskConnexionCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/FlaskConnexionCodegen.java
@@ -446,6 +446,15 @@ public class FlaskConnexionCodegen extends DefaultCodegen implements CodegenConf
         // remove dollar sign
         name = name.replaceAll("$", "");
 
+        // Add prefix/suffix before testing for reserved words/starts with number.
+        if (!StringUtils.isEmpty(modelNamePrefix)) {
+            name = modelNamePrefix + "_" + name;
+        }
+
+        if (!StringUtils.isEmpty(modelNameSuffix)) {
+            name = name + "_" + modelNameSuffix;
+        }
+
         // model name cannot use reserved keyword, e.g. return
         if (isReservedWord(name)) {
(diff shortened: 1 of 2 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/FlaskConnexionCodegen.java
```diff
@@ -654,7 +655,23 @@ public String toModelImport(String name) {
             if (!"".equals(modelPackage())) {
                 modelImport += modelPackage() + ".";
             }
-            modelImport += toModelFilename(name)+ " import " + name;
+
+            // with modelNamePrefix/modelNameSuffix, name already includes
```
frol: This is a bit ugly from the code design side of things, but since it works, l don't mind that much.

univac490: Agreed. I suspect other language generators suffer from the same bug (#7866), and it seems the general assumption is that the name parameter is unprocessed. Perhaps a better fix would be to make sure it actually is. However, doing so would require careful review of the 50+ other language generator files. Thanks @frol.

PR #7869, opened 2018-03-18
Fix #7866: Incorrect handling of --model-name-prefix/suffix (typescri…

### LINKED ISSUE #7866: [typescript-angular] Incorrect handling of model-name-prefix and model-name-suffix

univac490:
### Description of the PR

Fix import in *.service.ts when using --model-name-prefix or --model-name-suffix

java -jar swagger-codegen-cli.jar generate -l typescript-angular --model-name-prefix swg

Before:

import { SwgSwgError } from '../model/swgSwgError';

After:

import { SwgError} from '../model/swgError';

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngularClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngularClientCodegen.java
@@ -19,6 +19,8 @@ import io.swagger.codegen.utils.SemVer;
 import io.swagger.models.ModelImpl;
 import io.swagger.models.properties.*;
 
+import org.apache.commons.lang3.StringUtils;
+
 public class TypeScriptAngularClientCodegen extends AbstractTypeScriptClientCodegen {
     private static final SimpleDateFormat SNAPSHOT_SUFFIX_FORMAT = new SimpleDateFormat("yyyyMMddHHmm");
     private static final String X_DISCRIMINATOR_TYPE = "x-discriminator-value";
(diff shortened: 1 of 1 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngularClientCodegen.java
```diff
@@ -377,7 +379,22 @@ public String toModelFilename(String name) {
 
     @Override
     public String toModelImport(String name) {
-        return modelPackage() + "/" + toModelFilename(name);
+        // with modelNamePrefix/modelNameSuffix, name already includes
```
macjohnny: The implementation lookgs good and well-documented.
However, is there a reason why this isnt implemented in the ‚toModelFilename()‘ method?

univac490: Yes, if you implement it in toModelFilename, the file name will become Error.ts, but the import will stay the same:
import { SwgError} from '../model/swgError';

I did run a test to confirm.

gummibjorn:
awesome, this PR will fix my issues

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngularClientCodegen.java
```diff
@@ -377,7 +379,22 @@ public String toModelFilename(String name) {
...
+
+        if (!StringUtils.isEmpty(modelNamePrefix)) {
+            filename = StringUtils.removeStart(filename, modelNamePrefix);
+        }
+
+        if (!StringUtils.isEmpty(modelNameSuffix)) {
+            filename = StringUtils.removeEnd(filename, modelNameSuffix);
+        }
+
+        return modelPackage() + "/" + filename;
```
gummibjorn: Filename is capitalised. `import { SwgError} from '../model/SwgError';` leads to issues on case sensitive OS at least.

Issue #7845, opened 2018-03-15
3.0.0 "swagger-codegen-cli.jar meta" "can't load template codegen/pom.mustache"


jimcullenaus:
##### Description

Following the instructions in the README under "making your own codegen modules", running the initial step as follows produces the following result:

```bash
me@work:path-to/swagger-codegen$ java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar meta \
>   -o output/myGeneratorDir -n myGeneratorName -p my.company.mygeneratorname
[Thread-1] INFO io.swagger.codegen.cmd.Meta - writing to folder [/mnt/c/path-to/swagger-codegen/output/myGeneratorDir]
[Thread-1] ERROR io.swagger.codegen.AbstractGenerator - codegen/pom.mustache (No such file or directory)
[Thread-1] ERROR io.swagger.codegen.AbstractGenerator - can't load template codegen/pom.mustache
Exception in thread "Thread-1" java.lang.RuntimeException: can't load template codegen/pom.mustache
        at io.swagger.codegen.AbstractGenerator.readTemplate(AbstractGenerator.java:51)
        at io.swagger.codegen.cmd.Meta$1.convert(Meta.java:103)
        at io.swagger.codegen.cmd.Meta$1.convert(Meta.java:92)
        at ch.lambdaj.function.convert.ConverterIterator.next(ConverterIterator.java:37)
        at ch.lambdaj.Lambda.convert(Lambda.java:986)
        at ch.lambdaj.collection.LambdaIterable.doConvert(LambdaIterable.java:142)
        at ch.lambdaj.collection.LambdaList.convert(LambdaList.java:57)
        at io.swagger.codegen.cmd.Meta.run(Meta.java:79)
        at java.lang.Thread.run(Thread.java:748)
```

This is after cloning the repo, switching to the 3.0.0 branch, and running `mvn clean package`, which executed successfully.

This also occurs if I change to the default values like `output/myLibrary`, so it's not an issue of some parts of the codebase hardcoding those values.

##### Swagger-codegen version

3.0.0

Also, Apache Maven version 3.3.9, and JDK version 8.

The same error occurs whether I run in Windows Subsystem for Linux's bash shell, or Windows PowerShell.

jmini:
I don't think that the bootstrap project generated by the engine is up to date with all the changes made with version `3.0.0-SNAPSHOT` of `swagger-codegen`. I will try to have a look at it.

---

[...]

PR #7881, opened 2018-03-20
[3.0.0] Modify 'meta' generator

jmini:
Fix for #7845.

With this version of the 'meta' generator, I was able to follow the section [Making your own codegen modules](https://github.com/swagger-api/swagger-codegen#making-your-own-codegen-modules) without any problem.


jmini:
I think I will add a `bin/run-meta-generator.sh` script and commit the result into the `samples/` folder (similar to what is done for all `lang` generators (with the petstore example)).

jmini:
@HugoMario (or someone else in the team), what is your vision on this topic? Will you merge this?

@webron has told me that the vision is to separate livecycle of `swagger-codegen-generators` and of `swagger-codegen`. In this case, an other solution could be to have the "meta template" in `swagger-codegen-generators` and to call it from `cli`.

What do you think? Can we move forward on this topic (at least with the discussion part)?

HugoMario:
i'll definitely merge this, thanks a lot!!

Issue #7900, opened 2018-03-24
Python: proper support for binary typeMapping


aogier:
##### Description

Hi all,

`binary` typeMapping could probably be better handled by python 2/3 portable `six.binary_type`.

there's already a [TODO in master branch](https://github.com/swagger-api/swagger-codegen/blob/ea2474d7075b52d41547248df86275b574e23f0e/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PythonClientCodegen.java#L91) and I'd be happy to help with that but unfortunately I'm not (anymore) proficient in java language, nor on this codebase.  
Is that really trivial as it seems, that is just putting the proper python types in between quotes ?

Anyway, this could fix kubernetes-client/python#494 and so I came to beg for some help :)  
Otherwise I'll try to take a look on this after refreshing a bit my rusty java skillz.

Thank you, regards !

PR #7902, opened 2018-03-24
add binary type

### LINKED ISSUE #2305: [python] Encoding issue with Python 3 binary string response
When using binary string responses under Python 3, the encoding defaults to UTF8 which is incompatible with binary data. This looks to stem from Python 3 having `str` default to UTF8 and a new type called `bytes` which represents binary data. This is a significant departure from functionality in Python 2 and you can [read more about it](https://docs.python.org/3/whatsnew/3.0.html#text-vs-data-instead-of-unicode-vs-8-bit).

This issue does not exist in Python 2.

**Swagger Revision:** 18262c1b6132a28438e2c3d9426458c413f011f8 (master as of right now)

**Response schema:**
```yaml
schema:
  type: string
  format: binary
```

**Example Error:**
```
  File "./dr.py", line 13, in <module>
    "document_type": "pdf",                                         # pdf or xls or xlsx
  File "/usr/local/lib/python3.5/site-packages/docraptor/apis/doc_api.py", line 203, in create_doc
    callback=params.get('callback'))
  File "/usr/local/lib/python3.5/site-packages/docraptor/api_client.py", line 322, in call_api
    response_type, auth_settings, callback)
  File "/usr/local/lib/python3.5/site-packages/docraptor/api_client.py", line 149, in __call_api
    post_params=post_params, body=body)
  File "/usr/local/lib/python3.5/site-packages/docraptor/api_client.py", line 358, in request
    body=body)
  File "/usr/local/lib/python3.5/site-packages/docraptor/rest.py", line 208, in POST
    body=body)
  File "/usr/local/lib/python3.5/site-packages/docraptor/rest.py", line 171, in request
    r.data = r.data.decode('utf8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe2 in position 10: invalid continuation byte
```

I'm not sure what a good general fix would look like but I have a work-around for our use by delaying byte decoding until right before `JSON` deserialization and short-circuiting `str`. You can see [my fix directly on generated code](https://github.com/DocRaptor/docraptor-python/commit/0cd0b7a7af7ae72f37dd0ce825b1903e98824bd0).

Related Issues and PRs:
- https://github.com/DocRaptor/docraptor-python/issues/2
- https://github.com/DocRaptor/docraptor-python/pull/3

aogier:
Hello,   
this is my humble try to implement a portable python binary stream typemapping. As far as I can teel, this should be non-breaking because six, and could maybe fix #2305 and other py3 encoding-related issues.

For me, this is a sideproduct developed for unlocking kubernetes-client/python#494 as per #7900, altough specific patch for kubeclient will live in 676da33d2089fffb332140a35b48865d93d031e7 because local python generator choices.

Tests went ok, but maybe you could help me point out if more are required.

Thank you, regards

/cc @taxpon @frol @mbohlool @cbornet @kenjones-cisco

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PythonClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PythonClientCodegen.java
@@ -79,6 +79,7 @@ public class PythonClientCodegen extends DefaultCodegen implements CodegenConfig
         languageSpecificPrimitives.add("datetime");
         languageSpecificPrimitives.add("date");
         languageSpecificPrimitives.add("object");
+        languageSpecificPrimitives.add("binary_type");
 
         typeMapping.clear();
         typeMapping.put("integer", "int");
@@ -96,8 +97,8 @@ public class PythonClientCodegen extends DefaultCodegen implements CodegenConfig
         typeMapping.put("file", "file");
         // TODO binary should be mapped to byte array
         // mapped to String as a workaround
-        typeMapping.put("binary", "str");
-        typeMapping.put("ByteArray", "str");
+        typeMapping.put("binary", "binary_type");
+        typeMapping.put("ByteArray", "binary_type");
         // map uuid to string for the time being
         typeMapping.put("UUID", "str");
 


frol:
Given there is no influence on the generated client, the tests might just not cover this use-case. Can you reproduce #2305 in a form of a test?

aogier:
hi @frol , that's a great suggestion, I'll work on that !

HugoMario:
generator class was moved to a different repo, however, changes from this PR were migrated with https://github.com/swagger-api/swagger-codegen-generators/pull/510

PR #7905, opened 2018-03-25
[python] remove imports from models

tomplus:
### PR checklist

### Description of the PR

It resolves the issue https://github.com/swagger-api/swagger-codegen/issues/7541 (circular dependencies in Python models), based on @wy-z work, thanks.

PTAL @taxpon (2017/07) @frol (2017/07) @mbohlool (2017/07) @cbornet (2017/09) @kenjones-cisco (2017/11)


samples/client/petstore/python-asyncio/.swagger-codegen/VERSION
```diff
@@ -1 +1 @@
-2.4.0-SNAPSHOT
\ No newline at end of file
+2.3.0-SNAPSHOT
```
wy-z: It seems that the version should be `2.4.0`~

modules/swagger-codegen/src/main/resources/python/model.mustache
```diff
@@ -6,10 +6,6 @@ import pprint
 import re  # noqa: F401
 
 import six
-{{#imports}}{{#-first}}
```
kenjones-cisco: What about the scenario where the discriminator is used?
The lack of import will result in invalid code due to the missing import.

tomplus: Does this discriminator work if modules can't be imported due to circular dependencies ? 

Maybe there is another way to provide paths/data for the discriminator - I don't know, I don't use it. Such imports make more problems for standard usage of generated code.

kenjones-cisco:
LGTM! Thanks!

wy-z:
+1

modules/swagger-codegen/src/main/resources/python/model.mustache
Sarek-Wyona: What is the status of this PR? I have a similar fix and was about to submit it before stumbling into this PR. 
We are seeing models created with circular imports too which makes the generated code useless unless they are manually commented out!

Issue #7911, opened 2018-03-26
[Typescript-Fetch] - Default Configuration Class conflicts with Generated Class of same name


jawa-the-hutt:
##### Description

The default `Configuration` class that is used to pass in parameters to setup API security is the same name as a generated classes from the Swagger file.  This class is created in the `configuration.mustache` template but is then used in the `api.mustache` template.

##### Swagger-codegen version

2.4.0

##### Swagger declaration file content or url

```json
{
  "swagger": "2.0",
  "info": {
    "version": "3.0.0",
    "title": "Service API"
  },
  "host": "a.host.com",
  "basePath": "/apis/3.0",
  "paths":{
    "/service/tickets/{id}/configurations": {
      "post": {
        "tags": [
          "Tickets"
        ],
        "description": "Create Configuration Association",
        "consumes": [
          "application/json"
        ],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "id",
            "type": "integer",
            "format": "int32",
            "in": "path",
            "required": true
          },
          {
            "name": "configuration",
            "in": "body",
            "required": true,
            "schema": {
              "$ref": "#/definitions/ConfigurationReference"
            }
          }
        ],
      }
    }
  },
  "definitions": {
    "ConfigurationReference": {
      "type": "object",
      "properties": {
        "id": {
          "format": "int32",
          "type": "integer"
        },
        "deviceIdentifier": {
          "type": "string"
        },
        "_info": {
          "$ref": "#/definitions/Metadata",
          "description": "Metadata of the entity",
          "readOnly": true
        }
      }
    }
  }
}
```

##### Command line used for generation
 `generate -i test.json -l typescript-fetch -o out/`

##### Steps to reproduce

Generate the code and you should get warnings in the IDE about the generated class(es).

##### Suggest a fix/enhancement

Allow a command line option that gives you the ability to name the `Configuration` class to something of your choice to avoid conflict with any classes generated from your Swagger definition.  If nothing is passed in, use the default class named `Configuration`.  

But if you did something like this: 
`generate -i test.json -l typescript-fetch -o out/  --additional-properties apiConfigClassName=ApiConfiguration`, then the generated class name would be `ApiConfiguration`. 

This change would also use Mustache Lamda expressions to Lower and Upper case class names automatically so as to stick to standard naming conventions.

[...]

Issue #7595, opened 2018-02-06
[typescript-fetch] 2.3.0 changes how parameters are passed to methods


janslow:
##### Description

In 2.2.3 and below, the `typescript-fetch` generator provided accepted parameters in an object, but 2.3.0 has regressed/made a very breaking change so that all parameters have to be passed individually:

e.g.,

```typescript
# 2.2.3
public getFoo(params: {  "id": string; }, options?: any) { ... }

# 2.3.0
public deleteDataModel(dataModel: string, options?: any) { ... }
``` 

I couldn't find any reference to this as a deliberate change in any of the release notes

##### Swagger-codegen version

2.3.0 (regressed since 2.2.3)

##### Steps to reproduce

Generate any operation with parameters using `typescript-fetch`.

##### Related issues/PRs

I eventually managed to track the regression down to #6130, as `api.mustache` was completely rewritten when it was "copied" from [here](https://github.com/swagger-api/swagger-codegen/pull/6130/files#diff-f9274d8764d3f6292ed89c2f34d8d499) to [here](https://github.com/swagger-api/swagger-codegen/pull/6130/files#diff-5b00d4f9d86ecef17023b02beda653ea)

##### Suggest a fix/enhancement

Restore method signature to 2.2.3 form.

sbusch:
Besides the breaking change, I think it's a bad idea to use positional arguments and wonder what's the reasoning behind this.

@isman-usoh can you comment on this?

Depending on the position of arguments means a changed order could easily break the generated API 

Another problem: our API uses an "options" parameter. Because the `typescript-fetch`generator appends it's own "options?: any", there is a name collision:

```
Failed to compile.

Module parse failed: Argument name clash (5251:89)
You may need an appropriate loader to handle this file type.
|          * @throws {RequiredError}
|          */
|         findEntity: function (name, options, options) {
```

Method signature in TypeScript

```typescript
public findEntity(name: string, options?: string, options?: any) {
```

Issue #3620, opened 2016-08-20
[typescript-angular] Pass parameters in an object


DCstewieG:
I generated my first client and was disappointed to get a method with 15 arguments. This seems like it should be a single object conforming to an interface. It would be especially helpful when you have many optional arguments that you only want to pass one value for. It would also be less error prone if arguments were added, removed, or reordered.

Obviously this would break compatibility. And to be honest it's probably undesirable to people with services taking only a couple parameters. So it would need to be either an option or a second method. I like the idea of an option for code size and simplicity but a second method would allow flexibility for those calling the client.

I'm working on a change myself but would like feedback before submitting a pull request. The output looks like this:

    namespace API.Client {
        'use strict';

        interface IresourcesGetParams {
            /** This is a required parameter */
            requiredParam: string;
            /** This is an optional parameter */
            optionalParam?: string;
        }

        export class DefaultApi {
    
            ...

            /**
             * Retrieve a list of resources  
             * @param params Parameters.
             */
            public resourcesGet (params: IresourcesGetParams, extraHttpRequestParams?: any) : ng.IHttpPromise<ResponseModel> {
- The interface and params argument wouldn't be present if there are no parameters.
- IresourcesGetParams should be I**R**esourcesGetParams but I'm not sure the best way to go about that since I'm just grabbing the nickname.

wing328:
@DCstewieG thanks for the feedback. I'm no expert in Typescript so I cannot speak on behalf of the Typescript developer community on which way is preferred. As you've pointed out, owners of the spec with only a few parameters in the API endpoint would tend to keep the current approach.

I would suggest making the enhancement a non-breaking change by using an option so that developers can choose which method signature to use. (if you need more info on this, please PM me via my email address in my github profile page)

driesdroesbeke:
+1

lbcat:
👍

exytab:
Maybe, it can help. 

Interafaces: 
```
{{#operation}}
export interface I{{classname}}{{nickname}}Params {
	{{#allParams}}
	{{paramName}}{{^required}}?{{/required}}: {{{dataType}}};
	{{/allParams}}
}
{{/operation}}
```

Operations:
```
    {{#operation}}
        /**
         * {{&notes}}
         {{#summary}}
         * @summary {{&summary}}
         {{/summary}}
         {{#allParams}}
         * @param I{{classname}}{{nickname}}Params [params]
         {{/allParams}}
[...]

PR #7916, opened 2018-03-26
Feature/typescript fetch/api config class name

jawa-the-hutt:
### PR checklist

@TiFu @taxpon @sebastianhaas @kenisteward @Vrolijkx @macjohnny

### Description of the PR

Implements suggested changes for `typescript-fetch` client that allows you to change the default name of the `Configuration` class used to setup access to an API.  This would close issue https://github.com/swagger-api/swagger-codegen/issues/7911

This commit adds another _system property_ named `apiConfigClassName`   By adding this property, it will allow you to change the name of this class so as to avoid confusion with template generated classes and interfaces that also are named `Configuration`.  When passing in the new name, you can use camelCase or PascalCase strings and the Mustache Lambda feature will know when to upper/lowercase generated class names and references.

You can activate this property via this method:
```
-i http://petstore.swagger.io/v2/swagger.json \
-l typescript-fetch -o samples/client/petstore/typescript-fetch/builds/api-config-class-name \
-D apiConfigClassName=apiConfiguration
```

If this additional property is not passed in then the class name will default to `Configuration`.


jawa-the-hutt:
So, there is something I need to fix on this before it can be merged.  will update accordingly

.gitignore
```diff
@@ -34,6 +34,8 @@ packages/
 *.xml~
 *.t~
 
+
```
macjohnny: special character? should be removed

modules/swagger-codegen/src/main/resources/typescript-fetch/api.mustache
```diff
@@ -76,7 +76,7 @@ export class RequiredError extends Error {
  * {{&description}}{{/description}}
  * @export
  */
-export const {{classname}}FetchParamCreator = function (configuration?: Configuration) {
+export const {{classname}}FetchParamCreator = function ({{#lambda.lowercase}}{{apiConfigClassName}}{{/lambda.lowercase}}?: {{#lambda.titlecase}}{{apiConfigClassName}}{{/lambda.titlecase}}) {
```
macjohnny: why need to change the parameter name?

.gitignore
jawa-the-hutt: Not sure.  I'm doing all of my development via VS Code in the supplied Docker container.  When it mounts the container into the local path it appears that the  /.m2/repository folder showed up with the folder name having this special character.  Git tried to add all the files in it into source control. So, I excluded it in my .gitignore.  

Open to suggestions as to how to avoid this situation, but I haven't found a solution for it yet.

modules/swagger-codegen/src/main/resources/typescript-fetch/api.mustache
jawa-the-hutt: I'll reply here on this even thought this specific line is getting ready to change when I make a commit here in a few minutes.  `lambda.lowercase` is going to change to `lambda.camelcase_param`.

I agree it does make the template a bit harder to read, but the parameter name is changing to this as it will now be dynamic in nature based on what you pass into the generator. 

That's the basic point of this pull request.  There are swagger definitions that may auto-generate other parameters/variables named `configuration` in the same class.  When this happens you end up with invalid code as you can't have two parameters/variables with the same name.

.gitignore
jawa-the-hutt: FYI, here's how it appears in VS Code:  

capture

modules/swagger-codegen/src/main/resources/typescript-fetch/api.mustache
jawa-the-hutt: @macjohnny, if you haven't yet, take a look at https://github.com/swagger-api/swagger-codegen/issues/7911.  I include a sample swagger definition where if it's run against the generator, you will end up with multiple parameters/variables named `configuration`.

I would also add that the reason I'm putting the `lambda.camelcase_param` is to try and conform to standard Javascript/Typescript patterns that has Class/Interface names as PascalCase and parameters/variables in camelCase.    By using the lambda's, I'm allowing someone to pass in the dynamic Configuration class name as something like `ApiConfiguration` or `apiConfiguration` and the template will then know when to use PascalCase or camelCase automatically based on those standards.

jawa-the-hutt:
I have another set of changes I'd like to merge into the project that deal with solutions specifically for https://github.com/swagger-api/swagger-codegen/issues/7595, but could also be a pattern that is used in addressing https://github.com/swagger-api/swagger-codegen/issues/3620

It adds another system property option when calling the generator that allows you to set a flag that puts all the parameters in an object like it used to in 2.2.3 and below, or just leaves it the default way parameters are generated now from 2.3.0 and above.  I've gotten the changes coded and done but the code includes the code from this pull request in it.  I can open another pull request, but it would essentially supersede this one. 

So, my point is should I wait for this pull request to be fully merged and then submit the other pull request, or should I add those changes into this pull request and we get both at the same time?

macjohnny:
@jawa-the-hutt I think you should open a new PR with the changes for #7595, while leaving this one open, too. That way, the code can be reviewed and tested in parallel.

.gitignore
jawa-the-hutt: Wanted to update this with an issue I just opened up:  https://github.com/swagger-api/swagger-codegen/issues/7928

This appears to be an issue with Docker for Windows and a line in the `run-in-docker.sh` script.

jawa-the-hutt:
Ok, PR https://github.com/swagger-api/swagger-codegen/pull/7930 has been created with the other set of changes.

PR #7921, opened 2018-03-27
[Swift4] accept empty content with default client

### LINKED ISSUE #7850: [Swift4] 204 No Content returns error
##### Description
Swift 4 code-gen issue:
If a request returns a status code of 204 with no content, the completion handler will return an error and nil response.  Since a 204 with no content is expected, no error should be thrown.

##### Swagger-codegen version

Swagger version 2.3.1

##### Steps to reproduce

In swagger, create a request that returns a 204 and no content.  

An error should not be thrown.

d-date:
### PR checklist

### Description of the PR
To fix #7850, we accept `Void` response as empty body.  This behavior is same as `PromiseKit` and `RxSwift` response

### Breaking Change
By merging this PR, for using `default` client users,  they need to change response as follow warnings by compiler.

Before:

```swift
FakeAPI.deletePet(id: "1000") { (error) in
  if error == nil { // error will be tuple contain (response, error), so compiler will show `error == nil always true`
    // success
  }
}
```

After:

```swift
FakeAPI.deletePet(id: "1000") { (response, error) in
  if error == nil { 
    // still means success
  }

  if let _ = response {
   // this also means success
  }
}
```
or you can ignore response as below.

```swift
FakeAPI.deletePet(id: "1000") { (_, error) in
  if error == nil { 
    // still means success
  }
}
```


wing328:
Thanks for the PR.

cc @jgavris @ehyche @Edubits @jaz-ah for review

jgavris:
+1 We've actually been doing this, but didn't want to do a breaking change! What do the others think?

jaz-ah:
@jgavris I think it's ok to add this as a breaking change to 3.0.0 which already has a lot of other breaking changes...

PR #7933, opened 2018-03-28
[qt5cpp] Fix crash when API return a map container

MartinDelille:
### PR checklist

### Description of the PR

The client was crashing during the callback with a map container. This was due to an initialized pointer passed to the *setValue* method.

@ravinikam @stkrwork @fvarose @etherealjoy

### DIFF
--- a/modules/swagger-codegen/src/main/resources/qt5cpp/api-body.mustache
+++ b/modules/swagger-codegen/src/main/resources/qt5cpp/api-body.mustache
@@ -185,7 +185,7 @@ void
     QJsonObject obj = doc.object();
 
     foreach(QString key, obj.keys()) {
-        qint32 val;
+        {{returnBaseType}} val;
         setValue(&val, obj[key], "{{returnBaseType}}", QString());
         output->insert(key, val);
     }
--- a/modules/swagger-codegen/src/main/resources/qt5cpp/api-body.mustache
+++ b/modules/swagger-codegen/src/main/resources/qt5cpp/api-body.mustache
@@ -185,9 +185,9 @@ void
     QJsonObject obj = doc.object();
 
     foreach(QString key, obj.keys()) {
-        qint32* val;
+        qint32 val;
         setValue(&val, obj[key], "{{returnBaseType}}", QString());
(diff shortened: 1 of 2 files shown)


modules/swagger-codegen/src/main/resources/qt5cpp/api-body.mustache
```diff
@@ -185,9 +185,9 @@ void
     QJsonObject obj = doc.object();
 
     foreach(QString key, obj.keys()) {
-        qint32* val;
```
etherealjoy: Should be {{{returnBaseType}}}

etherealjoy: Not part of this PR, but in case the {{{returnBaseType}}} is dynamically allocated (Object, QString,QDateTime,QbyteArray ... etc ), it has to be wrapped with QObjectWrapper for later deletion.

MartinDelille: Could you tell me more about QObjectWrapper? I never used it.

etherealjoy: Ahh, it is a template class which holds the pointer to the memory allocated for dynamically allocated types like QString, Object, QByteArray etc.., for the parameter to be passed in the slot callback. In this file you can find it being used, so once the signal is emitted and the slot activated, the deleteLater ensures the object is deleted after the process event has finished.
Present in the template QObjectWrapper.mustache

auto objwrapper = new {{prefix}}QObjectWrapper<{{{returnBaseType}}}*> (o);
objwrapper->deleteLater();

etherealjoy: Like I mentioned, it is not part of the crash, but simply to avoid memory leaks for complex types. If you have the time it is a good thing to add, to remove memory leaks. But for the qt5cpp petstore example the cleanup is not working because the loop is exited on receiving the callback. In a real qt application the event loop would still be running and the object deleted.

PR #7963, opened 2018-04-04
Issue 7949 - Added in injected basePath observable to api service to allow it to respond to external updates to this setting post construction.

chrismellard:
Extended api service to allow optional injection of an observable basePath. This facilities configuration of basePath post service construction which is required in circumstances where runtime setting of basePath cannot be achieved using APP_INITIALIZER tokens in angular. One such use case is when using libraries such as ngrx effects that do not respect such angular APP_INITIALIZER tokens.


chrismellard:
Another option is to simply include the observable as part of the Configuration class rather than a separate dependency injectable token. If its preferable to move it inside that class I can do that.

PR #7968, opened 2018-04-04
[PHP] Fix deserialize ApiException as a Model

### LINKED ISSUE #7544: [PHP] Bug deserialize default responses as a Model
##### Description

I'm generating clients that make use of `default` responses for other HTTP codes (implying an error). It's defined as a Model (code, message) in `definitions` in the swagger file.

```yaml
...
  responses:
    '200':
      description: pet response
      schema:
        type: array
        items:
          $ref: '#/definitions/Pet'
    default:
      description: error payload
      schema:
        $ref: '#/definitions/ErrorModel'
...
```

After generating the client API with swagger-codegen, when catching an `ApiException`, the deserialized response object `ApiException::responseObject` is empty (it is an an instance of `ErrorModel` as expected, but with no content).

##### Swagger-codegen version

2.3.0

[gist: minimal swagger.json example](https://gist.github.com/ymilin/74a804dc5130dfda2c2c6ed349f22865)

```
java -jar bin/swagger-codegen-cli.jar generate -i swagger.json \
  -l php \
  -o build/ \
  --git-user-id "vendor" \
  --git-repo-id "ws-clients" \
  -DpackagePath=.,srcBasePath=lib,variableNamingConvention=camelCase,invokerPackage="MyCie\\Service",apiTests=false,apiDocs=false,modelTests=false,modelDocs=false
```

##### Steps to reproduce

With the generated client: 
- try to make a request for a non-existing pet
- Expect `$e->getResponseObject()` to be a fully hydrated instance of ErrorModel
- Got an empty instance of ErrorModel instead

```php
$petApi = new PetApi();

try {
  $pet = $api->getPetById(123);
  ...
} catch (ApiException $e) {
  $errorModel = $e->getResponseObject();
  print_r($errorModel); // empty ErrorModel
}
```

I checked the generated code and the bug seem to be located in the `catch` part of the `{{operationId}}WithHttpInfo` methods:

```php
try {
    ...
} catch (ApiException $e) {
    switch ($e->getCode()) {
        case 200:
            $data = ObjectSerializer::deserialize(
                $e->getResponseBody(),
                '\MyCie\Service\Model\Pet',
                $e->getResponseHeaders()
            );
            $e->setResponseObject($data);
            break;
        default:
            $data = ObjectSerializer::deserialize(
                $e->getResponseBody(), // <-- works with json_decode($e->getResponseBody())
                '\MyCie\Service\Model\ErrorModel',
                $e->getResponseHeaders()
            );
            $e->setResponseObject($data);
            break;
    }
    throw $e;
}
```

We can see that there is a specific test for this in the `try` part of same function in the `api.mustache` file, line 164

https://github.com/swagger-api/swagger-codegen/blob/36f69a034d2c6503b3e3cee00d10ab8944cc5be6/modules/swagger-codegen/src/main/resources/php/api.mustache#L163-L166

So a possible fix would be doing something similar when deserializing for `ApiException` objects.

I can work on a PR for this, as I really need it, but I have no experience with mustache templates and will not be able to test edge cases.

So I would need help for testing my PR.

ymilin:
### PR checklist

### Description of the PR

When defining `default` responses (`ApiException`) as a Schema Object in Swagger declaration files, the current implementation cannot `deserialize` the Model properly into `ApiException::$responseObject`.

Added a check for `$returnType` and `json_decode` appropriately (same logic as the try part of same function for success responses)

fix #7544


ymilin:
@jebentier @dkarlovi @mandrean @jfastnacht @ackintosh

ackintosh:
Thanks for the PR! I'll check it later. 😉

modules/swagger-codegen/src/main/resources/php/api.mustache
```diff
@@ -177,12 +177,16 @@ use {{invokerPackage}}\ObjectSerializer;
...
+            if ($returnType !== 'string') {
+                $content = json_decode($content);
+            }
             switch ($e->getCode()) {
         {{#responses}}
             {{#dataType}}
                 {{^isWildcard}}case {{code}}:{{/isWildcard}}{{#isWildcard}}default:{{/isWildcard}}
                     $data = ObjectSerializer::deserialize(
-                        $e->getResponseBody(),
+                        $content,
```
ackintosh: Each `case` section has own response type like below. ( `\Swagger\Client\Model\Pet`, `\Swagger\Client\Model\ErrorModel` in example)

```php
        } catch (ApiException $e) {
            switch ($e->getCode()) {
                case 200:
                    $data = ObjectSerializer::deserialize(
                        $e->getResponseBody(),
                        '\Swagger\Client\Model\Pet', // <---
                        $e->getResponseHeaders()
                    );
                    $e->setResponseObject($data);
                    break;
                default:
                    $data = ObjectSerializer::deserialize(
                        $content,
                        '\Swagger\Client\Model\ErrorModel', // <---
                        $e->getResponseHeaders()
                    );
                    $e->setResponseObject($data);
                    break;
            }
```

So I think it is needed to put in each `case` section like below. 💡 

```diff
        } catch (ApiException $e) {
            switch ($e->getCode()) {
                case 200:
+                  $content = $e->getResponseBody();
+                  if ('\Swagger\Client\Model\Pet' !== 'string') {
+                      $content = json_decode($content);
+                  }
                    $data = ObjectSerializer::deserialize(
                        $content,
                        '\Swagger\Client\Model\Pet', // <---
                        $e->getResponseHeaders()
                    );
                    $e->setResponseObject($data);
                    break;
                default:
+                  $content = $e->getResponseBody();
+                  if ('\Swagger\Client\Model\ErrorModel' !== 'string') {
+                      $content = json_decode($content);
+                  }
                    $data = ObjectSerializer::deserialize(
                        $content,
                        '\Swagger\Client\Model\ErrorModel', // <---
                        $e->getResponseHeaders()
                    );
                    $e->setResponseObject($data);
                    break;
            }
```

ymilin: Oh, I see... Thank you very much for the review :-)

I'll be able to try and submit a new patch on Monday I think.

Also, my apologies I forgot to move the commits to a git branch before submitting, let me know if you want me to change that (would I need to submit a new PR?)

ackintosh: No need to submit a new PR. 👍 Please add a new patch to this PR.

modules/swagger-codegen/src/main/resources/php/api.mustache
```diff
@@ -181,8 +181,12 @@ use {{invokerPackage}}\ObjectSerializer;
         {{#responses}}
             {{#dataType}}
                 {{^isWildcard}}case {{code}}:{{/isWildcard}}{{#isWildcard}}default:{{/isWildcard}}
+                    $content = $e->getResponseBody();
+                    if ('{{dataType}}' !== 'string') {
```
wing328: @ymilin what about other primitive types such as integer, boolean, etc? Do we need to check those as well?

ymilin: @wing328 Thanks for the review 😄

I went ahead and took a better look at the [`ObjectSerializer.mustache::deserialize`](https://github.com/swagger-api/swagger-codegen/blob/432b3589861d4a35e2d4a7a1f6a55b889645886f/modules/swagger-codegen/src/main/resources/php/ObjectSerializer.mustache#L218
) function. This PR seems to effectively break other primitive type support (especially with booleans as they get decoded as (int) 1 and null for true and false respectively).

Now it feels to me that the json_decode part should be handled in deserialize as the function has knowledge of both the data and its type.

The same approach is already used with "success" responses and is quite concerning for the same reasons, here 

https://github.com/swagger-api/swagger-codegen/blob/432b3589861d4a35e2d4a7a1f6a55b889645886f/modules/swagger-codegen/src/main/resources/php/api.mustache#L159-167

I'll spend some time and test with all the types from the OpenAPI spec and validate all this.

vladimir-shebastyuk: @wing328 I was trying to resolve the same problem and found out that the approach that was used here is correct, because php `json_encode` function properly handle raw types, e.g.:
```php
echo "json_decode('123')\n";
var_dump(json_decode('123'));

echo "json_decode('123.45')\n";
var_dump(json_decode('123.45'));

echo "json_decode('true')\n";
var_dump(json_decode('true'));

echo "json_decode('false')\n";
var_dump(json_decode('false'));
```

will return:
```
json_decode('123')
int(123)
json_decode('123.45')
double(123.45)
json_decode('true')
bool(true)
json_decode('false')
bool(false)
```
So each primitive type handled by `json_decode` properly (I think the same idea was behind handling success response with the same check only for string primitive type). I also generate client and test it with real API and got correct answers.

I believe it's safe to apply theses changes:
```php
$content = $e->getResponseBody();
if ('{{dataType}}' !== 'string') {
	$content = json_decode($content);
}
$data = ObjectSerializer::deserialize(
	$content,
	'{{dataType}}',
	$e->getResponseHeaders()
);
```

Is it possible to accept this PR or it is necessary  to do some extra actions? E.g. update code or create a new PR?

ymilin: Hi, this PR was ported and merged to the [OpenAPITools/openapi-generator](https://github.com/OpenAPITools/openapi-generator) fork of this project.

Link to the PR over there : [https://github.com/OpenAPITools/openapi-generator/pull/757](https://github.com/OpenAPITools/openapi-generator/pull/757)

PR #6735, opened 2017-10-18
Initial step towards an AOT-compatible API client.

sebastianhaas:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming langauge.

### Description of the PR

See discussion at #6722. Basically this is about making the codegen output ready for use in an AOT build (e.g. `ngc --build --aot`).

This is _not_ really ready for merging yet, I would love to gather feedback from other people first, but I thought it might be easier to discuss over some actual code.

@taxpon @kenisteward @TiFu @Vrolijkx and obviously also @JohannesHoppe who finally brought this up.

#### What is missing
- [ ] We need to make sure that `ngc` is invoked during automated tests to prevent problems at an early stage
- [ ] We need to include `dist` into the bundled npm package (result of `npm pack`). However, there is no `.npmignore` as of today so I guess it will use `.gitignore`, where we want to exclude the `dist` directory.

PR #7965, opened 2018-04-04
[Typescript][Angular] Added in injected basePath observable to api service to allow it to respond to external updates to this setting post construction.

chrismellard:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

Fixes #7949 

Extended api service to allow optional injection of an observable basePath. This facilitates configuration of basePath post service construction which is required in circumstances where runtime setting of basePath cannot be achieved using APP_INITIALIZER tokens in angular. One such use case is when using libraries such as ngrx effects that do not respect such angular APP_INITIALIZER tokens.

Another option is to simply include the injected observable as part of the Configuration class rather than a separate dependency injectable token. If its preferable to move it inside that class I can do so.

PR #7984, opened 2018-04-06
[typescript-angular] AOT-compatible API client (via ng-packagr and ngc)

### LINKED ISSUE #6722: [typescript-angular] generated package/ApiModule is not AOT-compatible
##### Description

Angular-Typescript code does break the AOT compilation of the consuming app. AOT compilation is going to be the default in near future.
I tested the generated code together with angular-cli 1.4.4 and angular 4.4.5.

`ng serve` is now complaining the following:

```
ERROR in Error: Error encountered resolving symbol values statically.
Calling function 'ApiModule', function calls are not supported.
Consider replacing the function or lambda with a reference to an exported function....
```

##### Swagger-codegen version

v3.0.0-20171009.075709-6 ([snapshot](https://oss.sonatype.org/content/repositories/snapshots/io/swagger/swagger-codegen-cli/3.0.0-SNAPSHOT/swagger-codegen-cli-3.0.0-20171009.075709-6.jar)) but also applies to v2.3.0

So I created an angular-typescript client (3.0.0-branch) with the following config:
```json
{
  "npmName": "my-client-package",
  "npmVersion": "0.0.1",
  "snapshot": true,
  "ngVersion": "4.4.5"
}
```
```
java -jar swagger-codegen-cli.jar generate -i http://localhost:5000/swagger/v1/swagger.json -l typescript-angular -c config.json -o client-ts/
cd client-ts
npm pack
```

##### Steps to reproduce

Install the package via `npm install my-client-package` to a fresh Angular-CLI application.
Include the generated module:

```ts
export function getApiConfig() {
  return new Configuration({
    basePath:  environment.API_BASE_PATH ,
  });
}

@NgModule({
  
  imports: [
    ApiModule.forConfig(getApiConfig)
  ]
})
export class AppModule { }
```

switch from `tsc` to `ngc`
see this: https://github.com/swagger-api/swagger-codegen/blob/026426e50cf411631b1ed40fe0852a71f46c1d98/modules/swagger-codegen/src/main/resources/typescript-angular/package.mustache#L13

Long story short: `tsc` is not enough. Code must be generated via `ngc` so that the package has metadata. If there is no metadata, the NgModule can't be included, and the compiler rejects with a misleading message. In a perfect world the result conforms to the Angular Package Format v4.0, too.
See http://blog.mgechev.com/2017/01/21/distributing-an-angular-library-aot-ngc-types/ for an introduction to the topic. 

##### Current workaround

Don't use the generated @NgModule.  Copy its content into the application, e.g:

```ts
import { NgModule, ModuleWithProviders } from '@angular/core';
import { CommonModule } from '@angular/common';
import { HttpClientModule } from '@angular/common/http';
import { Configuration, MyService } from 'my-client-package';

/*
 * Workaround that replaces the ApiModule from  'my-client-package'
 * --> makes code AOT-compatible again
 *
 * Reason:
 * tsc does not generate *.metadata.json which breaks compatibility with angular’s AOT compiler
 *
 * avoids:
 * ERROR in Error: Error encountered resolving symbol values statically.
 * Calling function 'ApiModule', function calls are not supported.
 * Consider replacing the function or lambda with a reference to an exported function
 */
@NgModule({
  imports:      [ CommonModule, HttpClientModule ],
  declarations: [],
  exports:      [],
  providers:    [ MyService ]
})
export class ApiWorkaroundModule {
    public static forConfig(configurationFactory: () => Configuration): ModuleWithProviders {
        return {
            ngModule: ApiWorkaroundModule,
            providers: [ {provide: Configuration, useFactory: configurationFactory}]
        };
    }
}
```

Now you can use:

```ts
  imports: [
    ApiWorkaroundModule.forConfig(getApiConfig)
  ]
```

Code compiles again, since `ApiWorkaroundModule` is now a part of the main compilation.

JohannesHoppe:
### PR checklist

- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. 
- [x] Filed the PR against the correct branch.
- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language. @macjohnny

### Description of the PR

This PR is a follow-up of PR #6735. It fixes #6722.
This solution simplifies the compilation drastically, by leveraging [ng-packagr](https://github.com/dherges/ng-packagr).
ng-packagr creates a complete new [angular-package](https://docs.google.com/document/d/1tdgcvdLKsYPHlgNBppGFrsaA1eINLxJi9C8KkyrH2sI/edit#heading=h.k0mh3o8u5hx) into `dist` - which can be easily published via `npm publish dist` (see generated README.md)

Depending on the angular version three different styles are generated:
- it uses `ngc` when targeting Angular 2 (as seen in #6735)
- it uses ng-packagr 1 when targeting Angular 4
- it uses ng-packagr 2 when targeting Angular 5

To stress this a little bit, I tested a package with Angular 6.0.0-beta.5 and it works, too. 

FYI
@taxpon @kenisteward @TiFu @Vrolijkx @sebastianhaas @macjohnny

TODO
We might want to fiddle around with the different TypeScript versions, each Angular Compiler has own requirements to the allowed range of TypeScript version. Right now compilation gives us a warnings, but the generated bundles are still fine.


JohannesHoppe:
@kenisteward I agree. But as far as I know this only leets to a nasty warning. After everything is bundled, the final webpack build of the consuming project does not care about typescript at all. It's "just" a devDependency of the angular-lib.

kenisteward:
For using ngc with 2.7 ts and ng 4 ( I think) aot compilation doesn't work
at all from when I was doing it by hand. So we should be careful to make
sure all of our settings generate proper libs is all.

JohannesHoppe:
Honestly, a lot of black-magic is done here! When using ng-packagr 1 it somehow creates those bundles for angular 4. I confirm that everything is nicely compilable via `npm run build`. Don't ask my why it works! 😉

kenisteward:
Interesting!

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngularClientCodegen.java
```diff
@@ -112,25 +122,20 @@ public void processOpts() {
...
         additionalProperties.put(NG_VERSION, ngVersion);
         additionalProperties.put("injectionToken", ngVersion.atLeast("4.0.0") ? "InjectionToken" : "OpaqueToken");
         additionalProperties.put("injectionTokenTyped", ngVersion.atLeast("4.0.0"));
         additionalProperties.put("useHttpClient", ngVersion.atLeast("4.3.0"));
         if (!ngVersion.atLeast("4.3.0")) {
             supportingFiles.add(new SupportingFile("rxjs-operators.mustache", getIndexDirectory(), "rxjs-operators.ts"));
         }
+
+        // the current text makes only sence when using it as an npm/angular package
+        supportingFiles.add(new SupportingFile("README.mustache", getIndexDirectory(), "README.md"));
```
macjohnny: > the current text makes only sence when using it as an npm/angular package

yes, that's true, but I would still generate it also for non-npm packages, because this README should be extended to be a Readme on how to use the generated code, too. 
E.g. in https://github.com/swagger-api/swagger-codegen/pull/7965 information is added on how to set the base path with an observable.

So I would revert this change and generate the README in any case

modules/swagger-codegen/src/main/resources/typescript-angular/README.mustache
```diff
@@ -2,42 +2,41 @@
...
 
-_unPublished (not recommended):_
+_without publishing (not recommended):_
 
 ```
-npm install PATH_TO_GENERATED_PACKAGE --save
+npm install PATH_TO_GENERATED_PACKAGE/dist --save
 ```
 
 _using `npm link`:_
```
macjohnny: add a note about an issue in angular CLI that prevents using linked libraries on Windows, see https://github.com/angular/angular-cli/issues/8284

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/TypeScriptAngularClientCodegen.java
JohannesHoppe: I agree. I will change that.

macjohnny:
@JohannesHoppe your article is excellent! you should add it as a Readme somewhere. @wing386 is there a place to publish language-specific help? wiki?

modules/swagger-codegen/src/main/resources/typescript-angular/README.mustache
JohannesHoppe: ✅

macjohnny:
@JohannesHoppe just wanted to point out that the newest angular cli version supports generating libraries: https://github.com/angular/angular-cli/releases/tag/v6.0.0-rc.2

maybe this could be useful here?

JohannesHoppe:
Yes, and no. 😄 

Yes: 
* This would be another option to generate a angular-library in the "angular package format"
* most people are similar with the cli

No:
* we would need to follow the angular-cli file structure, which means even more boilerplate files
* I'm not sure if this will work in older Angular versions (Angular 2)
* ng-packagr  == rollup, angular-cli == webpack. it is said that rollup is much more faster

Generally, we have a lot of different petstore samples, which shows the amount of complexity. I would vote to skip Angular 2 and Angular 4.0-4.2 and to skip the "no npm package" part sooner or later. (But not in  this PR!)

macjohnny:
@JohannesHoppe ok I agree, `ng-packagr` can be used with only one configuration file and without any additional changes, which is nice.

I would also agree to skip angular  `< 4.3`

macjohnny:
@JohannesHoppe thanks for the changes. 
I think you need to re-generate the samples, as you changed the `Readme.md`
But apart from that, the PR seems fine from my point of view.

JohannesHoppe:
I was hoping that nobody will notice that... 😎 I  will do it after work...

JohannesHoppe:
FYI: 

> `ng g library <name>`
> It generates a new folder inside an Angular CLI project, ready to be built with `ng-packagr`. 

https://twitter.com/juristr/status/984501406660022272

kenisteward:
Doesn't that requires us to refactor to make sure we fully confirm to the
generated package? Or do we already confirm?

I love me some angular cli haha. It makes our ut functional easier to
generate as well.

JohannesHoppe:
This means that the wo won't go for angular-cli as @macjohnny suggested. Because we would have this pipeline

1. you pull sagger-codegen and compile with `mnv clean package`
1. swagger-codegen generates an angular-cli project
2. `ng g library` generates ng-packagr project
3. ng-packagr project compiles NPM package in "Angular package format" 

With my PR we have:

1. you pull sagger-codegen and compile with `mnv clean package`
2. swagger-codegen generates ng-packagr project
3. ng-packagr project compiles NPM package in "Angular package format" 

... people will think we are completely crazy doing having 4 steps! 😆

kenisteward:
Hahaha

wing328:
If no further feedback/question on this PR, I'll merge it into master on the coming weekend.

damianmigo:
Did someone have to copy the metadata.json to the dist folder manually before publishing it?

I can’t build the project that depends on my client library with AOT because the decorators are missing.

hypery2k:
I've tried to use it with Angular 8 and AoT enabled. For this some change are need see [here](https://github.com/swagger-api/swagger-codegen-generators/pull/473) 

Any feedback is welcome

PR #7752, opened 2018-03-02
GO Plain string body type can not be set in generated API

Vad1mo:
fixes #7751

Can you review
@antihax  @bvwells

### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

(details of the change, additional tests that have been done, reference to the issue for tracking, etc)

PR #7987, opened 2018-04-08
Golang Client Refactor (body and model in errors, typed optional parameters)

antihax:
### PR checklist

- [X Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
@marians @bvwells @wing328 @arekkas @Vad1mo @jmini @mr1sunshine

### Description of the PR

This PR is for Swagger 2.0 and has a few changes; hopefully a step in the right direction. Please review and let me know if modifications need to be made. I have this running in production currently with a fairly large API.

- This PR includes fixes for #7752 from @Vad1mo with updated tests.
- Typed optional parameters, primitives checked at compile time.
- Assertable errors with response []byte body and models.

### Changes

Refactor to change a few features to be a bit more standard:

**Errors can be asserted to reveal model and body information:**
The current type is called GenericSwaggerError; but I request suggestions for names. It needs to be something which unlikely to collide with generated models.
```
	resp, r, err := client.PetApi.GetPetById(context.Background(), 999999999)
	if err != nil {
		e, ok := err.(sw.GenericSwaggerError)
		// e.Body() has a []byte of the response body
		// e.Model() is an interface{} to the custom response model,
		// or nil if one does not exist
	}
```

**Typed optional parameters:**
Based on some feedback; copied a similar approach to the AWS Go 2.0 API using optional structs that feature an `IsSet bool`. If an option is not needed, it is not specified. This allows us to check type at compile time again vs. runtime for primitive data types. 

At present, models will still use interface{}. This could be expanded upon later, but collisions need to be avoided somehow.

Optionals currently import from `"github.com/antihax/optional"`. Perhaps they should import from `"github.com/swagger-api/optional"`? Or some other neutral host? I could not find a simple solution to avoid collisions and properly import them using Go. Welcome to suggestions here also.

```
	r, err := client.PetApi.UpdatePetWithForm(context.Background(), 12830,
		&sw.UpdatePetWithFormOpts{
			Name:   optional.NewString("golang"),
			Status: optional.NewString("available"),
		})
```


arekkas:
Hey, really super mega awesome that you tackled this! Thank you so much! I haven't had an in-depth look yet but I will do so over the next days (it's a huge diff). One thing I saw is that we *could* use [pkg/errors](https://github.com/pkg/errors/) for returning errors? It would allow for stack traces but it would also imply that we have to do `errors.Cause(err).(*MyType)` for type assertions.

arekkas:
Which files do I have to check to see an exemplary generated code file as well as its usage?

antihax:
These two are the bulk:
https://github.com/antihax/swagger-codegen/blob/05717a2d2eb4ca7f59d5251e927269cc0ca4b90e/samples/client/petstore/go/go-petstore/client.go
https://github.com/antihax/swagger-codegen/blob/05717a2d2eb4ca7f59d5251e927269cc0ca4b90e/samples/client/petstore/go/go-petstore/api_pet.go

This shows use:
https://github.com/antihax/swagger-codegen/blob/05717a2d2eb4ca7f59d5251e927269cc0ca4b90e/samples/client/petstore/go/pet_api_test.go

I could look into `pkg/errors` but trying to keep dependencies minimal. Users can tweak the template for such things.

My main holdup currently is importing the `optional` package from a third party repo, but since i cannot determine how to predict the import path, i am at a loss.

arekkas:
> I could look into pkg/errors but trying to keep dependencies minimal. Users can tweak the template for such things.

I agree

> My main holdup currently is importing the optional package from a third party repo, but since i cannot determine how to predict the import path, i am at a loss.

We could include those in the a template?

antihax:
We could include them in the base package, but I worry about collisions.
If we make them a subpackage inside the main, I cannot predict the final import path the user will use.

samples/client/petstore/go/pet_api_test.go
```diff
@@ -58,7 +59,12 @@ func TestGetPetById(t *testing.T) {
 func TestGetPetByIdWithInvalidID(t *testing.T) {
 	resp, r, err := client.PetApi.GetPetById(context.Background(), 999999999)
```
arekkas: Are path parameters always their own argument? Can path parameters be optional? If so, how do we deal with this in the API?

arekkas:
> We could include them in the base package, but I worry about collisions.

Type name collisions or what exactly?

antihax:
We have a `NewString` function in optional. If an API also has a model `NewString` it will fail to compile.

samples/client/petstore/go/pet_api_test.go
antihax: Currently all required parameters are a separate argument and all optional fields are in the optional struct, it does not differentiate path, body, qs, etc.

antihax:
It would still risk collision with a user API. I will research a bit and see if there is a better way to split packages without knowing the import path. Vendoring may be an option.

samples/client/petstore/go/pet_api_test.go
arekkas: So what happens if qs and body have the same parameters? Is the body seen as a type always or will it get split if we have multiple arguments? I'm a bit confused :D

antihax: Same parameter will be put in both places. Interesting point, wonder what the other languages do.

arekkas: Have you seen my idea to have an embedded struct?

```go
type SomeRequestBody struct {
  Param1 NilString `json:"param1,omitempty"`
  Param2 string `json:"param2"`
  // ...
}
type SomeRequestPath struct {
  Param3 NilString `json:"param3,omitempty"`
  Param4 string `json:"param4"`
  // ...
}
type SomeRequestQuery struct {
  Param5 NilString `json:"param5,omitempty"`
  Param6 string `json:"param6"`
  // ...
}

type SomeRequest struct {
	*SomeRequestQuery
	*SomeRequestPath
	*SomeRequestBody 
}
```

That way it's possible to do:

```go
r := NewSomeRequest()
r.Param1 = ...
r.Param6 = ...
```

If there is a name collision, Go will print a compile error and you have to do:


```go
r := NewSomeRequest()
r.AType.Param1 = ...
r.BType.Param1 = ...
```

antihax: This is possible, perhaps for the 3.0.0 version we can go this way.
Looking at the other 2.0.0 clients: PHP, Python, Perl, and Java take arguments the same way and do not differentiate their locations.

arekkas: Sounds good

arekkas:
> Isn't it feasible to use *string to denote an optional string, for example? This would avoid the need of custom types.

Not really, types give various advantages (like method receivers). In this case it's absolutely justified.

arekkas:
From what I can tell this looks pretty solid so far!

arekkas:
So is this change actually backwards compatible to 2.3.x? If not, we should probably label it as such?

wing328:
@arekkas Done.

Issue #4973, opened 2017-03-08
[JavaScript (with promises)] Bug following $ref in object properties


advance512:
##### Description

Generated type file `ID` contains the following:

```
  /**
   * Constructs a <code>ID</code> from a plain JavaScript object, optionally creating a new instance.
   * Copies all relevant properties from <code>data</code> to <code>obj</code> if supplied or a new instance if not.
   * @param {Object} data The plain JavaScript object bearing properties of interest.
   * @param {module:model/ID} obj Optional instance to populate.
   * @return {module:model/ID} The populated <code>ID</code> instance.
   */
  exports.constructFromObject = function(data, obj) {
    if (data) {
      obj = obj || new exports();

    }
    return obj;
  }
```

`data` is an integer input. `obj` is null. Hence, the result is `exports()` instead of simply `data`.

##### Swagger-codegen version

Build package: io.swagger.codegen.languages.JavascriptClientCodegen
v2.2.2

##### Swagger declaration file content or url

```
definitions:

  # ===============================================================================
  # Fragments
  # ===============================================================================

  ID:
    description: An entity identifer
    type: integer
    format: int64
    readOnly: true

  # ===========================================================================
  # Users
  # ===========================================================================

  User:
    type: object
    required:
      - emailAddress
    properties:
      id:
        $ref: '#/definitions/ID'
      emailAddress:
        type: string
        format: email
        minLength: 6
        maxLength: 254
```
##### Command line used for generation

java -jar swagger-codegen-cli.jar generate -i ../source.yaml -l javascript --additional-properties usePromises=true -o ./javascript/

##### Steps to reproduce

Use the above definitions in any operation. You'll get a result object with an exports() objects instead of an ID integer.

##### Suggest a Fix

It tempts me to say:

```
  /**
   * Constructs a <code>ID</code> from a plain JavaScript object, optionally creating a new instance.
   * Copies all relevant properties from <code>data</code> to <code>obj</code> if supplied or a new instance if not.
   * @param {Object} data The plain JavaScript object bearing properties of interest.
   * @param {module:model/ID} obj Optional instance to populate.
   * @return {module:model/ID} The populated <code>ID</code> instance.
   */
  exports.constructFromObject = function(data, obj) {
    if (data) {
      obj = **data** || new exports();

    }
    return obj;
  }
```

but I actually think the issue is that swagger-codegen doesn't follow the $ref properly. Not certain

wing328:
[...]

PR #7991, opened 2018-04-09
Added vendorExtensions.x-isPrimitive.

satoshikumano:
Switch template for constructFromObject.

### PR checklist

### Description of the PR

When `$ref` is used to reference primitive types, The value is always empty object.
This forces users to modify the Swagger file even if the API definition is valid.

Reported in #4973

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
@@ -797,7 +797,7 @@ public class JavascriptClientCodegen extends DefaultCodegen implements CodegenCo
                 codegenModel.vendorExtensions.put("x-itemType", getSwaggerType(mm.getAdditionalProperties()));
             } else {
                 String type = mm.getType();
-                if (type != null && (type.equals("number") || type.equals("integer") || type.equals("string") || type.equals("boolean") || type.equals("null"))) {
+                if (isPrimitiveType(type)){
                     codegenModel.vendorExtensions.put("x-isPrimitive", true);
                 }
             }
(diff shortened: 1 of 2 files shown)


satoshikumano:
Could you review this PR?
@CodeNinjai @frol @cliffano

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java
```diff
@@ -795,6 +795,11 @@ public CodegenModel fromModel(String name, Model model, Map<String, Model> allDe
             if (mm.getAdditionalProperties() != null) {
                 codegenModel.vendorExtensions.put("x-isMap", true);
                 codegenModel.vendorExtensions.put("x-itemType", getSwaggerType(mm.getAdditionalProperties()));
+            } else {
+                String type = mm.getType();
+                if (type != null && (type.equals("number") || type.equals("integer") || type.equals("string") || type.equals("boolean") || type.equals("null"))) {
```
cliffano: To reduce cyclomatic complexity, could this line perhaps be replace with something like:
`String[] PRIMITIVE_TYPES = new String[] {"number", "integer","string","boolean", "null"};`
`Arrays.asList(PRIMITIVE_TYPES).contains(type)`

satoshikumano: Fixed. It also improves readability.

satoshikumano:
@cliffano Hi. I have fixed review indication. Just would like to send a reminder in case you haven't noticed yet.

cliffano:
@satoshikumano Thanks for that.

Overall I don't have any objection, but I was just wondering if it's worth having the primitive types array consistent to https://github.com/swagger-api/swagger-codegen/blob/cf4df630b70cadc1ee35deb87b2b895e9d492473/samples/server/petstore/spring-mvc-j8-localdatetime/src/main/java/io/swagger/configuration/SwaggerUiConfiguration.java#L30 , i.e. scoped the same way, and declared at the top.

What do you think @wing328 ? Should this be made consistent?

wing328:
@satoshikumano thanks for the PR, which has been merged into master.

@cliffano we don't enforce a standard on this and I agree declaring the array on the top seems more consistent with the overall code.

PR #7996, opened 2018-04-10
Fix Issue 7924 - Allow host and base path to be changed independently

### LINKED ISSUE #7924: [JAVA] Unable to modify the host without overwriting the whole base path
##### Description

The Java client base path includes the hostname (defaulting to `localhost`), as well as the rest of the base path defined in the yaml.  The only method to set the base path overwrites everything, so I can't set the host without overwriting the rest of the basePath which I want to keep.

##### Swagger-codegen version

2.4.0

```
basePath: /v2
```

##### Steps to reproduce

Online editor, Java client: https://editor.swagger.io/

Produces basePath of `http://petstore.swagger.io/v2` with no method available to change the `petstore.swagger.io` bit without also losing the `v2`

alango:
### PR checklist

@bbdouglas  @JFCote  @sreeshas @jfiala  @lukoyanov @cbornet @jeff9finger

### Description of the PR

Fixes #7924 
Separate the hostname and the base path so that they can be changed independently of one another

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache
@@ -53,7 +53,8 @@ import {{invokerPackage}}.auth.OAuth;
 
 public class ApiClient {
 
-    private String basePath = "{{{basePath}}}";
+    private String host = "{{host}}";
+    private String basePath = "{{{basePathWithoutHost}}}";
     private boolean debugging = false;
     private Map<String, String> defaultHeaderMap = new HashMap<String, String>();
     private String tempFolderPath = null;
@@ -102,9 +103,29 @@ public class ApiClient {
     }
 
     /**
+     * Get host
+     *
+     * @return host
+     */
+     public String getHost() {
+        return host;
+      }
+
+    /**
+     * Set host
+     *
+     * @param host (e.g {{{host}}})
+     * @return An instance of OkHttpClient
+     */
+    public ApiClient setHost(String host) {
+        this.host = host;
+        return this;
+      }
+
+    /**
      * Get base path
      *
-     * @return Baes path
+     * @return Base path
      */
     public String getBasePath() {
         return basePath;
@@ -113,7 +134,7 @@ public class ApiClient {
     /**
      * Set base path
      *
-     * @param basePath Base path of the URL (e.g {{{basePath}}}
+     * @param basePath Base path of the URL (e.g {{{basePathWithoutHost}}})
      * @return An instance of OkHttpClient
      */
     public ApiClient setBasePath(String basePath) {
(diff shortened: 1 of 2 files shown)


jeff9finger:
I like the concept. However, I think this would be more useful to provide more granularity here. Combining the schema, host and port into a property named "host" is a bit misleading. I'd like to see separate properties for schema, host, and port (or at least schema and host with the port combined with the host). An alternative would be to rename the property to something besides "host".

What does the community think?

alango:
Was erring on the side of making small changes, but agree it would make sense to split out the scheme as well and am happy to do that 🙂. (Alternatively, if there's already some precedent for this in swagger-codegen, then I'm also happy to follow that.)

alango:
Separated out the scheme, let me know if there's anything else you want from me.

modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache
```diff
@@ -53,7 +53,9 @@ import {{invokerPackage}}.auth.OAuth;
 
 public class ApiClient {
 
-    private String basePath = "{{{basePath}}}";
+    private String scheme = "{{scheme}}";
```
jeff9finger: Shouldn't this provide a default for "http" in a similar way that host does with "localhost"

jeff9finger:
Also, this seems useful for all the java client libraries. Would you consider adding it to the other Java client libraries?

modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache
alango: If I understand how this works correctly, then I believe this defaults to `https` if a scheme isn't specified (https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java#L98).

alango:
My project happens to use this library so that's what I've been testing with. I'd rather not edit the other Java client libraries since there are at least some where it looks my change won't easily apply, and I don't want to touch more code that I don't have time to test right now!

modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache
jeff9finger: Yes. that is correct. Thanks for pointing that out.

jeff9finger:
@alango I understand your time constraints. Thanks for the contribution.

@wing328 Can we put this request out to the community and see if someone else has time to add this to the other Java client libraries?

alango:
@jeff9finger Do you know if this is going to be merged soon, or are you planning to wait for the other libraries to be updated?

jeff9finger:
@wing328  Can we go ahead and get this merged?

alango:
@wing328 bump?

SpearThruster:
Any updates?

Issue #7876, opened 2018-03-20
[Java][okhttp-gson] Generates APIs do not clean com.squareup.okhttp.OkHttpClient#networkInterceptors


maxusoltsev:
##### Description
OKHttp documentation states what OkHttpClient instance should be shared and reused  
(https://square.github.io/okhttp/2.x/okhttp/com/squareup/okhttp/OkHttpClient.html)

But if we reuse swagger-codegen generated API (or APIClient), every async call with callback will add
new newworkInterceptor 
(look for apiClient.getHttpClient().networkInterceptors().add(new com.squareup.okhttp.Interceptor()  in generated API code)

This behavior does not allow to reuse APIClient or *API instance. 
##### Swagger-codegen version

2.3.1
##### Swagger declaration file content or url
https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/examples/v2.0/yaml/petstore-minimal.yaml

##### Command line used for generation
Online client generator at https://editor.swagger.io/ , java client

##### Steps to reproduce

1. get https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/examples/v2.0/yaml/petstore-minimal.yaml yaml file
2. Go to https://editor.swagger.io/ and generate java client for that yaml
3. Look into generated zip 
package io.swagger.client.api;
..
 public com.squareup.okhttp.Call petsGetCall(
...
  if(progressListener != null) {
            apiClient.getHttpClient().networkInterceptors().add(new com.squareup.okhttp.Interceptor() {
                @Override
                public com.squareup.okhttp.Response intercept(com.squareup.okhttp.Interceptor.Chain chain) throws IOException {
                    com.squareup.okhttp.Response originalResponse = chain.proceed(chain.request());
                    return originalResponse.newBuilder()
                    .body(new ProgressResponseBody(originalResponse.body(), progressListener))
                    .build();
                }
            });
        }
....

##### Suggest a fix/enhancement
Looks like we should change the way how APICallbacks are invoked (do not use apiClient.getHttpClient().networkInterceptors())

ngaya-ll:
This eventually causes a `StackOverflowError` as each interceptor in the chain gets called recursively.

Fix is probably to use a single interceptor that fans out to a list of progress listeners, rather than one interceptor per listener.

meilleureVie:
`if(progressListener != null) {
      this.apiClient.getHttpClient().networkInterceptors().add(new Interceptor() {
        ...
      });
[...]

PR #8053, opened 2018-04-21
[Java][okhttp-gson] Fix incorrect use of OkHttp interceptors 

ngaya-ll:
### PR checklist

- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language. @bbdouglas @JFCote @sreeshas @jfiala @lukoyanov @cbornet  @jeff9finger

### Description of the PR

Currently, the generated Java okhttp-gson client adds an interceptor to the underlying `OkHttpClient` for each async call.  The purpose of the interceptor is to wrap the response body to track download progress.  This implementation doesn't work correctly, for multiple reasons:

* The interceptor intercepts all requests to the client, not just the one it's trying to track.
* Interceptors are never removed from the client, so each async request adds another layer of interception for all subsequent requests.  Since the interceptor chain is invoked recursively, at some point all requests will start failing with a `StackOverflowError`.

With this code change, the client uses a single interceptor to decorate all async requests and send updates to the relevant listener only.

I also added a test case to demonstrate the issue, which fails on the current master and passes on this branch.

Fixes https://github.com/swagger-api/swagger-codegen/issues/7876


samples/client/petstore/java/okhttp-gson/src/test/java/io/swagger/client/api/PetApiTest.java
```diff
@@ -95,103 +92,104 @@ public void testCreateAndGetPetWithHttpInfo() throws Exception {
...
+        assertEquals("Not Found", exception.getMessage());
+        assertEquals("application/json", exception.getResponseHeaders().get("Content-Type").get(0));
+    }
 
-            @Override
-            public void onSuccess(Pet pet, int statusCode, Map<String, List<String>> responseHeaders) {
-                result.put("pet", pet);
-            }
+    @Test
+    public void testCreateAndGetMultiplePetsAsync() throws Exception {
```
ngaya-ll: This is the new test mentioned in the description.

samples/client/petstore/java/okhttp-gson/src/test/java/io/swagger/client/api/PetApiTest.java
```diff
@@ -269,6 +264,7 @@ public void testFindPetsByStatus() throws Exception {
     }
 
     @Test
+    @Ignore
```
ngaya-ll: This endpoint is deprecated on petstore.swagger.io and was returns a 500 status.

modules/swagger-codegen/src/main/resources/Java/libraries/okhttp-gson/ApiClient.mustache
```diff
@@ -138,6 +139,7 @@ public class ApiClient {
      */
     public ApiClient setHttpClient(OkHttpClient httpClient) {
         this.httpClient = httpClient;
+        addProgressInterceptor();
```
ngaya-ll: Note that if multiple `ApiClient`s are created with the same underlying `OkHttpClient` this will create duplicate progress updates for each request.  However, from reading the rest of the class (e.g. the `setDebugging()` method) I concluded that this is not a supported use case.

mattnworb:
It would be lovely to have this merged to fix #7876 as the current state renders it impossible to use any of the generated "async" methods in any sort of long-lived production server.

mtnourji:
> It would be lovely to have this merged to fix #7876 as the current state renders it impossible to use any of the generated "async" methods in any sort of long-lived production server.

+1
Any update on this ?

PR #8061, opened 2018-04-23
[python] Avoid creating unused ThreadPools

minrk:
Instead, create ApiClient.pool on first request for .pool property.

This avoids spawning n-cpus threads (the default for ThreadPool) at instantiation of every ApiClient, which can be very costly in, for example, container scenarios on nodes with many CPUs where multiple API Clients might be instantiated in one process.

The thread count is made configurable and defaults to 1 instead of N-CPUs, which I think makes more sense for the default behavior of a non-blocking API.


minrk:
cc @taxpon @frol @mbohlool @cbornet @kenjones-cisco for Python review

minrk:
Sorry, failed to commit it before. Looks like this had been missed a few times, since there are a few more changes in the results than mine.

minrk:
Had to fix a deserialization test that hadn't been updated since some changes in enums, I think. The failing test passed locally for me, hopefully Travis will find the same.

kenjones-cisco:
LGTM

minrk:
Test failure appears to have been an intermittent network issue. Anything I can to do help this progress?

minrk:
Resolved conflicts.

007:
This blocks on exit the majority of times when more than one API client is created (https://github.com/kubernetes-client/python/issues/411). What needs to be done to get this merged and released so the k8s Python API can be regenerated and fixed?

minrk:
Rebased and rerendered again

007:
Happy anniversary #8061 :tada:

noamgat:
Was this ever merged?

PR #8085, opened 2018-04-27
Camelize Kotlin model names and filenames

PatrickNLT:
Previously the case of the names could be inconsistent if there were nested object properties. Eg. 'MyNewPet_barking_type' now becomes 'MyNewPetBarkingType'.
Ideally, we should use nested classes to take advantage of the namespace of the parent class, but this is much more work.

### PR checklist

### Description of the PR

Previously the case of the names could be inconsistent if there were nested object properties. Eg. 'MyNewPet_barking_type' now becomes 'MyNewPetBarkingType'.

Ideally, we should use nested classes to take advantage of the namespace of the parent class, but this is much more work.

Technical committee member: @jimschubert

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractKotlinCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractKotlinCodegen.java
@@ -453,11 +453,20 @@ public abstract class AbstractKotlinCodegen extends DefaultCodegen implements Co
         String modifiedName = name.replaceAll("\\.", "");
         modifiedName = sanitizeKotlinSpecificNames(modifiedName);
 
+        // Camelize name of nested properties
+        modifiedName = camelize(modifiedName);
+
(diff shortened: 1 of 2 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractKotlinCodegen.java
```diff
@@ -453,13 +453,22 @@ public String toModelName(final String name) {
         String modifiedName = name.replaceAll("\\.", "");
         modifiedName = sanitizeKotlinSpecificNames(modifiedName);
 
+        // Camelize name of nested properties
+        modifiedName = camelize(modifiedName);
```
jimschubert: I think this is fine because it follows [kotlin naming conventions](https://kotlinlang.org/docs/reference/coding-conventions.html#naming-rules). However, this will potentially break for anyone using only the models of kotlin-client for non-JSON serialization that relies on model names for serialization, which I believe is what I was doing when I original wrote the kotlin client.

Issue #8090, opened 2018-04-27
[Python] Enhancement- Support http signing


twinforces:
##### Description

There's authorization based on headers or query params, but not based on: https://datatracker.ietf.org/doc/draft-cavage-http-signatures

Enhancement would be to support this as a code generation option, -Dhttpsig, with the auth_settings, then set to auth_setting['in'] == 'httpsig' to trigger it. 

##### Swagger-codegen version

This is an enhancement, filing issue prior to filing PR. 

##### Swagger declaration file content or url

N/A 

##### Command line used for generation

 ./bin/python-httpsig-petstore.sh  in associated PR. 

##### Steps to reproduce

N/A

##### Related issues/PRs

Will supply PR in moments. 

##### Suggest a fix/enhancement

Will supply PR in moments.

PR #8091, opened 2018-04-27
Httpsig

twinforces:
### PR checklist

- [*] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [*] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [*] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [*] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

See Issue #8090 for more details, or you can read the [RFC this is implementing](https://datatracker.ietf.org/doc/draft-cavage-http-signatures).

### DIFF
--- a/samples/client/petstore/java/feign/src/main/java/io/swagger/client/api/FakeApi.java
+++ b/samples/client/petstore/java/feign/src/main/java/io/swagger/client/api/FakeApi.java
@@ -8,6 +8,7 @@ import io.swagger.client.model.Client;
 import org.threeten.bp.LocalDate;
 import org.threeten.bp.OffsetDateTime;
 import io.swagger.client.model.OuterComposite;
+import io.swagger.client.model.User;
 
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -70,6 +71,52 @@ public interface FakeApi extends ApiClient.Api {
     "Accept: application/json",
   })
   String fakeOuterStringSerialize(String body);
+
+  /**
+   * 
+   * 
+    * @param body  (required)
+    * @param query  (required)
+   */
+  @RequestLine("PUT /fake/body-with-query-params?query={query}")
+  @Headers({
+    "Content-Type: application/json",
+    "Accept: application/json",
+  })
+  void testBodyWithQueryParams(User body, @Param("query") String query);
+
+  /**
+   * 
+   * 
+   * Note, this is equivalent to the other <code>testBodyWithQueryParams</code> method,
+   * but with the query parameters collected into a single Map parameter. This
+   * is convenient for services with optional query parameters, especially when
+   * used with the {@link TestBodyWithQueryParamsQueryParams} class that allows for
+   * building up this map in a fluent style.
+   * @param body  (required)
+   * @param queryParams Map of query parameters as name-value pairs
+   *   <p>The following elements may be specified in the query map:</p>
+   *   <ul>
+   *   <li>query -  (required)</li>
+   *   </ul>
+   */
+  @RequestLine("PUT /fake/body-with-query-params?query={query}")
+  @Headers({
+  "Content-Type: application/json",
+  "Accept: application/json",
+  })
+  void testBodyWithQueryParams(User body, @QueryMap(encoded=true) Map<String, Object> queryParams);
+
+  /**
+   * A convenience class for generating query parameters for the
+   * <code>testBodyWithQueryParams</code> method in a fluent style.
+   */
+  public static class TestBodyWithQueryParamsQueryParams extends HashMap<String, Object> {
+    public TestBodyWithQueryParamsQueryParams query(final String value) {
+      put("query", EncodingUtils.encode(value));
+      return this;
+    }
+  }
 
   /**
    * To test \&quot;client\&quot; model
--- a/samples/client/petstore/java/google-api-client/docs/FakeApi.md
+++ b/samples/client/petstore/java/google-api-client/docs/FakeApi.md
@@ -8,6 +8,7 @@ Method | HTTP request | Description
 [**fakeOuterCompositeSerialize**](FakeApi.md#fakeOuterCompositeSerialize) | **POST** /fake/outer/composite | 
 [**fakeOuterNumberSerialize**](FakeApi.md#fakeOuterNumberSerialize) | **POST** /fake/outer/number | 
 [**fakeOuterStringSerialize**](FakeApi.md#fakeOuterStringSerialize) | **POST** /fake/outer/string | 
+[**testBodyWithQueryParams**](FakeApi.md#testBodyWithQueryParams) | **PUT** /fake/body-with-query-params | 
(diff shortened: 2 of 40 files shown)


modules/swagger-codegen/src/main/resources/python/api_client.mustache
```diff
@@ -625,3 +640,124 @@ class ApiClient(object):
...
+        return digest
+    
+    def get_rsasig_b64encode(self, private_key_path, digest):
+        """
+        :param private_key_path : abs path to private key .pem file.
+        :param digest: digest
+        :return: instance of digest object 
+        """
+        
+        key = open(private_key_path, "r").read()
```
frol: Opening a key file on every request is a hit for performance.

Also there is no error handling at all. What if file reading fails?

modules/swagger-codegen/src/main/resources/python/api_client.mustache
```diff
@@ -132,14 +142,15 @@ class ApiClient(object):
...
-        self.update_params_for_auth(header_params, query_params, auth_settings)
-
+ 
         # body
         if body:
             body = self.sanitize_for_serialization(body)
 
-        # request url
+        # auth setting
+        self.update_params_for_auth(header_params, query_params, auth_settings, resource_path, body)
```
frol: I don't see why this line should be moved here.

modules/swagger-codegen/src/main/resources/python/api_client.mustache
```diff
@@ -625,3 +640,124 @@ class ApiClient(object):
...
+    
+        auth_str = auth_str + " " + "keyId=\"" + auth_settings['api_key_id'] + "\"," + "algorithm=\"" + auth_settings['digest_algorithm'] + "\"," + "headers=\"(request-target)"
+    
+        for key, _ in hdrs.items():
+            auth_str = auth_str + " " + key.lower()
+        auth_str = auth_str + "\""
+    
+        auth_str = auth_str + "," + "signature=\"" + signed_msg.decode('ascii') + "\""
+    
+        return auth_str
```
frol: See my comment to `prepare_str_to_sign`

modules/swagger-codegen/src/main/resources/python/api_client.mustache
```diff
@@ -625,3 +640,124 @@ class ApiClient(object):
...
+        length = len(hdrs.items())
+    
+        i = 0
+        for key, value in hdrs.items():
+            ss = ss + key.lower() + ": " + value
+            if i < length-1:
+                ss = ss + "\n"
+            i += 1
+        
+        return ss
```
frol: This whole function is unpythonic. Use string formatting and '\n'.join(...). String concatenation is both ugly and inefficient in Python.

modules/swagger-codegen/src/main/resources/python/api_client.mustache
```diff
@@ -625,3 +640,124 @@ class ApiClient(object):
...
+        :return: instance of digest object 
+        """
+        if body is None:
+            body = ''
+        else:
+            body = json.dumps(body)
+        
+        
+        target_host = urlparse(self.host).netloc
+        target_path = urlparse(self.host).path
```
frol: There is no reason to parse host twice and I don't expect host to contain netloc and path at all.

modules/swagger-codegen/src/main/resources/python/api_client.mustache
```diff
@@ -625,3 +640,124 @@ class ApiClient(object):
...
+        
+        if auth_settings['digest_algorithm'] != 'rsa-sha256':
+                raise ValueError('only rsa-sha256 supported for digest_algorithm in httpsig')
+        digest = self.get_sha256_digest(string_to_sign.encode())
+        
+        b64_signed_msg = self.get_rsasig_b64encode(auth_settings['private_key_file'], digest)
+        
+        auth_header = self.get_auth_header(auth_settings, headers, b64_signed_msg)
+        
+        self.set_default_header('Date', '{0}'.format(cdate))
```
frol: If you only need to make a string from it, just use `str` instead of `'{0}'.format`

modules/swagger-codegen/src/main/resources/python/api_client.mustache
twinforces: Constantly reading the file also bothered me too. I was thinking about how to handle this and the problem that auth_settings is coded to be a static, hard-coded set of values driven by the swagger spec; that's not typically how people setup security. So I'm going to do a bunch of changes to the code now that I've dug into a bit, so it will read the file once and cache it in a mutable auth-settings. That will also provide people the ability to store the cert data instead of having to have it at rest on disk. 


 (Wasn't my code, just got drafted to contribute back to the ecosystem, because I wanted to contribute something else). Feel free to give as much feedback as you want.

modules/swagger-codegen/src/main/resources/python/api_client.mustache
twinforces: Have to compute signature after all the data is there.

modules/swagger-codegen/src/main/resources/python/api_client.mustache
twinforces: Ok. Will change.

twinforces:
On the comment about the RFC being a draft, I feel like most RFCs live for years as a draft.

twinforces:
@frol thanks for suggestions, they should all be implemented now. 

@kenjones-cisco Would you like me to email the results from running the sample scripts? I get differences in all the samples for all languages, its not clear which should be checked in as checking in the generated code seems like an anti pattern. Or I can add a giant commit with all the updates.

kenjones-cisco:
It is just updating the generate code for the specific language only, ie python. Depending on the change it can be large.

twinforces:
Yes I’m just updating python but git reports changes in all languages. I can commit just the python diffs or all or none. Please advise.

twinforces:
Ok, see [commit](https://github.com/swagger-api/swagger-codegen/pull/8091/commits/c3bed66e050d7580e7b39ff6263e8b75452a10ef).

twinforces:
Also, my DE pointed out that AWS has a similar signing mechanism: https://docs.aws.amazon.com/general/latest/gr//signing_aws_api_requests.html so its possible someone other than myself could use this work as a basis of using Swagger to generate AWS APIs...

twinforces:
@ackintosh The change you made to EnumTest seems to be causing a unit test failure in Python, I was wondering if you could advise me?

PR #8105, opened 2018-04-30
[Clojure] Add util method to set the api-context globally

f-f:
### PR checklist

### Description of the PR

Before the change, to update the api-context (e.g. the URL of the API) it was necessary to wrap all the calls in `(with-api-context new-context api-call-here)`.
Since in some cases the new context needs to be shared globally in the app (e.g. we have a different URL for services in dev/test/production), it is useful to have a method to change the `api-context` once and for all.

Example usage:
```clojure
(set-api-context {:base-url "https://new-url-here"})
```


f-f:
Hi @wing328! Since CircleCI is failing for some unrelated reason than this commit, I was wondering if master is broken (which doesn't seem to, since the last `mvn -q verify -Psamples` passes there) or the build is just flaky and if we can do anything about it?

PR #8117, opened 2018-05-01
[Kotlin] Sanitize enum varname

### LINKED ISSUE #8116: [Kotlin] Invalid enum var names
##### Description

When enums name start with a number, the generated code won't compile

##### Swagger-codegen version

2.4.0-SNAPSHOT

```json
"enum":[
    "1name"
]
```

java -jar swagger-codegen-cli.jar generate    -i swagger.json    -c config.json    -l kotlin    -o tmp

##### Steps to reproduce

Add an enum parameter to swagger file; generate and compile.

The enumeration name can be prefixed by an underscore, it's done like this for the model name.

Another solution is to surround illegals names with backticks, but enum may be tricky to use (Status.`1available`). And what I understand of kotlin coding convention is backticks are acceptable only for names in tests.

Zomzog:
### PR checklist

### Description of the PR

Sanitize enumeration name for adding underscore when it's required.

It seems ok for all case but when "original" strategy is selected name can be deeply changed (add an underscore, remove space...) 

Must fix #8116

Technical committee member: @jimschubert


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractKotlinCodegen.java
```diff
@@ -507,6 +507,10 @@ private String sanitizeKotlinSpecificNames(final String name) {
         return word;
     }
 
+    private boolean startWithDigits(String modified) {
```
jimschubert: I don't think it's necessary to move such a simple regex to a function call. Also, the regex isn't digits (plural), it's a single digit.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractKotlinCodegen.java
```diff
@@ -402,10 +402,10 @@ public String toEnumVarName(String value, String datatype) {
         if (reservedWords.contains(modified)) {
             return escapeReservedWord(modified);
         }
-
-        return modified;
+        return sanitizeKotlinSpecificNames(modified);
```
jimschubert: Can you add a comment here about why this is necessary, considering this is called at the start of the method? Otherwise I think it would be confusing to maintainers.

Reading it, I see that the issue is an enum input of `1enum` becomes `_1enum` at the start of the method, then `camelCase` (the default) and `PascalCase` end up removing that beginning underscore, resulting in the invalid `1enum`. I had to look at it for a second to understand the need at the end of the method, but I'm not sure all contributors would pick up on the nuance even with the comments in the case statements.

modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractKotlinCodegen.java
Zomzog: The regex is so simple that I make an error on. I preferred method even for an easy regex to avoid any effort when reading the code. But ok I will remove this.

legzo:
Hi, I was about to implement a similar thing because I also have enums beginning with digits in my project. Is there any chance this PR gets merged ? It's 6+ months old, has passed all the tests, is there still something missing ?

I also want to add another little fix in the Kotlin generator but the status of this PR is not very engaging...

jimschubert:
@legzo I had approved this PR, but at the time there was a single person with commit access. He and I (and some others) have since forked this project to openapitools/openapi-generator, and I believe this test case passes there since shortly after the fork.

legzo:
@jimschubert Hi, thank you for your message, I did not know about this fork ! I'll try with the _new_ generator.

PR #8102, opened 2018-04-30
typescript-angular: Ng43 with interfaces petstore scripts

feikede:
Added petstore sample scripts and output for typescript-angular 4.3 withInterfaces generation

### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

@macjohnny Added scripts to generate the petstore examples for angular-typescript 4.3 like discussed in https://github.com/swagger-api/swagger-codegen/pull/7975#issuecomment-381109842

Also added the example output.

PR #8155, opened 2018-05-08
[typescript-angular] Added support for Angular v6 (by adding support for RxJS v6)

akehir:
### PR checklist

### Description of the PR
Update the .mustache templates for Angular 6 (by checking updating RxJS to v6 where appropriate). Also added test cases for Angular 5 & Angular 6, generated tests for Angular 6.

Travis CI: https://travis-ci.org/akehir/swagger-codegen/builds 

@macjohnny


modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -14,7 +14,12 @@ import { Response, ResponseContentType }                     from '@angular/http
...
 {{/useHttpClient}}
 
+{{^useRxJS6}}
 import { Observable }                                        from 'rxjs/Observable';
+{{/useRxJS6}}
+{{#useRxJS6}}
+import { Observable }                                    from 'rxjs';
+{{/useRxJS6}}
 {{^useHttpClient}}
 import '../rxjs-operators';
```
macjohnny: the `import '../rxjs-operators';` should be removed, as pipeable operators are already available as of RxJS 5.5

modules/swagger-codegen/src/main/resources/typescript-angular/package.mustache
```diff
@@ -40,9 +40,9 @@
     "@angular/platform-browser": "^{{ngVersion}}",{{#useNgPackagr}}
     "ng-packagr": {{#useOldNgPackagr}}"^1.6.0"{{/useOldNgPackagr}}{{^useOldNgPackagr}}"^2.4.1"{{/useOldNgPackagr}},{{/useNgPackagr}}
     "reflect-metadata": "^0.1.3",
-    "rxjs": "^5.4.0",
+    "rxjs": "{{#useRxJS6}}^6.1.0{{/useRxJS6}}{{^useRxJS6}}^5.4.0{{/useRxJS6}}",
     "zone.js": "^0.7.6",
-    "typescript": "^2.1.5"
+    "typescript": "~2.7.2"
```
macjohnny: since we are requiring angular cli, constraining to `"typescript": "^2.1.5"` should be fine and more flexible, especially when choosing between angular 5 or 6

modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache
macjohnny: the `import '../rxjs-operators';` should be removed, as pipeable operators are already available as of RxJS 5.5

samples/client/petstore/typescript-angular-v2/npm/package.json
```diff
@@ -34,7 +34,7 @@
     "reflect-metadata": "^0.1.3",
     "rxjs": "^5.4.0",
     "zone.js": "^0.7.6",
-    "typescript": "^2.1.5"
+    "typescript": "~2.7.2"
```
macjohnny: this is not good, we should not require typescript `2.7`here

modules/swagger-codegen/src/main/resources/typescript-angular/package.mustache
akehir: The problem is, with ^2.1.5, typescript 2.8.3 will be installed during the tests which is not compatible to angular core. I'll change it to >= 2.1.5 < 2.8, would that be more appropriate?

macjohnny: yes, this sounds better

modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache
akehir: Sorry, if I misunderstood, the rxjs-operators is imported in a {{^useHttpClient}} {{/useHttpClient}} - I (wrongly?) thought, this meant that these imports are only used if angular is older than 4.3, and that after 4.3 they are not imported anymore.

For RxJS 6.1, the operators would need to be imported differently, if they are used - see: https://github.com/ReactiveX/rxjs/blob/master/MIGRATION.md#howto-convert-to-pipe-syntax .

From what I can see, actually, of the operators, throw & catch are not used by the api service, only map is. However I just found out, that at least at one point map is used, which will need to be converted to a .pipe( map( ()=> {}), ) for rxjs 6. Let me update the PR and get back to you.

macjohnny:

modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache
akehir: Update: The map operator is also only generated when ```{{^useHttpClient}} {{/useHttpClient}}``` - therefore not used in the updated angular 6 code and should not need to be imported.

macjohnny: ah, you are right, I missed that it is in a `{{^useHttpClient}}{{/useHttpClient}}` block.

you don't need to convert the `.map()` to `.pipe(map())`, since it is only used is here: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache#L89

And since this line is only used for Angular <4.3, it should be fine as it is

akehir: My bad, I misread. You want to move it to the ```{{^useRxJS6}}{{/useRxJS6}}``` block, which makes sense. I'll update the PR.

macjohnny: actually I think we shouldn't move it to the `{{^useRxJS6}}{{/useRxJS6}}` block, since this would also affect Angular `>=4.3` and `<6`

akehir: After moving it, angular 5 test fails =(

macjohnny: yes, that's what I thought would happen. so you can simply ignore the initial review comment that stated

> the import '../rxjs-operators'; should be moved inside the {{^useRxJS6}}{{/useRxJS6}} block, see my general review comment.

samples/client/petstore/typescript-angular-v2/with-interfaces/api/user.serviceInterface.ts
```diff
@@ -11,8 +11,7 @@
  */
 import { Headers }                                           from '@angular/http';
 
-import { Observable }                                        from 'rxjs/Observable';
-
+import { Observable }                                    from 'rxjs/Observable';
```
macjohnny: did you re-generate the v2 sample files? this line should not change...

samples/client/petstore/typescript-angular-v2/with-interfaces/README.md
```diff
@@ -52,7 +52,7 @@ In your Angular project:
 
 ```
 // without configuring providers
-import { ApiModule } from '@swagger/angular2-typescript-petstore';
+import { ApiModule } from '';
```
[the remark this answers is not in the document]
akehir: I'm not sure about this one, I believe the package name should not disappear - but I was unfortunately not proficient enough to find the root cause of this. This behaviour is probably not caused by my pull request, though.

The same happens, if I ./bin/typescript-angular-petstore-all.sh``` on master. Was it incorrect to generate all angular examples with my updated ```.mustache``` templates?

Notabene, the same happened here: https://github.com/swagger-api/swagger-codegen/pull/8102/files#diff-ad35a6b2374dd58b5781c610802426c4

samples/client/petstore/typescript-angular-v2/with-interfaces/api/user.serviceInterface.ts
akehir: I re-generated by using ```./bin/typescript-angular-petstore-all.sh ```.  The padding was different in the apiInterface.mustache, I just committed a fix.

samples/client/petstore/typescript-angular-v2/with-interfaces/README.md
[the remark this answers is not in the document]
macjohnny: @akehir I remember that the package name appears and disappears frequently, so I think it is a problem somewhere else, certainly not cause by your PR. If I generate the samples on Windows, the package name appears. Maybe this is the difference?

keradus:
any chances to have this change release? :)

degahr:
Seems that this issue is linked to 2.4.0 ? What is up with this..  Are we going to see progress on this anytime soon?

macjohnny:
@degahr these changes are included in the latest release of https://github.com/openapitools/openapi-generator

rainmanhhh:
how to pass useRxJS6 flag to the template? config-help doesn't show it

keradus:
`java -jar openapi-generator-cli.jar generate -i swagger.json -l typescript-angular --additional-properties ngVersion=6`

rainmanhhh:
> `java -jar openapi-generator-cli.jar generate -i swagger.json -l typescript-angular --additional-properties ngVersion=6`

doesn't work, both 2.3.1 and 3.0.0-rc1 generate
import { Observable }                                        from 'rxjs/Observable';

keradus:
switch to openapi, I'm using 3.0.2 and it generates
`import { Observable } from 'rxjs';`

`wget -nc http://central.maven.org/maven2/org/openapitools/openapi-generator-cli/3.0.2/openapi-generator-cli-3.0.2.jar -O openapi-generator-cli.jar`

Issue #8188, opened 2018-05-15
[Swift] Very large numbers are not converted to string correctly


joeboyscout04:
##### Description
When working with very large numbers (in this case, timestamps in nanoseconds), the default printing by swift results in an exponent character in the string.  For example:

```
let largeNumber: Double = 1.52639622817888e+16
print("\(largeNumber)") // results in "1.52639622817888e+16"
```
Proposed solution is

##### Swagger-codegen version

2.4.1

##### Steps to reproduce

This is reproducible in an Swift project that uses Doubles for request parameters, as long as the input Double is sufficiently large. 

##### Related issues/PRs
TBD

##### Suggest a fix/enhancement
Proposed fix is to switch to using NumberFormatter()
```
let formatter = NumberFormatter()
let num = NSNumber(value: bigNumber)

print(formatter.string(from: num)) // results in "15263962281788800"

```

PR #8190, opened 2018-05-15
[Swift] Issue 8188

joeboyscout04:
### PR checklist

@ehyche

### Description of the PR

This PR changes the default behavior of converting numbers (Double, Int, etc) to String.  Previously, the string implementation was left up to Swift as ```"\(value)"```

However, when dealing with especially long numbers, this causes there to be exponential characters in the string output of a number, which seems wrong.  See https://github.com/swagger-api/swagger-codegen/issues/8188 for more detail.


joeboyscout04:
It seems like I'm getting some fairly large difference between this branch (master) and 2.3.0 when I build this from source.  The _id -> id thing seems to have happened recently, I'm not sure why the integration tests weren't working properly then.  is "_id" the path forward?

joeboyscout04:
This was a silly idea.  It still isn't that precise of a solution and doesn't work for parsing.

PR #8198, opened 2018-05-16
[Go] Fix API error handling

roycaihw:
### PR checklist

@wing328 @antihax
### Description of the PR
Previous implementation doesn't raise error properly. Arguably we could remove the if statement.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/go/api.mustache
+++ b/modules/swagger-codegen/src/main/resources/go/api.mustache
@@ -236,27 +236,29 @@ func (a *{{{classname}}}Service) {{{nickname}}}(ctx context.Context{{#hasParams}
 		}
 	}
 	{{/returnType}}
-
-	if localVarHttpResponse.StatusCode >= 300 {
-		newErr := GenericSwaggerError{
-			body: localVarBody,
-			error: localVarHttpResponse.Status,
-		}
-		{{#responses}}{{#dataType}}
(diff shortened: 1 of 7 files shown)


antihax:
this was intended, what is not working correctly?

`// If we succeed, return the data, otherwise pass on to decode error`

antihax:
I see.. we should likely look at removing the >= 300 on the line below and adding a case if there was a status response with no model.

roycaihw:
> we should likely look at removing the >= 300 on the line below and adding a case if there was a status response with no model.

I agree. And those checking StatusCode == 200 statements inside StatusCode > 300 blocks look confusing to me 
https://github.com/swagger-api/swagger-codegen/blob/4788eec0a8759eb2042688166ece908c8e305709/samples/client/petstore/go/go-petstore/api_pet.go#L714-L720

roycaihw:
StatusCode check removed: https://github.com/swagger-api/swagger-codegen/pull/8198/commits/34e016e8da1b8878cca8061ed7d247831593421e?w=1

The second commit is generated from `bin/go-petstore.sh` and `bin/security/go-petstore.sh`. Adding testcase now

roycaihw:
@antihax CI has passed. Actually it's not clear to me how to test this behavior. I guess I need to add a test in https://github.com/swagger-api/swagger-codegen/blob/master/samples/client/petstore/go/pet_api_test.go, but I don't know how to modify the test server to create a broken endpoint which talks to some existing client method.

PR #8245, opened 2018-05-29
Fix getter methods for Java model with additionalProperties

### LINKED ISSUE #4970: [Java] Runtime problem parsing response object defined with additionalProperties
##### Description

I'm seeing unexpected behavior when Swagger generated client code parses a response.  The response object is defined to have a set of fixed properties.  It also has a set of dynamic properties (not known until run time).  I'm using `additionalProperties` to indicate the object has dynamic properties.

For example, consider the following response object:

```json
{
  "@entryid": "xxx",
  "@noteid": "1139E",
  "fldText": "This is a test"
}
```

The first two properties (`@entryid` and `@noteid`) are a fixed part of the schema.  The third property (`fldText`) is dynamic.  I expect Swagger Codegen to generate getter methods for the first properties (which it does) and I expect to use the `HashMap` interface to read any dynamic properties.

The problem is the generated getter methods don't work.  I have to read all properties via the `HashMap`.  More details below in the *Swagger declaration* section.

##### Swagger-codegen version

Version 2.2.2.  I don't know if this is a regression.

Here's an excerpt from my OpenAPI spec:

```yaml
  viewEntry:
    type: object
    properties:
      '@entryid':
        description: |
          Position of the entry in the view or folder and the universal
          ID of any associated document.
        type: string
      '@noteid':
        description: |
          The note ID of the document associated with the entry, or an
          empty string if the entry is a category or total.
        type: string
    additionalProperties:
      type: object
```

When I generate a Java client I get a ViewEntry class.  As expected it extends `HashMap` so it can hold dynamic properties:

```java
public class ViewEntry extends HashMap<String, Object> {
```

Also as expected it includes getters and setters for the fixed properties:

```java
  public String getEntryid() {
    return entryid;
  }

  public void setEntryid(String entryid) {
    this.entryid = entryid;
  }
```

However, when the Swagger client receives a response it doesn't parse the fixed properties as expected.  In other words `entry.getEntryid()` always returns `null`.  To read the value of `@entryid` I have to use the `HashMap` interface -- `entry.get("@entryid")`. 

If I remove `additionalProperties` from the API definition and generate the client again, the getters do work as expected.  Of course, this isn't a suitable work-around because ViewEntry no longer extends `HashMap` and my dynamic properties are lost during response parsing.

`generate -i /{specs}/data.yaml -l java -o /{clients}/data/java`

##### Steps to reproduce

Since this is a runtime problem you need my API to reproduce.  Unfortunately, the API isn't publicly available.  Hopefully, you can diagnose the problem based on the information already provided. 

##### Related issues

I couldn't find a similar issue.

##### Suggest a Fix

One way to fix the problem would be to change the generated getters to delegate to the `HashMap`.  For example, instead of this:

```java
  public String getEntryid() {
    return entryid;
  }
```

Generate this:

```java
  public String getEntryid() {
    return get("@entryid");
  }
```

And of course make a similar change to the setters.

### LINKED ISSUE #5187: [JAVA][SPRING] Swagger codegen generates wrong java object for json objects with additional properties
##### Description
Swagger Codegen generates a java model object that extends HashMap for json objects that have the ` additionalProperties:` property set. This leads to jackson only serealizing the properties added to the hashmap and ignoring all additional properties that are set explicitly in the java file. 
##### Swagger-codegen version
2.2.2
```yaml
 Datavalue:
    type: object
    description: The individual dataelements 
    required:
      - id
    properties:
      id:
        type: string
      source:
        type: string
      target:
        type: string
    additionalProperties:
      type: string
```
This generates a model object that starts like this:
```java
public class Datavalue extends HashMap<String, String>  {
  @JsonProperty("id")
  private String id = null;

  @JsonProperty("source")
  private String source = null;

  @JsonProperty("target")
  private String target = null;

  public Datavalue id(String id) {
    this.id = id;
    return this;
  }
```

Serializing this objects leads to a json that is missing the explicitly set properties (id, source, target).
http://stackoverflow.com/questions/31320983/jackson-serialise-map-with-extra-fields

I think instead it should generate a java object that has a hashmap as a variable and the following annotated getter and setter methods, similar to this:
```java
public class Datavalue   {
  @JsonProperty("id")
  private String id = null;

  @JsonProperty("source")
  private String source = null;

  @JsonProperty("target")
  private String target = null;

  protected Map<String,String> otherProperties = new HashMap<String,String>();

  public Datavalue id(String id) {
    this.id = id;
    return this;
  }
  @JsonAnyGetter
  public Map<String, String> any() {
   return this.otherProperties;
  }

  @JsonAnySetter
  public void set(String name, String value) {
   this.otherProperties.put(name, value);
  }
```
as Explained here: http://www.cowtowncoder.com/blog/archives/2011/07/entry_458.html

### LINKED ISSUE #5259: [Spring] Codegen 2.2.2 does not render objects comprised by allOf and containing additionalProperties:
##### Description

When running Codegen 2.2.2 to render Spring objects, when the object has a definition using the allOf schema attribute and also the additionalProperties attribute, the rendered Java object contains no members at all.

##### Swagger-codegen version

swagger-codegen-cli-2.2.2.jar

```yaml
  DynamicAccountModel:
    allOf:
      - properties:
          accountId:
            type: string
            description: Account ID
      - additionalProperties:
          type: string
```

```
java -jar ../tools/swagger-codegen-cli-2.2.2.jar generate -i allof-no-render.yaml -l spring -o target
```
##### Steps to Reproduce

Execute above command.

Note that the class contains neither derives from HashMap (due to having the additionalParameters attribute) nor contains an accountId member. 

```
package com.foo.bar;

import java.util.Objects;
import java.io.Serializable;
import javax.validation.constraints.*;
/**
 * DynamicAccountModel
 */
@javax.annotation.Generated(value = "io.swagger.codegen.languages.SpringCodegen", date = "2017-03-30T07:57:06.421-04:00")

public class DynamicAccountModel  implements Serializable {

  @Override
    if (this == o) {
      return true;
    }
      return false;
    }
    return true;
  }

  @Override
  public int hashCode() {
    return Objects.hash();
  }

  @Override
  public String toString() {
    sb.append("class DynamicAccountModel {\n");

    sb.append("}");
    return sb.toString();
  }

  /**
   * Convert the given object to string with each line indented by 4 spaces
   * (except the first line).
   */
  private String toIndentedString(java.lang.Object o) {
    if (o == null) {
      return "null";
    }
    return o.toString().replace("\n", "\n    ");
  }
}
```

stianlik:
### PR checklist

- [ ] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`. **Did not generate petstore example because that would significantly increase the diff**
- [ ] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language. **No members left in the technical committee for Java, including @HugoMario because he has merged other PRs**

### Description of the PR

When a model class is configured with additionalProperties, getters for non-dynamic properties always return null because values have been set as properties of the map instead of the class. As a workaround, this commit ensures that we these values are retrieved from the underlying map.

This (partially) fixes #4970, fixes #5259, and fixes #5187

The issue of `additionalProperties` is not completely solved by this PR, however, we support the situation where `additionalProperties` is of type `object`, or all fields are of the same type as `additionalProperties`. I.e. it will not work if we have properties of type `integer` and `additionalProperties` of type `array` due to incorrect type casting.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/Java/pojo.mustache
+++ b/modules/swagger-codegen/src/main/resources/Java/pojo.mustache
@@ -140,7 +140,12 @@ public class {{classname}} {{#parent}}extends {{{parent}}} {{/parent}}{{#parcela
   {{{vendorExtensions.extraAnnotation}}}
 {{/vendorExtensions.extraAnnotation}}
   public {{{datatypeWithEnum}}} {{#isBoolean}}is{{/isBoolean}}{{getter}}() {
+    {{#parent}}
+    return {{name}} == null ? ({{{datatypeWithEnum}}}) this.get("{{name}}") : {{name}};
+    {{/parent}}
+    {{^parent}}
     return {{name}};
+    {{/parent}}
   }
   {{^isReadOnly}}
 


HugoMario:
thanks @stianlik !!!

fralalonde:
As I suggested in #5187, could the solution be to flag-enable this jackson-enhanced behavior to preserve compatibility with `gson` and `resteasy`?

PR #8284, opened 2018-06-05
Fixed template path resolvement to work also on Windows OS.

thomas-worm-datev:


modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java
```diff
@@ -1011,13 +1011,13 @@ private static String generateParameterId(Parameter parameter) {
 
     private com.github.jknack.handlebars.Template getHandlebars(String templateFile) throws IOException {
         if (templateFile.startsWith(config.templateDir())) {
-            templateFile = templateFile.replaceFirst(config.templateDir(), StringUtils.EMPTY);
+            templateFile = StringUtils.replaceOnce(templateFile, config.templateDir(), StringUtils.EMPTY);
         }
         TemplateLoader templateLoader = null;
         if (config.additionalProperties().get(CodegenConstants.TEMPLATE_DIR) != null) {
-            templateLoader = new FileTemplateLoader(config.templateDir(), ".mustache");
+            templateLoader = new FileTemplateLoader(config.templateDir().replace("\\", "/"), ".mustache");
```
HugoMario: hey @thomas-worm-datev, i think this could work but it's not the causes of the issue, i tested locally and when we use a custom templates it crash.

The real issue is this statement:
```
templateFile = templateFile.replace("\\", "/");
```
at line [568](https://github.com/swagger-api/swagger-codegen/blob/3.0.0/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java#L568)

if you're agree, please assign permissions to `HugoMario` to your repo and i can help sending the fix from your PR.

thomas-worm-datev: Hi @HugoMario,

I first tested with your suggested line, but that solved the problem only partially. I got error that V2\JavaSpring/api.mustache could not be found.

This was caused from the line
`final Handlebars handlebars = new Handlebars(templateLoader);`

So I found out that even the templateDir path has to be handled with a safe path handling.


Best Regards,
Thomas

HugoMario: I'm not sure if that line if the issue, try using a custom templates in a windows OS with your changes and you'll see my point.

HugoMario:
thanks a lot @thomas-worm-datev !!!

PR #8316, opened 2018-06-14
Update README.md

oss92:
Add fashionTrade to list of companies


HugoMario:
thanks a lot @oss92 !

Issue #8359, opened 2018-06-27
[JAVA][RESTEasy] ApiClient sets header-key as header-value


MarvGilb:
##### Description

I created a Java client with the `swagger-codegen-maven-plugin` and I am setting custom headers over the `ApiClient`. But when looking at the request, I see that the headers are set but use the header key as value parameter.

##### Swagger-codegen version

swagger-codegen-maven-plugin: 2.3.1
with:
```
<configuration>
  <inputSpec>...</inputSpec>
  <language>java</language>
  <output>...</output>
  ...
  <configOptions>
    <library>resteasy</library>
    <java8>true</java8>
      ...
  <configOptions>
<configuration>
```

##### Example

When performing a request with the configured client.
```
ApiClient defaultClient = Configuration.getDefaultApiClient();
defaultClient.addDefaultHeader("headerKey1", "headerValue1");
defaultClient.addDefaultHeader("headerKey2", "headerValue2");
..some request.
```

The reuqest has the following headers:

- headerKey1: headerKey1
- headerKey2: headerKey2

whereas it should be :

- headerKey1: headerValue1
- headerKey2: headerValue2

##### Problem file

ApiClient.java:

```
public <T> T invokeAPI(String path,  .... ) {
  ...
  for (Entry<String, String> defaultHeaderEnrty: defaultHeaderMap.entrySet()) {
    if (!headerParams.containsKey(defaultHeaderEnrty.getKey())) {

      ########################
      String value = defaultHeaderEnrty.getKey();  <=== This should use the value.
      ########################

      if (value != null) {
        invocationBuilder = invocationBuilder.header(defaultHeaderEnrty.getKey(), value);
      }
    }
  }
  ...
}
```

##### Pull Request

https://github.com/swagger-api/swagger-codegen/pull/8360

frantuma:
fixed in #8446

PR #8360, opened 2018-06-27
[Java][RESTEasy] Fixed setting of custom headers

MarvGilb:
Set request headers as key:value. Was key:key

See issue: #8359


frantuma:
included and replaced by #8446

Issue #8138, opened 2018-05-04
[Elixir] Bug in 2.3.1. Generating multi-part body instead of direct JSON content


chgeuer:
##### Description

The generated `Elixir` code from a Swagger definition *seems* to be wrong, but I'm not sure. 

I try to generate an Elixir SDK for the Azure Resource Management API. I wrote a [generator app](https://github.com/chgeuer/ex_microsoft_azure_management_generator), which uses uses `swagger-codegen-cli-2.3.1.jar` to generate Elixir code. 

The [swagger spec document](https://github.com/Azure/azure-rest-api-specs/blob/master/specification/resources/resource-manager/Microsoft.Resources/stable/2018-02-01/resources.json#L155-L162) says `"in": "body"`, and the API expects JSON structure directly in the body, but the Elixir code generates a multi-part body. 

```json
      "put": {
        "operationId": "Deployments_CreateOrUpdate",
        "summary": "Deploys resources to a resource group.",
        "parameters": [
          ...
          {
            "name": "parameters",
            "in": "body",
            "required": true,
            "schema": { "$ref": "#/definitions/Deployment" },
            "description": "Additional parameters supplied to the operation."
          },
```

The Elixir swagger generator creates this [code](https://github.com/chgeuer/ex_microsoft_azure_management/blob/master/lib/microsoft/azure/management/resources/api/deployments.ex#L95)

```elixir
  @spec deployments_create_or_update(Tesla.Env.client, String.t, String.t, Microsoft.Azure.Management.Resources.Model.Deployment.t, String.t, String.t, keyword()) :: {:ok, Microsoft.Azure.Management.Resources.Model.DeploymentExtended.t} | {:error, Tesla.Env.t}
  def deployments_create_or_update(connection, resource_group_name, deployment_name, parameters, api_version, subscription_id, _opts \\ []) do
    %{}
    |> method(:put)
    |> url("/subscriptions/#{subscription_id}/resourcegroups/#{resource_group_name}/providers/Microsoft.Resources/deployments/#{deployment_name}")
    |> add_param(:body, :"parameters", parameters)
    |> add_param(:query, :"api-version", api_version)
```

The parameter JSON is added as `|> add_param(:body, :"parameters", parameters)`, which results in 

```elixir
 body: %Tesla.Multipart{
    boundary: "6R3XOsKKCMYMatbOce3P7yRHKD6Zmm",
    content_type_params: [],
    parts: [
      %Tesla.Multipart.Part{
[...]

PR #8391, opened 2018-07-05
Fix Elixir :body handling (swagger-api/swagger-codegen#8138)

chgeuer:
Fix swagger-api/swagger-codegen#8138

When JSON payload needs to be in the body (`"in": "body"`), it must not generate a multi-part body, but embed directly in HTTP body.


chgeuer:
Is there something additional I need to do to get this in?

PR #8397, opened 2018-07-05
OAS 3.0 - Use OAS 3.0 annotations instead of old Swagger 1.5.x / OAS 2.x ones (@Api / @ApiXxx)

reta:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. - [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

Adding new configuration option and command-line argument to support generation of old OpenAPI 2.0 (Swagger) servers / clients. This is a supporting PR for https://github.com/swagger-api/swagger-codegen-generators/pull/101 since most command-line arguments and config properties are referenced from `CodegenConstants`.

@HugoMario could you please take a look? 
Thanks!

PR #8436, opened 2018-07-17
OAS 3.0 - Fixing the 'useOas2' option override in case of Codegen Maven plugin usage

reta:
### PR checklist

### Description of the PR

Found out that when Maven plugin is being used, the options may come already set from `additionalProperties` but the `DefaultGenerator` will override it with the default settings nonetheless. This small PR fixes that (followup for https://github.com/swagger-api/swagger-codegen/pull/8397)


reta:
@HugoMario Minor fix for the case when `useOas2` is set from Maven plugin `configOptions` (was overridden by default value).

HugoMario:
good! thanks @reta

PR #8463, opened 2018-07-20
added logic for newlines in comments for operation descriptions only

john-french:
Preserve new lines in comments in C#


modules/swagger-codegen/src/test/java/io/swagger/codegen/options/AspNetCoreServerOptionsProvider.java
```diff
@@ -10,7 +10,8 @@
     public static final String PACKAGE_NAME_VALUE = "swagger_server_aspnetcore";
     public static final String PACKAGE_VERSION_VALUE = "1.0.0-SNAPSHOT";
     public static final String SOURCE_FOLDER_VALUE = "src_aspnetcore";
-
+    public static final boolean PRESERVE_NEW_LINES = false;
+
```
frantuma: Could you replace all tabs with spaces in affected files (we have a CI check for that)?

PR #8625, opened 2018-08-24
Add parse options to V3 resolve call

john-french:
V3 specs weren't being resolved when the spec was passed to codegen instead of the spec URL. I added `ParseOptions` to the call to `SwaggerParser` for these specs to indicate that the parser should resolve and flatten these specs (previously `ParseOptions` was null).


frantuma:
replaced by #8629

Issue #6336, opened 2017-08-17
[JAVA-Spring] javax.validation.UnexpectedTypeException: HV000030: No validator could be found for constraint 'javax.validation.constraints.Pattern' validating type 'byte[]' 


Ghost93:
##### Description

Hi.

I'm using swagger-codegen to generate a spring boot server.
Recently I'm trying to add more constraints to input using patterns, and the `@Valid` annotation.

The problem occurs when I try to use validation on objects that contain references to models with fields that are `byte[]`. This is my problematic model:

```yaml
UploadedImage:
    type: object
    properties:
      data:
        type: string
        format: byte
      mime_type:
        type: string
        enum: [image/bmp,image/jpeg,image/png]
      url:
        type: string
```

and I see that Swagger-codegen has generated the following class for it:
```java
package com.github.ghost93.model;

import com.fasterxml.jackson.annotation.JsonCreator;
import com.fasterxml.jackson.annotation.JsonProperty;
import com.fasterxml.jackson.annotation.JsonValue;
import io.swagger.annotations.ApiModelProperty;
import java.util.Objects;
import javax.validation.constraints.Pattern;

/**
 * UploadedImage
 */
@javax.annotation.Generated(value = "io.swagger.codegen.languages.SpringCodegen", date = "2017-08-17T22:15:18.867+03:00")

public class UploadedImage {

  @JsonProperty("data")
  private byte[] data = null;
  @JsonProperty("mime_type")
  private MimeTypeEnum mimeType = null;
  @JsonProperty("url")
  private String url = null;

  public UploadedImage data(byte[] data) {
    this.data = data;
    return this;
  }

  /**
   * Get data
   *
   * @return data
   **/
  @ApiModelProperty(value = "")

  @Pattern(regexp = "^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$")
  public byte[] getData() {
    return data;
  }

  public void setData(byte[] data) {
    this.data = data;
  }

  public UploadedImage mimeType(MimeTypeEnum mimeType) {
    this.mimeType = mimeType;
    return this;
  }

  /**
   * Get mimeType
   *
   * @return mimeType
   **/
  @ApiModelProperty(value = "")

  public MimeTypeEnum getMimeType() {
    return mimeType;
  }

  public void setMimeType(MimeTypeEnum mimeType) {
    this.mimeType = mimeType;
  }

[...]

PR #8674, opened 2018-09-07
Wrong @Pattern validation for byte array

arkaev:
https://github.com/swagger-api/swagger-codegen/issues/6336

Just fixed mustache tepmplates. No petstore updates in this commit because of totally outdated and it's better new commit for update them


frantuma:
please build

PR #8535, opened 2018-08-04
[BUG] [CRITICAL] [PHP] Symfony server incorrect validation: min contstraint in max condition

4n70w4:
fix LessThan & LessThanOrEqual conditions

@dkarlovi @mandrean

PR #8730, opened 2018-09-20
[PHP] update php-symfony/services.mustache

4n70w4:
fix 2 deprecated messages:

"User Deprecated: Not quoting the scalar "%swagger_server.validator%" starting with the "%" indicator character is deprecated since Symfony 3.1 and will throw a ParseException in 4.0 on line 20."

"User Deprecated: Not quoting the scalar "%swagger_server.serializer%" starting with the "%" indicator character is deprecated since Symfony 3.1 and will throw a ParseException in 4.0 in "swagger/server-bundle/Resources/config/services.yml" on line 17."

@dkarlovi 
@mandrean


4n70w4:
also see https://github.com/swagger-api/swagger-codegen/pull/8535

PR #8752, opened 2018-09-27
fix ui and spec tags

frantuma:


frantuma:
please build

frantuma:
please build

PR #8769, opened 2018-10-02
Frozenstep kotlin download file fix

Frozenstep:
### PR checklist

### Description of the PR

This updates api.mustache and ApiClient.kt.mustache to handle when the response from the server is a file. It also makes changes in the generated sample code for kotlin-client-petstore

In addition 
-We fixed the default headers so they do not override generated headers
-The response Deserializer now looks at the response header content type instead assuming the response is the same as the accept header in the request.
-We added all accept types into the request header


HugoMario:
thanks a lot @Frozenstep !!!!!

PR #8822, opened 2018-10-17
Frozenstep kotlin void response type

Frozenstep:
### PR checklist

### Description of the PR

When our server responded with a 204 message, but the body had no json content, an EOF exception would be thrown because it tried to deserialize json that did not exist. 

Our solution was to follow a similar practice as some other languages, when the the expected response type of an API call is void (for kotlin, it's Unit), then simply return void from the response deserializer. 

@HugoMario Here's a similar PR


frantuma:
please build

PR #8847, opened 2018-10-24
Adding linquest support

umutozel:
Adding [linquest](https://github.com/jin-qu/linquest) typescript code generation support.


umutozel:
Will try again soon

PR #8850, opened 2018-10-24
Add TypeScript generation for Linquest

umutozel:
Add TypeScript code generation support for [linquest](https://github.com/jin-qu/linquest).


umutozel:
I'm not sure why Python tests fail. I'm not sure if I should mention @kenjones-cisco

micw523:
I think TravisCI is killing everybody's build with "W605 invalid escape sequence"

PR #8860, opened 2018-10-25
For polymorphism with Jackson, use the As.EXISTING_PROPERTY feature

charlescapps:
We were seeing duplicate fields serialized in the JSON from generated Swagger Codegen model classes, because in the newer versions of Jackson you have to use `As.EXISTING_PROPERTY` in the `@JsonSubTypes` annotation if the discriminant is also a normal serialized field.

These settings should be fine to fix the issue; the only caveat being that Jackson 2.5.0 or above must be used. I'll investigate if this is an issue. We should be able to ensure we never use an older version than that.

### PR checklist

### Description of the PR


charlescapps:
I'll close this for now. We're not using Java clients at my current position and hopefully everyone should be moving towards Swagger 3.0 anyway...

PR #8880, opened 2018-10-30
Spring RestTemplate: Encode request URL

JonasGroeger:


cck0001:
We are running into this same issue (basically URL parameters with unicode characters were not being properly encode). I see you closed this issue and we were going to do the exact same thing as you proposed here. What was the solution you ultimately adopted?

kingtran2112:
Hi, I also with the problem when have Germany character in URL parameters (more specific, it's ß character). The character is not encoded. I think we are get the same problem here and I also would like to know the solution for this. Thanks

edgardbr0:
this correction is not found in the master. Is there any way to activate this encode?

PR #8911, opened 2018-11-13
[Python client] Lazily initialize thread pool in API client

jdevera:
### PR checklist

### Description of the PR

We are using the swagger python client in parametrized tests with pytest to check the same operation on our API with all permission combinations. This results in thousands of tests, each with their own swagger client instance. We hit a wall after a certain number of tests, we get an error saying no more threads can be created.

The problem is in the creation of the swagger ApiClient object, which unconditionally starts a ThreadPool that is used for async calls. We do not use async calls.

I have tried to cleanup the ThreadPool but have achieved no success in reducing the number of threads the pytest process has opened.

However, the lazy initialization of the ThreadPool did work. Sync call only scenarios get zero penalty for the creation of threads and in the case of the use of async calls, the first call will have to wait a bit longer for the initialization of the pool.

#### From the commit

The Python API client thread pool is only used with async calls and
consumes system resources (threads). Many instances of the API client
can reach system limits even in cases where there will never be any
async call.

This change initializes the API client thread pool on its first use.

CC: @kenjones-cisco

### DIFF
--- a/modules/swagger-codegen/src/main/resources/python/api_client.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/api_client.mustache
@@ -60,7 +60,7 @@ class ApiClient(object):
             configuration = Configuration()
         self.configuration = configuration
 
-        self.pool = ThreadPool()
+        self._pool = None  # Use the pool property to lazily initialize the ThreadPool.
         self.rest_client = rest.RESTClientObject(configuration)
         self.default_headers = {}
         if header_name is not None:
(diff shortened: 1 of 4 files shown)


jdevera:
Is there a way to rerun the failed checks? They seem to have failed while building, I do not see how the failures relate to my changes.

jdevera:
Is this abandoned?

HugoMario:
@jdevera, just ping me any time you need help with a PR. i just miss some when reviewing.

frantuma:
please build

jdevera:
Okay then @HugoMario, thanks.  How do we get the tests to run again? the failures do not seem related to my changes.

Is there something else that is stopping this from moving forward?

HugoMario:
trusting in jenkins, crossing fingers !!!

PR #8946, opened 2018-11-26
[Fix #8858][Python] Discriminator value is no longer converted to lowercase

### LINKED ISSUE #8858: [Python] Discriminator value is converted to lower case
##### Description

Data deserialization based on a discriminator property does not seem to work correctly if the discriminator values are not all lower case. The discriminator_value_class_map maps the discriminator values as they are to class names (https://github.com/swagger-api/swagger-codegen/blob/7a049041e237099db112d3652ac67220a2566a0f/modules/swagger-codegen/src/main/resources/python/model.mustache#L51), while get_real_child_model converts the discriminator value to lower case and then tries to get the mapping (https://github.com/swagger-api/swagger-codegen/blob/7a049041e237099db112d3652ac67220a2566a0f/modules/swagger-codegen/src/main/resources/python/model.mustache#L172).

Using a swagger file that contains definitions similar to those below, the generated client does not succeed in deserializing Applications with 'PostgreSQL' type as PostgreSQL instances. Removing the call to lower from get_real_child_model seems to solve the problem.

I'm not sure If I'm missing something. Are the discriminators supposed to be all lower case? Is it intended behaviour? Clients generated in other languages do not seem to have the same problem. If this behaviour is not intended removing the call to lower from get_real_child_model could fix the issue.

##### Swagger-codegen version

2.3.1

```json
"definitions": {
	"Application": {
		"type": "object",
		"discriminator": "type",
		"properties": {
			"name": {
				"type": "string"
			}
		},
		"title": "Application"
	},
	"PostgreSQL": {
		"title": "PostgreSQL",
		"allOf": [
			{
				"$ref": "#/definitions/Application"
			},
			{
				"type": "object",
				"properties": {
					"name": {
						"type": "string"
					}
				},
				"title": "PostgreSQL"
			}
		]
	}
}
```

swagger-codegen generate -l python -i swagger -o client

Remove the call to lower from get_real_child_model (https://github.com/swagger-api/swagger-codegen/blob/7a049041e237099db112d3652ac67220a2566a0f/modules/swagger-codegen/src/main/resources/python/model.mustache#L172).

CristianPop:
### Fix #8858

The discriminator values were converted to lowercase prior to selecting the appropriate class to instantiate and no match was found in the class map.

https://github.com/swagger-api/swagger-codegen/blob/7a049041e237099db112d3652ac67220a2566a0f/modules/swagger-codegen/src/main/resources/python/model.mustache#L51-L54

https://github.com/swagger-api/swagger-codegen/blob/7a049041e237099db112d3652ac67220a2566a0f/modules/swagger-codegen/src/main/resources/python/model.mustache#L170-L173
As a result, the base class was always instantiated and not the subclasses corresponding to the discriminator values.

The model.mustache template has been modified and the discriminator values are no longer converted to lowercase. As subclasses override the discriminator to None and get_real_child_model is executed once for the base class and once for the subclass, get_real_child_model has been modified to return None in case the discriminator is None to avoid a KeyError.

https://github.com/CristianPop/swagger-codegen/blob/f48b669159990f644914e42e2abe18704c158a64/modules/swagger-codegen/src/main/resources/python/model.mustache#L170-L175

The subclasses are now appropriately selected and instantiated based on the discriminator.

The petstore sample has been updated.

@kenjones-cisco

### DIFF
--- a/modules/swagger-codegen/src/main/resources/python/model.mustache
+++ b/modules/swagger-codegen/src/main/resources/python/model.mustache
@@ -169,7 +169,9 @@ class {{classname}}({{#parent}}{{parent}}{{/parent}}{{^parent}}object{{/parent}}
 {{#discriminator}}
     def get_real_child_model(self, data):
         """Returns the real base class specified by the discriminator"""
-        discriminator_value = data[self.discriminator].lower()
+        discriminator_value = None
+        if self.discriminator is not None:
+            discriminator_value = data[self.discriminator]
         return self.discriminator_value_class_map.get(discriminator_value)
 
 {{/discriminator}}
--- a/samples/client/petstore-security-test/python/petstore_api/api/fake_api.py
+++ b/samples/client/petstore-security-test/python/petstore_api/api/fake_api.py
@@ -34,35 +34,35 @@ class FakeApi(object):
         self.api_client = api_client
 
     def test_code_inject____end__rn_n_r(self, **kwargs):  # noqa: E501
-        """To test code injection */ &#39; \&quot; &#x3D;end -- \\r\\n \\n \\r  # noqa: E501
+        """To test code injection */ ' \" =end -- \\r\\n \\n \\r  # noqa: E501
 
         This method makes a synchronous HTTP request by default. To make an
-        asynchronous HTTP request, please pass async=True
-        >>> thread = api.test_code_inject____end__rn_n_r(async=True)
+        asynchronous HTTP request, please pass async_req=True
+        >>> thread = api.test_code_inject____end__rn_n_r(async_req=True)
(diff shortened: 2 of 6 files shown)


safchain:
@cristianpop I'm facing the same issue, it would be nice to have this fix rebased and eventually merged

cristianpop:
@safchain Nobody responded to this pull request for a whole year. Try to use https://github.com/OpenAPITools/openapi-generator. You should not encounter this problem (I've tried it sometime in the past and it seemed to generate proper code in this situation). The project also seems to be better maintained.

safchain:
@cristianpop thanks for the pointer, worked for me

Issue #8836, opened 2018-10-22
[typescript-angular] error TS1345: An expression of type 'void' cannot be tested for truthiness


itsTeknas:
##### Description

The generated typescript client fails in angular 7 build when there seems to be any sort of file field.

Typescript version 

"typescript": "^3.1.3"

##### Swagger-codegen version

The latest online version

##### Swagger declaration file content or url

```yaml
parameters:
        - name: attachment1
          type: file
          in: formData
          required: false
```

Produces:

```
if (attachment1 !== undefined) {
            formParams = formParams.append('attachment1', <any>attachment1) || formParams;
}
```

The definition of append tracks down to 

```
let formParams: { append(param: string, value: any): void; };
```

<img width="654" alt="screenshot 2018-10-22 at 2 13 46 pm" src="https://user-images.githubusercontent.com/6565602/47284378-c2ba4280-d604-11e8-8b45-bbe369b6d0ad.png">

##### Steps to reproduce

Generate a new client with angular-typescript selected.

artwel:
Same problem in our environment, after updating angular from v6 to v7 it is not possible to compile the genereated swagger services with the same error as seen above.

Dviejopomata:
Also happens with 
```
import { Observable } from 'rxjs/Observable'
```
Which has to be :
```
import { Observable } from 'rxjs'
```
As a workaround you can use some bash after generating the code

```bash
sed -i 's/rxjs\/Observable/rxjs/' ./src/app/swagger/api/api.service.ts 
sed -i 's/formParams = formParams/formParams/' ./src/app/swagger/api/api.service.ts 
sed -i 's/ || formParams//' ./src/app/swagger/api/api.service.ts 
```

itsTeknas:
@Dviejopomata , this actually has been implemented, you just need to pass the ngversion parameter while generating. The online generator is generating for angular 4.3 as a default.

awalland:
> @Dviejopomata , this actually has been implemented, you just need to pass the ngversion parameter while generating. The online generator is generating for angular 4.3 as a default.

do you have any pointers to what to set it? i tried with `--additional-properties ngVersion=6.0.0` but that didn't seem to do anything.

Dviejopomata:
@awalland Make sure you are using v3.0.0 at least, i had to change from v2.3.1 to v3.0.0 and it worked fine:

Since i'm using docker this is what my command looks like:
```
[...]

PR #8968, opened 2018-12-03
On Upgrade to Angular 7, get a tslint error for truthiness check

navneetkarnani:
Check these issues for a trail:
https://github.com/Microsoft/TypeScript/issues/26262
https://github.com/Microsoft/TypeScript/commit/14d3c69c722793c7a007d53081977410fb521aca

### PR checklist

### Description of the PR

The Typescript 3.1 compiler (one that is used with Angular 7 onwards), has a check for void functions being checked for Truthiness. This change fixes the service function to have multiple return types if the HttpClient is being used, which is where the code gets generated.


jcollard:
:+1: Resolve my issue.

navneetkarnani:
Addresses: https://github.com/swagger-api/swagger-codegen/issues/8836

navneetkarnani:
@roni-frantchi , since you are the creator of typescript-angular, please review. I don't see a reviewer for typescript on the technical committee

navneetkarnani:
> Will the PR not create error when formParams is FormData type, for 'multipart/form-data'?
> 
> Since it forces the formParams to be HttpParams or void.

Changing recommendation to "any" instead of specific type to avoid having to specify all possible combinations ( and that I may not be aware of ).

roni-frantchi:
@navneetkarnani using `any` would solve, but as you've mention, we lose type safety.
Here's another way to go at it, which I think will solve it: #9012 .
What do you think?.

fireappsAdmin:
Is there any type of work around for this?

PR #9012, opened 2018-12-24
An expression of type 'void' cannot be tested for truthiness

roni-frantchi:
Closes #8968 .
See PR by @navneetkarnani for more details and trial.

### PR checklist

- [ ] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/typescript-angular-all-petstore.sh` and `./bin/security/{LANG}-petstore.sh`.

### Description of the PR

The Typescript 3.1 compiler (one that is used with Angular 7 onwards), has a check for void functions being checked for Truthiness.
This change drops the or (`||`) operator altogether, extracts the assignment code to runtime and utilizes the comma (`,`) operator instead.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache
+++ b/modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache
@@ -262,7 +262,8 @@ export class {{classname}} {
 {{#hasFormParams}}
         const canConsumeForm = this.canConsumeForm(consumes);
 
-        let formParams: { append(param: string, value: any): void; };
+        let formParams: FormParams;
+        let append = getAppender();
         let useForm = false;
         let convertFormParamsToString = false;
 {{#formParams}}
@@ -277,6 +278,7 @@ export class {{classname}} {
         } else {
 {{#useHttpClient}}
             formParams = new HttpParams({encoder: new CustomHttpUrlEncodingCodec()});
+            append = getAppender(true);
 {{/useHttpClient}}
 {{^useHttpClient}}
             // TODO: this fails if a parameter is a file, the api can't consume "multipart/form-data" and a blob is passed.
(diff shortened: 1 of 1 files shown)


tengis:
Hi Rony,

What's outstanding for this PR?

Cheers

roni-frantchi:
Hi @tengis .
This PR was made as a suggestion or an alternative to https://github.com/swagger-api/swagger-codegen/pull/8968#issuecomment-449731308 .
Haven't heard back from those who have made the original PR, nor the core maintainers... I'm waiting for a code review, really.

In the meantime I'm seeing yet another fix here https://github.com/swagger-api/swagger-codegen/pull/9065 by @marcbuils .
Maybe between the three of us we can bring one of these PRs forward?.. I'd love a review by anyone.

modules/swagger-codegen/src/main/resources/typescript-angular/api.service.mustache
```diff
@@ -262,7 +262,8 @@ export class {{classname}} {
 {{#hasFormParams}}
         const canConsumeForm = this.canConsumeForm(consumes);
 
-        let formParams: { append(param: string, value: any): void; };
+        let formParams: FormParams;
```
emmanuelgautier: formParams could be an `HttpParams` and need to not use the return as the same way

roni-frantchi: @emmanuelgautier `FormParams` type is defined [here](https://github.com/swagger-api/swagger-codegen/pull/9012/files#diff-5bdae26ba753c4986ceb5af35bada047R364) as a union type that includes HttpParams, so I think that would work. What am I missing?
What do you mean by "and need not use the return as the same way"?

emmanuelgautier: `FormData` and `HttpParams` append function does not have the same signature :
- [HttpParams](https://angular.io/api/common/http/HttpParams#append)
- [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append)

One return FormParams and the other not. When you use `formParams.append`, you have to check what is the exact variable type. Here, you do not make this check and could be a potential error when formParams is a HttpParams and you use HttpClient

It is almost the same thing that I done in this [lines](https://github.com/swagger-api/swagger-codegen/pull/9083/files#diff-5bdae26ba753c4986ceb5af35bada047R266).

roni-frantchi: @emmanuelgautier: They differ on the return type, that is the reason why in my method, I implicitly check the type (by checking the `useReturnValue `) - 
If it's `HttpParams` I return the return value of `append()`, if it isn't, I append, and return the same instance (see the comma `,` [there](https://github.com/swagger-api/swagger-codegen/pull/9012/files#diff-5bdae26ba753c4986ceb5af35bada047R361)).

emmanuelgautier: Here, if you use HttpClient, you do not use `useReturnValue` but formParams could be a `HttpParams` and return value.

emmanuelgautier: You are right. My bad 👍

sashaaro:
is there any news? Probably it was fixed in here https://github.com/OpenAPITools/openapi-generator/commit/0a45890709c2d4e7a9ed359b35789c4a223fb04d#diff-f61af8a4c4eb2720bfb15259e0dca419

PR #9151, opened 2019-02-07
Add support for java.time.Instant for java generators

stijnkoopal:
### Description of the PR

We are using `java.time.Instant` as Date(Time) type for all our backend services. Currently we achieve this by using a custom `dateLibrary` and custom `typeMappings` and `importMappings`. This is suboptimal however.

`java.time.Instant` is a valid type for code generation if you are using UTC timestamps only. Hence, this PR.


Sarev0k:
When can this be merged?  I'd like to use this feature.

HugoMario:
thanks @stijnkoopal for this PR!

PR #7841, opened 2018-03-14
Issue 6504

jacobweber:
### PR checklist

- [X] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).

- [X] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.

After running this, there were some changes in `samples` that didn't look like they were related to my change. I only committed the relevant ones.

- [X] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.

- [X] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.
@bbdouglas @JFCote @sreeshas @jfiala @lukoyanov @cbornet @jeff9finger

### Description of the PR

Similar to issue #6504, but for okhttp-gson.

For path parameters that specify a collectionFormat, the Java code was using toString, which resulted in something like `[a,b,c]`. I added a method to format this correctly, and called it when formatting the path parameters. I only did it for okhttp-gson, but you could probably do something similar for other libraries.

Issue #6504, opened 2017-09-15
[JAVA] Retrofit2 client doesn't support path parameters that are collections


jsilland:
##### Description

When declaring a `path` parameter that is an array with a `collectionFormat` of `csv`, the Retrofit2 code generator declares the parameter as a `List` rather than a `CSVParams`. When Retrofit generates the URL for the API call, the parameter is incorrectly serialized as `[foo,%20bar]`, which is the result of calling `toString` on the list.

This is admittedly a very much of an edge case but we're trying to retrofit an OpenAPI specification over an API that has existed for years and which makes use of this. We could probably work around this issue by tweaking the endpoint of the API on the server side but this is something I would rather avoid having to do. If the OpenApi specification allows for such parameters to be declared, code generators should support them.

##### Swagger-codegen version

2.2.3

##### Swagger declaration file content or url

This is the declaration of the endpoint — some parts omitted for brevity

```json
    "/activities/{id}/streams/{stream_types}": {
      "get": {
        "operationId": "getActivityStreams",
        "parameters": [
          {
            "name": "id",
            "in": "path",
            "type": "integer"
          },
          {
            "name": "stream_types",
            "in": "path",
            "type": "array",
            "items": {
              "type": "string",
              "enum": [
                "time",
                "distance",
                "latlng"
              ]
            },
            "collectionFormat": "csv",
            "minItems": 1
          }
      }
    }
```

##### Command line used for generation

`swagger-codegen generate -i myspec.json -c config.json -l java -o java`

The relevant parts of the config are:

```json
{
  "useRxJava2": true,
  "library": "retrofit2"
}
```

##### Steps to reproduce

- Declare an endpoint with a parameter that is an array with a collectionFormat
- Configure the java generator to use the retrofit2 templates
- Run the generator against the spec and the configuration

This is the generated code

```java
@GET("activities/{id}/streams/{stream_types}")
  Observable<StreamSet> getActivityStreams(
    @retrofit2.http.Path("id") Integer id, @retrofit2.http.Path("stream_types") List<String> streamTypes, @retrofit2.http.Query("key_by_type") Boolean keyByType
  );
```
 
##### Related issues/PRs

None found

##### Suggest a fix/enhancement

For query parameters, the template to generate query parameters (`modules/swagger-codegen/src/main/resources/Java/libraries/retrofit2/queryParams.mustache`) special-cases collections:

```
[...]

PR #9155, opened 2019-02-08
6504: Java-resttemplate: Support collection formats in pathParams

victor-xplore:
### PR checklist

### Description of the PR

Solves https://github.com/swagger-api/swagger-codegen/issues/6504
But instead of Java-Retrofit2 this is for Java-resttemplates
Similar to https://github.com/swagger-api/swagger-codegen/pull/7841


jpinkham1:
I just ran into this same issue myself before I found this issue.  
My fix was going to be an unconditional change to the mustache template to call parameterToString instead of just toString on the path parameters in the line that starts with .replaceAll.   I got as far as testing it and it seemed to work OK, but I couldn't just override the one template without making the whole batch part of my codebase... so I bailed on the whole idea.

I think that might be just a bit simpler than testing for the isCollection in the template and then having one path call collectionToString and the other just toString the way I think you have it here.   But I might not have read it carefully enough...

Anyway, I was halfway down the rabbit hole and decided to just change my API to a plain string and call StringUtils.split on my incoming array in the controller and call it a day instead.   Best of luck on this if you haven't already got past it.

victor-xplore:
We used my version for a bit and then switched to the similar https://github.com/OpenAPITools/openapi-generator instead

jpinkham1:
Thanks for that suggestion.
-- Jim

PR #9208, opened 2019-02-23
suffix enum name with classname to avoid conflict

CatTail:
fix #7300

### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [ ] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

When building a client with enums, it is possible to run into collisions using Go because the enum names are in the same name space.

Issue #7300, opened 2018-01-03
[Go] 2.3.0 enum support can lead to collisions


grokify:
##### Description

When building a client with enums, it is possible to run into collisions using Go because the enum names are in the same name space.

##### Swagger-codegen version

2.3.0 using new enum support.

##### Swagger declaration file content or url

* Official copy: https://apis.voicebase.com/v3/defs/v3-api.yaml
* GitHub copy: https://github.com/grokify/go-voicebase-v3/blob/master/swagger_voicebase_v3.yaml#L40

Collision example Open API 2.0 spec excerpt:

```yaml
  VbIncludeTypeEnum:
    type: string
    enum:
    - knowledge
    - metadata
    - prediction
    - spotting
    - streams
    - transcript
  VbCallbackTypeEnum:
    type: string
    enum:
    - analytics
    - transcript
    - media
```

Both of these have an enum value, `transcript` which gets converted to the same constant `TRANSCRIPT` in the `voicebase` package: `voicebase.TRANSCRIPT`:

`vb_include_type_enum.go`

```go
package voicebase

type VbIncludeTypeEnum string

// List of VbIncludeTypeEnum
const (
	KNOWLEDGE VbIncludeTypeEnum = "knowledge"
	METADATA VbIncludeTypeEnum = "metadata"
	PREDICTION VbIncludeTypeEnum = "prediction"
	SPOTTING VbIncludeTypeEnum = "spotting"
	STREAMS VbIncludeTypeEnum = "streams"
	TRANSCRIPT VbIncludeTypeEnum = "transcript"
)
```

`vb_callback_type_enum.go`

```go
package voicebase

type VbCallbackTypeEnum string

// List of VbCallbackTypeEnum
const (
	ANALYTICS VbCallbackTypeEnum = "analytics"
	TRANSCRIPT VbCallbackTypeEnum = "transcript"
	MEDIA VbCallbackTypeEnum = "media"
)
```

Compiling the client results in the following errors:

```bash
../../vb_include_type_enum.go:22:33: TRANSCRIPT redeclared in this block
	previous declaration at ../../vb_callback_type_enum.go:18:34
../../vb_task_status_enum.go:17:30: ACCEPTED redeclared in this block
	previous declaration at ../../vb_status_enum.go:17:26
../../vb_task_status_enum.go:18:28: FAILED redeclared in this block
	previous declaration at ../../vb_status_enum.go:18:24
[...]

PR #9234, opened 2019-03-05
[go] suffix enum name with classname to avoid conflict

frantuma:
refs #7300 - includes and replaces #9208 (rebase and updating samples cleaning up previous code)


b8kings0ga:
Im using the newest version but still running to this issue, what should I have for the option to use this patch?

PR #9278, opened 2019-03-19
Adding proxy header options for configuration and rest client

chowmean:
Added proxy headers option for urllib3. Added this in configuration object and Rest Client Object. 
This is regarding a PR in Kubernetes client and addresses this issue. https://github.com/kubernetes-client/python/issues/735. And related to this PR https://github.com/kubernetes-client/python/pull/779


vikramkute:
Any updates on this?

johnhodorowicz:
Any update here? Without this change we're unable to use a proxy - this is needed to be added each time we generate/regenerate an API.

PR #9289, opened 2019-03-21
Add option to generate overloaded methods for java client

### LINKED ISSUE #8939: Swagger generated Java Client doesn't have overloaded calls, even though the query parameter is marked as optional
I'm introducing an **optional**  "query parameter" (**with default value**) for one of my API endpoint ( i.e. /foo2) but hoping that existing API consumers don't have to make any changes in their application when they regenerate the java client stub from the new swagger file.

I was expecting that the generated java client will offer overloaded methods, to operate with and without a method parameter, so that my existing API consumers can work without any changes to their call. At the same time, they can switch to its more generic offering.

    public void foo2Get() throws ApiException {
    public void foo2Get(String type) throws ApiException {

**But it generates ONLY the latter**. I don't see the former.

At the same time, generated python client from same swagger throws a var arg and this way it more flexible and hence old and new offerings will work without any issues.

    def foo_get(self, **kwargs): 

(swagger sample used: attached)

##### Description
Is there a way to overcome this limitation with java code generator.

##### swagger-generator version

swagger 2.0

[sample.zip](https://github.com/swagger-api/swagger-codegen/files/2606716/sample.zip)

Hervian:
### Description of the PR
Fixes #8939 

This PR introduced a new option `methodsPerHttpRequestStrategy` to the java client codegeneration library (JavaClientCodegen.java).
When a call to create a java client library is made with the additional-property:
` "methodsPerHttpRequestStrategy": "overloading"`   
the logic will now not only create 1 method but instead several. Specifically, it will create 1 method which only accepts the required parameters (and if none, a method with no parameters) another method, which accepts the required parameters + the first encountered optional parameter, another method which accepts the required parameters + the first two optional parameters and so on. 
Please notice that the implementation is backwards compatible in two senses: the logic defaults to behave as before this option was introduced. Also, setting the new option to 'overloading' creates, among others, a signature which makes overloaded client generated code backwards compatible for users of that autogenerated library.

## Purpose/what this solves:
Currently adding an optional parameter to the swagger doc cause compile time errors when recompiling the java client code, since calls to this code is no longer valid due to the extra parameter being added. Choosing this overloading option solves this.

## Testing
Two unit tests asserts calls the generation logic and asserts that the generated class contains the expected methods. See `DefaultGeneratorTest.testJavaClientGeneration_useOverloading` and `DefaultGeneratorTest.testJavaClientGeneration_dontUseOverloading`.


Hervian:
Works locally, but the swagger codegen Jenkins server must have some peculiar setup that prevents Mapstruct from working.

Unable to make the build pass without swagger codegen team looking into why the build server will not accept autogenerated code from mapstruct.

Details: The new class OperationMapper is an interface marked with a special Mapstruct annotation that makes Mapstructs annotation processor pick it up and generate an implemenation of the interface compile time.
This works as expected when building locally, but does not work on the build server.

Hervian:
Please notice that I had to replace MapStruct with ModelMapper. Why the Jenkins server did not work with MapStruct I do not know, but I would suggest that someone look into it, since MapStruct is a better mapping framework in my opinion.

tryyebaobao:
hi, can I ask you does any codegen plugin version picks up your fix, this is exactly what I am looking for. could you please let me know how I can get this fix and how to use it? 

right now, I am using the plugin as

                <groupId>io.swagger</groupId>
                <artifactId>swagger-codegen-maven-plugin</artifactId>
                <version>2.3.1</version>

O330oei (commented):
Cont.

yapily-mofe:
Is there any update on this feature?

rogiller:
When do you think this PR could merge? It would reeeeealy nice to be able to introduce new Rest Api optional params and not break code that runs against the generate Java Client!!

Issue #7747, opened 2018-03-01
[JAVA][RestTemplate] Client: Plus character in query parameter not encoded


steffen-harbich-itc:
##### Description

I have the problem that in the generated client (language Java, library resttemplate) when I pass a plus character within a string query parameter then the client will send the plus unescaped to the server which interprets it as space. See also my [StackOverflow question](https://stackoverflow.com/questions/48906034).

##### Swagger-codegen version

CLI version 2.3.1

##### Swagger declaration file content or url

```yaml
swagger: '2.0'
info:
  title: Test for Issue
  version: v1
host: 'localhost:8080'
schemes:
  - http
basePath: /
tags:
  - name: test
paths:
  /test:
    get:
      tags:
        - test
      operationId: testMe
      produces:
        - application/json
      parameters:
        - name: filter
          in: query
          type: string
      responses:
        '200':
          description: OK
```

##### Command line used for generation

From my gradle script task:
```
args = ['generate',
            '-i', "${projectDir}/src/main/resources/swagger.yaml",
            '--api-package', 'com.example.demo.client.api',
            '--model-package', 'com.example.demo.client.model',
            '--invoker-package', 'com.example.demo.client',
            '--additional-properties', 'dateLibrary=java8',
            '--remove-operation-id-prefix',
            '--lang', 'java',
            '--library', 'resttemplate',
            '-o', "${buildDir}/generatedJava"]
```

##### Steps to reproduce

I made a minimal Spring boot 2.0 example with gradle including client generation:
[demo.zip](https://github.com/swagger-api/swagger-codegen/files/1770357/demo.zip)

To execute client generation start task "generateRestClient_sync". Call task "test" to start the REST server and the test will invoke the REST api via the generated client (see test class). Once the REST controller method is called, it prints the query parameter in log. There you can see that it replaced the + by space.

##### Related issues/PRs

none

##### Suggest a fix/enhancement

I assume that this is a bug of Spring's RestTemplate itself because it does escape all other characters but the plus sign. In my StackOveflow question, one answerer recommended changing the template but in my opinion it should be fixed not just for me. When I use the generated client I expect to not take care of any encoding issues. I can try to propose a fix once you confirm this is a bug.

steffen-harbich-itc:
Additionally, when I have a URL path variable and it contains a slash or colon then it won't be encoded by client, too. Is the generated client code expecting me to encode all parameters myself?

PR #9387, opened 2019-04-18
URL parameters in GET request would not ge properly url-encoded

kufi:
The + sign would not get encoded, but decoded, resulting in a space, which in turn resulted in invalid parameters on the server side.

This should fix this issue https://github.com/swagger-api/swagger-codegen/issues/7747.

As the encode() in RestTemplate only works when using expanded variables (https://github.com/spring-projects/spring-framework/issues/21577), instead of directly adding the parameters to the queryParams, this creates variables which are expanded for all inputted query params.


kufi:
Ok, so apparently this cannot be achieved in Sprint 4.3.9 (or any Spring 4 version for that matter). So, can the Rest-Template be updated to Spring 5? Or is this not wanted, as it should work with Spring 4?

kufi:
So, instead of trying to make it backwards-compatible I added a new config-parameter "spring5". If activated it generates the demo client with spring5. Per default it is deactivated in order to stay backwards compatible.

mangatmodi:
Hey guys, 
Any decision on this one, This is really important and breaks many things.

durcon:
Why is this fix after 5 years still not merged? Who can merge this pull request?

PR #9465, opened 2019-05-26
Support for current versions of RestSharp (Swagger 2.0 spec)

ulfsa:
Tried using RestSharp 106.6.9.

Without this fix, the build fails (wrong method signature)


mandrean (approved):
LGTM 👍

PR #9479, opened 2019-06-05
Java resttemplate - Registrer correct jackson module when using java 8 or joda-time

aaitmouloud:
### Description of the PR
Small change using the same logic as for `threetenbp`, registrer the correct Jackson modules (resp. `JodaModule` or `JavaTimeModule`) when using `joda-time` or `java8` as a time library.


user667:
Also in need of this fix because otherwise config option ```<dateLibrary>java8</dateLibrary>``` is ignored.

PR #9490, opened 2019-06-10
[JavaSpring] Fixed Issue 9250 - Codegen for file datatype

### LINKED ISSUE #9250: [spring] @RequestPart should use "{{basename}}" instead of "file" and observe "required" attribute
##### Description

When generating a spring-boot server for a POST processing multipart form data to upload one or several files, then the generated @RequestPart annotation always uses the fixed string "file" instead of the name of the part. Additionally, the required attribute is not considerd when building the @RequestPart annotation, so that file upload parameters are always required even if not defined to be so in the Swagger yaml file.

##### Swagger-codegen version

swagger-codegen-cli-2.4.2.jar

If you post the code inline, please wrap it with
```yaml
  /kpack/jsonld:
    post:
      operationId: "uploadJsonld"
      consumes:
        - multipart/form-data
      produces:
        - application/json
      parameters:
        - name: jsonld
          in: formData
          type: file
          description: "JSON-LD graph file"
          required: true
```

```bash
java -jar ./swagger-codegen-cli.jar generate \
	-i api.yaml \
	-l spring 
```

##### Steps to reproduce

Create code with command given above. Check generated Api and Controller.

In https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/JavaSpring/formParams.mustache change the fixed "file" and replace it by "{{basename}}". Additionally fill the "required" attribute of this annotation.

MGoyal92:
Updated mustache file to populate correct code inside @ApiParam annotation and inside @RequestParam annotation, and added "required" attribute (true|false) inside @RequestParam Annotation.

This should fix #9250


scara:
Hi @HugoMario and @frantuma,
is there any chance to speed up the review of this PR?

Small but **actually valuable**, to address the proper mapping of a multipart POST.

TIA,
Matteo

frantuma:
Thanks!

PR #9568, opened 2019-07-12
clearing temp directories from CodegenIgnoreProcessorTest

oleksabor:
### Description of the PR

seems that

````Java
@AfterTest
    public void afterTest() throws IOException {
````

is executed only once. But there are a lot of `CodegenIgnoreProcessorTest` instances. This causes only one temp directory to be deleted. I see a lot of test temp directories like `CodegenIgnoreProcessorTest3682269276225257646` after test is finished.

It would be nice to clean those folders
I've found only one way - to add static queue and poll it in the `afterTest`

I've tried it locally and found that all CodegenIgnoreProcessorTest temp folders are removed.


HugoMario:
thanks a lot @oleksabor !!

Issue #9203, opened 2019-02-21
[Java] Bug generating Boolean type property


bwangfdu:
##### Description

When generate model object with Boolean type property, the get method is generated with duplicated `is` as `isisPropertyName`

##### Swagger-codegen version

3.0.1, 3.0.4, 3.0.5, others 3.0.x versions doesn't test.

##### Swagger declaration file content or url

```
"OrderLineAudit": {
      "type": "object",
      "properties": {
        "id": {
          "format": "uuid",
          "type": "string"
        },
        "completeTS": {
          "type": "string"
        },
        "eventId": {
          "type": "string"
        },
        "processed": {
          "type": "boolean"
        }
      }
    },
```
The processed property get method generated as `isisProcessed`
##### Command line used for generation

swagger-codegen-maven-plugin

##### Steps to reproduce

run generate goal

frantuma:
fixed in swagger-api/swagger-codegen-generators#458

PR #9687, opened 2019-08-30
Test for Bug generating Boolean type property

gracekarina:
### Description of the PR
test for issue in  swagger-codegen-generators #9203
depends on https://github.com/swagger-api/swagger-codegen-generators/pull/456


frantuma:
replaced by #9697

Issue #8803, opened 2018-10-12
[Rust] Allow to choose client library between hyper and reqwest


bcourtine:
##### Description

Rust client codegen uses hyper library. This library is "low-level", and requires some tunning (to deal with https, for example).
Reqwest alternative https://github.com/seanmonstar/reqwest (which uses hyper under the hood), seems to be a good alternative.

##### Swagger-codegen version

master

##### Swagger declaration file content or url

Not applicable

##### Command line used for generation

Not applicable

##### Steps to reproduce

Not applicable

##### Related issues/PRs

Not found

##### Suggest a fix/enhancement

PR in preparation.

ctaggart:
Any reason why this wasn't reviewed and merged? Pull request #8804 has been open for 11 months. It would be great a reqwest option and may be a [surf](https://github.com/rustasync/surf) option for clients too.

bcourtine:
Hi @ctaggart.

Nobody looked at this PR here.
But I submitted the same PR on the Swagger-Codegen fork https://github.com/OpenAPITools/openapi-generator

Fork project is much more active, and the PR was merged a long time ago (with other stuff since this time). So, if you need a Rust client, I can suggest to use OpenAPI generator: Rust generator has more features (including a Reqwest based client).

PR #9699, opened 2019-09-06
upgrade to hyper 0.12 for rust (client)

ctaggart:
### PR checklist

### Description of the PR

I'm looking to upstream these changes for upgrading to hyper 0.12. It is a subset of the changes we have in our private repo. I haven't done testing yet on this subset without the rest of our changes. I'm still learning rust and there may be some things I didn't understand in the original templates, but that is why I'm opening a pull request.


modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -28,24 +21,29 @@ impl<C: hyper::client::Connect> {{{classname}}}Client<C> {
 pub trait {{classname}} {
 {{#operations}}
 {{#operation}}
-    fn {{{operationId}}}(&self, {{#allParams}}{{paramName}}: {{#isString}}&str{{/isString}}{{#isUuid}}&str{{/isUuid}}{{^isString}}{{^isUuid}}{{^isPrimitiveType}}{{^isContainer}}::models::{{/isContainer}}{{/isPrimitiveType}}{{{dataType}}}{{/isUuid}}{{/isString}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) -> Box<Future<Item = {{^returnType}}(){{/returnType}}{{#returnType}}{{{returnType}}}{{/returnType}}, Error = Error<serde_json::Value>>>;
+    fn {{{operationId}}}(&self, {{#allParams}}{{paramName}}: {{#isString}}&str{{/isString}}{{#isUuid}}&str{{/isUuid}}{{^isString}}{{^isUuid}}{{^isPrimitiveType}}{{^isContainer}}::models::{{/isContainer}}{{/isPrimitiveType}}{{{dataType}}}{{/isUuid}}{{/isString}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) -> Box<dyn Future<Item = {{^returnType}}(){{/returnType}}{{#returnType}}{{{returnType}}}{{/returnType}}, Error = Error>>;
```
ctaggart: I got rid of the generic of `Error<T> in our private repo and replaced it with just `Error`. I will either need to add those changes or change this back to `Error<serde_json::Value>`.

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -28,24 +21,29 @@ impl<C: hyper::client::Connect> {{{classname}}}Client<C> {
...
 }
 
-
-impl<C: hyper::client::Connect>{{classname}} for {{classname}}Client<C> {
+impl<'a, C: hyper::client::connect::Connect>{{classname}} for {{classname}}Client<'a, C> where C: 'static {
 {{#operations}}
 {{#operation}}
-    fn {{{operationId}}}(&self, {{#allParams}}{{paramName}}: {{#isString}}&str{{/isString}}{{#isUuid}}&str{{/isUuid}}{{^isString}}{{^isUuid}}{{^isPrimitiveType}}{{^isContainer}}::models::{{/isContainer}}{{/isPrimitiveType}}{{{dataType}}}{{/isUuid}}{{/isString}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) -> Box<Future<Item = {{^returnType}}(){{/returnType}}{{#returnType}}{{{.}}}{{/returnType}}, Error = Error<serde_json::Value>>> {
-        let configuration: &configuration::Configuration<C> = self.configuration.borrow();
+    fn {{{operationId}}}(&self, {{#allParams}}{{paramName}}: {{#isString}}&str{{/isString}}{{#isUuid}}&str{{/isUuid}}{{^isString}}{{^isUuid}}{{^isPrimitiveType}}{{^isContainer}}::models::{{/isContainer}}{{/isPrimitiveType}}{{{dataType}}}{{/isUuid}}{{/isString}}{{#hasMore}}, {{/hasMore}}{{/allParams}}) -> Box<dyn Future<Item = {{^returnType}}(){{/returnType}}{{#returnType}}{{{.}}}{{/returnType}}, Error = Error>> {
```
ctaggart: Same as above. Either add the `Error` changes, or change back to `Error<serde_json::Value>`.

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -120,27 +122,25 @@ impl<C: hyper::client::Connect>{{classname}} for {{classname}}Client<C> {
...
 
         // send request
         Box::new(
-        configuration.client.request(req)
+        self.configuration.client.request(req)
             .map_err(|e| Error::from(e))
             .and_then(|resp| {
                 let status = resp.status();
-                resp.body().concat2()
+                resp.into_body().concat2()
```
ctaggart: This is a key part of the upgrade.

modules/swagger-codegen/src/main/resources/rust/api.mustache
```diff
@@ -1,24 +1,17 @@
...
+use super::{Error, configuration, client};
 
-pub struct {{{classname}}}Client<C: hyper::client::Connect> {
-    configuration: Rc<configuration::Configuration<C>>,
+pub struct {{{classname}}}Client<'a, C: hyper::client::connect::Connect> {
+    configuration: &'a configuration::Configuration<C>,
 }
 
-impl<C: hyper::client::Connect> {{{classname}}}Client<C> {
-    pub fn new(configuration: Rc<configuration::Configuration<C>>) -> {{{classname}}}Client<C> {
```
ctaggart: I don't understand why `Rc` was being used. I updated it to be a plain reference.

ctaggart:
I'll run the script for the Petstore sample next week. I'm not sure who would be best to review, but may be @euank or @wing328.

ctaggart:
I run `bin/rust-petstore.sh` and the updated sample compiles and is committed.

modules/swagger-codegen/src/main/resources/rust/api.mustache
Luro02: The `Rc` was there to omit the lifetime of the struct. An `Rc` value isn't bound by any lifetimes and your change could cause lifetime issues, if the referenced config gets out of scope. I would suggest using an [Arc](https://doc.rust-lang.org/std/sync/struct.Arc.html)

ctaggart: Thank you @Luro02 for reviewing. I'm still a little fuzzy on the problems this may change. I read more about [Rc](https://doc.rust-lang.org/1.30.0/book/2018-edition/ch15-04-rc.html). It enables multiple ownership and I could use an example where that is useful here. Our use case is simple, we create a config and pass a reference in to create any of the clients. In this case `APIClient` creates all the other clients, passing the reference.
``` rust
let apicli = crate::apis::client::APIClient::new(&config);
```

wing328:
@ctaggart thanks for tagging me but I'm no longer involved in this particular project. I hope others will be able to help you out. Good luck.

ctaggart:
I see. Looks like we should look at https://github.com/OpenAPITools/openapi-generator instead for our Rust generators [as recommended](https://github.com/swagger-api/swagger-codegen/issues/8803#issuecomment-529965915).

modules/swagger-codegen/src/main/resources/rust/api.mustache
Luro02:
```rust
fn create_client() -> Client {
    let config = Config::new();
    let client = Client::new(&config);
    client
} // config gets dropped here

fn main() {
    let client = create_client();
}
```
There are 2 solutions, either return the Config + Client or wrapping the config in a Arc/Rc. I think the idiomatic solution would be to let Client take ownership of the Config and then internally wrapping it in an Arc/Rc.

Issue #9685, opened 2019-08-30
Swagger CodeGen APIClient.java not returning the headers from the rest response by calling resttemplate


ayushaccenture:
##### Description

Swagger CodeGen APIClient.java not returning the headers from the rest response by calling resttemplate
It is returning only the response body. Please have a look into this.
##### Swagger-codegen version

2.4.5

##### Swagger declaration file content or url

 addHeadersToRequest(headerParams, requestBuilder);
        addHeadersToRequest(defaultHeaders, requestBuilder);
        
        RequestEntity<Object> requestEntity = requestBuilder.body(selectBody(body, formParams, contentType));

        ResponseEntity<T> responseEntity = restTemplate.exchange(requestEntity, returnType);
        
        statusCode = responseEntity.getStatusCode();
        responseHeaders = responseEntity.getHeaders();

        if (responseEntity.getStatusCode() == HttpStatus.NO_CONTENT) {
            return null;
        } else if (responseEntity.getStatusCode().is2xxSuccessful()) {
            if (returnType == null) {
                return null;
            }
            return responseEntity.getBody();

thomasJoei:
Created a PR for this fix.
#39737

Issue #9222, opened 2019-03-01
[Java] - RestTemplate library - generated client thread safety


korlat:
##### Description

This is related to an old issue which was closed with documentation improvements: #1435.
A fix was applied to the Jersey library, for some parts of the thread safety issues - response headers, and response status codes: #7453, #7605

I would like to fix this issue for the RestTemplate library

##### Swagger-codegen version

Using 2.4.2 but applies to the current Master (HEAD) as well

##### Swagger declaration file content or url

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/Java/libraries/resttemplate/ApiClient.mustache#L95

##### Command line used for generation

Any - N/A

##### Related issues/PRs

#1435
#7453
#7605

##### Suggest a fix/enhancement

I am happy to provide a PR for the issue that perform the same changes as PR: #7605 but applied to the RestTemplate library.

Another option I was considering was to create a PR as above but also to go a step further and address some of the other thread safety issues raised in #1435.

This would include making the following fields immutable:

- debugging
- defaultHeaders
- basePath
- restTemplate
- authentications
- dateFormat

That would involve creating a builder class as suggested in #1435

Weather this would move the RestTemplate library to far from the other Java libraries may be an issue, thoughts?

Note: Either of the above approaches would be non backward compatible as #7605 was.

kmurthy0:
Subscribed - I'm interested in this change as well.

mjeffrey:
See https://github.com/OpenAPITools/openapi-generator/issues/2280 for a workaround.

thomasJoei:
Hi @korlat I created a PR regarding the response headers/status thread safety.
https://github.com/swagger-api/swagger-codegen/pull/9737

I might open another PR to create the builder class and enable users to provide thread safe `Authentication`, which something I had to work around in my current project, but it depends on my availabilities. 
Anyway you will at least be able to safely retrieve headers and status with `RestTemplate` once https://github.com/swagger-api/swagger-codegen/pull/9737 is merged :)

HugoMario:
hello @thomasJoei , please, ping me once you want your PR gets merged.

thomasJoei:
@HugoMario All checks have passed, PR is up to date with `master`. 
I don't know if you want to ping some other contributors to review it, but on my side it is ready to be merged :)

HugoMario:
great, i can review :)

sbilello:
Any update on this?

[...]

PR #7605, opened 2018-02-07
Java Jersey2: Make generated client code thread safe

tsiq-karold:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [x] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [x] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [x] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

This PR is to make the generated Jersey 2 client thread safe.  Currently, the only way to get to response headers and status is by inspecting some fields on the ApiClient which is not thread safe.  This PR changes the generated code to follow a similar pattern used by other generated clients with "withHttpInfo" flavours of each method.  It is backwards compatibility breaking as it removes the existing header and status fields.

Fixes https://github.com/swagger-api/swagger-codegen/issues/7453

cc @bbdouglas @JFCote @sreeshas @jfiala @lukoyanov @cbornet @jeff9finger

PR #9737, opened 2019-09-28
[Issue 9685] Make RestTemplate thread safe by using the withHttpInfo pattern used …

thomasJoei:
…by many other generated clients

### PR checklist

### Description of the PR

This PR is to make the generated **RestTemplate** client thread safe. Currently, we don't have any way of retrieving the response headers except by using  `getResponseHeaders()` method which returns the latest response headers, in the case multiple threads use the ApiClient we might end up getting the headers of the response sent to another thread.

This PR purpose is to add an additional method which not only return the expected DTO but also the `ResponseEntity` wrapping it, so user can easily retrieve headers by inspecting the `ResponseEntity`. The usual methods to directly get the DTO are conserved.

**This change breaks backward compatibility since (thread unsafe) attributes, `statusCode` and `responseHeaders`, are removed from ApiClient class.**

Fixes: #9685, #9222

PR mostly inspired by @tsiq-karold work on Jersey2 (https://github.com/swagger-api/swagger-codegen/pull/7605), and `withHttpInfo` pattern used in other generated clients.


HugoMario:
LGTM, thanks a lot @thomasJoei !!!

sliard:
We need the same for v3 :(

PR #9754, opened 2019-10-07
Fix dateLibrary=java8 handling in JacksonConfig.mustache #9753

### LINKED ISSUE #9753: [Java][JavaResteasyServerCodegen] dateLibrary=java8 is ignored and joda added instead
##### Description

When generating server code via ```io.swagger.codegen.v3.generators.java.JavaResteasyServerCodegen``` using ```dateLibrary=java8``` then ```com.fasterxml.jackson.datatype.joda.JodaModule``` and some is added to the ```JacksonConfig``` class instead of ```com.fasterxml.jackson.datatype.jsr310.JavaTimeModule```. This causes compilation to fail.

##### Swagger-codegen version

This happens when using version 3.* (3.0.12-SNAPSHOT) and 2.4.* (2.4.9-SNAPSHOT).

Any.

Using Maven to generate the code via https://gist.github.com/user667/57dd1c96de0fbaeacf850a99f9cd0749

##### Steps to reproduce

Simply generate the code via Maven from any valid YAML Swagger declaration file.

This problem stems from the fact that the mustache template files (```./modules/swagger-codegen/src/main/resources/JavaJaxRS/resteasy/JacksonConfig.mustache``` for version 2.4.* and ```swagger-codegen-generators/src/main/resources/handlebars/JavaJaxRS/resteasy/JacksonConfig.mustache``` in project generators for version 3.0.*) do not contain the switch for java8/joda.

user667:
Previously config option dateLibrary=java8 was ignored and joda
was added instead, causing compilation failure.

Fix #9753

### Description of the PR

See above.

### DIFF
--- a/modules/swagger-codegen/src/main/resources/JavaJaxRS/resteasy/JacksonConfig.mustache
+++ b/modules/swagger-codegen/src/main/resources/JavaJaxRS/resteasy/JacksonConfig.mustache
@@ -5,10 +5,15 @@ import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+{{#java8}}
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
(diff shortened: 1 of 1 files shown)


frantuma:
Thanks! included and replaced by #9760

Issue #9785, opened 2019-10-16
[typescript-inversify] generates unusable code


michalzubkowicz:
##### Description
typescript-inversify generates unusable code, with TypeScript warnings, also services are injected in a way that doesn't work with current version of inversify:
https://github.com/inversify/InversifyJS/issues/1004#issuecomment-541392992

##### Swagger-codegen version

2.4.9

##### Command line used for generation

swagger-codegen-cli.jar -l typescript-inversify

##### Suggest a fix/enhancement

1) Move DI classses from contructor to properties (api.service.mustache)
2) Replace type casting from <> to "as" (TS recommend that)
3) Replace == with === 
4) Remove redundant local variable basePath. 
5) fix quotemarks

HugoMario:
covered by #9786 , so closing. thanks again @michalzubkowicz for your help.

PR #9786, opened 2019-10-16
#9785

michalzubkowicz:
#9785 

Moved DI classses from contructor to properties (api.service.mustache)
Removed type casting when not needed (TS showed warning about that)
Replaced == with ===
Removed redundant local variable basePath.
fixed quotemarks


HugoMario:
sorry for delay @michalzubkowicz and thanks a lot for PR!!

Issue #9871, opened 2019-11-18
[typescript-fetch] Throws on HTTP 3xx responses


BernhardPosselt:
The generated code throws the response on HTTP 3xx responses such as not modified. Instead it should just throw on non 2xx and 3xx responses

Affects all versions

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/typescript-fetch/api.mustache#L267

PR #9873, opened 2019-11-18
wrap response in exception class, do not throw on http 3xx

### LINKED ISSUE #9872: [typescript-fetch] Exceptions impractical to catch
Since the response is thrown without wrapping it in a separate exception class you can not catch it in a try block but must resort to checking if certain fields exist on the object. That can also lead to false positives.

Instead it should be wrapped in some sort of ApiException class that also has the HTTP status.

https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/typescript-fetch/api.mustache#L270

BernhardPosselt:
### PR checklist

### Description of the PR
Fixes #9872 and #9871

@CodeNinjai


BernhardPosselt:
The sample files do not seem to be updated by rerunning the ./bin/typescript-fetch-petstore-all.sh script, any ideas?

PR #10002, opened 2020-01-22
[python] Cleanup ThreadPool with atexit rather than __del__

### LINKED ISSUE #9991: [Python] Generated code is out of contract - application hangs indefinitely
The generated code snippet from - https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/python/api_client.mustache#L73 is not allowed:

```python
    def __del__(self):
        if self._pool is not None:
            self._pool.close()
            self._pool.join()
```

The very simple program of creating a global client, and then just exiting will hang indefinitely on Python3.8. Initially raised as CPython bug under https://bugs.python.org/issue39360, but @pablogsal (core CPython developer) investigated and concluded that this way of cleaning up pools is not supported. Instead one should either allow the user to close or if the user did not do the generated code should do by registering https://docs.python.org/3.8/library/atexit.html callbacks.

The code in question seemed to be added within https://github.com/swagger-api/swagger-codegen/pull/6396 by @tuxbotix with approvers being @scottrw93 @taxpon @frol @mbohlool, so tagging them for context.

fabianvf:
### PR checklist

### Description of the PR

This removes the `__del__` function from the generated Python client,
and replaces it with a `cleanup` function. When a ThreadPool is created,
the cleanup function is registered with the `atexit` module.

This fixes #9991, where the API client could hang indefinitely at
garbage collection.

cc kenjones-cisco


fabianvf:
@gaborbernat that might be the fastest review I've ever gotten, got it in while I was writing up the PR description. I think I addressed your comments

gaborbernat:
I feel like maybe instead cleanup should be close and we should now also allow to call the client with the with statement, so it may cleanup automatically whenever the user no longer needs it.

fabianvf:
@gaborbernat renamed cleanup to close, think making it a context manager might be a bit out of scope for this bugfix but I can make a followup PR for it

anichari:
Hello, has this been merged and available in the latest codegen packages? I tried a few things with codegen versions but I still dont see this change in the api_client.py generated files.

1. I used 2.0.14-SNAPSHOT (built from master) and with that code, I still see the lazy init of ThreadPool rather than the newer code that creates the ThreadPool in __init__.
2. I used 3.0.19 (brew installed) and with that, the __del__ method is still there, no sign of above changes to api_client.py in that either.

gaborbernat:
Nope, we've moved on to use https://github.com/OpenAPITools/openapi-generator/pull/5094 instead, that contains the fix. Swagger never addressed this issue.

anichari:
> Nope, we've moved on to use [OpenAPITools/openapi-generator#5094](https://github.com/OpenAPITools/openapi-generator/pull/5094) instead, that contains the fix. Swagger never addressed this issue.

Thx much

Issue #10318, opened 2020-06-12
[JAVA] Bug generating a client with openfeign


tseale07:
##### Description

When attempting to generate a java client with open feign, the following error occurs when attempting to compile the code:

```
     [exec] [INFO] -------------------------------------------------------------
     [exec] [ERROR] COMPILATION ERROR :
     [exec] [INFO] -------------------------------------------------------------
     [exec] [ERROR] /Users/tseale/Documents/code/work/dapi-configuration/target/generated-sources/swagger/java/src/main/java/io/swagger/api/DefaultApiClient.java:[3,47] package org.springframework.cloud.openfeign does not exist
     [exec] [ERROR] /Users/tseale/Documents/code/work/dapi-configuration/target/generated-sources/swagger/java/src/main/java/io/swagger/api/DefaultApiClient.java:[6,2] cannot find symbol
     [exec]   symbol: class FeignClient
     [exec] [INFO] 2 errors
```

##### Swagger-codegen version

3.0.20, 2.4.14

##### Pom configuration used for generation

```
<plugin>
    <groupId>io.swagger</groupId>
    <artifactId>swagger-codegen-maven-plugin</artifactId>
    <version>${swagger-codegen-maven-plugin.version}</version>
    <executions>
        <execution>
            <id>generate-swagger-client-java</id>
            <phase>compile</phase>
            <goals>
                <goal>generate</goal>
            </goals>
            <configuration>
                <inputSpec>${project.build.directory}/swagger.json</inputSpec>
                <output>${swagger.codegen.output.dir.java}</output>
                <language>spring</language>
                <library>spring-cloud</library>
                <configOptions>
                     <generateForOpenFeign>true</generateForOpenFeign>
                    <groupId>${project.parent.groupId}</groupId>
                    <artifactId>${project.name}-java-client</artifactId>
                    <artifactVersion>${project.version}</artifactVersion>
                </configOptions>
            </configuration>
        </execution>
    </executions>
</plugin>
```

##### Steps to reproduce

1. Set the version to 2.4.14
2. Set the `language` to `spring` and the `library` to `spring-cloud`
3. Set the `generateForOpenFeign` option to `true`
4. auto generate the client with the above configurations
5. run the command `mvn clean install` in the auto generated project

##### Related issues/PRs

This is related to the support of openfeign in the 2.4.14 release: https://github.com/swagger-api/swagger-codegen/pull/9120

##### Suggest a fix/enhancement

[...]

PR #10319, opened 2020-06-12
fix(java): use correct dependency for spring-cloud openfeign [10318]

tseale07:
### Description of the PR

Fix for Issue https://github.com/swagger-api/swagger-codegen/issues/10318

the dependency `spring-cloud-starter-feign` was being used instead of `spring-cloud-starter-openfeign` when trying to autogen a client with openfeign.

This update will utilize the `isOpenFeign` boolean to include the correct dependency for openfeign


jfairley:
@tseale07 Your branch is out of date.

PR #10327, opened 2020-06-18
fix for security header html2 generator

gracekarina:
### Description of the PR
The HTML2 documentation does not include the header `Authorization: Bearer` (Bearer Authentication) in the curl requests like it does for Basic Authentication or API Keys, for example.

Depends on: 
https://github.com/swagger-api/swagger-codegen-generators/pull/702


JonathanParrilla:
please build

PR #10346, opened 2020-06-25
Fix details about template engine and location

UnHumbleBen:
The previous template engine `mustache` and file location don't apply for version 3.


HugoMario:
thanks @UnHumbleBen !

Issue #10444, opened 2020-09-01
[RUBY] Stop using deprecated `URI.encode`


kitop:
##### Description
Ruby's `URI.encode` has been [deprecated over 10 years ago](https://github.com/ruby/ruby/commit/238b979f1789f95262a267d8df6239806f2859cc). In addition, Ruby 2.7 now shows the warning in non-verbose mode, resulting in very noisy logs.

##### Swagger-codegen version
2.4.15, and previous

##### Swagger declaration file content or url
N/A

##### Command line used for generation
```
./run-in-docker.sh /gen/bin/ruby-petstore.sh
```

##### Steps to reproduce
Generate the Ruby code, see `URI.escape` used in `api_client.rb` and `configuration.rb` files

##### Related issues/PRs
Couldn't find.

##### Suggest a fix/enhancement
`URI.encode` is a very simplistic percent-encoding replacement. This is not always correct, as it may vary on differenet parts of the URI.
We could copy the code from `URI.escape` to a `Utils` module, or inline it in each class, but we would still be doing a non-optimal encoding.
Another solution is to add a dependency on the [`addressable`](https://github.com/sporkmonger/addressable) gem, and use `Addressable::URI.encode` which is part-aware. It adds one small dependency, but the escaping would be much better. That gem is also used by a lot of other gems, so might not be a big impact in the end user application.

kitop:
Closed via https://github.com/swagger-api/swagger-codegen/pull/10445

PR #10445, opened 2020-09-01
[RUBY] Replace deprecated `URI.encode` to support ruby 2.7 without warnings

kitop:
### PR checklist

### Description of the PR
Issue: https://github.com/swagger-api/swagger-codegen/issues/10444

Ruby's URI.encode has been deprecated over 10 years ago. In addition, Ruby 2.7 now shows the warning in non-verbose mode, resulting in very noisy logs.

This tackles the problem via the second solution proposed in the issue.
It adds a dependency on the addressable gem, and uses Addressable::URI.encode which is part-aware. 

Already tried this in a generated library with success: https://github.com/netlify/zuora-ruby-client/pull/7/files

/cc @zlx


kitop:
Tests are failing due to un unrelated change in Typhoeus: https://github.com/typhoeus/typhoeus/pull/575
`Typhoeus::Response::Header` no longer inherits from Hash, but now is a `DelegateClass(Hash)`. And since it's checking the type in https://github.com/kitop/swagger-codegen/blob/5236247dce2ac814a71a031763aa78959d42b424/samples/client/petstore/ruby/spec/pet_spec.rb#L82 it fails.

An option would be to change that check, or another one, to call `to_h` here: https://github.com/kitop/swagger-codegen/blob/5236247dce2ac814a71a031763aa78959d42b424/samples/client/petstore/ruby/lib/petstore/api_client.rb#L66

Thoughts?

kitop:
Pushed the second option to have the test passing. Let me know if you'd prefer an alternative, or a separate PR.

HugoMario:
Thanks a lot for PR @kitop!! Here is fine :)

PR #10464, opened 2020-09-17
Fix code generation for Angular 10

### LINKED ISSUE #10417: [typescript-angular] Bug ModuleWithProviders missing generic type for Angular 10
##### Description

ModuleWithProviders without a generic type is deprecated since Angular 9 and mandatory since Angular 10:
See: https://github.com/angular/angular/commit/20cc3ab37e2d79761993dfcec2cafd6393b4f414

##### Swagger-codegen version

3.0.21

##### Steps to reproduce

Using the generated code (without npm) in an Angular 10 project leads to a TypeScript error and the application not compiling.

Change `ModuleWithProviders` to `ModuleWithProviders<ApiModule>` to: https://github.com/swagger-api/swagger-codegen/blob/e6d69ab23a8ad4de1c57bea18a11c386bdd81022/modules/swagger-codegen/src/main/resources/typescript-angular/api.module.mustache#L21.

Edit:
Looks like the option to add a generic type was added in Angular 6.1.0, so my solution breaks earlier versions.
https://github.com/angular/angular/commit/60aeee7abfcc149d9f33ed9519b242b4bccea641#diff-968ca8dead1eca4223cb534b9f0b9c02R79

derkoe:
Make ModuleWithProviders generic for all Angular version >= 10 because it is mandatory.

Fixes #10417

Signed-off-by: Christian Köberl <christian.koeberl@porscheinformatik.com>

### PR checklist

### Description of the PR

This adds a generic type info to the ModuleWithProviders in the api.module.ts - this is needed because otherwise the generated code will not compile with Angular >= 10.


Michi-2142 (approved):

HugoMario:
please build

HugoMario:
thanks a lot @derkoe !!

Michi-2142:
When would the maven package 2.4.17 including this changes be pushed to maven central?

loutocky:
I need this fix in codegen cli 3.0.23 (https://repo1.maven.org/maven2/io/swagger/codegen/v3/swagger-codegen-cli/3.0.23/). How can I get it there?

Poitrin:
> I need this fix in codegen cli 3.0.23 (https://repo1.maven.org/maven2/io/swagger/codegen/v3/swagger-codegen-cli/3.0.23/). How can I get it there?

I think the problem lies here: https://github.com/swagger-api/swagger-codegen-generators/blob/master/src/main/resources/handlebars/typescript-angular/api.module.mustache

If I got it right, the code in this repo is used by the codegen CLI (3.0.23). There is already a PR for the -generators repo: https://github.com/swagger-api/swagger-codegen-generators/pull/730

And I agree with https://github.com/swagger-api/swagger-codegen/issues/10417#issuecomment-673570816, "The typescript-angular generator maintenance could be better".

jose-hernandez-iop:
I have added a comment to [swagger-api/swagger-codegen-generators#730](https://github.com/swagger-api/swagger-codegen-generators/pull/730) with a suggestion for backwards compatibility with versions of Angular earlier that 10 which is in line with the solution already merged-in for V2.

I would really appreciate it if someone was kind enough to merge these fixes in so that we can generate code for Angular 10 without the need for a custom local build of the code generator plug-in.

twopelu:
I have just tested version 2.4.17 and 18 of swagger-codegen-cli and it still generates the ApiModule as in previous versions, it does not fix the problem. Is that the correct version with the fix? Is is related to this other issue? swagger-api/swagger-codegen-generators#730?

Michi-2142:
Which version of angular are you using? Have you added `--additional-properties ngVersion=10` to your generate call? See my stackoverflow [answer](https://stackoverflow.com/a/66025640/2617281).

twopelu:
I have tested it with 2.4.19 (just released) and it works properly adding the `--additional-properties ngVersion=10`. Many thanks!

giovanni-bertoncelli:
Hi, I've installed the 3.0.25 swagger-codegen version and the issue remains...
The command run was:

```
swagger-codegen generate -i "http://127.0.0.1:${SERVER_PORT}${DOCS_PATH}/json" -l typescript-angular -o "${DESTINATION_ROOT}" --config "${PROJECT_ROOT}/sdk/sdk.config.json"
```

Where `sdk.config.json` is like this:

```json
{
  "ngVersion": "10.0.0",
  "supportsES6": true
}
```

Am I missing something?

Michi-2142:
Please try to set version via `--additional-properties ngVersion=10` or try to set angular version in your config file to `"ngVersion": "10"` and check whether issue remains.

giovanni-bertoncelli:
> Please try to set version via `--additional-properties ngVersion=10` or try to set angular version in your config file to `"ngVersion": "10"` and check whether issue remains.

@Michi-2142 Already tried both with no luck...

twopelu:
Just a question, does this fix avoids the need of setting --additional-properties ngVersion=10 to make it work?

Michi-2142:
> Just a question, does this fix avoids the need of setting `--additional-properties ngVersion=10` to make it work?

Sadly no, in this version you need to set settings explicitly.

raman-nbg:
@giovanni-bertoncelli did you resolve your issue? I got stuck with the same problem. `--additional-properties ngVersion=10` doesn't have any effect...

raman-nbg:
With which version is this change available? I'm using 3.0.24.

HighSoftWare96:
@raman-nbg I surrendered and replace the value inline with `sed`

raman-nbg:
I created a local copy of the mustache template (api.module.mustache) in my own repo and applied it with -t argument. That was the quickest workaround I could find.

Issue #3856, opened 2016-09-23
[JAXRS-SPEC] Enum values are generating "&quot;", "&#3D;" in the string putting imports all over the place


filpano:
##### Description

I wanted to generate API Definition with the `swagger-codegen-maven-plugin`, for jaxrs-spec so that I receive only the models and basic API implementations (i.e., no factories). For all models that have an enum however, it produces uncompilable source code.

##### Swagger-codegen version

The latest stable build, 2.2.1, although I have also tried 2.2.2-SNAPSHOT but it has not made a difference.

##### Swagger declaration file content or url

```yaml
  T:
    required:
      - a
      - c
    properties:
      a:
        type: number
        format: double
      c:
        type: string
      state: 
        type: string
        enum:
          - initial
```

##### Command line used for generation

`mvn clean io.swagger:swagger-codegen-maven-plugin:generate` with the following plugin entry in the `pom.xml`:

				<plugin>
					<groupId>io.swagger</groupId>
					<artifactId>swagger-codegen-maven-plugin</artifactId>
					<version>2.2.0</version>
					<configuration>
						<inputSpec>${project.basedir}/src/main/resources/api.yaml</inputSpec>
						<language>jaxrs-spec</language>
						<output>${project.build.directory}/generated-sources/payment</output>
						<configOptions>
							<sourceFolder>src/java/main</sourceFolder>
							<dateLibrary>java8</dateLibrary>
						</configOptions>
						<groupId>com.product</groupId>
						<artifactId>rest</artifactId>
						<modelPackage>com.product.rest.model</modelPackage>
						<invokerPackage>com.product.rest.invoker</invokerPackage>
						<apiPackage>com.product.rest.api</apiPackage>
					</configuration>
					<executions>
						<execution>
							<id>generate-server-stubs</id>
							<goals>
								<goal>generate</goal>
							</goals>
							<configuration>
							</configuration>
						</execution>
					</executions>
				</plugin>

##### Steps to reproduce

Attempt to generate server stubs using the model definition above.

The generated source code then looks similar to the following:

```
package com.product.rest.model;

import java.time.OffsetDateTime;



import io.swagger.annotations.*;
import java.util.Objects;


public class T {
  
  private Double amount = null;
  private String currency = null;

import javax.xml.bind.annotation.XmlEnum;
import javax.xml.bind.annotation.XmlType;

@XmlType(name="Transaction")
@XmlEnum
public enum Transaction {
[...]

PR #10533, opened 2020-10-31
Fixed outer enum issue on jaxrs-resteasy and jaxrs-resteasy-eap

GabriRuflex:
### PR checklist

### Description of the PR

* Fixed outer enum issue on jaxrs-resteasy and jaxrs-reasteasy-eap languages (similar to jaxrs-spec, see #3856)

### Ping
@HugoMario

### DIFF
--- a/modules/swagger-codegen/src/main/resources/JavaJaxRS/resteasy/eap/enumOuterClass.mustache
+++ b/modules/swagger-codegen/src/main/resources/JavaJaxRS/resteasy/eap/enumOuterClass.mustache
@@ -1,7 +1,20 @@
-public enum {{classname}} {
-  {{#allowableValues}}
-  {{#enumVars}}
-  {{{name}}}{{^-last}},{{/-last}}{{#-last}};{{/-last}}
-  {{/enumVars}}
-  {{/allowableValues}}
+/**
+ * {{^description}}Gets or Sets {{{name}}}{{/description}}{{#description}}{{description}}{{/description}}
+ */
+public enum {{#datatypeWithEnum}}{{{.}}}{{/datatypeWithEnum}}{{^datatypeWithEnum}}{{{classname}}}{{/datatypeWithEnum}} {
+  {{#allowableValues}}{{#enumVars}}
+  {{{name}}}({{{value}}}){{^-last}},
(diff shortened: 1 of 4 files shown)


GabriRuflex:
@HugoMario Hi Hugo, could you please take a look on this? Thanks a lot!

ellyxc:
Hi there, what is the status of this PR? i think this PR open for about 2 years already

GabriRuflex:
Hi maintainers 👋

This PR has now been open for **almost 6 years** (since October 2020), so here's a fresh status update:

**The bug is evident, trivially reproducible and blocking.** Any Swagger spec that defines a top-level (outer) enum currently produces broken code with these generators — you can reproduce it with a three-line definition:

```yaml
OrderStatus:
  type: string
  enum: [placed, approved, delivered]
```

- **`jaxrs-resteasy`**: the generated enum does not even compile — the template renders the raw `allowableValues` map, producing something like `public enum OrderStatus { {values=[placed, approved, delivered], enumVars=[...]} }`. This is the same class of bug that was fixed for `jaxrs-spec` in #3856.
- **`jaxrs-resteasy-eap`**: the enum compiles but loses the spec values — no value field and no `@JsonValue`, so Jackson serializes the Java constant name (`PLACED`) instead of the wire value (`placed`), breaking round-tripping with any conforming client.

**The fix is the missing half of code that is already on `master`.** Interestingly, `JavaResteasyServerCodegen` and `JavaResteasyEapServerCodegen` already contain the `postProcessModelsEnum` logic that adds the `com.fasterxml.jackson.annotation.JsonValue` import for outer enum models — but the templates were never updated to use it. This PR provides exactly that missing template side.

**Current state of the PR:**
- ✅ Rebased on the current `master`, no conflicts, mergeable
- ✅ Now includes regression tests (`JavaResteasyOuterEnumTest`, parametrized over both generators) that generate from a spec with a top-level enum and verify the emitted code — the tests **fail on `master`** and **pass with this PR**

Given that the change is two mustache templates plus tests, aligns these generators with what `jaxrs-spec` has done since 2019, and is now covered by regression tests, it would be great to finally get this reviewed and merged. Happy to address any feedback quickly. Thanks! 🙏

cc @ewaostrowska

Issue #10129, opened 2020-03-24
Error with response_type on Python client


juanpaps03:
Generating Python client with codegen, the code has an error in the default_api.py file.

In runtime I'm sending data with a client function, the error that I get is:
```python
AttributeError: module 'inbound_manager_client.models' has no attribute 'Object'
```
manually checking the code the error is in api_client.py/__deserialize function:
```python
            # convert str to class
            if klass in self.NATIVE_TYPES_MAPPING:
                klass = self.NATIVE_TYPES_MAPPING[klass]
            else:
                klass = getattr(inbound_manager_client.models, klass)
```
The error raise in the last line of the code shared above, the 'klass' variable value is "Object" (with uppercase O). I think that the root of the problem is on:

```python
        return self.api_client.call_api(
            "/patient",
            "POST",
            path_params,
            query_params,
            header_params,
            body=body_params,
            post_params=form_params,
            files=local_var_files,
            response_type="Object",  # noqa: E501
            auth_settings=auth_settings,
            async_req=params.get("async_req"),
            _return_http_data_only=params.get("_return_http_data_only"),
            _preload_content=params.get("_preload_content", True),
            _request_timeout=params.get("_request_timeout"),
            collection_formats=collection_formats,
        )
```
`response_type` is where the klass variable is being settled.
This is part of a function generated based on an API endpoint, the function name is:
```python
    def add_patient_patient_post_with_http_info(self, body, **kwargs):  # noqa: E501
```
the endpoint function in the original python client is `add_patient`.

I change the reponse_type manually to test if the error is actually there, and if I change it for "object" or for a valid model class name, it works like a charm.

I was looking how this response_type is being generated in order to make any necessary change in the openapi.json used to generate this, but I didn't found the proper value to set on openapi.

From my point of view, this could be a bug in codegen, but maybe is something wrong with my openapi.json, if someone can guide me to make the changes on the openapi, it would be nice, but if this is a bug on codegen, I can't do more than wait 

swagger-cogen version:  3.0.18

##### Command line used for generation
On linux:
java -jar swagger-codegen-cli.jar generate -i http://localhost:8000/openapi.json -l python -c swagger-config-inbound-manager-client.json
On mac:
swagger-codegen generate -i http://127.0.0.1:8000/openapi.json -l python -c swagger-config-inbound-manager-client.json
(same problem with both versions)

##### Openapi.json file link

[...]

PR #10907, opened 2021-02-11
Handle "Object" response type in Python client

colecloudtostreet:
Hypothetical fix for https://github.com/swagger-api/swagger-codegen/issues/10129. Not sure if there's a bigger issue at hand.


bkelley17:
I think the root cause of this bug is that it camel cases the open API type "object" to "Object".

colecloudtostreet:
Closing - I believe bkelley17's PR https://github.com/swagger-api/swagger-codegen-generators/pull/930 is the real fix

Issue #10313, opened 2020-06-09
[maven-plugin] Prevent always re-generating code


ancoron:
##### Description

The codegen plugin currently always re-generates code, even though neither the plugin configuration nor the input specification file changed. This results in later Maven build steps to be also re-executed for no reason, which finally make build times much longer than necessary for rebuilds.

##### Swagger-codegen version

3.0.18

##### Swagger declaration file content or url

_not relevant_

##### Command line used for generation

Plugin configuration:

```xml
<plugin>
    <groupId>io.swagger.codegen.v3</groupId>
    <artifactId>swagger-codegen-maven-plugin</artifactId>
    <version>3.0.18</version>
    <executions>
        <execution>
            <id>generate-api-interfaces</id>
            <goals>
                <goal>generate</goal>
            </goals>
            <configuration>
                <inputSpec>${basedir}/api/openapi.yaml</inputSpec>
                <artifactId>my-service</artifactId>
                <output>${basedir}/target/generated-sources</output>
                <language>spring</language>
                <apiPackage>com.mycorp.svc.api</apiPackage>
                <modelPackage>com.mycorp.svc.model</modelPackage>
                <generateApis>true</generateApis>
                <generateSupportingFiles>false</generateSupportingFiles>
                <configOptions>
                    <sourceFolder>swagger</sourceFolder>
                    <interfaceOnly>true</interfaceOnly>
                    <library>spring-boot</library>
                    <configPackage>com.mycorp.svc</configPackage>
                    <dateLibrary>legacy</dateLibrary>
                </configOptions>
            </configuration>
        </execution>
    </executions>
</plugin>
```

##### Steps to reproduce

```bash
mvn clean package
mvn package
```

##### Related issues/PRs

_none_

##### Suggest a fix/enhancement

The plugin could write out a "stamp" file at a well-known location, containing a hash of the plugin configuration. The timestamp of the file itself serves as a marker for the last generation run and can be compared to the file specified as `inputSpec` parameter.

The plugin should then only re-generate code if either of the following statements is true:
* the `inputSpec` file is newer than our marker file
* the plugin configuration hash does not match

teleological:
See also #8539

stoll-it:
Related changes from the openapi-generator project: https://github.com/OpenAPITools/openapi-generator/issues/1501 and https://github.com/OpenAPITools/openapi-generator/pull/1502

Philzen:
This was implemented a long time ago in 4.x:

[...]

PR #10936, opened 2021-03-07
Do not overwrite unchanged generated files

lavr:
Attempt to solve #10313


stoll-it:
Checking the source file for modifications (which of course comes with other caveats, so it should probably not be enabled by default) like the openapi-generator might also be worth a try: https://github.com/OpenAPITools/openapi-generator/issues/1501

PR #10949, opened 2021-03-23
[Python] Fix #10948 wrong mode opened file fixed

### LINKED ISSUE #10948: [Python] response.data has type "str" but file opened with "wb" mode, need be "w"
##### Swagger-codegen version 

3.0.25

```yaml
swagger: '2.0'
info:
  title: Some API
  description: Some API
  version: "1.0.0"
schemes:
  - http
produces:
  - application/json
paths:
  /export:
    get:
      operationId: export_op
      x-swagger-router-controller: source
      summary: Some export
      tags:
        - Sources
      responses:
        200:
          description: done
          schema:
            type: file

```

```bash
docker run --rm \
    -v $(pwd)/some.yml:/some.yml \
    -v $(pwd)/result:/tmp/result \
    swaggerapi/swagger-codegen-cli-v3:3.0.25 \
    generate -i /some.yml -l python -o /tmp/result
```
##### Steps to reproduce

Just execute command

I'll try fix this

##### BUG

in `result/swagger_client/api_client.py` file
in `__deserialize_file` method:
```python
def __deserialize_file(self, response):
        """Deserializes body to file

        Saves response body into a file in a temporary folder,
        using the filename from the `Content-Disposition` header if provided.

        :param response:  RESTResponse.
        :return: file path.
        """
        fd, path = tempfile.mkstemp(dir=self.configuration.temp_folder_path)
        os.close(fd)
        os.remove(path)

        content_disposition = response.getheader("Content-Disposition")
        if content_disposition:
            filename = re.search(r'filename=[\'"]?([^\'"\s]+)[\'"]?',
                                 content_disposition).group(1)
            path = os.path.join(os.path.dirname(path), filename)
        
        with open(path, "wb") as f:
            f.write(response.data)

        return path
```

Error in block with writing to file
`response.data` has type "str" but file opened with "wb" mode, need be "w"

goDeni:
### PR checklist

- [x] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).1

### Description of the PR


goDeni:
@kenjones-cisco please review this PR

goDeni:
I'm tired of doing dirty hacks for these methods to work
Community need this mini-fix

HugoMario:
Hello @goDeni,

Thanks for catching this issue and file a PR, however i think we should not replace the `wb` modes without leaving an option to the others 36 developers that worked on this and considered it as a valid option.

So, my propose is, add an option to the generator, and when this option is present use the `w` mode instead of `wb`. wdyt?

goDeni:
> 
> 
> Hello @goDeni,
> 
> Thanks for catching this issue and file a PR, however i think we should not replace the `wb` modes without leaving an option to the others 36 developers that worked on this and considered it as a valid option.
> 
> So, my propose is, add an option to the generator, and when this option is present use the `w` mode instead of `wb`. wdyt?




Option to the generator is solve problem, but this take out implementation details 


There is only one case when this method is called
is called when `response_type='file'` and `_preload_content=True`

and response.data always has `str` type

https://github.com/swagger-api/swagger-codegen/blob/068b1ebcb7b04a48ad38f1cadd24bb3810c9f1ab/modules/swagger-codegen/src/main/resources/python/rest.mustache#L213
image


i don't see case when this method work correctly (for python 3)

if you can show me this case, i'll glad


now i see my solution won't work for python 2. i'll fix this

Issue #9254, opened 2019-03-11
[Typescript-Angular] Generated client code for "Cookie Authentication" is incorrect.


ChrisJ87:
##### Description

I want to use the "Cookie Authentication" as described [HERE](https://swagger.io/docs/specification/authentication/cookie-authentication/) with the cookies transferred within the headers.  
Everything looks well in the Swagger-Editor/-UI but the generated code does not create any header nor does it put a cookie in it for any request.
Instead the generated code sets the cookie to the "queryParameters" if the appropriate ApiKey is available. 
If the request does not define any query parameters, the generated source code assigns the cookie to the uninitialized variable "queryParameters".

Thus, there are two issues:

1. No HTML header will be generated with the cookie in it.
2. The generated code tries to transfer the ApiKey within uninitialized "queryParameters" variable.

##### Swagger-codegen version

Swagger-Editor Version: 3.6.25

##### Swagger declaration file content or url

The YAML code:

```yaml
openapi: 3.0.0

paths: 
  '/entity/{uuid}/program':
    post:
      operationId: program
      parameters:
        - in: path
          name: uuid
          schema:
            type: string
          required: true
      responses:
        '202':
          description: Accepted
          content:
            text/plain:
              schema:
                $ref: '#/components/schemas/UUID'          
        '401':
          $ref: '#/components/responses/UnauthorizedError'
        default:
          description: Unexpected error.

components:
  securitySchemes:
    cookieAuth:
      type: apiKey
      in: cookie
      name: SESSIONID
  schemas:
    UUID:
      type: string
      description: Unique ID of the targeted blast group.
      example: 936DA01F-9ABD-4D9D-80C7-02AF85C822A8
  responses:
    UnauthorizedError:
      description: API key (BLASTSESSIONID) is missing or invalid
      headers:
        WWW_Authenticate:
          schema:
            type: string

security:
  - cookieAuth: []
```

The generated code:

     public program(uuid : string, observe?: 'body', reportProgress?: boolean): Observable<string>;
     public program(uuid : string, observe?: 'response', reportProgress?: boolean): Observable<HttpResponse<string>>;
     public program(uuid : string, observe?: 'events', reportProgress?: boolean): Observable<HttpEvent<string>>;
     public program(uuid : string, observe: any = 'body', reportProgress: boolean = false ): Observable<any> {

        if (uuid === null || uuid === undefined) {
            throw new Error('Required parameter uuid was null or undefined when calling program.');
        }

[...]

PR #11101, opened 2021-07-22
Make python codegen respect cookie setting

prius:
### PR checklist

### Description of the PR

There is an issue described here https://github.com/swagger-api/swagger-codegen/issues/9254
The problem is that in swagger doc https://swagger.io/docs/specification/authentication/cookie-authentication/ there is a description of Cookie auth. In reality there is nothing in the code that actually implements this cookie auth.

This PR introduces an idea of how to fix it.


prius:
I'm not sure about the implementation details though, as there are few things I'm not certain about:
* In 3.0.0 branch I don't see those files at all, so I've made a PR to master branch
* If we implement change for `DefaultCodegen.java` this means we automatically should change all the templates for all languages, which might require extensive testing. I'm not sure what kind of testing should be done here, though
* The way I modify cookie header is just a pure concatenation, which might not work in some cases

So if somebody can guide me here - would be nice.

Issue #11154, opened 2021-09-02
[Javascript] [2.x] Stackoverflow (infinite loop) from recursive model with multiple tags


mjschuetze102:
##### Description

When generating a Javascript client with a recursive model, an infinite loop occurs. Note, this only happens when there are multiple tags provided in the yaml spec, i.e.

```
/dog/:
    post:
      tags:
        - dog
      ...
  /cat/:
    get:
      tags:
        - cat
      ...
```

**Detailed explanation:**
Within the [DefaultGenerator](https://github.com/swagger-api/swagger-codegen/blob/90319bee8aef8be041e003499ea3cfffef612a40/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java#L477), each tag is being looped through. Within that loop, the [postProcessOperationsWithModels](https://github.com/swagger-api/swagger-codegen/blob/90319bee8aef8be041e003499ea3cfffef612a40/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java#L1030) method within the JavascriptClientCodegen class is called. For the first tag, the code is run with no problems. The outcome is that each model ends up with a vendor extension "x-codegen-model" which links back to the model. Upon second iteration of this loop the list of models is [converted to a HashSet](https://github.com/swagger-api/swagger-codegen/blob/90319bee8aef8be041e003499ea3cfffef612a40/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java#L1061), but the hashCode method of CodegenModel (and CodegenProperties because of the recursive nature) takes into account the vendorExtensions as part of the hash. Which means that in order for the hashCode to be generated for the model, it needs to generate the hashCode of itself, which introduces the loop.

##### Swagger-codegen version

Version in use: 2.4.12
Version bug was introduced: 2.4.9
Latest version still has bug: Yes (2.4.21, 2.4.22-snapshot)

##### Swagger declaration file content or url

https://gist.github.com/mjschuetze102/a60535fa28635c585e5711f5400aa94e

##### Command line used for generation

`swagger-codegen generate -i https://gist.githubusercontent.com/mjschuetze102/a60535fa28635c585e5711f5400aa94e/raw/322b6f1c05dcfb752a5451674d0aba553ab638f3/recursive_model_multiple_tags.yaml -l javascript -o ./tmp`

##### Steps to reproduce

Generate javascript client with the above swagger definition.

[...]

PR #11155, opened 2021-09-02
[Javascript] [2.x] Removed HashMap creation which caused infinite loop

mjschuetze102:
### PR checklist

### Description of the PR

Details: Removed the creation of a HashSet object which caused infinite loop
Fixes: #11154

Same change as: https://github.com/swagger-api/swagger-codegen-generators/commit/2d472e5151dc4b04aec09da86531c43abfd14b25, but for v2

**Detailed explanation:**
Within the [DefaultGenerator](https://github.com/swagger-api/swagger-codegen/blob/90319bee8aef8be041e003499ea3cfffef612a40/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java#L477), each tag is being looped through. Within that loop, the [postProcessOperationsWithModels](https://github.com/swagger-api/swagger-codegen/blob/90319bee8aef8be041e003499ea3cfffef612a40/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java#L1030) method within the JavascriptClientCodegen class is called. For the first tag, the code is run with no problems. The outcome is that each model ends up with a vendor extension "x-codegen-model" which links back to the model. Upon second iteration of this loop the list of models is [converted to a HashSet](https://github.com/swagger-api/swagger-codegen/blob/90319bee8aef8be041e003499ea3cfffef612a40/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavascriptClientCodegen.java#L1061), but the hashCode method of CodegenModel (and CodegenProperties because of the recursive nature) takes into account the vendorExtensions as part of the hash. Which means that in order for the hashCode to be generated for the model, it needs to generate the hashCode of itself, which introduces the loop.

The new collection isn't used anywhere, and the old one isn't modified at all, which means the new object creation is irrelevant anyways. If a new collection must be created, a List should be used instead of a Set. Still, if uniqueness was the reasoning for using a set, the values of the HashMap are paired with the classname of the model, which means that the values would all be unique anyway. Any sharing the same name would cause a collision and would have been overwritten, and if the classname wasn't the same, then it's a different object anyway.


HugoMario:
THanks for PR @mjschuetze102 !!
Really sorry for so much delay

PR #11183, opened 2021-09-22
Use mavenCentral instead of jcenter

nkiesel:
This simply replaces the retired `jcenter()` with `mavenCentral()`


pumano:
jcenter is shutted down more than half of year ago. Looks like should be merged

PR #11219, opened 2021-10-11
Fix for ENUM requires https://github.com/swagger-api/swagger-codegen-generators/pull/978

sbilello:
### PR checklist

- [X] Read the [contribution guidelines](https://github.com/swagger-api/swagger-codegen/blob/master/CONTRIBUTING.md).
- [ ] Ran the shell script under `./bin/` to update Petstore sample so that CIs can verify the change. (For instance, only need to run `./bin/{LANG}-petstore.sh` and `./bin/security/{LANG}-petstore.sh` if updating the {LANG} (e.g. php, ruby, python, etc) code generator or {LANG} client's mustache templates). Windows batch files can be found in `.\bin\windows\`.
- [X] Filed the PR against the correct branch: `3.0.0` branch for changes related to OpenAPI spec 3.0. Default: `master`.
- [ ] Copied the [technical committee](https://github.com/swagger-api/swagger-codegen/#swagger-codegen-technical-committee) to review the pull request if your PR is targeting a particular programming language.

### Description of the PR

This PR needs to be merged after https://github.com/swagger-api/swagger-codegen-generators/pull/978

@frantuma 

```
mvn test

[INFO] Reactor Summary for swagger-codegen-project 3.0.30-SNAPSHOT:
[INFO] 
[INFO] swagger-codegen-project ............................ SUCCESS [  0.498 s]
[INFO] swagger-codegen (core library) ..................... SUCCESS [ 13.882 s]
[INFO] swagger-codegen (executable) ....................... SUCCESS [  2.700 s]
[INFO] swagger-codegen (maven-plugin) ..................... SUCCESS [  3.798 s]
[INFO] swagger-generator .................................. SUCCESS [  5.670 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  26.865 s
[INFO] Finished at: 2021-10-11T12:06:20-07:00
[INFO] ------------------------------------------------------------------------
```
It needs the `mvn clean install` from 
https://github.com/swagger-api/swagger-codegen-generators/pull/978
to make it work.

If you encounter any problem please provide specific steps to replicate it

Thanks

PR #11250, opened 2021-11-05
Fix for ENUM that works only by merging PR https://github.com/swagger-api/swagger-codegen-generators/pull/978

frantuma:
includes and replaces #11219


sbilello:
@frantuma When are we going release a new version?

Issue #11719, opened 2022-02-22
[Templates] Undefined readOnly variable instead of the right isReadOnly


fedepad:
##### Description

This issue regards at least client documentation automatic generation (plus for some languages also not just doc) with `swagger-codegen` because of a bug in many languages templates.
Precisely, it regards the use of the undefined template variable `readOnly` to check if the variable/parameter is read-only while the right variable to use in the templates is `isReadOnly` as defined by: https://github.com/swagger-api/swagger-codegen/commit/a6df340cdc80f71cc4bdb88aa7174013a5d41c5c
##### Swagger-codegen version

I am using version **3.0.30** but I don't see this fixed in the current master.
##### Command line used for generation

`swagger-codegen generate -i <your yaml file with models that have readonly attributes> -l python -o client`

Note: it does not apply only to python clients.

##### Steps to reproduce

Just generate for example a python client.
You will see that for the generated models documentation, models' attributes that are readonly are not annotated as readonly because of the problem mentioned above.

##### Related issues/PRs

I haven't found any but please feel free to point that out if I missed something.

##### Suggest a fix/enhancement

I already have forked the project and in my branch have a fix for that.
I already tested in python client generation because I use it in one of my project and now the readonly variables are annotated properly in the client docs.
Let me know if I should just open a PR.

fedepad:
Opened: https://github.com/swagger-api/swagger-codegen/pull/11720
Will create a separate one for branch 3.0.0

fedepad:
Opened https://github.com/swagger-api/swagger-codegen-generators/pull/1016

PR #11720, opened 2022-02-22
Fix/client model doc template readonly

fedepad:
This commit fixes occurrences of the undefined `readOnly` in multiple templates with the right template variable `isReadOnly`.

### PR checklist

### Description of the PR

The change regards several templates for which an undefined template variable `readOnly` was used instead of `isReadOnly` as defined here: https://github.com/swagger-api/swagger-codegen/commit/a6df340cdc80f71cc4bdb88aa7174013a5d41c5c
This PR replaces `readOnly` with `isReadOnly` so that the client docs can be generated correctly.
Related issue: https://github.com/swagger-api/swagger-codegen/issues/11719


fedepad:
This should have targeted 3.0.0 branch. Doc is very confusing.

Issue #10912, opened 2021-02-17
[PYTHON] ref properties do not propagate


pnhowe:
##### Description

Generated python code does not propagate the ref properties.  I have checked other languages and they look right,  for Go the propagated values are a part of the struct definition, and for Java and Typescript, the class definitions inherit the from a parent class with the propagated properties.  As far as I can tell Python does neither.

##### Swagger-codegen version

using   swagger-codegen-cli-2.4.18.jar  and editor.swagger.io

##### Swagger declaration file content or url

```
swagger: "2.0"
info:
  version: 1.0.0
  title: Inheritance test
paths:
  /User:
    get:
      summary: Get User
      responses:
        '200':
          description: The User
          schema:
            $ref: '#/definitions/UserModel'

definitions:
  ModelBase:
    type: object
    additionalProperties: false
    required:
      - id
    properties:
      id:
        type: string
        format: guid

  UserModel:
    type: object
    additionalProperties: false
    allOf:
      - $ref: '#/definitions/ModelBase'
    required:
      - username
    properties:
      username:
        type: string

```

##### Command line used for generation

java -jar swagger-codegen-cli.jar generate -i test.yaml -l python -o test_api --additional-properties packageName=test_api --additional-properties packageVersion=1.0

This results in 
```
class UserModel(object):
    """NOTE: This class is auto generated by the swagger code generator program.

    Do not edit the class manually.
    """

    """
    Attributes:
      swagger_types (dict): The key is attribute name
                            and the value is attribute type.
      attribute_map (dict): The key is attribute name
                            and the value is json key in definition.
    """
    swagger_types = {
        'username': 'str'
    }

    attribute_map = {
        'username': 'username'
    }

    def __init__(self, username=None):  # noqa: E501
        """UserModel - a model defined in Swagger"""  # noqa: E501

        self._username = None
        self.discriminator = None

        self.username = username
```
the self.id field is missing and the UserModel does not inherit from ModelBase.

##### Steps to reproduce

The above yaml file into either editor.swagger.io or codegen v2.4.18

##### Related issues/PRs

I found a number of past issues 

##### Suggest a fix/enhancement

I'm not sure if python should be doing the object inheritance like Java/Typscript or just copy the properties in like Go.  My preference would probably more along the lines of the Go implementation.

Issue #11225, opened 2021-10-21
[Python] Swagger 2.0 - 2.4.23 cannot generate allOf types


MikaelUmaN:
##### Description

I generate REST clients against swagger/openapi v2.0

- Using swagger-codegen-cli:2.4.8 we get a class "DataDaily" (see json below) with all attributes.
- Using 2.4.23 we only get the properties immediately under "DataDaily", not respecting the properties in "DataRef"

##### Swagger-codegen version

- Using 2.4.23
- This is a regression from version 2.4.8

##### Swagger declaration file content or url

```
    "DataDaily": {
      "allOf": [
        {
          "$ref": "#/definitions/DataRef"
        },
        {
          "type": "object",
          "required": [
            "timestamp",
            "asOfDate"
          ],
          "properties": {
            "seriesId": {
              "title": "seriesId",
              "type": "string"
            },
            "timestamp": {
              "type": "string",
              "format": "date-time"
            },
            "asOfDate": {
              "type": "string",
              "format": "date-time"
            }
          }
        }
      ]
    },
```

##### Command line used for generation

docker run --rm -v ${PWD}:/local swaggerapi/swagger-codegen-cli:2.4.23 generate -i http://internalserver/swagger/v2.0/swagger.json -l python -o /local -c /local/generate/config.json

config.json only contains information on package version.

##### Steps to reproduce

- Create a route returning an allOf type
- Generate Python client using command above

##### Related issues/PRs

Not that I could find.

##### Suggest a fix/enhancement

Previous behavior (e.g. 2.4.8) seemed reasonable.

**Edit: I found the breakage occurs between 2.4.14 and 2.4.15**

radarsh:
This seems to be affecting more than just Python codegens. I am using JavaClientCodegen and I'm noticing the same behaviour. 

If I have the below definitions in my Swagger 2,

```yaml
Foo:
  type: object
    properties:
      revision:
        type: string
        format: 'date-time'
Bar:
  allOf:
    - type: object
      properties:
        id:
          type: string
    - $ref: "#/definitions/Foo"
```

With codegen 2.4.14, both `id` and `revision` properties are generated. With 2.4.15 up to 2.4.23 the `id` property gets ommitted.

PR #11968, opened 2022-11-11
Python Client [v2.X] Added inheritance (allOf) to python client

mjschuetze102:
### PR checklist

### Description of the PR

Fixes: #10912, #11225


modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/PythonClientCodegen.java
```diff
@@ -260,6 +260,13 @@ public String toModelImport(String name) {
         return modelImport;
     }
 
+    @Override
+    protected void addImport(CodegenModel m, String type) {
+        if (m.parent != null && m.parent.equals(type)) {
+            super.addImport(m, type);
+        }
+    }
+
```
mjschuetze102, line 269: Only adds the parent class for importing as the python client does not need to know the type of the values it works with, but does need to import the parent class

modules/swagger-codegen/src/main/resources/python/model.mustache
```diff
@@ -34,15 +36,15 @@ class {{classname}}(object):
...
-{{/vars}}
+{{/allVars}}
     }
 
     attribute_map = {
-{{#vars}}
+{{#allVars}}
         '{{name}}': '{{baseName}}'{{#hasMore}},{{/hasMore}}
-{{/vars}}
+{{/allVars}}
```
mjschuetze102, line 47: Due to the way the deserializer works in the api_client, the parent values must be present in `swagger_types` and `attribute_map`

modules/swagger-codegen/src/main/resources/python/model.mustache
```diff
@@ -52,7 +54,7 @@ class {{classname}}(object):
     }
 {{/discriminator}}
 
-    def __init__(self{{#vars}}, {{name}}={{#defaultValue}}{{{defaultValue}}}{{/defaultValue}}{{^defaultValue}}None{{/defaultValue}}{{/vars}}, _configuration=None):  # noqa: E501
+    def __init__(self{{#allVars}}, {{name}}={{#defaultValue}}{{{defaultValue}}}{{/defaultValue}}{{^defaultValue}}None{{/defaultValue}}{{/allVars}}, _configuration=None):  # noqa: E501
```
mjschuetze102, line 57: Similarly because the model is defined with the constructor rather than using setters for each property, the parent vars must be passed in the constructor as well as the properties belonging to this model

modules/swagger-codegen/src/main/resources/python/model.mustache
```diff
@@ -73,6 +75,10 @@ class {{classname}}(object):
 {{/required}}
 {{/vars}}
 
+{{#parent}}
+        super().__init__({{#parentModel.vars}}{{name}}{{#hasMore}}, {{/hasMore}}{{/parentModel.vars}})  # noqa: E501
+
+{{/parent}}
```
mjschuetze102, line 81: The parent constructor is called with the values passed in to the constructor of the child

jcgentr:
Can this be merged? This would be extremely helpful for Python devs using this tool.

modules/swagger-codegen/src/main/resources/python/model.mustache
jcgentr: I believe this should go up to the first line below the __init__ function so that calling the parent constructor does not wipe out any properties that were overridden by the current child class.

mjschuetze102: I believe I had reasoned that it didn't matter much because the code was using `vars` for the child class properties and `parentModel.vars` for properties on the parent. Since the child class never worked with `parentModel.vars` directly it would be fine. However, the discriminator (and any other code that is added in the future) might not realize this, so I changed the order of events

PR #12212, opened 2023-08-11
[python] Fix basic auth always being used

### LINKED ISSUE #9430: [Python] Generated client uses basic auth when username and password are not set
##### Description

I have a spec that allows either use of basic auth or an API-Key header. The generated python client always uses basic auth when making requests even when the username and password are empty (""). The following header is added: Authorization: Basic Og==

##### Swagger-codegen version

I was using the version on https://editor.swagger.io/

My security definitions are:
```json
"securityDefinitions" : {
    "APIKeyHeader" : {
      "type" : "apiKey",
      "in" : "header",
      "name" : "API-Key"
    },
    "basicAuth" : {
      "type" : "basic"
    }
  }
```
I am applying these schemes globally:
```json
"security" : [ {
    "APIKeyHeader" : [ ]
  }, {
    "basicAuth" : [ ]
  } ]
```

corytodd:
### PR checklist

@kenjones-cisco 

### Description of the PR
The Python client configuration will always generate a basic auth header regarless of username and password actually being set. This is a problem if the consumer is trying to use another authorization method that uses the same header name, e.g. Authorization. This bug is only triggered when the api spec's security definitions list BasicAuth after an auth type that uses a conflicting header.

Unfortunately, RFC 2617 [1] says that _both_ the username and password are allowed to be empty. This is how golang implements basic auth [2] which means golang codegen is also affected. Java codegen requires both username and password to be set [3]. C# requires only one of them to be set [4]. Java and C# are therefore not affected. Change python codegen to require either the username or password to be set as a compromise between being broken and rfc compliant.

[1]: https://datatracker.ietf.org/doc/html/rfc2617#section-2
[2]: https://github.com/golang/go/blob/master/src/net/http/client.go#L426
[3]: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/Java/auth/HttpBasicAuth.mustache#L45
[4]: https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/csharp/api.mustache#L407

fixes #9430


HugoMario:
thanks @corytodd for this PR!

PR #12289, opened 2023-11-24
New attribute isJson in parameters

luca-vercelli:
This PR relates to issue 1173 in `swagger-codegen-generators`:

https://github.com/swagger-api/swagger-codegen-generators/issues/1173

We add a new boolean argument `isJson` in class `CodegenParameter`.

This argument claims that parameter content-type is some kind of JSON. This info can be set and used by generators.


HugoMario:
Thanks @luca-vercelli !

PR #12330, opened 2024-01-31
typescript-fetch: nullify url search

simeonborko:
I had a problem with strict mode in TypeScript compilation because of lines `delete localVarUrlObj.search`

`localVarUrlObj` has type `Url`. Field `search` is defined here as this: `field: string | null` which means it's required, cannot be undefined and thus cannot be deleted. TypeScript compiled error was: "TS2790: The operand of a  delete  operator must be optional."

Source code for the `Url` type: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/url.d.ts

I have also removed a semicolon after BaseAPI (unneccesary after classes) and rerun `bin/{,security/}typescript-fetch-petstore.sh` which were not run since 2.4.0-SNAPSHOT, so there are few more changes.


funivan:
👏

HugoMario:
Thanks a lot @simeonborko for this PR

Issue #12158, opened 2023-05-23
Base Image in Dockerfile has Changed


ptMcGit:
##### Description

Docker fails to build image from Dockerfile.

##### Swagger-codegen version

n/a

##### Swagger declaration file content or url
n/a

##### Command line used for generation

n/a

##### Steps to reproduce

``` shell
docker build ./                                  
[+] Building 0.4s (6/16)                                                                                                                    
 => [internal] load .dockerignore                                                                                                      0.0s 
 => => transferring context: 310B                                                                                                      0.0s 
 => [internal] load build definition from Dockerfile                                                                                   0.0s 
 => => transferring dockerfile: 1.17kB                                                                                                 0.0s 
 => [internal] load metadata for docker.io/library/maven:3-eclipse-temurin-11                                                          0.0s 
 => CACHED [ 1/12] FROM docker.io/library/maven:3-eclipse-temurin-11                                                                   0.0s 
 => [internal] load build context                                                                                                      0.4s 
 => => transferring context: 10.87MB                                                                                                   0.3s 
 => ERROR [ 2/12] RUN set -x &&     apk add --no-cache bash                                                                            0.4s 
------                                                                                                                                      
 > [ 2/12] RUN set -x &&     apk add --no-cache bash:                                                                                       
#0 0.320 + apk add --no-cache bash                                                                                                          
#0 0.320 /bin/sh: 1: apk: not found                                                                                                         
------                                                                                                                                      
Dockerfile:3                                                                                                                                
--------------------                                                                                                                        
   2 |                                                                                                                                      
   3 | >>> RUN set -x && \                                                                                                                  
   4 | >>>     apk add --no-cache bash                                                                                                      
   5 |                                                                                                                                      
--------------------                                                                                                                        
ERROR: failed to solve: process "/bin/sh -c set -x &&     apk add --no-cache bash" did not complete successfully: exit code: 127 
```

##### Related issues/PRs

None found.

##### Suggest a fix/enhancement

The base image [maven:3-eclipse-temurin-11](
https://hub.docker.com/layers/library/maven/3-eclipse-temurin-11/images/sha256-5e024ec8ab22c38e4994300094231067c49e1785eabea9e4fcf9082ca5b4f961?context=explore) [here](https://github.com/swagger-api/swagger-codegen/blob/master/Dockerfile#L1) is not alpine so `apk` is not found.

Update the line with `maven:3-eclipse-temurin-11-alpine` to explicitly ask for the alpine image.

[...]

PR #12381, opened 2024-04-23
Dockerfile fix

dreglad:
Dockerfile base image had changed from Alpine to Ubun, breaking Docker image build

### PR checklist

### Description of the PR

Fixes in Dockerfile:

#### 1. Updated base image
As described in #12158, the Docker build is broken since the base image was changed from Alpine to Ubuntu.

#### 2. Added `GEN_DIR` as Build Argument

The `WORKDIR` was previously set by an environment variable `GEN_DIR` at build time. This setup doesn't make sense as it always resolves to the same default value, which makes the Dockerfile harder to read and prone to errors if anyone actually tries to customize this value 
 Assuming there's a valid reason for allowing this to be dynamically set, I've added a build argument (`ARG` instruction) before `ENV`, so it can actually be used effectively.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM maven:3-eclipse-temurin-11
+FROM maven:3-eclipse-temurin-11-alpine
 
 RUN set -x && \
     apk add --no-cache bash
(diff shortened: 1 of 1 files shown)


frantuma:
Thanks!

Issue #9843, opened 2019-11-11
Swagger3 Authentication is not generated correctly


lindgrenfredrik:
#9786 ### Description

Swagger3 code generator has removed authMethod section for each Operation that was available in swagger2
There is a general AuthMethod section but for oauth2 it contains all scopes so it can't be used when generating scope checks for specific operations

##### Swagger-codegen version

Tested with 3.0.13 & 2.4.8 so see the difference

##### Swagger declaration file content or url

Swagger2 & 3 files to reproduce the problem:
https://gist.github.com/lindgrenfredrik/0eb76d8ee2cac7066bfefd914632a635

##### Command line used for generation

java -jar -DdebugOperations=True swagger-codegen-cli.jar generate -l jaxrs-resteasy -i pet_example_2_0.json -o tmp2

 java -jar -DdebugOperations=True swagger-codegen-cli-3.0.13.jar generate -l jaxrs-resteasy -i pet_example_3_0.yaml -o tmp3
<!-- if it is a bug, a json or yaml that produces it.

##### Steps to reproduce

Run command-line args above
Source generated in file src/gen/java/io/swagger/api/PetApi.java doesn't generate same output for 
@SecurityRequirement scopes section

Note that problem is not possible to solve by only updating template files.
This since only sections still containing scope values are for the security definition where all scopes are listed

##### Suggest a fix/enhancement

Problem seems to have occurred when refactoring for swagger 3 was done in processOperation (DefaultGenerator.java) where code for authMethods was changed.

lindgrenfredrik:
For some reason part of the text I wrote was lost:
##### Steps to reproduce

Run commands from Command line section

Note this is not only a template issue, scope group per operation isn't available

##### Suggest a fix/enhancement
Issues seems to have occurred as part of refactoring of processOperation function in DefaultGenerator.java
It seems that authMethod being set there is not getting same value as it was in swagger2

lindgrenfredrik:
might be duplicate of this defect
https://github.com/swagger-api/swagger-codegen/issues/9498

ashish211985:
has this issue been resolved .. i m trying to add OAuth2 scopes using OpenApi 3.0 but when the server stub is getting generated all the scopes are getting attached to all the operation functions. Using spring & java ? @lindgrenfredrik

lindgrenfredrik:
[...]

PR #12410, opened 2024-06-06
#9843: Sort out OAuth scopes for individual endpoints

ionel-sirbu-crunch:
This is a fix for issue #9843 .

According to the [OpenAPI spec](https://swagger.io/docs/specification/authentication/oauth2/), one is supposed to define the superset of all OAuth scopes available globally for an API under the `securitySchemes` block, then reference only a required subset from those scopes for an individual endpoint. Example [here](https://swagger.io/docs/specification/authentication/oauth2/#scopes-extra).

As described in issue #9843 , the generated data model that is passed to the templating engine, regardless of language chosen, always uses the full list of OAuth scopes defined globally, ignoring the subset of references used for individual endpoints.

For context, I'm using `io.swagger.codegen.v3:swagger-codegen-maven-plugin` for my requirements, converting to static html (API documentation).


ionel-sirbu-crunch:
Is there anything I need to do here? I don't quite understand from the auto-generated comment above...
I hope this project is still maintained.

michaelcrunch (approved):

PR #12502, opened 2024-12-18
prepare release 2.4.44

frantuma:


fgreinacher:
@frantuma I saw the [Docker build failing](https://github.com/swagger-api/swagger-codegen/actions/runs/12389035407/job/34581266257) because the new base image only supports amd64 :/

image

I see two quick and easy options:

1. Roll back the base image change
2. Do not build the generator image for all these platforms
   - We might upgrade to a more recent Java version to at least have support for ARM64:
     image

PR #12611, opened 2025-09-19
feat: prevent path traversal attacks

ewaostrowska:


modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
```diff
@@ -3562,6 +3563,11 @@ public String apiTestFilename(String templateName, String tag) {
     }
 
     public boolean shouldOverwrite(String filename) {
+        try {
+            SecureFileUtils.validatePath(filename);
+        } catch (SecurityException e) {
+            return true;
```
daniel-kmiecik, line 3569: Why we want to continue the process?

modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultGenerator.java
```diff
@@ -801,6 +804,7 @@ public Reader getTemplate(String name) {
                     .compile(template);
 
             writeToFile(adjustedOutputFilename, tmpl.execute(templateData));
+            SecureFileUtils.validatePath(adjustedOutputFilename);
```
daniel-kmiecik, line 807: It's already validated at line 790

modules/swagger-codegen/src/main/java/io/swagger/codegen/utils/SecureFileUtils.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.IOException;
+
+
+/**
+ * Utility class for secure file operations that prevent path traversal attacks.
+ * Uses a simplified approach focusing on canonical path validation and allowlist-based security.
+ */
+public class SecureFileUtils {
+
+    private SecureFileUtils() {
```
daniel-kmiecik, line 13: Please add logs before each exception is thrown

daniel-kmiecik (approved):

PR #12627, opened 2025-10-06
next snapshot 3.0.75-SNAPSHOT

djankows:


ewaostrowska:
What about pom.java11.xml and pom.java8.xml?

ewaostrowska (approved):

Issue #8263, opened 2018-06-01
[Python] Swagger API Clients will log to stderr by default


asqui:
Hello,

I am using the [OpsGenie SDK](https://github.com/opsgenie/opsgenie-python-sdk), which is autogenerated using Swagger, and was most surprised to find printing urlib3 retry notices out to stderr in my application.

The culprit appears to be the default behaviour of configuring a stderr `StreamHandler` here:
https://github.com/swagger-api/swagger-codegen/blob/74f70a19243a4bf022786c7b381579b4485d73a8/modules/swagger-codegen/src/main/resources/python/configuration.mustache#L133

This is in contradiction to the guidance in the Python logging docs: https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library

Most notably, these docs say:
> If the using application does not use logging, and library code makes logging calls, then (as described in the previous section) events of severity WARNING and greater will be printed to sys.stderr. This is regarded as the best default behaviour.

So if the Swagger template did not set up this `StreamHandler`, the default behaviour for an application that does not configure logging would be the same.

Instead, what is happening now is that this `StreamHandler` interferes with applications that do configure the desired logging handlers, and causes duplicate warnings to be logged to stderr, in addition to the ones that are being captured by the other handlers configured by the application.

Also from the Python logging docs linked above:
> Note It is strongly advised that you do not add any handlers other than NullHandler to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements

To make this a bit more concrete, in our application we never expect any output to stderr, and as a fallback we capture any and email it. An unintended consequence of adopting a Swagger-generated API is that now the application has started generating stderr under certain circumstances, which is both surprising and undesirable.

This seems similar to issue #3687, which appears to be the same bug for Go.

Hope that makes sense? Keen to hear thoughts...

PR #12640, opened 2025-10-20
[python] Don't set a default logging stream handler

nickbroon:
This fixes the issue of the generated python library setting a default stream handler logger, which can subvert an applications intention and goes against python coding best practice described in #8263.

The solution takes the same approach as that in https://github.com/OpenAPITools/openapi-generator/pull/480

Closes: #8263


nickbroon:
Not sure who the appropriate people are to seek review from... @ewaostrowska, @djankows @daniel-kmiecik are you able to help?

PR #12648, opened 2025-11-04
prevent polynomial regular expression used on uncontrolled data secur…

ewaostrowska:
### Description of the PR

- replaceAll uses a backtracking engine. The pattern ([A-Z]+)([A-Z][a-z][a-z]+) had a greedy quantified group ([A-Z]+) followed by another subpattern that starts with an overlapping class [A-Z] and then [a-z][a-z]+. Inputs like "A...A a...a" (many uppercase, then many lowercase) could have had caused an issue. 
- String.replaceAll did compile the regex on each call. Using Pattern allows to compile regex once and reuse it. 

The outcome of the method should be the same

### DIFF
--- a/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
+++ b/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
@@ -3212,28 +3212,20 @@ public class DefaultCodegen {
 
     /**
      * Underscore the given word.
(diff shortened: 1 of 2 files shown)


modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java
djankows: not anymore ;]

daniel-kmiecik (approved):

PR #12651, opened 2025-11-17
fix: remove file from fileMap in online generator

daniel-kmiecik:
### PR checklist

### Description of the PR

- System.out.println statements were replaced with a logger
- logic for reading file bytes and deleting the file was extracted into separate methods
- a null check for g and its filename was added to return a 404 response instead of throwing an exception when the file does not exist in the map
- remove downloaded file from fileMap after download


ewaostrowska:
Please consider updating the documentation with a sentence about 
- possible 404 when no file 
- no authentication needed

